Issue #8, opened 2015-09-25
Store state in RocksDB


srderson:
The world state needs to be persisted in RocksDB. @manish-sethi and I are discussing and implementing this feature.

Planned implementation
- In the database, state is grouped by chaincode
- When modifying the state, an in memory map is made of all changes
- When a block is added (after consensus) the state is committed to the DB
- A hash of each modified chaincode state must be calculated
- A global hash of all chaincode hashes is calculated and stored in the block

@manish-sethi and I discussed possibly not storing the global hash in the block and only storing the hash of chaincodes that had been modified in the given block. @binhn said that this method would require too many comparisons across the network during consensus. The pro to this method is that the hash calculation per block would be faster and for private chaincodes, it would be possible to not store encrypted data on peers that do not have access. @binhn said he wanted the encrypted data to be on ALL nodes so that it would be distributed and recoverable in the event of a data loss across a large number of nodes.

PR #26, opened 2015-09-28
Initial pass at storing state in RocksDB. Enhancement #8

srderson:


srderson:
This is a very simple implementation for milestone 1. @manish-sethi is completing the actual implementation, but we're not sure if it will be ready for the milestone yet.

PR #29, opened 2015-09-28
Changed openchain.yaml's peer.Dockerfile entry to include RocksDB ins…

### LINKED ISSUE #21: openchain.yml has references to hub.jazz.net
openchain.yml contains the following line

RUN go install hub.jazz.net/openchain-peer && cp src/hub.jazz.net/openchain-peer/openchain.yaml ./bin

The repository and project name need to be fixed.

### LINKED ISSUE #28: Building Peer from Dockerfile needs to incorporate RocksDB

jeffgarratt:
…tall.  Also fixed references to old project.  Closes #21, Closes #28.

### DIFF
--- a/openchain.yaml
+++ b/openchain.yaml
@@ -20,9 +20,20 @@ peer:
 
     Dockerfile:  |
         from golang:1.5.1
+        # Install RocksDB
+        RUN cd /opt  && git clone --branch v3.13.1 --single-branch --depth 1 https://github.com/facebook/rocksdb.git && cd rocksdb
+        WORKDIR /opt/rocksdb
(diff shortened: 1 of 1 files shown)


srderson:
When is this run? Compiling RocksDB is not a fast process, so if it's something we're running every time a chaincode is deployed, we need to open an issue for putting this in the base docker image.

jeffgarratt:
@srderson  This will only build RocksDB 1 time as the docker system will cache these portions(images) of the dockerfile commands.  It is planned to be used for creating the images for the openchain peer itself to be used during development.

PR #37, opened 2015-10-01
Merging initial Openchain API changes

angrbrd:
Hi @srderson and @jeffgarratt.

This the initial Openchain API changes introducing some of the APIs related to the blockchain. I'd appreciate if Sheehan you could take a look at the Go code and the associated unit tests that would be much appreciated. Jeff if you could take a look at the protobuf code for any comments.

Thanks!


angrbrd:
@srderson , I know I need to add the header to these new files. I am wondering if I need to add it only on the .go files? Or should I also add it on the .proto files too?

Thanks!

srderson:
It should probably be in the proto files if that does not break anything.

srderson:
Looks good. There's a case where you call
`s.blockchain.db.GetLastBlock()`
This should probably be
`s.blockchain.GetLastBlock()`

@manish-sethi has hidden access to the DB in his branch at https://github.com/manish-sethi/obc-peer/tree/issue/8/state_persistence and provided most of the APIs you will use directly in blockchain.go. We can update your code when merging his PR.

Also, the size checks should probably be done in blockchain.go or the DB so not everyone has to think about that.

angrbrd:
@srderson , I have added the Apache license header to all the .proto files and the two new .go files in commit 4c9e714907da9777c5357a7902904bd8d53c517c

Thanks!

PR #52, opened 2015-10-04
Issue/8/state persistence

manish-sethi:
Changes for issue#8. Main changes include
1) Introduced a configuration in openchain.yaml for db location (default /var/openchain/db)
2) Keeping state changes in memory and committing atomically along with persisting block
3) Computation of state hash based on high-level idea discussed
4) Changes to files api.go and api_test.go for making them compatible with the changes


srderson:
I tried this quickly and it seems a number of tests are failing with the following error.

```
--- FAIL: TestChain-2 (0.00s)
    state_test.go:101: Error in creating test db. Error = [IO error: /var/openchain/db: No such file or directory]
```

On my VM, there is a `/var` directory, but there is no `/var/openchain` directory.

manish-sethi:
Hi @srderson  - Can you try with changing the directory from /var/openchain/db to /tmp/openchain/db. This could be because of permission problems. I had tested this with /tmp/openchain/db value in openchain.yaml.
I am not sure what final value we should keep in openchain.yaml. Please let me know if you face further issues

srderson:
Changing the dir to /tmp solved the problem. Thanks! I've updated the default to be /tmp. We probably don't want to leave that as the default when we ship, so I will open an issue.

PR #76, opened 2015-10-18
Issue/44/async indexing and delta state

### LINKED ISSUE #42: Improve database design
The database design needs to support two modes of operation.
1. Running transactions and consensus
2. Queries from clients
- Consensus will primarily be performed by validating peers. It involves
  - Updating the state
  - Calculating the state hash
  - Storing a block on the blockchain
  - Sending transaction results and state deltas to peers that do not have access to a private transaction
- Queries from clients
  - Get block by number
  - Get block by hash
  - Get transactions in a block
  - Get transaction by index in a block
  - Get transaction by hash
  - Get transaction by address
  - Get transaction by identity
  - Get chaincodes I have access to

Currently the database format is
`Block number -> Block bytes`
which is fast for consensus, but my be slow for queries. We could consider

```
Block number -> block hash
block hash -> block bytes` (where block bytes contains an array of transaction hashes
transaction hash -> transaction bytes
```

but this may be slower for consensus. However, this could reduce the overall size of the DB as we do not need to store duplicate transactions.

Should we consider two different database schemas, one optimized for validating peers and one optimized for non-validating peers?

### LINKED ISSUE #44: Store state deltas in the database
Peers need to store the last X state deltas between blocks in the database so another peer can catch up with the blockchain.

Example
- A peer requests a download of the entire state. See issue #43.
- While download the state, the blockchain continues to progress.
- The peer must now be able to ask for state deltas for subsequent blocks so that it can obtain state change information for private transactions that it may not have the ability to run.

X, the number of state deltas stored by a peer, should be configurable in openchain.yaml.

manish-sethi:
Signed-off-by: Manish Sethi manishsethi@in.ibm.com

Changes include
1) Added code for storing state deltas for last few blocks.
2) Two implementations for indexers (sync and async) for indexing block-data. Default configured for using sync implementation
Configurations for both above are in the code. However, we can expose these in yaml file if we think suitable.
This closes #42 and closes #44

### DIFF
--- a/openchain/blockchain.go
+++ b/openchain/blockchain.go
@@ -57,12 +57,19 @@ func GetBlockchain() (*Blockchain, error) {
 			}
 			blockchainInstance.previousBlockHash = previousBlockHash
 		}
+		err = startIndexer()
+		if err != nil {
+			return nil, err
+		}
 	}
 	return blockchainInstance, nil
 }
 
 // GetLastBlock get last block in blockchain
 func (blockchain *Blockchain) GetLastBlock() (*protos.Block, error) {
+	if blockchain.size == 0 {
+		return nil, nil
+	}
 	return blockchain.GetBlock(blockchain.size - 1)
 }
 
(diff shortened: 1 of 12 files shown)


manish-sethi:
Signed-off-by: Manish Sethi manishsethi@in.ibm.com

ghost:
Thank you for your contribution! This is a read only mirror, however. Please submit your change using gerrit.

you would need to check out the repo here: https://gerrit.hyperledger.org/r/#/admin/projects/fabric and make your edits, pushing them to gerrit

here is a walkthrough for zephyrproject which is much the same WRT setting up an LFID, adding ssh keys, etc:
https://www.zephyrproject.org/doc/1.1.0/collaboration/code/gerrit_accounts.html

feel free to ask on rocket.chat - https://chat.hyperledger.org/ - discuss in #fabric

one note: when you set up your LFID do not use social logins the first time, create the account and you can add social logins later

PR #77, opened 2015-10-19
issues#67 - partial fix: add start/stop containers

muralisrini:
added start and stop to the vmcontroller framework. "start" will create a docker exec context and then start . This process assumes the container was created earlier (during build). The created container will currently exist forever. "start" will create the container if one doesn't exist with the new args passed at this time. Perhaps in this case we should delete the container after the exec.


muralisrini:
Signed-off-by: Srinivasan Muralidharan muralisr@us.ibm.com

PR #78, opened 2015-10-19
Introducing the Openchain REST service

angrbrd:
-- Introduce the Openchain REST service, ServerOpenchainREST.
-- Expose two service endpoints: /chain and /chain/blocks/{Block}.
-- Modify main.go and openchain.yaml to incorporate the new REST
service, listening on port 3000.
-- Begin compiling the REST interface documentation in rest_api.json
that will be readable through Swagger.

### DIFF
--- a/main.go
+++ b/main.go
@@ -38,6 +38,7 @@ import (
 	"google.golang.org/grpc/grpclog"
 
 	"github.com/openblockchain/obc-peer/openchain"
+	"github.com/openblockchain/obc-peer/openchain/rest"
 	pb "github.com/openblockchain/obc-peer/protos"
 )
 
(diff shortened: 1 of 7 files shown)


angrbrd:
@jeffgarratt,

These are the initial stubs for the Openchain REST service. I have exposed the endpoints of /chain as well as /chain/blocks/{Block}. The /chain endpoint retrieves information on the current state of the blockchain and the /chain/blocks/{Block} endpoint retrieves the information on individual blocks within the blockchain. I have also created an file for Swagger documentation.

I'd appreciate if you could pull this down and make sure that it is working for you before you merge. As you remember I had some issues getting this to work with the GetBlockchain() function that retrieved the blockchain singleton object. I believe I sorted that out, but it would be great if you could double check. Also, take a glance at the Swagger file, any comments you have are welcome.

This submission will close issue#47.

Thanks!

srderson:
Some minor issues.
1. Add github.com/gocraft/web to Gom file
2. Remove mapping of port 5000 -> 4000 in vagrant file
3. http://127.0.0.1:3000/chain/blocks/test returns 200. Should be 400. Need a test case for this
4. Weirdness now with DB because test case and peer run in two different processes. Will be resolved once test case builds sample chain using sendTransaction rest API in peer.

Will merge then we can fix these

Issue #74, opened 2015-10-15
send() function on consensus framework


binhn:
We need a function to send a message to the validators

kchristidis:
Implemented as `broadcast()` in the current milestone.

Issue #71, opened 2015-10-15
receive() for consensus algorithm


kchristidis:
The implementation-specific "**consensus algorithm**" module (q.v. issue #70 for context) should have a `receive(message)` function as part of its interface.

For now at least, this `message` is a client request originating from the **P2P** layer, and passed to us by the generic **consensus** module.

This message will either be packaged into a block in the **consensus algorithm** module, or prompt the transmission of a consensus phase message (e.g. `PRE-PREPARE`, `PREPARE`, or `COMMIT` for PBFT).

kchristidis:
For the current sprint ending tomorrow, we are handling this by means of a validator/leader FSM and a  `messageHandler()` function.

Proper abstraction still necessary.

PR #86, opened 2015-10-21
Add definitions for consensus messages, fixes #40

### LINKED ISSUE #40: Consensus messages
Define the message structure for the consensus messages that will be passed between the "**consensus**" and the "**consensus algorithm**" modules. (q.v. issue #70 for context)

kchristidis:
I'm not signing this off (DCO, etc.) until I'm sure it's good to go.

@binhn, @muralisrini, and @jeffgarratt please review and suggest fixes if necessary.

@jeffgarratt: I'm wondering whether we need to define RPC service interfaces on those messages. (See #71 and #74.)


kchristidis:

Issue #89, opened 2015-10-21
peer fails to start if DB does not exist


srderson:
If the DB does not exist, the peer will fail to start with the following error

```
vagrant@vagrant-ubuntu-trusty-64:/opt/gopath/src/github.com/openblockchain/obc-peer$ ./obc-peer peer
17:11:31.886 [main] serve -> DEBU 001 Running as peer
Error opening DB Invalid argument: /var/openchain/db: does not exist (create_if_missing is false)
panic: Could not open openchain db error = [Invalid argument: /var/openchain/db: does not exist (create_if_missing is false)]

goroutine 1 [running]:
github.com/openblockchain/obc-peer/openchain/db.GetDBHandle(0x4dea76)
    /opt/gopath/src/github.com/openblockchain/obc-peer/openchain/db/db.go:89 +0x19e
github.com/openblockchain/obc-peer/openchain.fetchBlockchainSizeFromDB(0xc20810b830, 0x0, 0x0)
    /opt/gopath/src/github.com/openblockchain/obc-peer/openchain/blockchain.go:200 +0x37
github.com/openblockchain/obc-peer/openchain.(*Blockchain).init(0xc20810b890, 0x0, 0x0)
    /opt/gopath/src/github.com/openblockchain/obc-peer/openchain/blockchain.go:58 +0x34
github.com/openblockchain/obc-peer/openchain.GetBlockchain(0x7f427aab8ca8, 0x0, 0x0)
    /opt/gopath/src/github.com/openblockchain/obc-peer/openchain/blockchain.go:48 +0x74
github.com/openblockchain/obc-peer/openchain.NewOpenchainServer(0xc2080ea7d0, 0x0, 0x0)
    /opt/gopath/src/github.com/openblockchain/obc-peer/openchain/api.go:41 +0x3a
main.serve(0x0, 0x0)
    /opt/gopath/src/github.com/openblockchain/obc-peer/main.go:315 +0x75e
main.func·001(0xf79ce0, 0xfa2088, 0x0, 0x0)
    /opt/gopath/src/github.com/openblockchain/obc-peer/main.go:62 +0x1f
github.com/spf13/cobra.(*Command).execute(0xf79ce0, 0xfa2088, 0x0, 0x0, 0x0, 0x0)
    /opt/gopath/src/github.com/spf13/cobra/command.go:501 +0x82f
github.com/spf13/cobra.(*Command).Execute(0xf79b20, 0x0, 0x0)
[...]

PR #90, opened 2015-10-21
Db create

manish-sethi:
Signed-off-by: Manish Sethi manishsethi@in.ibm.com
db gets created if the db path does not exist on file system or the path is empty.
However, if a non-empty exists, we do not create a db and give control to rocks db. If the path happens to be a valid rocksdb, the db can be used else rocks db throws error (thought, strangely the error from rocks db says that path does not exist)

### DIFF
--- a/openchain.yaml
+++ b/openchain.yaml
@@ -82,7 +82,7 @@ peer:
         validator:
             enabled: false
         leader:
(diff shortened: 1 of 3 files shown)


manish-sethi:
closes #89

PR #92, opened 2015-10-21
Introduce error check when parsing URL

### LINKED ISSUE #83: GET /chain/blocks/test returns block 0
GET /chain/blocks/test returns a 200 with block 0. It should return a 400 as it's not a valid block

angrbrd:
Introduce an error check when parsing the REST URL to insure that the
Block id is specified as an unsigned integer.

Fixes #83.


srderson:
Could you add a test case to ensure /chain/blocks/foobar returns 400?

PR #99, opened 2015-10-22
Add leader FSM, update PBFT message definition

kchristidis:


kchristidis:
@muralisrini Note that the last 3 callbacks in `openchain/leader/leader.go` will need to be modified so that they're match what you're doing in `openchain/validator.go`.

Issue #43, opened 2015-10-02
Support state download


srderson:
state.go needs to provide API for fetching the entire state of the current block. This API will
- Capture a read only snapshot of the entire state so the peer can continue adding new blocks.
- Provide a stream of the read only snapshot so that a requesting peer can download the entire state.

PR #111, opened 2015-10-22
Ability to create snapshot of the global state. Issue #43

srderson:

This is code for creating a state snapshot. @manish-sethi Please review if you have a chance.

### DIFF
--- a/openchain/ledger/state_snapshot.go
+++ b/openchain/ledger/state_snapshot.go
@@ -1,0 +1,66 @@
+/*
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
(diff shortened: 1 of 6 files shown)


openchain/ledger/state_snapshot.go
```diff
@@ -0,0 +1,66 @@
...
+
+// newStateSnapshot creates a new snapshot of the global state for the current block.
+func newStateSnapshot() (*stateSnapshot, error) {
+	chain, err := getBlockchain()
+	iterator, snapshot := db.GetDBHandle().GetStateCFSnapshotIterator()
+	if err != nil {
+		return nil, err
+	}
+	blockNumber := chain.getSize()
+	stateSnapshot := &stateSnapshot{blockNumber, iterator, snapshot}
```
manish-sethi, line 41: For safety, better to pull highest block number from the snapshot. New blocks may be committed after snapshot have been created. 
(Though, we need to work on synchronizing the access to variables of blockchain struct (e.g., size, previousBlockHash etc.) for other cases as well, but in the case of snapshot - I guess, we can pull the block number from the snapshot)

openchain/ledger/state_snapshot_test.go
```diff
@@ -0,0 +1,66 @@
...
+	state.set("chaincode1", "key7", []byte("value7"))
+	state.set("chaincode2", "key8", []byte("value8"))
+	commitTestState(t, 3)
+
+	snapshot, err := state.getSnapshot()
+
+	if err != nil {
+		t.Fatalf("Error fetching snapshot: %s", err)
+	}
+
```
manish-sethi, line 54: Can we add a few keys after the snapshot creation as well to make sure that they do not appear in snapshot iterator?

cclauss:
@christo4ferris Your review please?

PR #124, opened 2015-10-26
Remove '/' from chaincode ID and transaction key

srderson:


srderson:
@jeffgarratt Merge for fetching state key that contains deploy chaincode transaction UUID.

PR #141, opened 2015-10-28
Clean up protos package, closes #140

### LINKED ISSUE #140: Clean-up protos package
Actions needed:
- [x] Message definitions are all over the place; re-group as necessary.
- [x] Remove unnecessary imports.
- [x] Add package-level logger.
- [x] Document exposed functions.

kchristidis:
There are three more issues as far as I can tell:
1. There are undocumented structures. If you've created a structure definition and it's not immediately obvious how/where it should be used, and what its fields stand for, please add some documentation to explain.
2. I think the test cases for `block_test.go` and `transaction_test.go` may need some further refining.
3. The [protocol buffer style guide](https://developers.google.com/protocol-buffers/docs/style#message-and-field-names) suggests that we use underscore_separated_names for field names. `Block` is the obvious offender here, but I'm not touching it so as not to break compatibility with all the layers it's interacting with.

I've labeled all the pending items with `TODO`.

The first issue is the most important one; please proceed with proper documentation when you find some time.


kchristidis:
Note: the only test that fails is the `consensus_test` one, which I've fixed in #142.

(New pull request coming after we merge this one with the master branch.)

Issue #132, opened 2015-10-28
Implement ExecTXs in consensus helper


muralisrini:
Basically move the meat of `validator.executeTransactions` to `consensus/helper.ExecTxs`.

binhn:
@muralisrini when you select a milestone to assign an issue, pay attention to the date on the milestone. Milestone 2 has been done.

PR #175, opened 2015-11-02
Move validator functionality to consensus layer, fixes #115, #132

### LINKED ISSUE #115: Refactor validator into PBFT implementor to use the consensus abstraction
The `pbft` package should implement the abstract `Consenter` interface to receive the call whenever a message of type `ConsensusMessage` arrives.

The consenter will receive the payload of the `ConsensusMessage` structure.

See #108 for more details.

kchristidis:
Do not merge, pushing this for review by @binhn, @muralisrini.

This is a proposal for the consensus abstraction. The aim is to provide a clearer abstraction than before and a basis for developers to build on.

A few pieces (such as the `streamReceiver` for instance) will probably have to be pushed back into the communications/peer layer, but until the APIs for that layer are finalized, it's best to minimize our dependence on it.

Further changes needed: `main.go` will have to be modified so as to call `consensus.go` instead of `validator.go`. `validator.go` and `leader.go` will have to be deleted.


binhn:
Here are my comments:
1. FSM should probably be left to the plugin rather than part of streamReceiver object. Some plugin might not need FSM as they would have a simple protocol (eg proof of work)
2. I meant to ask you about this Consenter.GetParam(param string) use-case. Would we need this to call this func in the abstract layer? If not, I would leave it to the plugin
3. In consensus.go, I noticed a few structs and interfaces are lower-case, did you want them to be as internal to consensus package?

kchristidis:
We discussed this in person, but I'm recording my answers here for reference as well:
1. A valid point. In that case, I'll move it down to `streamReceiver.plugin`.
2. I added this in the `consensus` layer so that the `plugin` dev can have easy access to it. (I may as well have added it to the `utils` package.) I think it is a handy function to have as it can read any config file (i.e. not just `plugin/config.yaml`). I will move it down to the `plugin` package until we see the need to move it up again.
3. As a general rule, I keep all structs/interfaces/functions lowercase until I see an explicit need to export them. Several of these definitions will be uppercased when we finalize on the code from the other modules that interacts with them.

muralisrini:
I may have more comments after some of the higher level code comes in. 
1. Why do we need "stream openchain.PeerChatStream" on streamReciever ? It implements "handleMessage"
   method with OpenchainMessage passed it. So implies stream is outside the scope of streamReceiver with messages just flowing through it.  (I could be missing something...)
2. many comments use the word "peer" but feel should be "validator" or "validating peer" ? For instance "GetReceiver(stream openchain.PeerChatStream): Attaches a `streamReceiver`
   +// for that stream to the peer's `consensusLayer` thus allowing them to transact
   +// on this link."

kchristidis:
@muralisrini, thank you for the comments.
1. Look at the definition of `sendMessage` for example. (It is also part of the `messageHandler` interface, as is `handleMessage`.) When you do `streamReceiver.sendMessage(msg)`, you are effectively calling `streamReceiver.stream.Send(msg)`. That's why you need it. Now, you can argue that we shouldn't be dealing with streams at all in consensus. I agree with that. I do however need a model to get the ball rolling and start iterating without waiting on other packages
2. You're right, I'll make those edits.

kchristidis:
Passes all tests, but do note there are a few TODOs that I need to take care of before #182 is considered closed.

I have moved my abstraction to a new temporary package called `newconsensus`, to avoid an import cycle caused by the code that is already in the `consensus` package.

Pushing this as requested by @binhn so that he can work on his issue.

Issue #165, opened 2015-10-30
Expose state delta APIs in ledger.go


srderson:
This is needed for synchronizing the state.

binhn:
done per murali

PR #205, opened 2015-11-04
Expose state delta API. Issue #165

srderson:
This code is for @muralisrini. I am not signing off at this point because I need to add test cases.


binhn:
@muralisrini please take a look at see if you can accept this PR then sign off and merge. Thanks.

PR #229, opened 2015-11-12
Make CPI oblivious to transaction format

corecode:

### DIFF
--- a/openchain/consensus/consensus.go
+++ b/openchain/consensus/consensus.go
@@ -20,10 +20,7 @@ under the License.
 package consensus
 
 import (
-	pb "github.com/openblockchain/obc-peer/protos"
-
 	"github.com/op/go-logging"
-	"golang.org/x/net/context"
 )
 
 // =============================================================================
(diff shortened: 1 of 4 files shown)


kchristidis:
You are combining two changes into one commit here.

Besides making CPI oblivious to the transaction format (which is a good change), you are also changing the way `ExecTXs()` works.

It used to return an array of errors, now it returns one error. There's a logic behind returning an array of errors. Discuss with @muralisrini, and submit your suggestion for a new `ExecTXs()` as a separate PR if necessary.

Until then, you may want update this PR by going back to the previous form of `ExecTXs()`. (Feel free to do the change where its argument is a `transaction []byte` though, as this is inline with what you suggest in this PR.)

corecode:
I prefer to discuss the reason for multiple errors.

binhn:
@corecode I am closing this PR due to @kchristidis's comment above. The reason we need multiple errors to eliminate bad tx's. Bring up during scrum if you need further discussion.

corecode:
Does closing this PR mean that you are not interested in any of the proposed changes?  Typically the PR gets updated based on feedback.

I believe the error results of executed transactions is important, but too specific for the consensus plugin:  the errors need to be handled elsewhere in the stack.  If you believe that the consensus plugin should be exposed to the transaction errors, please provide guidance on how the consensus plugin should handle these errors.

I think we should think about and discuss issues before talking about them during scrum - only that way we can make sure that people can actually think about the issues, instead of deciding based on a gut feeling.

binhn:
@corecode well, you didn't ask for feedback but signed off already.  You just come onboard and lack of the history on this. We have discussed and concluded how to implement, and that's why the function was implemented the way it is. Changing existing function without consulting and get agreement with the tech leads is not welcomed.

corecode:
Unfortunately I do not have any access to previous design discussions.  I will wait with code until I get more details on existing design decisions.

kchristidis:
@corecode: I have already pointed you to the right direction. @muralisrini is whom you should be talking to about `ExecTXs()`.

You also wrote that "the PR gets updated based on feedback". I gave you concrete feedback and invited you to update the PR. You ignored it.

Your preference to "discuss the reason for multiple errors" does not address my feedback. Your changing the `ExecTXs()` does **not** belong to this commit, or this PR.

Split these two changes (rendering the CPI oblivious to transaction formats + changing ExecTXs) into two separate PRs, explain the thinking behind them, and those responsible for the original code will chime in with their thoughts.

PR #230, opened 2015-11-12
Remove FSM from PBFT plugin

corecode:
Builds on #229.


kchristidis:
This is a good change since we're no longer using the FSM.

Let's remove the dependency from #229?

PR #240, opened 2015-11-15
No consensus consensus

### LINKED ISSUE #179: Transactions should work without consensus
Make sure the consensus abstraction work as default without any implementation plugin. This is useful for development and test.

Default plugin will call the chaincode without consensus.

binhn:
close #179


binhn:

PR #251, opened 2015-11-17
Take2 implement query/issues/153

muralisrini:
query implementation - 
    main implementation and test case
   race condition fixes
   error handling


muralisrini:

PR #254, opened 2015-11-18
Added support for rolling back a tx if chaincode execution fails

manish-sethi:
Introduced two methods in Ledger.go. 
TxBegin(txUuid string)  and TxFinished(txUuid string, txSuccessful bool). Any state changes API should be called after calling TxBegin(txUuid string) in order to associate changes with a tx. If txSuccessful flag in TxFinished method is true then only changes made by the tx are retained, otherwise - the changes are discarded.
@srderson Please see if the changes in ledger package looks fine.
@muralisrini Please review that the changes made in exectransaction.go are fine.

### DIFF
--- a/openchain/api_test.go
+++ b/openchain/api_test.go
@@ -23,12 +23,11 @@ import (
 	"bytes"
 	"testing"
 
-	"google/protobuf"
-
 	"github.com/openblockchain/obc-peer/openchain/ledger"
 	"github.com/openblockchain/obc-peer/openchain/util"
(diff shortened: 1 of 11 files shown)


manish-sethi:
Signed-off-by: manishsethi@in.ibm.com

srderson:
I reviewed the code changes and it looks good. I also ran the test cases and they passed.

srderson:
I spoke with @muralisrini and he said this is good to merge.

Issue #253, opened 2015-11-18
Peer should use stream for all API calls (including "self" calls in dev mode/validator mode)


muralisrini:
CLI deploy, invoke and query calls are grpc driven and land as a devops service call. Currently, the service directly drives chain code execution in devmode whereas in true "peer" mode (where the validator is on a different node) the grpc call is sent over a stream to the validator. We should not have two methods but just have one so we can leverage the full functionality on the pipeline(such as any security mechanisms we'd implement on the validator).

muralisrini:
The code for this issue is complete (devops streams CLI grpc requests both "locally" (dev mode) or "remote" (non-validating peer mode). Will close it after making sure TestServer_Chat test case works (issue #256)

markparz:
sprint 4, related to issue #300

muralisrini:
TestServer_Chat test case (#256) is fixed. Plus this code has been in the works for a while. Closing

Issue #256, opened 2015-11-18
TestServer_Chat test case in peer_test.go.go fails


srderson:
The test case TestServer_Chat in peer_test.go fails with the following error

```
chainlet env using [startuptimeout-20000, chaincode run mode-net_mode, peer address-0.0.0.0:30303]
2015/11/18 17:16:03 setting Number of procs to 2, was 2
2015/11/18 17:16:03 transport: http2Client.notifyError got notified that the client transport was broken EOF.
--- FAIL: TestServer_Chat (0.03s)
    peer_test.go:76: Received message: DISC_HELLO
    peer_test.go:72: stream closed with unexpected error: rpc error: code = 13 desc = "transport is closing"
FAIL
FAIL    github.com/openblockchain/obc-peer/openchain/peer   0.070s
```

The peer exits with the following error

```
17:16:03.853 [peer] beforeHello -> DEBU 00a Received DISC_HELLO from endpoint=
17:16:03.854 [peer] beforeHello -> DEBU 00c Received DISC_HELLO, sending back DISC_HELLO
panic: runtime error: invalid memory address or nil pointer dereference
    panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xb code=0x1 addr=0x0 pc=0x5645d7]

goroutine 24 [running]:
github.com/openblockchain/obc-peer/openchain/peer.getHandlerKey(0x7ff581cb3128, 0xc820061c00, 0x0, 0x0, 0x0, 0x0)
    /opt/gopath/src/github.com/openblockchain/obc-peer/openchain/peer/peer.go:209 +0x1d7
github.com/openblockchain/obc-peer/openchain/peer.(*Peer).DeregisterHandler(0xc8201d6f90, 0x7ff581cb3128, 0xc820061c00, 0x0, 0x0)
    /opt/gopath/src/github.com/openblockchain/obc-peer/openchain/peer/peer.go:235 +0x4f
github.com/openblockchain/obc-peer/openchain/peer.(*Handler).Stop(0xc820061c00, 0x0, 0x0)
    /opt/gopath/src/github.com/openblockchain/obc-peer/openchain/peer/handler.go:101 +0x97
github.com/openblockchain/obc-peer/openchain/peer.getHandlerKey(0x7ff581cb3128, 0xc820061c00, 0x0, 0x0, 0x0, 0x0)
    /opt/gopath/src/github.com/openblockchain/obc-peer/openchain/peer/peer.go:209 +0x1d7
[...]

PR #257, opened 2015-11-18
Skip test case. Need to be reworked after issue 253 completed

srderson:
Skip test case based on info from @jeffgarratt and @muralisrini. This test case will need to be reworked after issue #253 is completed.


srderson:
Removing PR. Test case can not be skipped as it's causing problems running a non-validating peer based on a comment in issue #256

PR #271, opened 2015-11-19
Including crypto-hashes for state-deltas caused by each Tx in a Tx-batch

manish-sethi:
@binhn @muralisrini @kchristidis In this PR, I have added a new method in ledger.go
GetTempStateHashWithTxDeltaStateHashes() ([]byte, map[string][]byte, error)
first return value is same as in existing method 'GetTempStateHash' (i.e., global state hash)
second value is a map of (txUuid -> hashOf(stateDelta caused by the tx) that can be used to find a 'non-deterministic' chaincode if a mismatch in statehash occurs.

@srderson Please have a look if changes looks OK.

### DIFF
--- a/openchain/ledger/ledger.go
+++ b/openchain/ledger/ledger.go
@@ -117,6 +117,14 @@ func (ledger *Ledger) GetTempStateHash() ([]byte, error) {
 	return ledger.state.getHash()
 }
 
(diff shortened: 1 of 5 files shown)


manish-sethi:

srderson:
This looks good and it's not breaking any test cases that are not failing already. Will merge.

PR #298, opened 2015-11-20
deploy command should not be treated as a query

muralisrini:
this completes the work needed for bring up validators with noops and doing transactions and queries with them


muralisrini:

PR #307, opened 2015-11-23
removed system wide chain properties into properties per chain. Also …

muralisrini:
…added viper config to api_test (it appears to have been leaning on viper config being read from chaincode. When chaincode properties were passed in api_test lost config properties)

### DIFF
--- a/main.go
+++ b/main.go
@@ -20,14 +20,14 @@ under the License.
 package main
 
 import (
-	"bufio"
 	"encoding/json"
 	"errors"
 	"fmt"
 	"net"
-	"os"
 	"runtime"
+	"strconv"
 	"strings"
+	"time"
 
 	"golang.org/x/net/context"
 
@@ -335,8 +335,9 @@ func serve(args []string) error {
 	// Register the Admin server
 	pb.RegisterAdminServer(grpcServer, openchain.NewAdminServer())
 
-	// Register ChaincodeSupport server
-	pb.RegisterChaincodeSupportServer(grpcServer, chaincode.NewChaincodeSupport(peer.GetPeerEndpoint))
+	// Register ChaincodeSupport server...
+	// TODO : not the "DefaultChain" ... we have to revisit when we do multichain
+	registerChaincodeSupport(chaincode.DefaultChain, grpcServer)
 
 	// Register Devops server
 	serverDevops := openchain.NewDevopsServer()
@@ -364,9 +365,6 @@ func serve(args []string) error {
 		return err
 	}
 	logger.Info("Starting peer with id=%s, network id=%s, address=%s, discovery.rootnode=%s, validator=%v", peerEndpoint.ID, viper.GetString("peer.networkId"), peerEndpoint.Address, rootNode, viper.GetBool("peer.consensus.validator.enabled"))
-	if len(args) > 0 && args[0] == "cli" {
-		go doCLI()
-	}
 	grpcServer.Serve(lis)
 	return nil
 }
@@ -396,6 +394,24 @@ func stop() {
 	status, err := serverClient.StopServer(context.Background(), &google_protobuf.Empty{})
 	logger.Info("Current status: %s", status)
 
+}
+
+func registerChaincodeSupport(chainname chaincode.ChainName, grpcServer *grpc.Server) {
+	//get user mode
+	userRunsCC := false
+	if viper.GetString("chaincode.chaincoderunmode") == "user_mode" {
+		userRunsCC = true
+	}
+
+	//get chaincode startup timeout
+	tOut, err := strconv.Atoi(viper.GetString("chaincode.startuptimeout"))
+	if err != nil { //what went wrong ?
+		fmt.Printf("could not retrive timeout var...setting to 5secs\n")
+		tOut = 5000
+	}
+	ccStartupTimeout := time.Duration(tOut) * time.Millisecond
+
+	pb.RegisterChaincodeSupportServer(grpcServer, chaincode.NewChaincodeSupport(chainname, peer.GetPeerEndpoint, userRunsCC, ccStartupTimeout))
 }
 
 func checkChaincodeCmdParams(cmd *cobra.Command) error {
@@ -558,50 +574,6 @@ func chaincodeInvokeOrQuery(cmd *cobra.Command, args []string, invoke bool) {
 		} else if resp != nil {
 			logger.Info("Trying to print as string: %s", string(resp.Msg))
 			logger.Info("Raw bytes %x", resp.Msg)
-		}
-	}
-}
-
-func doCLI() {
-	wio := bufio.NewWriter(os.Stdout)
-	rio := bufio.NewReader(os.Stdin)
-	help := "deploy <chaincodeid> <version> <funcname> <arg> <arg>...\n"
-	for {
-		wio.WriteString(">")
-		wio.Flush()
-		line, _ := rio.ReadString('\n')
-		if line == "q\n" {
-			break
-		}
-		toks := strings.Split(line, "\n") //get rid of /n
-		toks = strings.Split(toks[0], " ")
-		if toks == nil || len(toks) < 1 {
-			wio.WriteString("invalid input " + line + "\n")
-			continue
-		}
-		if toks[0] == "help" {
-			wio.WriteString(help)
-			continue
-		}
-		switch toks[0] {
-		case "deploy":
-			fallthrough
-		case "d":
-			if len(toks) < 5 {
-				wio.WriteString(help)
-				continue
-			}
-			chaincodeID := toks[1]
-			version := toks[2]
-			spec := &pb.ChaincodeSpec{Type: 1, ChaincodeID: &pb.ChaincodeID{Url: chaincodeID, Version: version}, CtorMsg: &pb.ChaincodeInput{Function: toks[3], Args: toks[4:]}}
-			_, err := openchain.DeployLocal(context.Background(), spec)
-			if err != nil {
-				wio.WriteString("Error transacting : " + fmt.Sprintf("%s", err) + "\n")
-			} else {
-				wio.WriteString("Success\n")
-			}
-		default:
-			wio.WriteString(help)
 		}
 	}
 }
--- a/openchain/api_test.go
+++ b/openchain/api_test.go
@@ -20,6 +20,7 @@ under the License.
 package openchain
 
 import (
+	"fmt"
 	"bytes"
 	"testing"
 
@@ -28,8 +29,24 @@ import (
 	"github.com/openblockchain/obc-peer/openchain/ledger"
 	"github.com/openblockchain/obc-peer/openchain/util"
 	"github.com/openblockchain/obc-peer/protos"
+	"github.com/spf13/viper"
 	"golang.org/x/net/context"
 )
+
+func TestMain(m *testing.M) {
+	setupTestConfig()
+}
+
+
+func setupTestConfig() {
+	viper.SetConfigName("openchain") // name of config file (without extension)
+	viper.AddConfigPath("./")   // path to look for the config file in
+	viper.AddConfigPath("./..")   // path to look for the config file in
+	err := viper.ReadInConfig()      // Find and read the config file
+	if err != nil {                  // Handle errors reading the config file
+		panic(fmt.Errorf("Fatal error config file: %s \n", err))
+	}
+}
 
 func TestServerOpenchain_API_GetBlockchainInfo(t *testing.T) {
 	// Construct a ledger with 0 blocks.
--- a/openchain/chaincode/chaincode_support.go
+++ b/openchain/chaincode/chaincode_support.go
@@ -23,7 +23,6 @@ import (
 	"bytes"
 	"fmt"
 	"io"
-	"strconv"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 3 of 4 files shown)


muralisrini:

docs/source/membership/membership.md
```diff
@@ -191,35 +311,77 @@ ORG2.MSP does exist in the network definition.*
...
+*The figure above shows the subfolders in a local MSP on the file system*
 
-As you can see, there are nine elements to an MSP. It's easiest to think of these elements
-in a directory structure, where the MSP name is the root folder name with each
-subfolder representing different elements of an MSP configuration.
+* **config.yaml:**  Used to configure the identity classification feature
+  in Fabric by enabling "Node OUs" and defining the accepted roles.
 
-Let's describe these folders in a little more detail and see why they are important.
+* **admincerts (Deprecated from Fabric v1.4.3 and higher):** This folder
```
nikhil550: deprecated feels a bit strong, given that you can still use them. I would say optional instead.

docs/source/membership/membership.md
```diff
@@ -50,100 +93,175 @@ multiple MSPs and name them accordingly, e.g., `ORG2-MSP-NATIONAL` and
...
-node or user. However, as channel MSPs are available to all nodes in the
+node. However, as channel MSPs are available to all nodes in the
 channel, they are logically defined once in the channel configuration. However,
 **a channel MSP is also instantiated on the file system of every node in the
 channel and kept synchronized via consensus**. So while there is a copy of each
 channel MSP on the local file system of every node, logically a channel MSP
 resides on and is maintained by the channel or the network.
 
-### MSP Levels
+## MSP Levels
```
nikhil550: I think this section contains the same information as the channel MSP and the local MSP sections above, but only less clearly. I think the section can be removed, including the diagram.

docs/source/membership/membership.md
```diff
@@ -50,100 +93,175 @@ multiple MSPs and name them accordingly, e.g., `ORG2-MSP-NATIONAL` and
...
-to reflect these separate lines of business. When a CA issues X.509 certificates,
-the `OU` field in the certificate specifies the line of business to which the
-identity belongs.
-
-We'll see later how OUs can be helpful to control the parts of an organization who
-are considered to be the members of a blockchain network. For example, only
-identities from the `ORG1-MANUFACTURING` OU might be able to access a channel,
-whereas `ORG1-DISTRIBUTION` cannot.
+### Organizational Units (OUs) and MSPs
+
```
nikhil550, line 86: When reading this section, I was left a bit confused on the difference between OU's and Node OU's. I was also left wondering what the relationship between the affiliation of an identity and the Node OU was.

docs/source/membership/membership.md
```diff
@@ -1,43 +1,86 @@
...
 
-This is where a **Membership Service** Provider (MSP) comes into play ---
-**it identifies which Root CAs and Intermediate CAs are trusted to define
-the members of a trust domain, e.g., an organization**, either by listing the
+Because Fabric is a permissioned network, blockchain participants need a way to
+prove their identity to the rest of the network in order to transact on the
+network. If you've read through the documentation on [identity](../identity/identity.html)
+you've seen how a PKI can provide verifiable identities through a chain
+of trust. How is that chain of trust used by the blockahin network?
+Recall in the credit card scenario that the Certificate Authority is
```
joealewine: blockchain

joealewine: Also --- "recall in the credit card scenario" reads a bit funny. Maybe something like "Recall from the credit card scenario in the Identity topic that..."

docs/source/membership/membership.md
```diff
@@ -1,43 +1,86 @@
...
+like a card provider — it dispenses many different types of verifiable
+identities. An MSP, on the other hand, is like the list of card providers
+accepted by the store, determining which identities are the trusted
+members (actors) of the store  payment network.
```
joealewine: Maybe "An MSP, on the other hand, determines which credit card providers are accepted at the store. In this way, the MSP turns an **identity** (the credit card) into a **role** (the ability to buy things at the store)."

docs/source/membership/membership.md
```diff
@@ -1,43 +1,86 @@
...
+*Identities are similar to your credit cards that are used to prove
+you can pay. The MSP is similar to the list of accepted credit cards.*
+
+Or consider a consortium of banks that each operate peer
+nodes that both endorse and validate transactions. They could also optionally
+operate ordering nodes. The Bank A account holders have identities affiliated
+with Bank A, and the Bank B account holders have identities affiliated
+with Bank B. But the account holders don't operate nodes, they simply
+interact with the system from their mobile or web application.
+
```
joealewine: I think we're missing a transitional paragraph here that takes from the credit card analogy into the blockchain network without diving straight into more detailed minutia like what's below. Something like:

"This ability to turn verifiable identities into roles is fundamental to the way Fabric networks function, since it allows organizations, nodes, and channels the ability establish MSPs that determine who is allowed to do what at the organization, node, and channel level."

joealewine: Then we can go into an example like your bank example of how this separation of MSPs work in practice. 

I get where you're going with this bank example, but it's a bit out of place where you have it since we're still establishing the fundamentals of MSPs here. I might say something a bit simpler here like:

"Consider a network consisting of three banks on a channel. The channel MSP would define which banks had administrative rights over that channel (for example, which banks have to sign in order to add new banks to the channel). Relatedly, Bank A would have an organizational MSP identifying who is allowed to make decisions on behalf of that organization. Note that while the other members of a channel will have the ability to decide whether Bank A has administrative rights over the channel, Bank A has the exclusive right to decide who its own admins are through its organization MSP. 

Similarly, the nodes belonging to the various banks will have their own local MSPs, determining who can, for example, upgrade the node to a new version. While in some scenarios it might be true that node admins will be the same entities who are also organization admins, this separation of roles allows the admins of Bank A, for example, to define a node admin to be a lower level employee of Bank A whose administrative duties are confined just to the maintenance of the peer."

I realize you go into some of this later, but this might serve as an introduction to these concepts in a way that continues the theme you've established above.

docs/source/membership/membership.md
```diff
@@ -1,43 +1,86 @@
...
+a permissioned blockchain network. To transact on a Fabric network you
+need to:
+
+- Have an identity issued by a CA that is trusted by the network.
+- Become a member of an _organization_ that is recognized and approved
+by the network members. The MSP is how the identity is linked to the
+membership of an organization. Membership is achieved by adding your
+public key (also known as certificate, signing cert, or signcert) to
+your organization’s MSP.
+
```
joealewine: I think the "you" here is confusing. Nodes have MSPs and identities and aren't a "you".

docs/source/membership/membership.md
```diff
@@ -1,43 +1,86 @@
...
+## What is an MSP?
+
+Despite its name, the Membership Service Provider does not actually provide
+anything. Rather, the implementation of the MSP requirement is a folder
+that contains a list of related identities and is used to store users's
+certificates or in the case of an organization or channel, related identities.
```
joealewine: This is confusing. You say that it's a folder that contains a list of related identities or it's a list of related identities. The distinction you're trying to draw is not coming through. 

What I think you're trying to say here is that organizational and node MSPs list identities whereas channel MSPs primarily list organizations (who have the ability to decide which identities are node admins). I sort of mention this above as well.

docs/source/membership/membership.md
```diff
@@ -1,43 +1,86 @@
...
-permissions over a particular component (who has the ability to install chaincode on
-a peer, for example).
+member of a channel. It is the MSP that turns an identity into a **role** by
+identifying specific roles an actor might play within the scope of the
+organization the MSP represents (e.g., admins, or as members of a
+sub-organization group), and sets the basis for defining **access
+privileges** in the context of a network and channel e.g., channel admins,
+readers, writers). Also, when a user is registered with a CA, a role of
+admin, peer, client, orderer, or member must be associated with the
+user.
```
joealewine: Most of this material should, ideally, be covered in the "Why" section (other than the discussion of sub-organizational groups and admins/readers/writers, which should either be here or later). 

What would be more powerful here imo is a visual look at the folders themselves, which you have later.

docs/source/membership/membership.md
```diff
@@ -50,62 +94,200 @@ multiple MSPs and name them accordingly, e.g., `ORG2-MSP-NATIONAL` and
...
+file and contain a list of organizational units whose members are considered
+to be part of the organization represented by this MSP. This is particularly
+useful when you want to restrict the members of an organization to the ones
+holding an identity (signed by one of MSP designated CAs) with a specific
+Node OU role in it. For example, an endorsement policy can specify that
+only members of ORG1 can endorse a transaction. But when Node OU roles are
+enabled, a more granular restriction can be implemented,
+requiring that only peers in ORG1 can endorse a transaction instead of
+any member of ORG1.
+
```
joealewine: As usual, Fabric takes a useful word ("roles") and turns it into a part of the code, making it difficult for us to convey concepts to users. In other words, we might have to use a different conceptual model than "turns identities into roles at different levels of administration". Maybe "authority" could work. "Turns identity into authority". I kind of like that. What do you think?

docs/source/membership/membership.md
```diff
@@ -50,62 +94,200 @@ multiple MSPs and name them accordingly, e.g., `ORG2-MSP-NATIONAL` and
...
-### Local and Channel MSPs
+## MSP domains
 
-MSPs appear in two types of places in a blockchain network:
+MSPs occur in two domains in a blockchain network:
 
-* In channel configuration (**channel MSP**)
 * Locally on an actor's node (**local MSP**)
+* In channel configuration (**channel MSP**)
+
```
joealewine: This whole section is higher level conceptual stuff than the node OU material above it, imo. Users need to understand these MSPs and the various places they sit in the code before they understand identity classification in configtx.yaml, as a for example. 

Generally speaking, the most helpful way I find to organize is to lay out all of the necessary concepts and then get into how it's achieved (implementation details). Of course there are exceptions to any rule.

joealewine: There is also an organization MSP, which was discussed earlier. Another reason why this material is a bit awkwardly placed, imo.

docs/source/membership/membership.md
```diff
@@ -50,62 +94,200 @@ multiple MSPs and name them accordingly, e.g., `ORG2-MSP-NATIONAL` and
...
-**Local MSPs are defined for clients (users) and for nodes (peers and orderers)**.
-Local MSPs define the permissions for that node (who the peer admins are, for example).
-The local MSPs of the users allow the user side to authenticate itself in its transactions
-as a member of a channel (e.g. in chaincode transactions), or as the owner of a
-specific role into the system (an org admin, for example, in configuration transactions).
+**Local MSPs are defined for clients and for nodes (peers and
+orderers)**.
+Local MSPs define the permissions for that node (who the peer admins
+are, for example). The local MSPs of clients (the account holders in
+the banking scenario above), allow the user to authenticate
```
joealewine: The original reference to account holders is pretty far up. Might want to link to where it was. However, if you move this section higher to the top it might be more naturally co-located.

docs/source/membership/membership.md
```diff
@@ -1,341 +1,197 @@
...
-
-### Local and Channel MSPs
-
-MSPs appear in two types of places in a blockchain network:
+# Membership Service Provider (MSP)
+
+## Why do I need an MSP?
+
+Because Fabric is a permissioned network, blockchain participants need a way to prove their identity to the rest of the network in order to transact on the network. If you've read through the documentation on [identity](../identity/identity.html)
+you've seen how a PKI can provide verifiable identities through a chain of trust. How is that chain of trust used by the blockchain network? Recall from the credit card scenario in the Identity topic that the Certificate Authority is like a card
```
nikhil550: I would replace `a PKI` with `Public Key Infrastructure`

docs/source/membership/membership.md
```diff
@@ -1,341 +1,197 @@
...
+
+This ability to turn verifiable identities into roles is fundamental to the way Fabric networks function, since it allows organizations, nodes, and channels the ability establish MSPs that determine who is allowed to do what at the organization, node, and channel level.
+
+![MSP1a](./membership.msp.diagram.png)
+
+*Identities are similar to your credit cards that are used to prove you can pay. The MSP is similar to the list of accepted credit cards.*
+
+Or consider a consortium of banks that operate a blockchain network. Each bank operates peer and ordering nodes, and the peers endorse transactions submitted to the network. However, each bank would also have departments and account holders. The account holders would belong to each organization, but would not run nodes on the network. They would only interact with the system from their mobile or web application. So how does the network recognize and differentiate these identities? MSP's are used to define the organizations that belong to the network, whose members are trusted by other network members. MSPs are also the mechanism that provide members with a set of roles and permissions within the network.
```
nikhil550: Would not start a sentence with or here ...

docs/source/membership/membership.md
```diff
@@ -1,341 +1,197 @@
...
+
```
nikhil550, line 19: MSP's are used to define the organizations that belong to the network, whose members are trusted by other network members.  > MSP's are used to define the organizations that are trusted by other network members.

docs/source/membership/membership.md
```diff
@@ -1,341 +1,197 @@
...
+Finally, consider if you want to join an _existing_ network, you need a way to turn your identity into something that is recognized
```
nikhil550: I would make this sentence longer.. Emphasize that you already might have the ability to create identities, but like the card example, those identities cant just be issued. the need to be recognized.

docs/source/membership/membership.md
```diff
@@ -1,341 +1,197 @@
...
+by the network. The MSP is the mechanism that enables you to participate on a permissioned blockchain network. To transact on a Fabric network a member needs to:
+
+- Have an identity issued by a CA that is trusted by the network.
```
nikhil550: These ideas are similar. I would separate the CAs belonging to organizations in one bullet, and then what the members of those organizations can do in another bullet.

docs/source/membership/membership.md
```diff
@@ -1,341 +1,197 @@
...
-network administration concerns** while **MSPs at a lower level handle
-identity for the administration of private resources**. MSPs are mandatory
-at every level of administration --- they must be defined for the network,
-channel, peer, orderer, and users.
+![MSP1d](./ChannelMSP.png)
+
+*Snippet from a channel config.json file that includes two organization
+MSPs.*
+
+**Channel MSPs identify who has authorities at a channel level**.
```
nikhil550, line 63: I expected information about channel MSPs and organization definitions in the updating a channel topic. Maybe something to add with future work.

docs/source/membership/membership.md
```diff
@@ -1,341 +1,197 @@
...
+
+**The system channel MSP includes the MSPs of all the organizations that participate in an ordering service.** An ordering service will likely include ordering nodes from multiple organizaitons and collectively these organizations run the ordering service, most importantly managing the consortium of organizations and the default policies that are inherited by the application channels.
+
+**Local MSPs are only defined on the file system of the node or user** to which they apply. Therefore, physically and logically there is only one local MSP per
+node. However, as channel MSPs are available to all nodes in the channel, they are logically defined once in the channel configuration. However, **a channel MSP is also instantiated on the file system of every node in the channel and kept synchronized via consensus**. So while there is a copy of each channel MSP on the local file system of every node, logically a channel MSP resides on and is maintained by the channel or the network.
```
nikhil550, line 71: This section is alot better. I finally understand what that picture is for. However, that makes me wonder if we could create a better one.

docs/source/membership/membership.md
```diff
@@ -1,341 +1,197 @@
...
+
+### Organizational Units (OUs) and MSPs
+
+An organization can also be divided into multiple **organizational units**, each of which has a certain set of responsibilities. Think of an OU as a department
+inside an organization. For example, the `ORG1` organization might have both `ORG1.MANUFACTURING` and `ORG1.DISTRIBUTION` OUs to reflect these separate lines of business. When a CA issues X.509 certificates, the `OU` field in the certificate specifies the line of business to which the identity belongs. These values can then be used in policy definitions in order to restrict access or in smart contracts for attribute-based access control.
+
```
nikhil550, line 88: I would maybe include a sentance or two on how this is different from creating multiple MSPs as mentioned above

docs/source/membership/membership.md
```diff
@@ -1,341 +1,197 @@
...
+Specifying OUs is optional. If OUs are not used, all of the identities that are part of an MSP --- as identified by the Root CA and Intermediate CA folders --- will be considered members of the organization.
+
+### Node OU Roles and MSPs
+
+Additionally, there is a special kind of OU, sometimes referred to as a `Node OU`, that can be used to confer a role onto an identity. These Node OU roles are defined in the `$FABRIC_CFG_PATH/msp/config.yaml` file and contain a list of organizational units whose members are considered to be part of the organization represented by this MSP. This is particularly useful when you want to restrict the members of an organization to the ones holding an identity (signed by one of MSP designated CAs) with a specific Node OU role in it. For example, an endorsement policy can specify that only members of ORG1 can endorse a transaction. But when Node OU roles are enabled, a more granular restriction can be implemented, requiring that only peers in ORG1 can endorse a transaction instead of any member of ORG1.
```
nikhil550: I would just say `config.yaml`

docs/source/membership/membership.md
```diff
@@ -1,341 +1,194 @@
...
-
-### Local and Channel MSPs
-
-MSPs appear in two types of places in a blockchain network:
+# Membership Service Provider (MSP)
+
+## Why do I need an MSP?
+
+Because Fabric is a permissioned network, blockchain participants need a way to prove their identity to the rest of the network in order to transact on the network. If you've read through the documentation on [identity](../identity/identity.html)
+you've seen how a Public Key Infrastructure (PKI) can provide verifiable identities through a chain of trust. How is that chain of trust used by the blockchain network? Recall from the credit card scenario in the Identity topic that the Certificate Authority is like a card
```
joealewine: "Identity" should be capitalized in the link.

PR #309, opened 2015-11-24
implement dev command with options to override user run mode and cons…

muralisrini:
…ensus algorithm

### DIFF
--- a/main.go
+++ b/main.go
@@ -70,6 +70,26 @@ var peerCmd = &cobra.Command{
 	},
 }
 
+var peerDevModeCmd = &cobra.Command{
+	Use:       "dev",
+	Short:     fmt.Sprintf("Run the peer in dev mode"),
+	Long:      fmt.Sprintf("Run the peer in dev mode.\nBy default this will setup noops consensus and have user run chaincode.\nUser can override the two with user-runs-chaincode and consensus-algorithm respectively"),
+	ValidArgs: []string{"1"},
+	Run: func(cmd *cobra.Command, args []string) {
+		viper.Set("peer.mode", "dev")
+		if userRunsChaincode {
+			fmt.Printf("Setting chaincode.chaincoderunmode to dev_mode\n")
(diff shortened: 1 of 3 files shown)


muralisrini:

binhn:
I ran and it worked fine, but a couple of comments:

1) The last line on the terminal out (see below) says "validator=false", which is incorrect. In this case, the peer is VP using noops.

vagrant@vagrant-ubuntu-trusty-64:/opt/gopath/src/github.com/openblockchain/obc-peer$ ./obc-peer peer dev
Setting chaincode.chaincoderunmode to dev_mode
Setting peer.consensus.plugin to noops
18:52:41.243 [main] serve -> DEBU 001 Running in DEV mode, using NOOPS handler factory
18:52:41.243 [peer] chatWithPeer -> DEBU 002 Starting up the first peer
18:52:41.244 [chaincode] NewChaincodeSupport -> INFO 003 Chaincode support using peerAddress: 0.0.0.0:30303
18:52:41.245 [db] createDBIfDBPathEmpty -> DEBU 004 Is db path [/var/openchain/db] empty [false]
18:52:41.253 [server] NewOpenchainServer -> INFO 005 
Creating new OpenchainServer...
18:52:41.255 [main] serve -> INFO 006 Starting peer with id=name:"jdoe" , network id=dev, address=0.0.0.0:30303, discovery.rootnode=, validator=false

2) When I started the chaincode, I got this line among others:
19:18:03.224 [chaincode] notifyDuringStartup -> DEBU 00c readyNotify is nil!!!!
It is screaming for attention with the "!". Do we need attention here or is this just status reporting and no harms?

muralisrini:
After discussion with @jeffgarratt and @binhn this needs to be implemented with a "--dev" (or maybe an explicit "--chaincodedev") command. Further in that mode we'll only allow "user runs chaincode" and "noops" consensus to simplify user experience (allowing "pbft" will make it complex to use). Retracting this PR to implement those changes.

PR #312, opened 2015-11-24
Added a separate config file (.yaml) for tests in ledger and db packages

manish-sethi:
This allows the tests to use a separate DB folder for each package. Otherwise, the tests would use the same DB folder as the peer process running on the same machine.


manish-sethi:

Issue #201, opened 2015-11-04
Support ability to build code (chaincode) from local folder or remote repository


prjayach:
Current implementation only permits chaincode to be placed in a local folder. No support for specifying a remote location where the chaincode is available. This is needed for multi-validator scenario where each validator builds the code themselves.

binhn:
@prjayach which milestone should this one be in?

prjayach:
Lets put it for Milestone 4 (milestone 3 ends in a week, right?). Next week we'll focus on query and chaincode-invoke-chaincode as those are higher priority. Will get to this the week after that.

muralisrini:
merged. Closing.

PR #315, opened 2015-11-25
Issue201/remote chaincode build

prjayach:
#201 Support for allowing remote URL for go chaincode.

Fix for ledger lock issue in chaincode tests.
Rebased with latest master. All tests pass except the peer tests which are still failing in master.

### DIFF
--- a/openchain/chaincode/chaincode_support.go
+++ b/openchain/chaincode/chaincode_support.go
@@ -139,7 +139,15 @@ func getChaincodeID(cID *pb.ChaincodeID) (string, error) {
 	if cID == nil {
 		return "", fmt.Errorf("Cannot construct chaincodeID, got nil object")
 	}
-	return cID.Url + ":" + cID.Version, nil
+	var urlLocation string
+	if strings.HasPrefix(cID.Url, "http://") {
+		urlLocation = cID.Url[7:]
+	} else if strings.HasPrefix(cID.Url, "https://") {
+		urlLocation = cID.Url[8:]
+	} else {
+		urlLocation = cID.Url
+	}
+	return urlLocation + ":" + cID.Version, nil
 }
 
 func (chaincodeSupport *ChaincodeSupport) registerHandler(chaincodehandler *Handler) error {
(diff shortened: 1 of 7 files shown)


prjayach:
Signed-off by: Praveen Jayachandran prjayach@in.ibm.com

muralisrini:
Looks good for most part Praveen. Time has come though to promote "getChaincodeID" ? I think @srderson also mentioned a need for such a function elsewhere. If we move it into an util or something, we can reuse that code. Noticed the code is duped in a few places already.

prjayach:
@muralisrini Yes, I considered this. I have consolidated multiple places where we were constructing VM name into one function GetVMName in container/controller.go. The chaincodeID is a substring of this name and I think its best to keep a separate function for this. I didn't want to import package vm within package chaincode, so left this method within chaincode. If there is a better suggestion of where we can place these methods, I am all for it.

srderson:
There's a util package with utils.go. That could be a possible location.

core/container/externalbuilder/copy.go
```diff
@@ -0,0 +1,106 @@
...
+)
+
+// MoveOrCopyDir attempts to copy src to dest by firstly trying to move, then copy upon failure.
+func MoveOrCopyDir(logger *flogging.FabricLogger, srcroot, destroot string) error {
+	mvErr := os.Rename(srcroot, destroot)
+	if mvErr == nil {
+		return nil
+	}
+
+	logger.Warnf("Failed to move %s to %s: %s, try copy instead", srcroot, destroot, mvErr)
```
jyellick: Warning seems a bit severe here, no?  I would expect that in most production environment the TMPDIR will not be on persistent storage while the chaincode persistence dir necessarily would be.  Maybe turn this down to Debug?

core/container/externalbuilder/copy.go
```diff
@@ -0,0 +1,106 @@
...
+
+	info, err := os.Stat(srcroot)
+	if err != nil {
+		return errors.WithMessagef(err, "failed to stat")
```
jyellick: Very much a nit, but no parameters and a `WithMessagef`.  Wouldn't have even commented, but at line 33 below you choose a `WithMessage`

PR #317, opened 2015-11-25
Development testnet configuration and document

### LINKED ISSUE #300: devops chats need to send peer address on HELLO
in non-Validator mode, the chat initiates a regular "HELLO" message in peer.SendTransactionsToPeer. This would need to send peer address just as other Chat initiations do (example, chat initiation in NewPeerHandler). 

@jeffgarratt :  Not sure if duplicating marshal-peer-address-and-send whereever we send HELLO is the right way to do this.... assigning it to you if you are already looking to clean up this area (related issue #256)

binhn:
close #300


binhn:

PR #320, opened 2015-11-25
update setting up network section to use the devnet document

binhn:


binhn:

PR #322, opened 2015-11-25
Change host port to match rest port in openchain.yaml

ecb:
The rest service port for the peer defaults to 5000 via the openchain.yaml. The rest_api.json looks for the service on port 3000.

### DIFF
--- a/openchain/rest/rest_api.json
+++ b/openchain/rest/rest_api.json
@@ -5,7 +5,7 @@
         "description": "Interact with the enterprise Blockchain through Openchain API",
         "version": "1.0.0"
     },
-    "host": "127.0.0.1:3000",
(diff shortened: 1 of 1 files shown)


angrbrd:
Hi @ecb , thanks for noticing this. But this is intended to be a different port. Here is why. If you are running inside of Vagrant, yes you will be listening for REST requests on port 5000. As you note, this is specified inside of openchain.yaml. Though you will not be running your Swagger inside of Vagrant, but rather on the local host. I meant the difference to illustrate the fact that you must set up port forwarding from the local host to Vagrant. I describe that inside the readme files in multiple places. So I would prefer not to change right now, as tons of places in docs will need to be changed. We may consider to change later. Hopefully this makes sense.

Thanks.

PR #325, opened 2015-11-27
implement --peer-chaincodedev option

muralisrini:
@jeffgarratt , @binhn 

Implemented --peer-chaincodedev mode option. With this option we no longer need a "peer.mode" property in openchain.yaml. The option sets the following properties to put the peer in "chaincode developer" mode
                 "peer.validator.enabled" - true
                 "peer.validator.consensus" -  "noops"
                 "chaincode.chaincoderunmode" -  "dev_mode"

Also peer.validator.enabled is used in peer.go to determine of a request needs to be looped back or sent to a validator. With this change, the following configurations appear to work correctly
               CLI -> VP -> 3 VPs
               CLI -> NVP -> 4 VPs

### DIFF
--- a/docs/SandboxSetup.md
+++ b/docs/SandboxSetup.md
@@ -10,20 +10,16 @@ Go to obc-peer directory:
 
     cd $GOPATH/src/github.com/openblockchain/obc-peer
 
-Edit openchain.yaml to set the peer mode “mode” to “dev”, as shown below:
+Also edit openchain.yaml to set the “mode” to “dev” under the "chaincode" property, as shown below:
 
-	 mode: dev
-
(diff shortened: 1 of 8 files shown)


muralisrini:

binhn:
@muralisrini I am not clear on how to start up. Is it ./obc-peer peer --dev ?
I would also suggest changing "chaincode.chaincoderunmode" - "dev_mode"  to "chaincode.mode=dev" (avoid repeating the same words over and over)

muralisrini:
@binhn it is ./obc-peer peer --peer-chaincodedev 

Other peer command flags begin with "--peer-". We could have "--peer-dev" or "--peer-chaincodedev". The later is more explicit that this flag is for chaincode developers ?

"chaincode.mode=dev"  does look better. Will change.

Issue #299, opened 2015-11-20
Implement block and state synchronization protcol


srderson:
Need to implement the block and state synchronization protocol.
- Every peer needs to broadcast each time a new block is added to their chain.

@kchristidis Can we link into consensus here? Does this broadcast already take place? Is there a built in mechanism to detect falling behind in consensus

If it is determined that a peer has fallen behind

Path A) If the peer is at block 0
- Broadcast a request for the full state
- Download the full state from a responding peer

Path B) Else if the peer is at block 0 + X where X is > 0
- Broadcast a request for the  state delta of the peer's latest block number + 1
- Download the state delta from a responding peer and repeat for each missing block. If no peer has a state delta available, the full state must be requested.

For both paths
- Broadcast a request for the latest block and download from f +1 peers
- Confirm all state hashes in the blocks match
- Hash downloaded state and confirm it matches that in the block
- Broadcast a request for all older blocks
- Download blocks from a responding peer
- Confirm hashes

@binh - let me know if you have any comments. I think we should meet Monday to discuss.

kchristidis:
> Can we link into consensus here? Does this broadcast already take place?

PBFT has this notion of checkpointing every `K` blocks. You're effectively broadcasting a new `checkpoint` message right after executing the K transactions that says, "here's my global hash right after I executed transaction number `X`". If you get `2f+1` of those messages with you have a so-called **stable certificate** and you know that this checkpoint can be trusted.

So by setting `K` to 1 we can make this work for every block.

Refer to Section 4.4 (pg. 409-410) of the paper for more info.

> Is there a built in mechanism to detect falling behind in consensus?

I find at least two such references in the paper.

One, when the network is switching to a new view (elect a new leader when it's suspected to be faulty). A reconfiguration of sorts take place and the primary (leader) or the backups (the rest of the validating peers) may find out that they're missing certain checkpoints that are needed ("New-View Message Processing", pg. 413-414). In that case, they need to initiate a state transfer (#291).

Two, a replica will generally initiate a state transfer (again, #291) when it learns about a stable checkpoint with sequence number greater than "the high water mark in its log" (i.e. the sequence numbers it is currently processing or set to process -- do note that this is a rough and almost certainly incorrect explanation). See Section 6.2.2 "State Transfer", pg. 427-429 for more. I think this is the closest that you'll get to a "yes" to your question.

[...]

PR #335, opened 2015-12-01
Ability to put a raw block on to the chain. Issue #299

srderson:

@manish-sethi Please review.

### DIFF
--- a/openchain/ledger/blockchain.go
+++ b/openchain/ledger/blockchain.go
@@ -251,8 +251,12 @@ func (blockchain *blockchain) persistBlock(block *protos.Block, blockNumber uint
 	writeBatch := gorocksdb.NewWriteBatch()
 	writeBatch.PutCF(db.GetDBHandle().BlockchainCF, encodeBlockNumberDBKey(blockNumber), blockBytes)
 
-	sizeBytes := encodeUint64(blockNumber + 1)
-	writeBatch.PutCF(db.GetDBHandle().BlockchainCF, blockCountKey, sizeBytes)
+	// Need to check as we suport out of order blocks in cases such as block/state synchronization. This is
+	// really blockchain height, not size.
+	if blockchain.getSize() < blockNumber+1 {
+		sizeBytes := encodeUint64(blockNumber + 1)
+		writeBatch.PutCF(db.GetDBHandle().BlockchainCF, blockCountKey, sizeBytes)
+	}
 
 	if persistState {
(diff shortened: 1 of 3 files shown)


openchain/ledger/blockchain.go
```diff
@@ -254,7 +254,9 @@ func (blockchain *blockchain) persistBlock(block *protos.Block, blockNumber uint
 	sizeBytes := encodeUint64(blockNumber + 1)
```
manish-sethi: Now with PutRawBlock method - can the blocks arrive out of order? If yes, then "sizeBytes := encodeUint64(blockNumber + 1)" needs a revisit.
Remaining changes look fine.

srderson: Thanks for catching that. It is possible for blocks to arrive out of order. I'll edit this so it returns the highest block number. For example, if you've received blocks 9, 3, and 1, the size will be 9, not 3.

manish-sethi: Even if we maintain the highest block number seen, there could be some blocks missing at a point in time. I think, we need to think through how this affect other functionality. For instance, either we need to put peer in 'catching up mode' and hence does not answer ledger queries (because, this may return inconsistent results). Or, we may need to create a snapshot before it goes into 'catching up mode' and answer ledger queries from that snapshot only (which may be stale but still consistent).
Also, I am not sure if this is already been discussed if some other peer may connect to a peer for catch-up which itself is in 'catching-up mode' and related issues.

srderson: I think a peer that is in 'catch up mode' will not be able to perform any other tasks. Trying to support other functionality while in this mode would be difficult.

manish-sethi: Yes, sure. Does this gets handled at grpc layer or we need to maintain some flag at ledger level?
I mean that do we need to offer another APIs at ledger level. CatchupModeOn/Off? And we do not allow any other methods on ledger in between these two calls (other than rawPut methods). And at CatchupModeOff, we perform some checks such as all the blocks have been received etc.?

srderson:
Height issue should be resolved now.

openchain/ledger/blockchain.go
srderson: This is something I think we'll need at both grpc and ledger level, but I think it needs to be at grpc level first.

manish-sethi:
The changes look good.

PR #337, opened 2015-12-01
Ability to set the raw state on the chain. Issue #299

srderson:

Ability to set the raw state. @manish-sethi please review. I'm not sure hash calculation will be handled correctly if this is used. Does it depend on deltas?

### DIFF
--- a/openchain/ledger/state.go
+++ b/openchain/ledger/state.go
@@ -187,17 +187,12 @@ func (state *state) addChangesForPersistence(blockNumber uint64, writeBatch *gor
 	stateLogger.Debug("state.addChangesForPersistence()...finished")
 }
 
-// addRawStateKeysAndValuesForPersistence commits raw keys and values to the
-// state. This should only be used during state synchronization.
-func (state *state) addRawStateKeysAndValuesForPersistence(keys, values [][]byte) error {
-	if keys == nil || values == nil || len(keys) != len(values) {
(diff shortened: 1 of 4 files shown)


srderson:
@manish-sethi I have updated the code with the changes we discussed. It now uses state deltas.

openchain/ledger/state.go
```diff
@@ -187,6 +187,21 @@ func (state *state) addChangesForPersistence(blockNumber uint64, writeBatch *gor
 	stateLogger.Debug("state.addChangesForPersistence()...finished")
 }
 
+func (state *state) applyStateDelta(delta *stateDelta) error {
+
```
manish-sethi, line 191: Need to add here following two lines.
state.recomputeHash = true
state.GetHash() // this is when the hash changes are made which would then be applied in the later call 'state.statehash.addChangesForPersistence(writeBatch)'

In test, better to add a check for hash
hash1 = 'after adding initial data'
hash2 = 'After removing initial data'
hash3 = 'After applying stateDelta'
hash1 and hash3 should be same while hash2 would be different.

In fact, for slightly stronger test - we can make two state deltas delat1 and delta2 (each containing 2 key-values in the test example) and apply those in any sequence. Now, there would be four hashes and only final hash should match hash1

srderson:
Merged in PR #346

PR #338, opened 2015-12-01
kill and remove containers in addition to stopping - issue #327

### LINKED ISSUE #327: Need to kill containers after running tests and when chaincode init fails
Currently, the tests leave a bunch of containers in stopped state, which cause successive running of the tests to fail. Need to cleanup and remove the containers after the tests are complete. The same needs to be done when chaincode init/deploy fails.
Solution: container.stop should kill container and not just stop.

muralisrini:
This fixes #327 by killing and removing containers after they are stopped. This will also help keep users's environment clean by reducing containers that are left around after tests.


muralisrini:

Issue #119, opened 2015-10-24
Genesis block for chaincodes


binhn:
related #120 
We need a genesis block to capture properties for chaincodes, especially system chaincodes. The chaincodes may only read the properties from the genesis block. The genesis block is created during start up and configured with a `genesis` section in the `openchain.yaml` file.

```
genesis:
    property1:
    property2:
```

markparz:
Murali, Jeff and Sheehan to get together on call.

PR #339, opened 2015-12-01
Function to create genesis block. Issue #119

srderson:
@jeffgarratt @muralisrini  This is the function to create the genesis block. One issue, where should we call genesis.MakeGenesis(). I'd like to call it in main. If you look at the genesis_test.go, it includes `go grpcServer.Serve(lis)` which I believe needs to be running to deploy the chaincode. However, in main we don't run `grpcServer.Serve(lis)` in a goroutine so it blocks. Could we run it in a gorountine?

### DIFF
--- a/main.go
+++ b/main.go
@@ -43,6 +43,7 @@ import (
 	"github.com/openblockchain/obc-peer/openchain"
 	"github.com/openblockchain/obc-peer/openchain/chaincode"
 	"github.com/openblockchain/obc-peer/openchain/consensus/helper"
+	"github.com/openblockchain/obc-peer/openchain/ledger/genesis"
 	"github.com/openblockchain/obc-peer/openchain/peer"
 	"github.com/openblockchain/obc-peer/openchain/rest"
 	pb "github.com/openblockchain/obc-peer/protos"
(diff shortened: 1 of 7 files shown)


muralisrini:
@srderson : the test case had to create grpc server as the test process was not a peer. The actual validator would already be wired to chaincode support. In theory, you should be able to call ledger.MakeGenesis directly (in main if that's what you want to do). As you are by-passing consensus and calling deploy and execute directly, it should work I think.. or we should make it work.

srderson:
I tried adding the following code to the `serve` function main.go

```
makeGeneisError := genesis.MakeGenesis()
if makeGeneisError != nil {
  return makeGeneisError
}
```

If I add it above `grpcServer.Serve(lis)` it will timeout trying to deploy the chaincode into the genesis block. If I add it below `grpcServer.Serve(lis)`, it will never run.

muralisrini:
a quick sanity test would be to run Serve(lis) in a go routine and then call MakeGenesis to see if that is all it takes or if there are bigger issues in trying to run Execute chaincode locally.

srderson:
Confirmed that running Server(lis) in a groutine resolves the issue. @jeffgarratt Would like your input on whether this is the right approach.

jeffgarratt:
I see no problem with the approach, but I just want to make sure we are clear and concise on timing related startup issues (if any)

Issue #414, opened 2015-12-23
Ledger.ApplyDelta should support commit and rollback functionality


srderson:
@jyellick pointed out that the Ledger.ApplyDelta function does not allow him to rollback a delta if the state hash turns out to be incorrect. It would be useful to have the following API available in the ledger.

Ledger.ApplyDeltaBegin(delta)
<test state hash>
Ledger.CommitDelta()
or
Ledger.RollbackDelta()

srderson:
This API is now available.

PR #418, opened 2015-12-28
Ability to rollback applied deltas. Issue #414

srderson:
@manish-sethi This PR is for issue #414. It allows for applied deltas (used during sync) to be rolled back by calling `ledger.ApplyStateDelta` followed by either `ledger.CommitStateDelta` or `ledger.RollbackStateDelta`. This is similar to `ledger.BeginTxBatch`, `ledger.RollbackBatch`, and `ledger.CommitTxBatch`. However, I found a problem.

See the `ledger_test.TestLedgerRollbackWithHash` test in this PR. It does the following
1. Begin a TX batch and manipulate the state
2. Get the hash using `ledger.GetTempStateHash()`
3. Rollback the TX batch
4. Get the hash again using `ledger.GetTempStateHash()`

The problem is that we're not recalculating the hash in step 4. If you remove step 2, it works. It seems we're too aggressive in caching the hash. At first I thought the solution was to modify `state.ClearInMemoryChanges` and add `state.updateStateImpl = true`, but the bucket tree implementation ignores an empty delta in `state_impl.PrepareWorkingSet` so the state hash is still not recalculated. Do you have thoughts on how this bug should be resolved? I was trying to do so without modifying the state implementations, but I'm not sure that is possible. I only tested the bukettree and have not yet investigated if the trie has this same issue.

### DIFF
--- a/openchain/ledger/ledger.go
+++ b/openchain/ledger/ledger.go
@@ -230,7 +230,7 @@ func (ledger *Ledger) CommitStateDelta(id interface{}) error {
 	if err != nil {
 		return err
 	}
-	defer ledger.resetForNextTxGroup()
+	defer ledger.resetForNextTxGroup(true)
 	return ledger.state.CommitStateDelta()
 }
 
@@ -241,7 +241,7 @@ func (ledger *Ledger) RollbackStateDelta(id interface{}) error {
 	if err != nil {
 		return err
 	}
-	ledger.resetForNextTxGroup()
+	ledger.resetForNextTxGroup(false)
 	return nil
 }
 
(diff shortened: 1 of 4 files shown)


manish-sethi:
@srderson - Thanks for unearthing this bug. I looked at this issue. Initially, I thought of removing the caching of hash as it's not going to help much caching the hash as such. However, I feel it would be important for keeping part (which is more often retrieved for computing hash) of the data structure in cache going forward.
I have made some changes in the code and pushed a new branch (https://github.com/manish-sethi/obc-peer/tree/issues/414/stateImplCacheBug)
The changes primarily include introducing a flag in method `ClearWorkingSet(changesPersisted bool)` in the interface `statemgmt.HashableState`. Based on this flag, the implementations can decide whether to update the cache contents (for now, only the hash value - as an illustration).
Please see if you are fine with the changes.

srderson:
Issue should be resolved in PR #426

srderson:

Code was reviewed with @manish-sethi

PR #419, opened 2015-12-30
Ability to configure state data structure. Resolves #385

### LINKED ISSUE #385: Ability to configure state storage implementation
There are currently two state storage implementations

1) Bucket tree #244 
2) Trie #243 

Currently, OBC is hard coded to use the bucket tree. We should add a configuration setting that allows us to choose an implementation before the genesis block is created.

We need to decide
1) Should this be a configuration setting managed by the user?
2) What recommendations do we provide?

These questions may be answered by issue #245

srderson:
This commit allows the state storage implementation to be managed from openchain.yaml.

### DIFF
--- a/openchain.yaml
+++ b/openchain.yaml
@@ -90,7 +90,7 @@ peer:
         state:
             snapshot:
                 # Channel size for readonly syncStateSnapshot messages channel for receiving state deltas for snapshot from oppositie Peer Endpoints.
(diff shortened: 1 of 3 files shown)


manish-sethi:
@srderson - Changes look good to me. However, I do not seem to have write permission to github repo so can you merge this?

PR #425, opened 2016-01-05
change dev instructions to use name instead of path & version

muralisrini:


muralisrini:

binhn:
it looks much simpler

PR #429, opened 2016-01-05
Some cleanup; TLS for Docker

ebuchman:
Here are a few little things, mostly just some cleanup and comments from digging into the code. Great project! Forgive me if I've overstepped bounds by removing things that don't appear to be used, but may be later.

### DIFF
--- a/openchain.yaml
+++ b/openchain.yaml
@@ -215,8 +215,19 @@ vm:
     # Endpoint of the vm management system.  For docker can be one of the following in general
     # unix:///var/run/docker.sock
     # http://localhost:4243
+    # tcp://localhost:2376
     endpoint: unix:///var/run/docker.sock
-
+        
+    # settings for docker vms
+    docker:
+        tls:
+            enabled: false
+            cert:
+                file: /path/to/server.pem
+            ca:
+                file: /path/to/ca.pem
+            key:
+                file: /path/to/server-key.pem
 
 ###############################################################################
 #
--- a/openchain/admin.go
+++ b/openchain/admin.go
@@ -44,7 +44,7 @@ func NewAdminServer() *ServerAdmin {
 type ServerAdmin struct {
 }
 
-func worker(id int, die chan bool) {
+func worker(id int, die chan struct{}) {
 	for {
 		select {
 		case <-die:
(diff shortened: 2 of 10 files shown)


srderson:
Thanks, all contributions are welcome!

+1 for the state changes. I also ran all tests and confirm they pass.

@muralisrini - can you review the chaincode changes?

@ebuchman - you'll have to sign off before the PR can be accepted. You just need to add a comment with "Signed-off-by: John Doe john.doe@domain.com". You can read details at https://github.com/openblockchain/obc-getting-started under the contribution section.

ebuchman:
Great, thanks!

Signed-off-by: Ethan Buchman ethan@coinculture.info

mrshah-at-work:
Should we keep the docker TLS configuration to match the format of other TLS configurations in the yaml file? For example, as follows:

```
vm:

    # Endpoint of the vm management system.  For docker can be one of the following in general
    # unix:///var/run/docker.sock
    # http://localhost:4243
    endpoint: unix:///var/run/docker.sock
    docker:
        tls:
            enabled: true
            cert:
                file: /chaincode_host/cert.pem
            ca:
                file: /chaincode_host/ca.pem
            key:
                file: /chaincode_host/key.pem
```

ebuchman:
Good point, will update. Note about the config, the "validator" section seems to be duplicated. Once under "peer" and once at the top level. The top level instance does not appear to be used anywhere. We should remove one

muralisrini:
I'm OK with chaincode changes (chaincode_support.go).

The change to remove "MessageHandler" parameter from handlerFactory in peerImpl.go (and corresponding changes in related files) would need @jeffgarratt attention..... I think this may be part of (future) handler chaining strategy. Perhaps revoke them for now and include them in a future PR so as to expedite the docker-SSL changes ?

@mrshah-at-ibm : I assume the docker-SSL changes match changes you were testing (except for yaml file configuration format you suggest above) ?

mrshah-at-work:
Correct @muralisrini, they are similar (not exactly same) to my changes at https://github.com/mrshah-at-ibm/obc-peer/pull/1

The differences are as under:
1. Checking for availability of the certificate files to throw an appropriate error.
2. Using a different function from fsouza's docker library. There is an option to not pass the CA. This will skip certificate verification. I thought someone might need this.
3. Implemented the changes in vm.go and controller.go as both the files create docker client.

muralisrini:
@ebuchman : could you remove the MessageHandler changes mentioned in my previous comment/request so we can accept the pull request by Monday, 01/11 as we need the docker/https fix (which @mrshah-at-ibm has been working on as well/waiting for)

ebuchman:
Hi, sorry for the delay. I removed the commit (https://github.com/eris-ltd/obc-peer/commit/cde337849be084bb22e064a58953fcc2d36da026 now gone from the pull request)

Note the commit in question removed the `done chan` from ConsensusHandler because it blocks (https://github.com/openblockchain/obc-peer/blob/master/openchain/consensus/helper/handler.go#L177). I will leave it out for now, but presumably someone intended to do something with that channel.

muralisrini:
had to merge to bring it up-to-date.

thanks @ebuchman

PR #439, opened 2016-01-06
Return ChaincodeID when deploying through REST API

masterDev1985:
The rest api should return the chaincode ID that is generated whenever a user deploys code, or they won't be able to invoke or query it.


masterDev1985:
@angrbrd Per your suggestion, I removed that variable declaration and fixed my text alignment.  I'll see if I can fix my environment to avoid those types of issues in the future.

angrbrd:
@masterDev1985 I've taken a look. I have a few more comments. I'll ping you today to discuss. Thanks.

angrbrd:
I pulled this change and gave it a test. It looks good, merging.

Thanks!!

PR #440, opened 2016-01-06
Added a fix for broken chaincode dependencies

masterDev1985:
During the process of collecting the chaincode specified in the "path" parameter of the POST request.  A temporary directory is created to store the chaincode in before it is built.  This chaincode depends on obc-peer code (specifically the shim code) in order to work, which triggers a pull of the obc-peer repo.  Because the repository is private, the pull request will query the command line for user credentials, which will of course never be entered, causing the deployment process to go off into the ether, as far as the user deploying code is concerned.

All of this can be avoided by copying the local copy of obc-peer into the temporary directory before the go get command.  This is the change that I have implemented.  This is obviously not an optimal solution, but it's a good workaround.

@muralisrini

### DIFF
--- a/openchain/container/hashfromcode.go
+++ b/openchain/container/hashfromcode.go
@@ -118,18 +118,28 @@ func getCodeFromHTTP(path string) (codegopath string, err error) {
 	}
 
 	//ignore errors.. _usercode_ might exist. TempDir will catch any other errors
-	os.Mkdir(newgopath, 0755)
+	os.Mkdir(newgopath, 0755)    
 
 	if codegopath, err = ioutil.TempDir(newgopath, ""); err != nil {
 		err = fmt.Errorf("could not create tmp dir under %s(%s)", newgopath, err)
 		return
 	}
+	
+	// If we don't copy the obc-peer code into the temp dir with the chaincode,
+	// go get will have to redownload obc-peer, which requires credentials.
+	cmd := exec.Command("cp", "-r", origgopath + "/src", codegopath + "/src")
+	cmd.Env = env
+	var out bytes.Buffer
+	cmd.Stdout = &out
+	err = cmd.Run()
+	if err != nil {
+		return
+	}
 
 	env[gopathenvIndex] = "GOPATH=" + codegopath
 
-	cmd := exec.Command("go", "get", path)
+	cmd = exec.Command("go", "get", path)
 	cmd.Env = env
-	var out bytes.Buffer
 	cmd.Stdout = &out
 	err = cmd.Run()
 	if err != nil {


masterDev1985:

muralisrini:
@masterDev1985 : while the workaround looks like a good idea, I hesitate to put it out - even as a workaround .... we are violating user intent by using a different obc code base. Worse, the original "go get" command is broken as implemented. Let us get together and figure out the correct way to fix this.

masterDev1985:
@muralisrini sounds good.  Thanks for the feedback.  I'm not sure I understand what you mean by the go get command being broken.  I'll close this pull request, but I do want to stress that the REST API does not currently work without this fix.

core/chaincode/platforms/golang/list.go
```diff
@@ -68,16 +70,15 @@ func gopathDependencyPackageInfo(goos, goarch, pkg string) ([]PackageInfo, error
...
+		return nil, errors.Wrapf(err, "listing deps for package %s failed", pkg)
 	}
-	decoder := json.NewDecoder(stdout)
-
-	err = cmd.Start()
-	if err != nil {
-		return nil, err
+	// on Windows make sure we use forward slashes
+	if os.PathSeparator == '\\' {
+		output = []byte(filepath.ToSlash(string(output)))
```
sykesm, line 79: Modifying the json before unmarshaling isn't a safe operation. It's being fed far more than a single path.

core/chaincode/platforms/golang/list.go
```diff
@@ -68,16 +70,15 @@ func gopathDependencyPackageInfo(goos, goarch, pkg string) ([]PackageInfo, error
 	cmd := exec.CommandContext(ctx, "go", "list", "-deps", "-f", packageListFormat, pkg)
 	cmd.Env = append(os.Environ(), "GOOS="+goos, "GOARCH="+goarch)
 
-	stdout, err := cmd.StdoutPipe()
+	output, err := cmd.Output()
```
sykesm, line 73: We're increasing memory load for little reason by reading all output up front.  A decoder was originally used to process the stream as it was written to reduce memory overhead (streaming) and to support multiple objects in the list output.

core/chaincode/platforms/golang/list.go
```diff
@@ -7,12 +7,14 @@ SPDX-License-Identifier: Apache-2.0
 package golang
 
 import (
+	"bytes"
```
sykesm, line 10: I don't understand why any changes are needed in list. We are just capturing the output from the tool to generate a model.

I do agree we need to make changes in how we deal with the model.

core/chaincode/platforms/golang/list.go
```diff
@@ -68,16 +70,15 @@ func gopathDependencyPackageInfo(goos, goarch, pkg string) ([]PackageInfo, error
...
-		return nil, wrapExitErr(err, "'go list -deps' failed")
+		return nil, errors.Wrapf(err, "listing deps for package %s failed", pkg)
 	}
-	decoder := json.NewDecoder(stdout)
-
-	err = cmd.Start()
-	if err != nil {
-		return nil, err
+	// on Windows make sure we use forward slashes
+	if os.PathSeparator == '\\' {
```
sykesm, line 78: If we do need to make changes here, no need to check path separators; just use ToSlash - it's idempotent.

core/chaincode/platforms/golang/list_test.go
```diff
@@ -9,6 +9,7 @@ package golang
 import (
```
sykesm, line 9: Assuming there are no changes required in list.go, there shouldn't' be any changes required here either (beyond fixing the spelling).

PR #454, opened 2016-01-08
Fix cc dependencies [Issue #455]

masterDev1985:
@muralisrini 

I've added a timeout to the process of pulling chaincode from a repository.  This fixes the issue where a peer will get stuck waiting for github repository credentials forever, messing up the peers environment variables (GOPATH gets modified during this process).  Instead of hanging, the timeout will cause getCodeFromHTTP to throw an error.  The client's chaincode still won't deploy if they have any dependencies in private repositories, but at least they will get an error instead of nothing.

I will remove the commented sections with another commit to this request once I have your approval.

Thanks!

### DIFF
--- a/openchain.yaml
+++ b/openchain.yaml
@@ -244,9 +244,9 @@ chaincode:
 
     #timeout in millisecs for starting up a container and waiting for Register to come through. 1sec should be plenty for chaincode unit tests
     startuptimeout: 1000
-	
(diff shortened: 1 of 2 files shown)


masterDev1985:

masterDev1985:
A discussion on why the channel is buffered, for @muralisrini 
http://stackoverflow.com/questions/11886531/terminating-a-process-started-with-os-exec-in-golang

muralisrini:
Thanks @masterDev1985

PR #486, opened 2016-01-15
Update of ECA, TCA, TLSCA

thkzrl:
- all CAs are now listening at one port that can be configured in the yaml file (#404)
- re-registering a client w/ the ECA returns an error (#415)
- CA certificates have been externalized for manual signing by some root CA; automated signing is scheduled for the next release (#443)
- ECA now returns certificate pairs, one for signing w/ ECDSA and one for encryption w/ RSA (#446)
- the path to where the CAs keep state is now configurable in the yaml file
- plenty of bug fixes and internal optimizations, some API consolidations

Issue #415, opened 2015-12-23
ECA Fails To Return Error on Repeated Registration


angrbrd:
Opening this issue after a discussion with Angelo. 

When registering a client with the ECA a second time, after deleting the original login cert/key, the ECA does not return an error. The ECA allows the repeated registration and places the cert/key once again into the local directory. However, when the client proceeds to execute any of deploy/invoke/query with these new keys, the peer crashes with the errors and stacks below.

Thanks.

17:06:54.961 [CRYPTO.client.user1] loadEnrollmentCertificate -> ERRO 05f Failed checking enrollment certificate against enrollment key [Signature not valid.].
17:06:54.961 [CRYPTO.client.user1] init -> ERRO 060 Failed initiliazing crypto engine [Signature not valid.].
17:06:54.962 [CRYPTO] InitClient -> ERRO 061 Failed client initialization [user1]: [Signature not valid.].
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xb code=0x1 addr=0x28 pc=0x8117a5]

goroutine 48 [running]:
github.com/openblockchain/obc-peer/openchain/crypto.closeClientInternal(0x0, 0x0, 0x0, 0x0)
        /opt/gopath/src/github.com/openblockchain/obc-peer/openchain/crypto/client.go:127 +0x45
github.com/openblockchain/obc-peer/openchain/crypto.CloseClient(0x0, 0x0, 0x0, 0x0)
        /opt/gopath/src/github.com/openblockchain/obc-peer/openchain/crypto/client.go:102 +0x74
github.com/openblockchain/obc-peer/openchain.(_Devops).Deploy(0xc8201b9d70, 0x7fcfb9032738, 0xc8201ee1b0, 0xc82037a0c0, 0x0, 0x7fcfb8fdf000, 0xc8201a2820)
        /opt/gopath/src/github.com/openblockchain/obc-peer/openchain/devops.go:122 +0x653
github.com/openblockchain/obc-peer/protos._Devops_Deploy_Handler(0xc4f0e0, 0xc8201b9d70, 0x7fcfb9032738, 0xc8201ee1b0, 0xc8201ee330, 0x0, 0x0, 0x0, 0x0)
        /opt/gopath/src/github.com/openblockchain/obc-peer/protos/devops.pb.go:198 +0x132
[...]

PR #487, opened 2016-01-15
Fix for issue #415. Alignment between obbca.yaml and openchain.yaml missing in PR #486.

adecaro:
The ECA enforces that a pair (enrollID,enrollPWD) can be used only once. A test has been added to check this behavior.

The PR contains also an alignment between obbca.yaml and openchain.yaml that was missing in the PR #486.
Some minor cleanups are also in.

### DIFF
--- a/obc-ca/obcca.yaml
+++ b/obc-ca/obcca.yaml
@@ -10,4 +10,7 @@ server:
 
 eca:
         users:
-#                 donq: 9gvZQRwhUq9q
+                donq: 7avZQLwcUe9q
+                nepumuk: 9gvZQRwhUq9q
+                jim: AwbeJH2kw9qK
+                lukas: NPKYL39uKbkj
--- a/openchain.yaml
+++ b/openchain.yaml
@@ -139,9 +139,9 @@ peer:
         eca:
             paddr: localhost:50051
         tca:
-            paddr: localhost:50551
+            paddr: localhost:50051
         tlsca:
-            paddr: localhost:50951
+            paddr: localhost:50051
 
     # Peer discovery settings.  Controls how this peer discovers other peers
     discovery:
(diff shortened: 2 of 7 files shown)


adecaro:

cmd/configtxgen/main.go
```diff
@@ -101,7 +101,11 @@ func doOutputAnchorPeersUpdate(conf *genesisconfig.Profile, channelID string, ou
 		return errors.Errorf("org '%s' does not have any anchor peers defined", asOrg)
 	}
 
+	// Workaround for existing configs that don't contain these keys
```
jyellick, line 104: It's not really a workaround for existing configs.  It's a workaround for the fact that this function "guesses" what the state of your channel looks like, and, by adding this element to configtx.yaml, it now guesses that the element exists in the application channel, but it doesn't for any newly created channel.

discovery/support/config/testdata/configtx.yaml
```diff
@@ -72,6 +72,15 @@ Organizations:
             - Host: peer0.org1.example.com
               Port: 7051
 
+        # PrivateDataImplicitCollection stores the configuration for this org's implicit
```
jyellick, line 75: It's done, so, I'm not willing to say there's harm in including it.  But, you'll note that this yaml isn't actually consumed/used anywhere in the integration tests, and I expect for almost every user of `configtxgen` we'll never actually want/need to specify this.  You could simplify this PR considerably by simply removing the `configtxgen` support for it.  I'll comment more generally on the approach in the overall comments.

gossip/service/eventer.go
```diff
@@ -112,6 +114,10 @@ func (ag *appGrp) AnchorPeers() []*peer.AnchorPeer {
 	return ag.anchorPeers
 }
 
+func (ag *appGrp) PrivateDataImplicitCollection() *peer.PrivateDataImplicitCollection {
```
jyellick, line 117: Why is this needed? It feels to me like if gossip needs a custom wrapper for the app org data, they should be declaring their own interface and using that.  Here, we're adding a function to a local implementation of an interface to satisfy a method of that interface that is never invoked by the gossip code, it seems.

PR #503, opened 2016-01-17
add nil security helper to test case

muralisrini:


muralisrini:
recent security work calls for security helper to chaincode support. Pass nil to the genesis test case.

muralisrini:

PR #506, opened 2016-01-18
Container fixes

### LINKED ISSUE #504: Chaincode container re-config on chaincode/peer restart
Chaincode container should always be re-configured with the peer's ip address and port when the chaincode/peer is restarted.

When the peer container (with same ID) is restarted or moved to some other VM, and the chaincode is queried/invoked, the peer tries to start the old container, which tries to connect to the old address of the peer and keeps failing.

muralisrini:
fixes #504 
- on start, container management will attempt to create container if not found (and a new Test for this)
- on start, validator will pass ip address to container (as opposed to baking it at container creation time)
- sundry log message and formatting 
- bug fixes
  - chaincode transaction context leak


muralisrini:

srderson:
Reviewed and confirmed tests are passing.

PR #509, opened 2016-01-18
set default command line peer address to empty string

muralisrini:
The fix is needed so chaincode with use ENV var if peer address is not passed in


muralisrini:

PR #511, opened 2016-01-18
treat path's with schemes as remote and let them pass through

### LINKED ISSUE #510: checkGolangSpec() doesn't fails even if it is a valid http(s) path
Chaincode Deployment with a valid git repository path (github.com / hub.jazz.net) fails as the checkGolangSpec() function returns error.

muralisrini:
fixes #510


muralisrini:

PR #521, opened 2016-01-20
Implement confidential query

binhn:


binhn:
closes #522

Issue #526, opened 2016-01-20
Error running test case crypto_test.go


srderson:
I hit the below error when running the entire test suite using the command

```
cd $GOPATH/src/github.com/openblockchain/obc-peer
go test -timeout=20m $(go list github.com/openblockchain/obc-peer/... | grep -v /vendor/)
```

It looks like it's creating a database under `.obc-peer/db`. It should not be using that path. There is a `/var/openchain/test/` path within that vagrant image for test case databases.

```
2016/01/20 15:29:19 15:29:19.649 [db] createDBIfDBPathEmpty [db.go:181] -> DEBU 6c4 Is db path [.obc-peer/db] empty [true]
2016/01/20 15:29:19 15:29:19.649 [db] CreateDB [db.go:58] -> DEBU 6c5 Creating DB at [.obc-peer/db]
Error opening DB Invalid argument: Column family not found: : blockchainCF
--- FAIL: TestValidatorDeployTransaction (0.65s)
    crypto_test.go:503: TestValidatorDeployTransaction with [%!s(int=0)]
panic: Could not open openchain db error = [Invalid argument: Column family not found: : blockchainCF] [recovered]
    panic: Could not open openchain db error = [Invalid argument: Column family not found: : blockchainCF]

goroutine 290 [running]:
testing.tRunner.func1(0xc820196ab0)
    /usr/local/go/src/testing/testing.go:450 +0x171
github.com/openblockchain/obc-peer/openchain/db.GetDBHandle(0x0)
    /opt/gopath/src/github.com/openblockchain/obc-peer/openchain/db/db.go:106 +0x2cf
github.com/openblockchain/obc-peer/openchain/ledger.fetchBlockchainSizeFromDB(0x200, 0x0, 0x0)
    /opt/gopath/src/github.com/openblockchain/obc-peer/openchain/ledger/blockchain.go:304 +0x28
github.com/openblockchain/obc-peer/openchain/ledger.newBlockchain(0x0, 0x0, 0x0)
    /opt/gopath/src/github.com/openblockchain/obc-peer/openchain/ledger/blockchain.go:52 +0x2e
github.com/openblockchain/obc-peer/openchain/ledger.newLedger(0xc8203ffb48, 0x0, 0x0)
    /opt/gopath/src/github.com/openblockchain/obc-peer/openchain/ledger/ledger.go:70 +0x28
github.com/openblockchain/obc-peer/openchain/ledger.GetLedger.func1()
[...]

PR #532, opened 2016-01-21
Error running test case crypto_test.go #526

diegomasini:
We fixed the errors and changed the way the logging was handled in the validity_period test to avoid problems.

@srderson explained to me that the tests require to have a peer running in a separated window. That conflicts with the peer we are already launching on the validity_period test, so for now we are skipping it. The test is working if it is running alone without launching the additional peer in a separated window (removing the skip sentence). This is a temporary solution to have all the tests running until we figure it out how to solve it. 

Adding @raullaprida  to the conversation since he was working on this.

### DIFF
--- a/obc-ca/obcca/validity_period_test.go
+++ b/obc-ca/obcca/validity_period_test.go
@@ -74,6 +74,7 @@ func setupTestConfig() {
 }
 
 func TestValidityPeriod(t *testing.T) {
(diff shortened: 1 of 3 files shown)


diegomasini:

srderson:
I ran all test cases and they pass. Thanks @diegomasini and @raullaprida

Issue #528, opened 2016-01-20
Move run-time artifacts out of obc-ca code folder


binhn:
The folder .obcca is default to where the code is, which causes git and editor to show them. We should use the same convention as openchain.yaml fileSystemPath: /var/openchain/production.  This includes unit test yaml files.
@srderson

thkzrl:
Can you please elaborate a bit on this as I'm not sure what the issue is.  We can rename “rootpath” in "obcca.yaml" to “SystemPath” for the sake of consistency and default it to “/var/openchain/production”, but aside from that the CAs need their own subdirectory.  Thanks.

PR #539, opened 2016-01-22
CA Update.

thkzrl:
- Root path changes to /var/openchain/production (#528)
- ECA double-checks encryption key during certificate creation
- TCAA.ReadCertificateSet implemented
- Added roles to users distinguishing peers, validators, and auditors.

Signed by: Thorsten Kramp thk@zurich.ibm.com


thkzrl:
Signed-off by: Thorsten Kramp thk@zurich.ibm.com

thkzrl:
Changed common name in certificates to user id. This change has been urgently requested for the forthcoming demo!

Signed-off by: Thorsten Kramp thk@zurich.ibm.com

binhn:
@thkzrl 
I see `Panic.Panicln(err)` and that prompts question whether we handle the error right because it will shutdown the CA and render the entire chain inoperable.  Is that the right thing to do here?

thkzrl:
The CA panics only during startup, that is, when it cannot create its own key pairs or certificates. There was one panic for an impossible error condition left which I've just removed.

binhn:
@thkzrl thanks

PR #544, opened 2016-01-22
serialize sends; all Sends must be via peer handler's SendMessage

muralisrini:
@jeffgarratt : can you review please ?

Also, I suspect we no longer need "syncBlocksMutex" in handler.go....but didn't want to make that change in the spirit of minimal change (as it didn't do any harm really). If you agree, we can remove it in another PR.


muralisrini:

PR #554, opened 2016-01-24
Fixing Broken Doc Links

angrbrd:

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,17 +64,16 @@ To run a specific test use the `-run RE` flag where RE is a regular expression t
     go test -test.v -run=TestGetFoo
 
 ## Writing Chaincode
-Since chaincode is written in Go language, you can set up the environment to accommodate the rapid edit-compile-run of your chaincode. Follow the instructions on the [Sandbox Setup](https://github.com/openblockchain/obc-peer/blob/master/docs/SandboxSetup.md) page, which allows you to run your chaincode off the blockchain.
+Since chaincode is written in Go language, you can set up the environment to accommodate the rapid edit-compile-run of your chaincode. Follow the instructions on the [Sandbox Setup](https://github.com/openblockchain/obc-docs/blob/master/api/SandboxSetup.md) page, which allows you to run your chaincode off the blockchain.
 
 ## Setting Up a Network
 
-To set up an Openchain network of several validating peers, follow the instructions on the [Devnet Setup](https://github.com/openblockchain/obc-peer/blob/master/docs/DevnetSetup.md)
-page. This network leverage Docker to manage multiple instances of validating peer on the same machine, allowing you to quickly test your chaincode.
+To set up an Openchain network of several validating peers, follow the instructions on the [Devnet Setup](https://github.com/openblockchain/obc-docs/blob/master/dev-setup/DevnetSetup.md) page. This network leverage Docker to manage multiple instances of validating peer on the same machine, allowing you to quickly test your chaincode.
 
 
 ## Working with CLI, REST, and Node.js
 
-When you are ready to start interacting with the Openchain peer node through the available APIs and packages, follow the instructions on the [API Documentation](https://github.com/openblockchain/obc-peer/blob/master/docs/Openchain%20API.md) page.
+When you are ready to start interacting with the Openchain peer node through the available APIs and packages, follow the instructions on the [API Documentation](https://github.com/openblockchain/obc-docs/blob/master/api/Openchain%20API.md) page.
 
 ## Configuration
 


docs/source/deployment_guide_overview.md
```diff
@@ -0,0 +1,151 @@
...
+
+This deployment guide will not go through every iteration and potential network configuration, but does give common guidelines and rules to consider.
+
+## Step two: Set up a cluster for your resources
+
+Generally speaking, Fabric is agnostic to the method used to deploy and manage it. It is possible, for example, to deploy and manage a peer from a laptop. For a number of reasons, this is likely to be unadvisable, but there is nothing in Fabric to prohibit it.
+
+As long as you have the ability to deploy Docker containers, whether locally (or behind a firewall), or in a cloud, it should be possible to stand up components and connect them to each other. However, Kubernetes features a number of helpful tools that have made it a popular container management platform for deploying and managing Fabric networks. For more information about Kubernetes, check out [the Kubernetes documentation](https://kubernetes.io/docs).
```
pamandrejko: Gari mentioned he prefers that we  limit the scope to the binaries. I think we can mention this but maybe say we will provide instructions around using the binaries that can be applied when using a Docker deployment or Kubernetes?

docs/source/deployment_guide_overview.md
```diff
@@ -0,0 +1,151 @@
...
+
+### Managing your infrastructure
+
+The exact methods and tools you use to manage your backend will depend on the backend you choose. However, here are some considerations worth noting if you choose some of the more popular deployment methods (Docker containers running in a Kubernetes cluster).
+
+* Using secret objects to securely store important configuration files in your cluster. For information about Kubernetes secrets, check out [Kubernetes secrets](https://kubernetes.io/docs/concepts/configuration/secret/).
```
pamandrejko: See previous comment

docs/source/deployment_guide_overview.md
```diff
@@ -0,0 +1,151 @@
...
+
+## Step three: Set up your CAs
+
+The first component that must be deployed in a Fabric is a CA. This is because the configuration of a component (such as a peer, an ordering node, or the MSP representing an organization) must include at least one user identity that is authorized to operate it before it can be deployed.
+
+**In a production network, it is recommended to deploy two CAs per organization to deploy all of the certificates associated with an organization.** For example, if you deploy three peers that are associated with one organization and an ordering node that is associated with another organization, you will need four CAs. Two of the CAs will be for the peer organization (generating the certificates for the peer, admins, communications, and the folder structure of the MSP representing the organization) and the other two will be for the orderer organization (generating the certificates for the ordering node, admins, and the folder structure of the ordering organization).
+
```
pamandrejko: I was confused reading this and trying to do the math without understanding the logic behind it- I think it would be clearer just to say you are going to create 2 kinds of CAs for each org - an enrollment CA and a TLS CA. I think you were trying to build up to that because the explanation comes later, but I think the explanation should come first and then you tell them you need 2 CAs for the org - it's totally clear then why.

docs/source/deployment_guide_overview.md
joealewine: This should be harmless --- if there are other types of secrets or ways to achieve I'm happy to mention them. I don't think this mention violates Gari's advice.

docs/source/deployment_guide_overview.md
```diff
@@ -0,0 +1,151 @@
...
+## Step three: Set up your CAs
+
+The first component that must be deployed in a Fabric is a CA. This is because the configuration of a component (such as a peer, an ordering node, or the MSP representing an organization) must include at least one user identity that is authorized to operate it before it can be deployed.
+
+**In a production network, it is recommended to deploy two CAs per organization to deploy all of the certificates associated with an organization.** For example, if you deploy three peers that are associated with one organization and an ordering node that is associated with another organization, you will need four CAs. Two of the CAs will be for the peer organization (generating the certificates for the peer, admins, communications, and the folder structure of the MSP representing the organization) and the other two will be for the orderer organization (generating the certificates for the ordering node, admins, and the folder structure of the ordering organization).
```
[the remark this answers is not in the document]
joealewine: Added a bit more about the TLS CA per org being optional (though recommended).

docs/source/deployment_guide_overview.md
```diff
@@ -0,0 +1,151 @@
...
+
+These two CAs have a related role in generating the certificates associated with the organization.
+
+* One CA is used to generate (through a process called "enrollment") the certificates of the admin of an organization, the MSP of that organization, and any nodes owned by that organization. This CA will also generate the certificates for any additional users. Because of its role in "enrolling" identities, this CA is sometimes called the "enrollment CA" or the "ecert CA".
```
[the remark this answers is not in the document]
joealewine: Added a line about this, but certainly willing to hear more if people want more about this.

docs/source/deployment_guide_overview.md
```diff
@@ -0,0 +1,151 @@
...
...he admin of the CA. Attributes and affiliations can also be given to the identity (for example, a `role` of `admin`). After the identity has been registered, it can be enrolled by using the username and password. The CA will generate two certificates for this identity --- a public certificate (also known as a signcert) known to the other members of the network, and a private certificate (stored in the `keystore` folder) used to sign actions taken by the identity. The CA will also generate an MSP file containing the public certificate of the CA issuing the certificate and the root of trust for the CA (this may or may not be the same CA). This MSP can be thought of as defining the organization associated with the identity of the admin. For an example of how this process looks, check out the [this example of how an admin is enrolled](https://hyperledger-fabric-ca.readthedocs.io/en/latest/operations_guide.html#enroll-org1-s-admin). In cases where the admin of the org will also be an admin of a node (which will be typical), **you must create the org admin identity before creating the local MSP of a node, since the certificate of the node admin must be used when creating the local MSP**.
```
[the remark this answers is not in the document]
joealewine: I'd rather have specific links. We can change them when we have better links to the specific docs.

docs/source/deployment_guide_overview.md
```diff
@@ -0,0 +1,151 @@
...
+### Create a peer
+
+Before you can create a peer, you will need to customize the configuration file for the peer. In Fabric, this file is known as `core.yaml`. You can find a sample `core.yaml` configuration file [in the sampleconfig directory of Hyperledger Fabric](https://github.com/hyperledger/fabric/blob/master/sampleconfig/core.yaml).
+
+As you can see in the file, there are quite a number of parameters you either have the option to set or will need to set for your node to work properly. In general, if you do not have the need to change a tuning value, leave it alone. You will, however, likely need to adjust the various addresses, specify the database type you want to use, as well as to specify where the MSP for the node is located.
+
+You have two main options for tuning your configuration.
+
+1. Edit the YAML file bundled with the binaries.
+2. Use environment variable overrides when deploying (if using Docker).
```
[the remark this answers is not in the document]
joealewine: As this is a launch point for full docs for each node, I think we should repeat them.

docs/source/deployment_guide_overview.md
```diff
@@ -0,0 +1,173 @@
...
+
+There are two main types of CAs, each with a related role in generated the certificates associated with an organization.
+
+* One CA (or more, if you are using intermediate CAs) is used to generate (through a process called "enrollment") the certificates of the admin of an organization, the MSP of that organization, and any nodes owned by that organization. This CA will also generate the certificates for any additional users. Because of its role in "enrolling" identities, this CA is sometimes called the "enrollment CA" or the "ecert CA".
```
denali49: 2-step process is 'Register' then 'Enroll'.  Boostrapping the CA server with the start command `fabric-ca-server start -b admin:adminpw` is equivalent to the 'Register' step.  Since this is usually scripted in a container, most users rarely do it manually and skip straight to the 'enroll' step.  This 2-step process is more apparent to the user when issuing commands in the fabric-ca-client because they always have to 'Register' then 'Enroll' and identity using the same user id and secret.

docs/source/deployment_guide_overview.md
```diff
@@ -0,0 +1,173 @@
...
...ons as a way of preventing "man in the middle" attacks. Note that the TLS CA is only used for issuing certificates and can be shut down when that activity is completed. Users have the option to use one way (server only) TLS as well as two way (server and client) TLS, with the latter also known as "mutual TLS". Note that mutual TLS is necessary for [service discovery](discovery-overview.html) and [private data collections](private-data/private-data.html) to work. **Because specifying that your network will be using TLS (which is recommended) should be decided before deploying the "enrollment" CA (the YAML file specifying the configuration of this CA has a field for enabling TLS), you should deploy your TLS CA first and use it to generate a certificate that will be used when bootstrapping your enrollment CA.** Note that organizations can decide to share a TLS CA for the generation of TLS certificates if they wish. However, this increases the exposure of the TLS CA in case. If all of the organizations in a network use a single TLS CA and this CA is compromised, it might not be possible to change the configuration of the network to enable new TLS certificates to be used on the network.
```
denali49: Or you can transfer TLS CA certs 'out of band' to other members of the network.

docs/source/deployment_guide_overview.md
```diff
@@ -0,0 +1,173 @@
...
+### Create a peer
+
+Before you can create a peer, you will need to customize the configuration file for the peer. In Fabric, this file is known as `core.yaml`. You can find a sample `core.yaml` configuration file [in the sampleconfig directory of Hyperledger Fabric](https://github.com/hyperledger/fabric/blob/master/sampleconfig/core.yaml).
+
+As you can see in the file, there are quite a number of parameters you either have the option to set or will need to set for your node to work properly. In general, if you do not have the need to change a tuning value, leave it alone. You will, however, likely need to adjust the various addresses, specify the database type you want to use, as well as to specify where the MSP for the node is located.
+
+You have two main options for tuning your configuration.
+
+1. Edit the YAML file bundled with the binaries.
+2. Use environment variable overrides when deploying (if using Docker).
```
denali49: You can also specify container environment variables in the Kubernetes yaml files. Probably worth noting here.

docs/source/deployment_guide_overview.md
```diff
@@ -0,0 +1,172 @@
...
+
+The former option has the advantage of persisting your changes whenever you bring down and bring back up the node. The downside is that you will have to port the options you customized to the new YAML when upgrading to a new binary version (you should use the latest YAML when upgrading to a new version).
+
+Either way, here are some values in `core.yaml` you must review.
+
+* `localMspID`: this is the name of the local MSP, generated by your CA, of your peer organization. This is where your peer organization admins will be listed.
+
+* `mspConfigPath`: the place where the local MSP for the peer is located. Note that it is a best practice to mount this volume external to your container.
+
+* `peer/address`: represents the endpoint to other peers in the same organization, an important consideration when establishing gossip communication within an organization.
```
jyellick: It seems like you might mean `peer.address`? Though you do not give fully qualified paths for the other entries.

docs/source/deployment_guide_overview.md
```diff
@@ -0,0 +1,172 @@
...
+
+* `TLS enabled`, `Enabled: true`: This is where you specify whether you want to [enable TLS](enable_tls.html). If you set this value to `true`, you will have to specify the locations of the relevant TLS certificates. Note that all of the nodes in a network (both the peers and the ordering nodes, must either all have TLS enabled or not enable. For production networks, it is highly recommended to enable TLS). As with your MSP, it is a best practice to mount this volume external to your container.
```
jyellick: None of our keys have spaces in them, so `TLS enabled` doesn't make sense.  This could be `peer.tls.enabled`.

docs/source/deployment_guide_overview.md
```diff
@@ -0,0 +1,172 @@
...
+As you can see in the file, there are quite a number of parameters you either have the option to set or will need to set for your node to work properly. In general, if you do not have the need to change a tuning value, leave it alone. You will, however, likely need to adjust the various addresses, specify the database type you want to use, as well as to specify where the MSP for the node is located.
+
+You have two main options for tuning your configuration.
+
+1. Edit the YAML file bundled with the binaries.
+2. Use environment variable overrides when deploying.
+
+The former option has the advantage of persisting your changes whenever you bring down and bring back up the node. The downside is that you will have to port the options you customized to the new YAML when upgrading to a new binary version (you should use the latest YAML when upgrading to a new version).
+
+Either way, here are some values in `core.yaml` you must review.
```
jyellick, line 117: In general, it's not clear to me what you're supposed to do here.  I would think maybe use fully qualified key paths.  It seemed at first that the user should search `core.yaml` for the indiciated key, but this breaks down, especially for keys which are not globally unique, plus some oddities noted below.

docs/source/deployment_guide_overview.md
```diff
@@ -0,0 +1,172 @@
...
+
+* `TLS`: this is where you specify whether you want to [enable TLS](enable_tls.html). If you set this value to `true`, you will have to specify the locations of the relevant TLS certificates. Note that all of the nodes in a network (both the peers and the ordering nodes, must either all have TLS enabled or not enable. For production networks, it is highly recommended to enable TLS). As with your MSP, it is a best practice to mount this volume external to your container.
+
+* `FileLedger`: although ordering nodes do not have a state database, they still all carry copies of the blockchain, as this allows them to verify permissions using the latest config block. Therefore the ledger fields should be customized with the correct endpoints.
+
+* `Cluster`: these values are important for ordering service nodes that communicate with other ordering nodes, such as in a Raft based ordering service.
+
+* `BootstrapFile`: this is the name of the genesis block you will generate for this ordering service.
```
jyellick: This was renamed from GenesisFile to BootstrapFile to help indicate that this can refer to a later config block which is not the genesis block.

docs/source/deployment_guide_overview.md
```diff
@@ -0,0 +1,172 @@
...
+
+* `BootstrapMethod`: the method by which the bootstrap block is given. For now, this can only be `file`, in which the file in the `BootstrapFile` is specified.
```
jyellick: For 2.0,  you can also specify `none` to simply start the orderer without bootstrapping.  If the system is already bootstrapped, this has no (actually, limited) effect, but if the system is not bootstrapped, it no longer crashes.

docs/source/deployment_guide_overview.md
```diff
@@ -0,0 +1,172 @@
...
+
+* `LocalMSPDir`: the place where the local MSP for the ordering node is located. Note that it is a best practice to mount this volume external to your container.
+
+* `ListenAddress` and `ListenPort`: represents the endpoint to other ordering nodes in the same organization, an important consideration when establishing gossip communication among ordering nodes.
+
+* `TLS`: this is where you specify whether you want to [enable TLS](enable_tls.html). If you set this value to `true`, you will have to specify the locations of the relevant TLS certificates. Note that all of the nodes in a network (both the peers and the ordering nodes, must either all have TLS enabled or not enable. For production networks, it is highly recommended to enable TLS). As with your MSP, it is a best practice to mount this volume external to your container.
+
+* `FileLedger`: although ordering nodes do not have a state database, they still all carry copies of the blockchain, as this allows them to verify permissions using the latest config block. Therefore the ledger fields should be customized with the correct endpoints.
```
jyellick: correct 'endpoints'? What is an endpoint for a ledger?

docs/source/deployment_guide_overview.md
```diff
@@ -0,0 +1,172 @@
...
+
+Either way, here are some values in `orderer.yaml` you must review. You will notice that some of these fields are the same as those in `core.yaml` only with different names.
+
+* `LocalMSPID`: this is the name of the local MSP, generated by your CA, of your orderer organization. This is where your orderer organization admins will be listed.
+
+* `LocalMSPDir`: the place where the local MSP for the ordering node is located. Note that it is a best practice to mount this volume external to your container.
+
+* `ListenAddress` and `ListenPort`: represents the endpoint to other ordering nodes in the same organization, an important consideration when establishing gossip communication among ordering nodes.
+
+* `TLS`: this is where you specify whether you want to [enable TLS](enable_tls.html). If you set this value to `true`, you will have to specify the locations of the relevant TLS certificates. Note that all of the nodes in a network (both the peers and the ordering nodes, must either all have TLS enabled or not enable. For production networks, it is highly recommended to enable TLS). As with your MSP, it is a best practice to mount this volume external to your container.
```
jyellick: TYpo "or not enable" should be "or not enabled.

Also, there are multiple TLS sections, and it's not immediately obvious what volume you're talking about.  You can specify paths to TLS CA files, but you often don't have to. (The orderer also extracts this information from the bootstrap block).

docs/source/deployment_guide_overview.md
```diff
@@ -0,0 +1,172 @@
...
+
+The former option has the advantage of persisting your changes whenever you bring down and bring back up the node. The downside is that you will have to port the options you customized to the new YAML when upgrading to a new binary version (you should use the latest YAML when upgrading to a new version).
+
+Either way, here are some values in `orderer.yaml` you must review. You will notice that some of these fields are the same as those in `core.yaml` only with different names.
+
+* `LocalMSPID`: this is the name of the local MSP, generated by your CA, of your orderer organization. This is where your orderer organization admins will be listed.
+
+* `LocalMSPDir`: the place where the local MSP for the ordering node is located. Note that it is a best practice to mount this volume external to your container.
+
+* `ListenAddress` and `ListenPort`: represents the endpoint to other ordering nodes in the same organization, an important consideration when establishing gossip communication among ordering nodes.
```
jyellick: Orderers do not gossip.  ListenAddress and ListenPort are simply the address/port to bind to.  Some externally routable address which ultimately points to this bind will be referenced in the channel config, but it's definitely not gossip.

docs/source/deployment_guide_overview.md
```diff
@@ -0,0 +1,172 @@
...
+You have two main options for tuning your configuration.
+
+1. Edit the YAML file bundled with the binaries.
+2. Use environment variable overrides when deploying.
+
+The former option has the advantage of persisting your changes whenever you bring down and bring back up the node. The downside is that you will have to port the options you customized to the new YAML when upgrading to a new binary version (you should use the latest YAML when upgrading to a new version).
+
+Either way, here are some values in `orderer.yaml` you must review. You will notice that some of these fields are the same as those in `core.yaml` only with different names.
+
+* `LocalMSPID`: this is the name of the local MSP, generated by your CA, of your orderer organization. This is where your orderer organization admins will be listed.
```
jyellick: This is just a string.  There's no way to list orderer organization admins

docs/source/deployment_guide_overview.md
```diff
@@ -0,0 +1,172 @@
...
+
+* `externalBuilders`: this field is important to set when using [Chaincode as an external service](cc_service.html).
+
+When you're comfortable with how your peer has been configured, how your volumes are mounted, and your backend configuration, you can run the command to launch the peer.
+
+### Create an ordering node
+
+Unlike the creation of a peer, you will need to create a genesis block (or reference a block that has already been created, if adding an ordering node to an existing ordering service) and specify the path to it before launching the ordering node.
+
+In Fabric, this configuration file for ordering nodes is known as `orderer.yaml`. You can find a sample `orderer.yaml` configuration file [in the sampleconfig directory of Hyperledger Fabric](https://github.com/hyperledger/fabric/blob/master/sampleconfig/orderer.yaml). Note that `orderer.yaml` is different than the "genesis block" of an ordering service. This block, which includes the initial configuration of the orderer system channel, must be created before an ordering node is created because it is used to bootstrap the node.
```
jyellick: Is it known as `orderer.yaml`? It is named orderer.yaml, and you don't really have a choice what it's called.

docs/source/deployment_guide_overview.md
joealewine: Sorry, I meant file path here.

Issue #520, opened 2016-01-20
State not propagated on invoke in multiple peer setup - noops


mrshah-at-work:
# Summary:

Invoke transaction is not propagated to all the peers when chaincode is invoked on one peer
# Observed In:

Had a similar bug before. Current test is with Git pull request - https://github.com/openblockchain/obc-peer/pull/519
# Priority:

**Highest**
# Description:
## Abstract:

The objective of this test is to deploy, invoke & query the example chaincode. This issue is seen on a multi-peer network only. Rest APIs are used to deploy, invoke and query.
### openchain.yaml:

Attached at the end of the report.
### Environment variables set:

`OPENCHAIN_PEER_ADDRESS`, `OPENCHAIN_VM_DOCKER_TLS_ENABLED`, `OPENCHAIN_VM_DOCKER_TLS_CERT_FILE`, `OPENCHAIN_VM_DOCKER_TLS_CA_FILE`, `OPENCHAIN_VM_DOCKER_TLS_KEY_FILE`, `OPENCHAIN_VM_ENDPOINT`, `OPENCHAIN_PEER_ID`, `OPENCHAIN_PEER_ADDRESSAUTODETECT=false`, `OPENCHAIN_PEER_LISTENADDRESS`
## Scenario:
1. Setup the network with 5 peers as mentioned above.
2. Call _deploy_ api with the following configuration:
   
   URL: `http://<peer_ip>:<peer_port>/devops/deploy`
   METHOD: `POST`
   BODY: 
   
   ```
   {
   "type": "GOLANG",
   "chaincodeID": {
     "path": "https://hub.jazz.net/git/averyd/cc_ex02/chaincode_example02"
   },
   "ctorMsg": {
     "function": "init",
     "args": [
       "a", "100", "b", "200"
     ]
   }
   } 
   ```
3. Call _query_ api with the following configuration on all peers:
   URL: `http://<peer_ip>:<peer_port>/devops/query`
   METHOD: `POST`
   BODY: 
   
   ```
   {
   "chaincodeSpec": {
   "type": "GOLANG",
   "chaincodeID": {
     "name": "09c08b4770c13cbbd877eb96728a4644068f77be1b85489f9b51700c046429af214d43c8d4428422f7a5c785d5afdc3837e9e86abaab3da1f5786d353418d307"
   },
   "ctorMsg": {
     "function": "query",
     "args": [
       "a"
     ]
   }
   }
   } 
   ```
[...]

Issue #500, opened 2016-01-16
fix wrinkles in chaincode FSM handling


muralisrini:
Chaincode state change - both on validator and chaincode side - can be triggered by callbacks from a previous state change. As these callbacks execute go routines to accomodate non-blocking "Query" requests, there is a race condition between the previous state completion and new state changed triggered by the callbacks. This causes occasional hiccups. 

The fix is to serialize the new FSM state change to coincide with the FSM changes triggered by network events.

muralisrini:
this is quite stable now... closing.

PR #555, opened 2016-01-24
serialize sends and make corresponding tweaks; also use short form of…

muralisrini:
… uuid to prefix log messages

This should complete the fixes for #500 based on findings from #520


muralisrini:

PR #557, opened 2016-01-26
Changed key usage for TLS certificates

diegomasini:
After checking with David we changed the Key usage attribute of the TLS certificates to KeyUsageDigitalSignature.

@jeffgarratt, @binhn this is the right key usage to use for TLS certificates


diegomasini:

PR #575, opened 2016-01-29
Tls issue574

adecaro:
This PR is the result of a combined effort of Almaden and Zurich to adds support for TLS connections between any node and the PKI. 
This can be configured by setting the following properties in openchain.yaml:

```
peer:
    # PKI member services properties
    pki:
        tls:
            enabled: false
            rootcert:
                file: tlsca.cert
            # The server name use to verify the hostname returned by TLS handshake
            server-host-override:
```


srderson:
Thanks @adecaro. I completed a quick review and tests pass. One minor comment, I noticed there are a couple of commented out code blocks in `node_ks.go`. Will these be uncommented soon or can they be removed?

PR #576, opened 2016-01-29
Dont lock in read write loop#520

muralisrini:
In progress, for code review.

### DIFF
--- a/openchain/consensus/consensus.go
+++ b/openchain/consensus/consensus.go
@@ -38,7 +38,7 @@ type Inquirer interface {
 
 // Communicator is used to send messages to other validators
 type Communicator interface {
-	Broadcast(msg *pb.OpenchainMessage) error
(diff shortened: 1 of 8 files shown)


binhn:
func (i *Noops) notifyBlockAdded() is to tell the connected NVPs the newly created block. The noops NVPs don't need states nor payload since they never execute txs. 

I don't see the changes to the code in this function to only send the notification to the NVPs. Are there reasons for that or we haven't done that yet?

muralisrini:
@binhn : We haven't done that yet but can do that easily once we verify the root cause fix without perturbing message flows too much.

On the other hand it might make sense to keep them around even if unused so we can catch these kind of issues.

muralisrini:
noops now broadcasts SYNC_BLOCK_ADDED to non-validators only.

muralisrini:
@kchristidis : yes. @binhn is going to create a issue for you though

Issue #580, opened 2016-02-01
peer login fails with TLS failed loading external cert error


tuand27613:
Assigning to @jeffgarratt  since it looks like a TLS thing.

Running obc-peer inside a Docker container.

When I do a CLI login, I get back a 
"error on client login: open: no such file or directory"

Also, the log has this :
21:59:41.460 [CRYPTO.client.satoshi] registerCryptoEngine -> INFO 040[0m Registering node crypto engine...
[36m21:59:41.460 [CRYPTO.client.satoshi] initTLS -> DEBU 041[0m Initiliazing TLS...
[36m21:59:41.461 [CRYPTO.client.satoshi] loadExternalCert -> DEBU 042[0m Loading external certificate at []...
[31m21:59:41.461 [CRYPTO.client.satoshi] loadExternalCert -> ERRO 043[0m Failed loading external certificate: [open : no such file or directory].
[31m21:59:41.461 [CRYPTO.client.satoshi] initTLS -> ERRO 044[0m Failed loading TLSCA certificates chain [open : no such file or directory].
[31m21:59:41.461 [CRYPTO.client.satoshi] registerCryptoEngine -> ERRO 045[0m Failed initliazing TLS [open : no such file or directory].
[31m21:59:41.461 [CRYPTO.client.satoshi] register -> ERRO 046[0m Failed registering crypto engine [open : no such file or directory].

adecaro:
Hi, I issued a PR to fix the bug :)

tuand27613:
thanks @adecaro ... will wait for #586 to merge.

workaround is to get a copy of tslca.cert and set OPENCHAIN_PEER_PKI_ROOTCERT_FILE

tuand27613:
@adecaro ,
with the workaround ^^^ , I'm able to login. But, now seeing this error: 
<
[36m16:22:18.778 [peer] HandleMessage -> DEBU 085[0m Handling OpenchainMessage of type: DISC_HELLO 
[36m16:22:18.778 [peer] beforeHello -> DEBU 086[0m Received DISC_HELLO, parsing out Peer identification
[...]

PR #586, opened 2016-02-02
If TLS to obc-ca is disabled then tls root cert must not loaded.

adecaro:
This PR is to address Issue 580. Namely, when TLS connections to obc-ca are disabled then tls root cert is not loaded.


binhn:
@tuand27613 please check this out in your test scenario

tuand27613:
original problem from #580 fixed as I can start peer and login without tlsca.cert.

However, seeing new crypto.GetEnrollmentCert() error message.  documented in #580

PR #601, opened 2016-02-04
docker-compose files to create a 4 validating peer consensus network

tuand27613:
signed by Tuan tdang@us.ibm.com


tuand27613:
Signed-off: Tuan Dang tdang@us.ibm.com

common/metrics/gendoc/options.go
```diff
@@ -167,7 +167,7 @@ func populateOption(lit *ast.CompositeLit, target interface{}) (interface{}, err
 				field.Set(labelNamesValue)
 
 			// simple strings
-			case "Namespace", "Subsystem", "Name", "Help", "StatsdFormat":
+			case "Namespace", "Subsystem", "Name", "Help", "StatsdFormat", "SourceComponent":
```
sykesm: Names are always difficult, but `SourceComponent` does not reflect the intent. It's also confusing to see `parentComponent` and `SourceComponent` used as synonyms within the same pull request.

The request was for us to identify whether a metric is emitted by the peer, the orderer, or both (common). Using a field on the metric is fine but it also seems like the namespace would imply the grouping. (I'll come back to this in a bit.)

common/metrics/gendoc/table.go
```diff
@@ -21,11 +21,12 @@ import (
...
-	Type                     // Type is the type of meter option.
-	Description              // Description is the help text from the meter option.
-	Labels                   // Labels is the meter's label information.
-	Bucket                   // Bucket is the statsd bucket format
+	Name            Field = iota // Name is the meter name.
+	Type                         // Type is the type of meter option.
+	Description                  // Description is the help text from the meter option.
+	Labels                       // Labels is the meter's label information.
+	Bucket                       // Bucket is the statsd bucket format
+	SourceComponent              // peer or orderer
```
sykesm: This isn't exclusive. Some metrics (eg. comm layer, logging counts) are common to both; others may also be present in the CA.

common/metrics/gendoc/table.go
```diff
@@ -51,6 +52,7 @@ func NewPrometheusTable(cells Cells) Table {
 			{Field: Name, Name: "Name", Width: max(cells.MaxLen(Name)+2, 20)},
 			{Field: Type, Name: "Type", Width: 11},
 			{Field: Description, Name: "Description", Width: 60},
+			{Field: SourceComponent, Name: "Parent Component", Width: 20},
```
sykesm: Here's where we use `SourceComponent` and "Parent Component" as synonyms. Between the inconsistency and the possible inaccuracy, I wonder if there's a better breakdown?

PR #602, opened 2016-02-04
Add iterator for range query

srderson:
This adds an iterator to the range query API on chaincode shim. The chaincode writer is able to call API on the iterator such as Next() and HasNext(). Calls to the peer to fetch additional state keys/values are transparent.

### DIFF
--- a/obc-ca/obcca/tlsca_test.go
+++ b/obc-ca/obcca/tlsca_test.go
@@ -20,27 +20,30 @@ under the License.
 package obcca
 
 import (
-	"golang.org/x/net/context"
 	"io/ioutil"
 	"net"
 	"os"
 	"testing"
 	"time"
 
+	"golang.org/x/net/context"
+
 	"github.com/spf13/viper"
 	"google.golang.org/grpc"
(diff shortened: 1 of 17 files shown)


srderson:

muralisrini:
@srderson :  handler.Lock/Unlock in handler.go is used to to protect data structure access. Unlike other uses of the mutex, the Lock/Unlock in lines 602,603 seems to span a potentially larger path length. This would hold other access (high frequency ones at that) such as creating tx contexts at bay,  Can we model its use along the lines of other handler.Lock uses so as to reduce lock contention ? For example can we use it around wrapper methods for accessing rangeQueryIteratorMap ?

srderson:
@muralisrini Thanks, I have added wrapper functions to reduce lock contention.

muralisrini:
@sheehan: the integration with chaincode code looks good to me. Range Query States basically can be called in both invoke and query and are modelled after GET_STATE calls.  I'll leave out the merging as I have to sign off soon today (won't have time to test).

Issue #453, opened 2016-01-08
CLI should print outputs, not indirectly via logger


corecode:
In order to automate operations via CLI (and testing), the CLI should print results in a consistent way.  The logger is not useful for this purpose.

Additionally, logger messages produce interference with output parsing.  For CLI operations, logger messages should probably be suppressed below the `Error` level, and also go to `stderr` instead of `stdout`.

bcbrock:
I agree. This should be a top priority as it will quickly enable simple automated testing, and should not be hard (for an experienced GO/JSON programmer) to implement.

bcbrock:
Hi @angrbrd, @binhn I am about to submit a PR that fixes this. I should have commented when I started working it.

angrbrd:
@bcbrock , OK thanks for letting me know. I can review the PR if you would like. Thanks.

bcbrock:
@angrbrd - Please see pull request #613. Thanks.

angrbrd:
@bcbrock , my understanding is that the usage of logger was supposed to be changed for Error level messages only, correct? The Info level messages remain as they are. If I am understanding this correctly, I just want to make sure that all logger.Error instances have been modified accordingly. What about lines 430 and 447? Should those also be changed?

angrbrd:
Closed by pull request #613 in commit 2d59bcb3047894be789e3267b3017f72f726b25c

PR #613, opened 2016-02-05
CLI Changes to support scripting

bcbrock:
This PR addresses #453.

New documentation (to be added in a separate pull request to obc-docs):

---

To facilitate its use in scripted applications, the `obc-peer` command always
produces a non-0 return code in the event of command failure. Upon success,
many of the subcommands produce a result on **stdout** as shown in the table
below:

| Command | **stdout** result in the event of success |
| --- | --- |
| `peer` | N/A |
| `status` | String form of ServerStatus_StatusCode |
| `stop` | String form of ServerStatus_StatusCode |
| `login` | N/A |
| `vm` | String form of ServerStatus_StatusCode |
| `chaincode deploy` | The chaincode container name (hash) required for subsequent `chaincode invoke` and `chaincode query` commands |
| `chaincode invoke` | The transaction ID (UUID) |
| `chaincode query` | By default, the query result is formatted as a printable string. Command line options support writing this value as raw bytes (-r, --raw), or formatted as the hexadecimal representation of the raw bytes (-x, --hex). If the query response is empty then nothing is output. |

---

Although many lines of main.go are changed, the changes are primarily
cosmetic:

The commands now use the Cobra 'RunE' interface because they return errors to
top-level command processing.

Since Cobra command processing prints the usage message and error string when
commands fail, several places where the code was doing this explicitly have
been removed.

In a few cases logging messages were converted to error objects and returned.

Things are written to **stdout** as documented.

Note that many of the commands still do ad-hoc error processing, e.g., they
call `panic` in the event of certain errors. I tried to change as little as
possible to implement the required function here so did not change these - I
did not feel comfortable to completely "rewrite" error handling.

### DIFF
--- a/main.go
+++ b/main.go
@@ -574,10 +574,11 @@ func checkChaincodeCmdParams(cmd *cobra.Command) (err error) {
 		}
 	}
 
-	// Check that non-empty chaincode parameters contain exactly 2 keys of the
-	// proper form. To better understand what's going on here with JSON
-	// parsing take a look at http://blog.golang.org/json-and-go - Generic
-	// JSON with interface{}
+	// Check that non-empty chaincode parameters contain only Function and
+	// Args keys. Type checking is done later when the JSON is actually
+	// unmarshaled into a pb.ChaincodeInput. To better understand what's going
+	// on here with JSON parsing see http://blog.golang.org/json-and-go -
+	// Generic JSON with interface{}
 	if chaincodeCtorJSON != "{}" {
 		var f interface{}
 		err = json.Unmarshal([]byte(chaincodeCtorJSON), &f)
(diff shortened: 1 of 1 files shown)


bcbrock:

bcbrock:
Following @angrbrd suggestion, I made the user interface more consistent by replacing all of the `logging.Error()` calls in `main()` with the creation and return of `error` objects.

In doing/testing this I noticed that the code in `checkChaincodeCmdParams()` that
tried to check the JSON arguments to chaincodes for consistency did not actually work; So
I fixed this as well. This leads to a question though: Is it correct to provide empty JSON arguments `{}` to chaincode delay/invoke/query? I did not change this aspect of the original code.

angrbrd:
@bcbrock, this looks good. I went ahead and merged this change. I agree with the above comment that providing an empty JSON argument to chaincode deploy/invoke/query should be disallowed. However, I think we should put a separate commit in for that. We are already fixing several issues with this merge. It's a two line change, if you'd like to put that check into `checkChaincodeCmdParams()` that's fine with me, otherwise assign to me and I will do it myself. Thanks!

bcbrock:
@angrbrd , Thanks for merging this. Please go ahead and fix the empty JSON problem, I'll work on other things.

PR #615, opened 2016-02-05
populate PkiID for every create of PeerEndpoint, even for self

tuand27613:
If you call method `func (p *PeerImpl) GetPeerEndpoint() (*pb.PeerEndpoint, error)`  ,  your PkiID field is not populated, especially when you call the method on yourself.

Problem was found in consensus/helper/Helper.go where a crypto Verify() fails because PeerEndpoint.PkiID is nil.


tuand27613:
Signed-off: Tuan Dang tdang@us.ibm.com

tuand27613:
@jeffgarratt  please review

Issue #604, opened 2016-02-04
Remove deployment and invocation of validity period chaincode


diegomasini:
Until the execution of system chaincode could be done in a secure manner (reported in issue #603) we will remove the deployment and invocation of the update validity period chaincode.

Working with @raullaprida on this.

PR #616, opened 2016-02-05
Disable validity period

diegomasini:
According to the issue reported in #604 the deployment, execution and verification of the validity period is disabled for now.

Copying @binh, @raullaprida, @adecaro in this PR.


adecaro:
@diegomasini, @binh, @raullaprida: I was reading in other threads that the code was supposed to be removed. Does disabling the functionality suffice?

binhn:
@adecaro I think it is sufficient to disable the code. Whoever wants to deploy it (only us) would know the implication.

Issue #608, opened 2016-02-05
large memory leak executing http://<peeraddress>:5000/chain


muralisrini:
running curl http://127.0.0.1:5000/chain in a loops causes obc-peer to lose memory rapidly and crash. The underlying method is "s.ledger.GetBlockchainInfo()" called in api.GetBlockchainInfo. Commenting that out and returning pb,BlockchainInfo{Height: 1} shows no mem leak. Assigning to @srderson to assign appropriately.

This was found while investigating #589

muralisrini:
Should have mentioned - @jeffgarratt found this leak and may have additional insight

srderson:
I'm continuing to investigate, but I've discovered that the reproducibility of this bug seems to depend on the last block in the chain. If the last block contains a single deploy transaction, the crash will happen after around 300 GET /chain invocations. If the last block contains a single invoke transaction, it's possible to run 100,000+ GET /chain invocations without issue.

srderson:
I should also mention that I'm using the default buckettree configuration to reproduce the issue.

PR #619, opened 2016-02-06
Free slice. Fixes /chain memory leak. Issue #608

srderson:
This fixes the /chain memory leak in issue #608. I was able to run 50,000+ GET requests to /chain without seeing a significant increase in memory with this fix.

@jeffgarratt @muralisrini - Please check that this resolves the issue in your environment.

The root cause of this issue is that RocksDB Go wrapper requires that Free() be called on returned slices to free up the allocated memory. This was not being done in the low level functions called by /chain requests.

@manish-sethi Please review the change. There are at least several more locations we need to investigate/fix. They are
buckettree/db_helper.go
buckettree/range_scan_iterator.go
buckettree/snapshot_iterator.go
trie/range_scan_iterator.go
trie/snapshot_iterator.go
Let me know if you find more locations.

### DIFF
--- a/openchain/ledger/statemgmt/buckettree/range_scan_iterator.go
+++ b/openchain/ledger/statemgmt/buckettree/range_scan_iterator.go
@@ -56,13 +56,11 @@ func (itr *RangeScanIterator) Next() bool {
 	}
 
 	for itr.dbItr.Valid() {
-		keySlice := itr.dbItr.Key()
-		defer keySlice.Free()
-		valueSlice := itr.dbItr.Value()
-		defer valueSlice.Free()
-		// TODO Why do we copy the key, but not the value?
-		keyBytes := statemgmt.Copy(keySlice.Data())
-		valueBytes := valueSlice.Data()
+
+		// making a copy of key-value bytes because, underlying key bytes are reused by itr.
+		// no need to free slices as iterator frees memory when closed.
+		keyBytes := statemgmt.Copy(itr.dbItr.Key().Data())
+		valueBytes := statemgmt.Copy(itr.dbItr.Value().Data())
 
 		dataNode := unmarshalDataNodeFromBytes(keyBytes, valueBytes)
 		dataKey := dataNode.dataKey
--- a/openchain/ledger/statemgmt/buckettree/range_scan_iterator.go
+++ b/openchain/ledger/statemgmt/buckettree/range_scan_iterator.go
@@ -56,8 +56,13 @@ func (itr *RangeScanIterator) Next() bool {
 	}
 
 	for itr.dbItr.Valid() {
-		keyBytes := statemgmt.Copy(itr.dbItr.Key().Data())
-		valueBytes := itr.dbItr.Value().Data()
+		keySlice := itr.dbItr.Key()
+		defer keySlice.Free()
+		valueSlice := itr.dbItr.Value()
+		defer valueSlice.Free()
+		// TODO Why do we copy the key, but not the value?
+		keyBytes := statemgmt.Copy(keySlice.Data())
+		valueBytes := valueSlice.Data()
 
 		dataNode := unmarshalDataNodeFromBytes(keyBytes, valueBytes)
 		dataKey := dataNode.dataKey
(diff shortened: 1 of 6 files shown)


srderson:

srderson:
@manish-sethi I added fixes for
buckettree/db_helper.go
buckettree/range_scan_iterator.go
buckettree/snapshot_iterator.go
trie/range_scan_iterator.go
trie/snapshot_iterator.go
also have some TODO comments with questions in the code.

openchain/ledger/statemgmt/buckettree/range_scan_iterator.go
```diff
@@ -56,8 +56,13 @@ func (itr *RangeScanIterator) Next() bool {
 	}
 
 	for itr.dbItr.Valid() {
-		keyBytes := statemgmt.Copy(itr.dbItr.Key().Data())
-		valueBytes := itr.dbItr.Value().Data()
+		keySlice := itr.dbItr.Key()
+		defer keySlice.Free()
+		valueSlice := itr.dbItr.Value()
+		defer valueSlice.Free()
+		// TODO Why do we copy the key, but not the value?
```
manish-sethi: I made copy of the key because, I had observed that rocks db reuses the underlying byte-array for key. So, if someone stores the returned key - it gets corrupted by mixing the content of the next key. Hence, I was making a copy of the key bytes. I did not find value bytes behaving in this manner. But now since we need to free the slice, the copy of value would be required.

openchain/ledger/statemgmt/buckettree/snapshot_iterator.go
```diff
@@ -45,8 +45,13 @@ func (snapshotItr *StateSnapshotIterator) Next() bool {
 
 // GetRawKeyValue - see interface 'statemgmt.StateSnapshotIterator' for details
 func (snapshotItr *StateSnapshotIterator) GetRawKeyValue() ([]byte, []byte) {
-	keyBytes := statemgmt.Copy(snapshotItr.dbItr.Key().Data())
-	valueBytes := snapshotItr.dbItr.Value().Data()
+	keySlice := snapshotItr.dbItr.Key()
+	defer keySlice.Free()
+	valueSlice := snapshotItr.dbItr.Value()
+	defer valueSlice.Free()
+	// TODO Why do we copy the key, but not the value?
```
manish-sethi: Since we need to free the slice, the copy of value would be required.

openchain/ledger/statemgmt/trie/range_scan_iterator.go
```diff
@@ -49,8 +49,13 @@ func (itr *RangeScanIterator) Next() bool {
 		return false
 	}
 	for ; itr.dbItr.Valid(); itr.dbItr.Next() {
-		trieKeyBytes := itr.dbItr.Key().Data()
-		trieNodeBytes := itr.dbItr.Value().Data()
+		trieKeySlice := itr.dbItr.Key()
+		defer trieKeySlice.Free()
+		trieNodeSlice := itr.dbItr.Value()
+		defer trieNodeSlice.Free()
+		// TODO Do we need to copy the key and value?
```
manish-sethi: Since we need to free the slice, the copy of value would be required.

openchain/ledger/statemgmt/trie/snapshot_iterator.go
```diff
@@ -45,8 +45,13 @@ func newStateSnapshotIterator(snapshot *gorocksdb.Snapshot) (*StateSnapshotItera
 func (snapshotItr *StateSnapshotIterator) Next() bool {
 	var available bool
 	for ; snapshotItr.dbItr.Valid(); snapshotItr.dbItr.Next() {
-		trieKeyBytes := snapshotItr.dbItr.Key().Data()
-		trieNodeBytes := snapshotItr.dbItr.Value().Data()
+		trieKeySlice := snapshotItr.dbItr.Key()
+		defer trieKeySlice.Free()
+		trieNodeSlice := snapshotItr.dbItr.Value()
+		defer trieNodeSlice.Free()
+		// TODO Do we need to copy the key and value?
```
manish-sethi: Since we need to free the slice, the copy of value would be required.

manish-sethi:
@srderson please see the comments above in the code where you had marked TODO. In an independent search, I could not find any additional places where a `Slice` needs to be free. 

In addition - An alternate way could be such that - we return a `struct` say `keyvalue` which has a `release / free` api. This would require caller (mostly obc code though) remembering the release. However, it could save us an allocation and a copy operation. Noting down just to keep in mind - in case we notice a high amount of allocation cost going forward.

PR #627, opened 2016-02-09
Check incoming PBFT messages for correctness of self-reported replica_id value, closes #568

### LINKED ISSUE #568: Check incoming PBFT messages for correctness of self-reported replica_id value
Messages exchanged during PBFT include a `replica_id` field. See `messages.proto` in `obcpbft` package. This field is set by the sending validator. This information is not however checked by the receiving replicas.

Now that we are signing and verifying hello messages (see #537), we are in a position to check whether a message does indeed originate from the referenced `replica_id`, and reject it if that's not the case.

I am working on this.

kchristidis:
Also:
- Includes @jyellick's fix for `TestNewViewTimeout` failing (sporadically)
- Addresses an issue with `recvViewChange()`, spotted by @corecode 

Passes all tests (along with the `behave` ones).


kchristidis:
@srderson: Can you also have a look at this one?

Merged it with the latest `master` and it passes all the tests, incl. the `behave` ones. (Oddly enough.)

srderson:
Confirmed unit tests and behave tests are passing for me also.

PR #633, opened 2016-02-09
Fix error in handling of error cases in CommitTxBatch. Fixes #632

### LINKED ISSUE #632: Bug in Ledger CommitTxBatch error case handling
I believe there is a programming error in the `CommitTxBatch` function of `ledger.go`.

```
success := true
defer ledger.resetForNextTxGroup(success)
defer ledger.blockchain.blockPersistenceStatus(success)
stateHash, err := ledger.state.GetHash()
if err != nil {
  success = false
  return err
}
```

The intent of this code is to have the defer functions run with `false` as the input parameter, but that will not happen according to the following example https://play.golang.org/p/7pQHdeDp-l

srderson:
Resolves the bug described in issue #632.

### DIFF
--- a/openchain/ledger/ledger.go
+++ b/openchain/ledger/ledger.go
@@ -116,13 +116,10 @@ func (ledger *Ledger) CommitTxBatch(id interface{}, transactions []*protos.Trans
 		return err
 	}
 
-	success := true
-	defer ledger.resetForNextTxGroup(success)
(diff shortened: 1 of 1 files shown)


manish-sethi:
@srderson changes look fine to me. Thanks for catching this.

srderson:
Merging per review from @manish-sethi

PR #634, opened 2016-02-10
Update BDD test timeouts

srderson:
@jeffgarratt The BDD tests regularly fail on my system due to timeouts. I had to change docker-compose-4-consensus-classic.yml from 30 -> 60 and docker-compose-4-consensus-batch.yml from 30 -> 90.

Not sure if this is normal. Perhaps we should investigate the slowness rather than increasing timeouts, but I'm submitting this PR for awareness.

### DIFF
--- a/openchain/peer/bddtests/peer_basic.feature
+++ b/openchain/peer/bddtests/peer_basic.feature
@@ -2,10 +2,10 @@
 # Test openchain Peers
 #
 # Tags that can be used and will affect test internals:
-#   
+#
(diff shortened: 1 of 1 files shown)


tuand27613:
i ran ok with the default wait times under vagrant on my thinkpad w520/16M.

tuand27613:
also, any way we can cross-post this discussion to obc-test ?

srderson:
I'm closing this based on the discussion in issue #646. It appears there is a bug in consensus that needs to be resolved.

PR #639, opened 2016-02-10
extraneous view-changes in obcpbft-classic as pbft messages were broadcasted before request

tuand27613:
Primary processing request before broadcasting request.

This caused pre-prepare/prepare/commit/execute to happen before the backups receive the request. The backup then starts its request timer which expires since there's nothing left to do. This then causes a view-change

### DIFF
--- a/openchain/consensus/obcpbft/config.yaml
+++ b/openchain/consensus/obcpbft/config.yaml
@@ -13,7 +13,11 @@ general:
     mode: classic
 
     # Maximum number of validators/replicas we expect in the network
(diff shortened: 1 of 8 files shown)


tuand27613:
master seems to be missing some fixes I thought were already there.  Re-doing

kchristidis:
@tuand27613: Just a heads up that you can keep pushing to your branch and this PR will be updated. No need to close and re-open. (Just post a message here saying that the PR is not ready yet.) Let me know if you need help.

PR #643, opened 2016-02-10
broadcast request before doing pbft messages when doing classic consensus

tuand27613:
updated obcClassic.RecvMsg() to broadcast the request before doing pbft processing.

Previously, the backups were getting the requests after we had finished with pre-prepare/prepare/commit/execute. This causes the request timer to start and expire (since that request was finished ) causing many view-changes to occur.


tuand27613:

PR #656, opened 2016-02-12
Removed RSA to have a consistent use of elliptic curve crypto.

thkzrl:
Encryption enrollment certificates now use EC, replacing RSA and thereby having a consistent use of EC crypto in OBC.

Signed-off by: Thorsten Kramp thk@zurich.ibm.com

### DIFF
--- a/obc-ca/obcca.yaml
+++ b/obc-ca/obcca.yaml
@@ -10,7 +10,13 @@ server:
 #              certfile: "/var/openchain/production/.obcca/tlsca.cert"
 #              keyfile: "/var/openchain/production/.obcca/tlsca.priv"
 
-
+logging:
+        trace: 0
+        info: 1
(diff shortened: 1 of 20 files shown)


thkzrl:
Signed-off-by: Thorsten Kramp thk@zurich.ibm.com

thkzrl:
Allowed logging levels to be turned on/off via yaml settings.

adecaro:
@thkzrl "golang.org/x/crypto/hkdf" is missing from the vendor folder and obc-peer doesn't compile inside vagrant. May you add the dependency using govendor following the instructions specified here https://github.com/openblockchain/obc-peer#adding-or-updating-a-go-packages ?

Thanks

thkzrl:
Added  golang.org/x/crypto/hkdf to the vendors libs.

PR #672, opened 2016-02-14
SendBlock() loop index underflow

tuand27613:
[openchain/peer/handler.go function `sendBlock()`](https://github.com/openblockchain/obc-peer/blob/master/openchain/peer/handler.go#L432) has a loop to process blocks in reverse order high to low.

This loop uses a uint var as the index so decrementing below 0 makes the var jump to uint.MaxValue which causes the loop to continue forever which causes an `out of memory` error.

Changed the loop condition to check that the index stays between upper and lower bounds.


tuand27613:
Waiting for master to be re-opened before signing off

tuand27613:
Closing this PR since I need to re-fork my repo after obc-peer became public. Will re-submit in a bit.

PR #678, opened 2016-02-15
`sendBlocks()` loop index underflow causes out of memory crash

tuand27613:
`sendBlock()` uses a uint64 index to reverse walk through a set of blocks, e.g. 10 to 0. 
When the index decrements from 0, we wrap to uint.MaxValue and the loop continues until we run out of memory.

Update the loop condition to check that the index is within the correct start and end values.


tuand27613:

Issue #646, opened 2016-02-10
Test performed on the pull request https://github.com/openblockchain/obc-peer/pull/636 failed


rameshthoomu:
Issue: One of the test scenario failed in behave tests.

Summary: 
Latest code pulled from the https://github.com/openblockchain/obc-peer/pull/636 pull request and performed the behave test.

Failing scenarios:
peer_basic.feature:133  chaincode example02 with 4 peers and 1 obcca, issue #567 -- @1.3 Consensus Options
0 features passed, 1 failed, 0 skipped
7 scenarios passed, 1 failed, 0 skipped
96 steps passed, 1 failed, 7 skipped, 0 undefined
Took 3m15.279s

[Behave_#518_Failed_.txt](https://github.com/openblockchain/obc-peer/files/125625/Behave_.518_Failed_.txt)
[go test.txt](https://github.com/openblockchain/obc-peer/files/125626/go.test.txt)

srderson:
This is a timeout. I hit these all the time on my machine. I increased the timeouts in PR #634

srderson:
Looks like PR #634 was merged. @rameshthoomu could you retest please?

rameshthoomu:
I pulled the #634 build today and ran the behave tests. I Could see "Scenario Outline: chaincode example02 with 4 peers and 1 obcca, issue #567 -- @1.3 Consensus Options" scenario is with 90 secs timeout.

[BDD_Timeout_#634_Failed.txt](https://github.com/openblockchain/obc-peer/files/126755/BDD_Timeout_.634_Failed.txt)

tuand27613:
@rameshthoomu , can you
- re-run only that `chaincode example02 with 4 peers and 1 obcca, issue #567 -- @1.3 Consensus Options` test and gather the vp0,1,2,3 logs ?
- try on a different machine ?  I cannot reproduce on either my OSX or windows machines

rameshthoomu:
I ran the specific failed scenario @1.3 Consensus Options and no issues observed. As requested by @tuand27613, I will perform the tests for each scenario and will share the results.

srderson:
[...]

PR #681, opened 2016-02-16
Fix off by 1 error which could cause panic in state transfer, closes #676

### LINKED ISSUE #676: state transfer fails with `panic: Our blockchain is already higher than a sync target, this is unlikely, but unimplemented` message
Scenario is
- 4 validating peers
- security enabled
- VP0-2 start at beginning of test,  VP3 manually started after VP0-2 have processed chaincode deploy and 5 TXs
- after VP3 starts, invoke 5 more TXs. This will cause VP3 to initiate state transfer

Result:
- VP3 initiates state transfer, tries to sync with VP1,VP2 but stops on panic message.

Attaching VP3 debug log:
[vp3_err.txt](https://github.com/openblockchain/obc-peer/files/131339/vp3_err.txt)

jyellick:
Fixes an off by one error.  The state transfer code checked if the target block number was less than the block height, to determine if this update would require deleting a block, and subsequently panic-ed if so.  This check is incorrect, as the block height is not the highest block number, it is the highest block number + 1.

Added test which generated that panic scenario, which now passes after the fix.

Increases the default channel size for blocks and state deltas from 1 and 5 to 20 and 40 respectively.  This is necessary especially for the behave tests, as the extremely low latency network and constrained CPU of a docker-ized environment can cause blocks and deltas to be discarded.

The combination of these two changes closes #676 

Passes all go tests.

Passes all behave tests, with the exception of

```
Failing scenarios:
  peer_basic.feature:133  chaincode example02 with 4 peers and 1 obcca, issue #567 -- @1.3 Consensus Options
  peer_basic.feature:134  chaincode example02 with 4 peers and 1 obcca, issue #567 -- @1.4 Consensus Options

0 features passed, 1 failed, 0 skipped
6 scenarios passed, 2 failed, 0 skipped
88 steps passed, 2 failed, 14 skipped, 0 undefined
Took 2m7.120s
```

Which are known to be failing due to issue #646 and should be addressed with PR #677 

### DIFF
--- a/openchain.yaml
+++ b/openchain.yaml
@@ -126,7 +126,7 @@ peer:
             # blocks from oppositie Peer Endpoints.
             # NOTE: currently messages are not stored and forwarded, but rather
             # lost if the channel write blocks.
-            channelSize: 1
+            channelSize: 10
         state:
             snapshot:
                 # Channel size for readonly syncStateSnapshot messages channel
@@ -140,7 +140,7 @@ peer:
                 # Peer Endpoints.
                 # NOTE: currently messages are not stored and forwarded,
                 # but rather lost if the channel write blocks.
-                channelSize: 5
+                channelSize: 20
 
     # Validator defines whether this peer is a validating peer or not, and if
     # it is enabled, what consensus plugin to load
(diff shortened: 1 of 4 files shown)


jyellick:
@binhn Could you take a look?

PR #691, opened 2016-02-17
Fix README typo

alandotcom:


binhn:
@lumberj i know the change is obvious, but could you please sign off to follow the process.  Thanks

alandotcom:
Signed-off-by: Alan Cohen alan@ripple.com

srderson:
Thanks!

docs/source/msp.rst
```diff
@@ -309,9 +309,11 @@ Two ways to handle this:
   Configuration of that MSP would consist of a list of root CAs,
   intermediate CAs and admin certificates; and membership identities would
   include the organizational unit (``OU``) a member belongs to. Policies can then
-  be defined to capture members of a specific ``OU``, and these policies may
-  constitute the read/write policies of a channel or endorsement policies of
-  a chaincode. A limitation of this approach is that gossip peers would
+  be defined to capture members of a specific ``OU`` (should be one of: peer, admin,
```
jyellick: This isn't entirely accurate.  There are four different types of identity principals: https://github.com/hyperledger/fabric-protos/blob/3bc53193070ecef4edd4ccc361c728d818c98128/msp/msp_principal.proto#L41-L49

The two that are being conflated here, are OU based ones, and Role based ones.

When 'NodeOUs' are enabled for an MSP, it ties three OUs to four roles.  These roles are Peer, Orderer, Client, and Admin, and typically, we use an OU of the same name.

You may additionally specify identities in policies directly by OU (for instance, maybe you want to create an OU for each region, or business unit of your clients).  Since these identities are more free-form, creating sane tooling to deal with this is challenging, and doesn't exist yet.

So, you may absolutely specify identity roles defined by NodeOUs, but these are roles, not OUs (even if the OU is of the same name).  There is no way to specify identity OUs via configtx.yaml, though the underlying support exists in the MSP.

Issue #694, opened 2016-02-17
Make PBFT log size multiplier configurable


jyellick:
Today, the PBFT log size is fixed at `L=2*K` as prescribed in the paper.

However, the choice of `2` is arbitrary, as the log size `L` is used as parameter and no special properties about it (other than its size being at least `2*K` is relied upon.

For certain deployments, especially high volume high latency ones, the default log multiplier of 2 may be insufficient.  As seen in issue #687, we are observing behavior where replicas in a normally functioning pbft network are occasionally receiving sequence numbers outside of the watermarks, increasing the log size `L` could help improve the operation of such networks.

Therefore, we should introduce a configuration variable, maybe `logMultiplier`, which will default to 2.

@kchristidis Can you tag as appropriate and possibly assign to me?

jyellick:
@kchristidis @tuand27613 if @tuand27613 wants this one, it's fine by me

tuand27613:
merged pull request #694

PR #698, opened 2016-02-17
add logmultiplier config variable. use to calculate log size. For issue #694

tuand27613:
Add new entry `logmultiplier` in consensus/obcpbft/config.yaml

Use to set log size as described in #694 

All unit tests run.

All behave tests run.

### DIFF
--- a/openchain/consensus/obcpbft/config.yaml
+++ b/openchain/consensus/obcpbft/config.yaml
@@ -25,12 +25,9 @@ general:
     # overall throughput in normal case operation.
     K: 10
 
-    # The log size is calculated as K * logmultiplier
-    # The log size is the maximum sequence number value ( also referred to as the high watermark )
-    # for which a replica will log information. In high volume/high latency environments, a replica might
-    # receive sequence numbers outside of its watermarks and have to wait for a stable checkpoint. Increasing
-    # the log size could improve operations in these cases.
-    logmultiplier: 10
+   # Affects the log size which is K * logmultiplier
+   # For high volume/high latency environments, a higher log size may increase throughput
+    logmultiplier: 2
 
     # How many requests should the primary send per pre-prepare when in "batch" mode
     batchsize: 2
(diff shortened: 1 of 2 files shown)


tuand27613:
@jyellick  @kchristidis  please take a look.

tuand27613:

jyellick:
@tuand27613 I think a default multiplier of 10 might be excessive, the paper suggests 2 as a default, which I think we might want to stick with, but if we wanted to increase it, 3 might be more reasonable? @kchristidis ?

Also, in the description, it specifies that the log size is the 'high water mark', but more accurately, the high water mark is the sequence number of the last stable checkpoint plus the log size.

For high volume / high latency environments, a higher log size may increase throughput.

kchristidis:
Right, was about to comment on the "high water mark" issue. @jyellick is also right that `10` is excessive. (Jason, what would the side-effect of a high number be though?)

This bit in the text is a bit problematic:

> In high volume/high latency environments, a replica might receive sequence numbers outside of its watermarks and have to wait for a stable checkpoint. Increasing the log size could improve operations in these cases.

"Improve operations" doesn't clarify anything.

I would just go with the following (or something along these lines):

```
# Affects the log size which is equal to: K * logmultiplier
# For high volume / high latency environments, a higher log size may increase throughput.
```

Looks good otherwise.

kchristidis:
This bit in the text is a bit problematic:

> In high volume/high latency environments, a replica might receive sequence numbers outside of its watermarks and have to wait for a stable checkpoint. Increasing the log size could improve operations in these cases.

"Improve operations" doesn't clarify anything.

I would just go with the following (or something along these lines):

```
# Affects the log size which is equal to: K * logmultiplier
# For high volume / high latency environments, a higher log size may increase throughput.
```

tuand27613:
yes, what happens if you don't clean up after testing ...

I've reset the default to 2 and copied the description text from @kchristidis

Issue #692, opened 2016-02-17
Don't have crypto sign()/verify() methods print out all message content in debug log.


tuand27613:
for example ,  [here](https://github.com/openblockchain/obc-peer/blob/master/openchain/crypto/node_sign.go#L43)  or [here](https://github.com/openblockchain/obc-peer/blob/master/openchain/crypto/node_sign.go#L28)

If you run with logging level=debug,  this increases log size by 10x or more, (e.g. it will print out a 17MB chaincode) and throw off readability of that log.

I'd say only print on error.

adecaro:
@tuand27613 what do you think if I rather introduce the property logging.crypto (openchain.yaml) to setup the logging level of the crypto module?

tuand27613:
@adecaro , logging.crypto will work for me. Thanks !

adecaro:
This has been addressed by PR #705.

Issue #684, opened 2016-02-16
performance issue with security and privacy enabled


dongmingh:
code level:
commit e338d6a25865db96c4e6374168473965a2b39054
Merge: 67e2a10 15dcc91
Author: Binh Q Nguyen binhn@us.ibm.com
Date:   Mon Feb 15 12:10:54 2016 -0500

```
Merge pull request #675 from bcbrock/docs

A couple of comments for outside-of-Vagrant builds
```

security and privacy: enabled
test: invoke a to b, then invoke b to a, then repeat continuously
chaincode can only process 5 invoke transactions per second.

Below is the information I collected during the test.  Detailed log is in the attached file, sec_test0216.txt
[sec_test0216.txt](https://github.com/openblockchain/obc-peer/files/133048/sec_test0216.txt)
1. There are many selectNextCert messages through out the log:
   15:53:53.548 [CRYPTO.client.lukas] selectNextTCert -> ERRO 03f No TCert in cache found.
2. an error message when deploy after successfully deploy and creating VM

15:53:54.543 [rest] Deploy -> INFO 044 Successfuly deployed chainCode: bb540edfc1ee2ac0f5e2ec6000677f4cd1c6728046d5e32dede7fea11a42f86a6943b76a8f9154f4792032551ed320871ff7b7076047e4184292e01e3421889c.
15:53:55.627 [container] newDockerClient -> INFO 045 Creating VM with endpoint: unix:///var/run/docker.sock
15:54:03.169 [container] newDockerClient -> INFO 046 Creating VM with endpoint: unix:///var/run/docker.sock
15:54:03.358 [ledger] sendProducerBlockEvent -> ERRO 047 Error unmarshalling deployment transaction for block event: proto: can't skip unknown wire type 7 for protos.ChaincodeDeploymentSpec
1. it spends 12 seconds on selectNextCert from 15:59:48 to 16:00:00

15:59:48.222 [rest] Invoke -> INFO 1bff REST invoking chaincode...
15:59:48.223 [rest] Invoke -> INFO 1c00 Local user 'lukas' is already logged in. Retrieving login token.
15:59:48.224 [CRYPTO] InitClient -> INFO 1c01 Initializing client [lukas]...
15:59:48.224 [CRYPTO] InitClient -> INFO 1c02 Client already initiliazied [lukas]. Increasing counter from [149]
15:59:48.224 [CRYPTO] InitClient -> INFO 1c01 Initializing client [lukas]...
[...]

PR #705, opened 2016-02-19
New TCertPool for better performance at client side

adecaro:
This PR addresses the "No TCert in cache found" issue #684 by providing a new infrastructure that loads the TCert from the TCA in background and deliver the TCerts to the application as soon as possible in a streaming fashion.  

@binhn @muralisrini For better performance, a client should be initialized once and kept active as much as possible to avoid to close and reinitialize it. 

The PR also addresses issue #692 by using the property logging.crypto, it is possible to configure the logging level of the crypto layer.

@bmos299 In this PR, I have also added more logging information related to the servers the crypto layer is connecting to. I hope it helps.


muralisrini:
@adecaro : thanks.. will test today.

muralisrini:
this might be related to #714 if logging config changes in this PR is causing the crash.

muralisrini:
ref #714 : we were not sure if the problem was in go-logging or in the way we were using it. It is only with this PR that we see this go-logging related crash. So whatever we are doing with that package prior to this PR is correct till proven otherwise.

Commenting out initialize-log-per-client code avoids the problem. We could try to figure out the right way to maintain that semantics. I suspect that's going to be a bit of work and may not be worth the effort ? An alternative might be to initialize log once at the beginning and write logs for all clients to it, thus avoiding write-read collisions. What do you think @adecaro ?

adecaro:
@muralisrini, @dongmingh I updated the PR to have a centralized log. I took the opportunity to introduce in main.go a line to initialize the crypto layer.

Please, @dongmingh, may you test the updated PR, if the performance are still 'good'. I will merge. Thanks :)

Issue #710, opened 2016-02-20
consensus helper ExecTx ignoring errors from transaction batch execution


muralisrini:
ExexTxs ignores error array from chaincode.ExecuteTransactions. This will result in transactions getting committed whether they succeeded or not.

jyellick:
This is pending to be fixed by issue #579 and you can find additional information there.

@tuand27613 We can still address this with a stopgap solution, if you would like to implement it in `helper.go`, it should be a fairly straightforward fix, or feel free to assign to me, and I can quickly handle it.

This is all related to a discussion that occurred between myself, @srderson, @kchristidis, and @binhn.  In short, `ExecTxs` used to return a `[]error` from the transactions, and it was consensus' responsibility to remove transactions from the commit list which had encountered an error.  There were two big problems with this, first, every consensus implementation must duplicate this logic, which increases the surface for bugs.  Secondly, consensus should be agnostic to the details of transactions, it does not make sense that two consensus algorithms, who pick the same order for transactions, could result in two different blocks.

Finally, there are also multiple competing ideas as to whether failing transactions should be stored or not.  Some use cases involve many many failing transactions, which could produce an excessively long blockchain with few valid transactions, and so wish to have invalid transactions excluded from the block.  In other use cases, it may be very important that for auditing purposes all attempted transactions are recorded, regardless of their ability to be executed.  This sort of complexity does not belong in consensus, and more likely belongs with the chaincode, or possibly some configuration.

@muralisrini Sorry for the verbose explanation, but I wasn't sure if you had any of the above on your radar.

tuand27613:
@muralisrini , your thoughts ?

jyellick:
@muralisrini Is this on your radar? We'd like to get this closed ASAP.

muralisrini:
@jyellick @tuand27613 @binh

To capture the main point - ExecTxs has access to chaincode generated errors but ignores them today. (Summarizing @jyellick comment) this is because we do not know how to treat errors...we just save everything today.

Shoud we just add a "not started/succ/fail" enum to proto.Transaction ? ExecTx will set this flag to succ/fail
- Consensus can make use of this if it wants to
- it will be stored in the block (just an an additional int)... available for audit
- event clients have information about failed transaction too (SDK could provide additional mechanism to filter out succ or failed transactions from the block)
- minimal change (and we don't duplicate code by exposing errors to higher layers)

The above won't work if we think we shoud NEVER save a block with erred transactions in it.

[...]

PR #711, opened 2016-02-20
resurrect deploy tx when re-launching chaincode

muralisrini:
This is needed for encryption/decrypion after relaunching chaincode.
Also has the side effect of preventing invokes/queries from getting launched before deploy completes successfully.
It has a work around for #710


muralisrini:

Issue #728, opened 2016-02-24
panic in crypto test unit tests


muralisrini:
@adecaro .. assigning this to you but do move it to the right person if incorrect please.

Running 
go test -timeout=20m $(go list github.com/openblockchain/obc-peer/... | grep -v /vendor/)

I see the following panic every time. Not sure if this is something in my environment (but I cannot think of why that should be...). 

--- FAIL: TestSPI (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
        panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xb code=0x1 addr=0x38 pc=0x50be8e]

goroutine 6 [running]:
panic(0x703c60, 0xc82000e0d0)
        /opt/go/src/runtime/panic.go:464 +0x3e6
testing.tRunner.func1(0xc82007c090)
        /opt/go/src/testing/testing.go:467 +0x192
panic(0x703c60, 0xc82000e0d0)
        /opt/go/src/runtime/panic.go:426 +0x4e9
crypto/ecdsa.randFieldElement(0x0, 0x0, 0x7f0a5af794f0, 0xc820018330, 0x0, 0x0, 0x0)
        /opt/go/src/crypto/ecdsa/ecdsa.go:84 +0x4e
crypto/ecdsa.GenerateKey(0x0, 0x0, 0x7f0a5af794f0, 0xc820018330, 0x0, 0x0, 0x0)
        /opt/go/src/crypto/ecdsa/ecdsa.go:100 +0x66
github.com/openblockchain/obc-peer/openchain/crypto/ecies/generic.TestSPI(0xc82007c090)
        /opt/gopath/src/github.com/openblockchain/obc-peer/openchain/crypto/ecies/generic/spi_test.go:19 +0xe1
testing.tRunner(0xc82007c090, 0x90f4a0)
        /opt/go/src/testing/testing.go:473 +0x98
created by testing.RunTests
        /opt/go/src/testing/testing.go:582 +0x892
FAIL    github.com/openblockchain/obc-peer/openchain/crypto/ecies/generic       0.014s

adecaro:
@muralisrini, I guess, it happens when running the tests for my PR, right? If this is the case, I have just updated the PR with a fix. Thanks.

adecaro:
@muralisrini Oh, just realized that my previous PR has been merged. I have issued another one to fix the bug. Sorry for the inconvenient.

muralisrini:
fixed by #730

PR #730, opened 2016-02-25
Fix for bug #728

adecaro:


muralisrini:
again assigning to self to unblock failing unit (and possibly behave) tests as @srderson is out.

muralisrini:
Don't see a panic from unit tests.  #728 appears to be fixed. In behave 1 step out of 310 failed but that's not because of a panic.

Merging.

Issue #734, opened 2016-02-25
.obc-ca/tca.db gets included in container image


muralisrini:
When we create chaincode images, we tar the src/ dir and if tca.db keeps changing under its feet while that;s happening, the tar will fail (its header will not indicate the tca.db file size when its copied as its changing). Also, we really don;t want to copy dynamic data into chaincode image

We can filter out “.obc-ca” out when creating the container image if that’s the best way or we can move .obc-ca it outside of src.

@adecaro , if the right thing to do is to filter it out at chain code image creation, move the bug back to me please.

muralisrini:
fixed by #735

PR #735, opened 2016-02-25
Fix for issue #734

adecaro:
With this PR the data produced when testing the crypto layer are stored in system temp directory.
This fix Issue #734 .


muralisrini:
Let me test and unblock given this is a unit test issue.

muralisrini:
I'm still seeing a panic from #728 but I see there's a PR for that out there for that. The problem for #734 appears to have gone away.

Issue #739, opened 2016-02-25
Error `nil amount for a` when running sieve state transfer `behave` test


tuand27613:
Error happens on the final query for value of "a" .

It only happens for sieve. Classic and batch tests pass.

It only happens when I run the entire peer_basic.feature bucket. If I run `behave --name "State transfer"` so that only the state transfer tests run then all 3 tests pass.

jyellick:
This is essentially happening because state transfer is completing, and pbft is executing outstanding requests via its `stateTransferCompleted` method before sieve's `Completed` method is invoked.  Because state transfer is done, sieve attempts to execute requests but the block number is out of sync.  I am working on a broader fix for this, because a direct fix would leave some other unfortunate corner cases.

I'll need to discuss some of the specifics with @corecode to make sure that these changes don't break any assumptions of sieve.

muralisrini:
@jyellick : "`nil amount for a` "  will happen if a query comes in between chaincode image creation and chaincode deploy initialization transaction (i.e., query is able to sneak in and start up an uninitialized chain code which should have really been started only by the deploy initialization transaction). There's a known race condition in chain code that can cause this (@corecode mentioned this). Wonder if that is at the bottom of this as well ?

jyellick:
@muralisrini It's possible that this is involved, but there is definitely a bug in the interaction between sieve state transfer and standard pbft state transfer which is visible in the logs for this issue.  There is a race between the two, with a number of options to resolve the race, I'll be discussing with @kchristidis and @corecode more today as the mechanism of how to handle it could have some significant overall design impacts.

muralisrini:
@jyellick : I take that comment back about race condition. The expected race condition was one where the chain code will never recover once the "nil amount " error appeared and would indicate an issue in chain code framework.

I could recreate the "nil amount" when querying in a loop while deploying chain code but it did recover from it once the deploy went thru. The received "nil amount"  (I get it one time in the query loop) is thus an expected behavior - for my test case. @corecode - can you confirm please ? 

If you don't expect a "nil amount" to show up due to the way your scenario is expected play out, that does seem to be a different problem that needs your investigation.

rameshthoomu:
Observed the same error while executing behave scripts `inside vagrant` or `outside of vagrant` environments.

tuand27613:
verify after #751 is merged

bmos299:
@rameshthoomu can you verify behave passes and @RicHernandez2 can you manually verify?

[...]

PR #740, opened 2016-02-25
Pbft behave tests

tuand27613:
Add code to make validating peer simulate byzantine faults

Add behave tests to verify consensus when one peer is byzantine


tuand27613:

tuand27613:
unit tests all pass.

behave tests pass except for issue #739. But the state transfer tests pass when I run them separately.

Issue #741, opened 2016-02-25
Inconsistency between docker environment and dev-env


ghaskins:
It was discovered that the docker environment was inadvertently based on debian:jessie rather than ubuntu trusty.  This should be based on ubuntu to match the vagrant environment, ideally driven by the same configuration

ghaskins:
@jeffgarratt @srderson @muralisrini @jyellick

ghaskins:
Gah, I started with opening an issue and then seem to have become confused and wrote it like a PR...

srderson:
@ghaskins closing as I believe you have resolved this issue with the openblockchain/baseimage image.

PR #742, opened 2016-02-25
peer modifications to support new packer-based dev-env

ghaskins:
This patch addresses #741 and works in conjunction with https://github.com/openblockchain/obc-dev-env/pull/48 to unify the development and runtime environments behind a singular definition based on ubuntu:trusty.  There are two primary benefits to this patch to consider:

1) We now are using prebuilt and/or pre-installed binaries for everything such that it is now inexpensive to create a new environment (e.g. vagrant up + BuildImage_Peer).  In my testing, building a new docker image takes about 1 minute, down from over 15 previously.

2) We have a system in place (via packer) where one definition is used to build both our development environment and the docker environments, which should encourage a more stable environment going forward.

I have tested this patch in conjunction with https://github.com/openblockchain/obc-dev-env/pull/48 and it passes all unit tests.  I also ran behave, and everything passed with one exception:

```
Failing scenarios:
  peer_basic.feature:218  chaincode example02 with 4 peers and 1 obcca, issue #680 (State transfer) -- @1.3 Consensus Options

0 features passed, 1 failed, 0 skipped
16 scenarios passed, 1 failed, 1 skipped
310 steps passed, 1 failed, 19 skipped, 0 undefined
```

It is my understanding that this failure is an outstanding issue not related to this patch.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,14 +1,11 @@
-from golang:1.6
-# Install RocksDB
-RUN cd /opt  && git clone --branch v4.1 --single-branch --depth 1 https://github.com/facebook/rocksdb.git && cd rocksdb && make shared_lib
-ENV LD_LIBRARY_PATH=/opt/rocksdb:$LD_LIBRARY_PATH
-RUN apt-get update && apt-get install -y libsnappy-dev zlib1g-dev libbz2-dev
+from openblockchain/baseimage:latest
+
 # Copy GOPATH src and install Peer
 RUN mkdir -p /var/openchain/db
 RUN mkdir -p /var/openchain/production
 WORKDIR $GOPATH/src/github.com/openblockchain/obc-peer/
 COPY . .
-RUN CGO_CFLAGS="-I/opt/rocksdb/include" CGO_LDFLAGS="-L/opt/rocksdb -lrocksdb -lstdc++ -lm -lz -lbz2 -lsnappy" go install && cp $GOPATH/src/github.com/openblockchain/obc-peer/openchain.yaml $GOPATH/bin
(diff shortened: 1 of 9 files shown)


binhn:
@ghaskins could you add a doc on how to build and upload pieces that the script needs. I want to make sure there's instruction for someone to be able to do this.

@jeffgarratt any comments?

ghaskins:
@binhn I added some doc to the baseimage/README.md included in the patch:

https://github.com/ghaskins/obc-dev-env/blob/packer-baseimage/baseimage/README.md

Let me know if this isn't what you had in mind.

binhn:
@ghaskins I am looking for info on how the pieces (eg rocksdb) may be built and uploaded as shown in the diagram. It isn't clear to me.  Thanks.

ghaskins:
@binhn ah, ok.  I will add some more documentation for that process then

ghaskins:
@binhn @srderson it might also be the case where we would be wise to keep the packaging in a repo somewhere (e.g. openblockchain/third-party/rocksdb), as I did have to develop the .deb packaging metadata.  It is tracked (to a degree) within the launchpad environment, but its not super obvious (especially for someone coming up to speed) to figure that out...TBD.

ghaskins:
Another idea is we could just put any third party stuff right in dev-env (e.g. openblockchain/obc-dev-env/third-party/... )

binhn:
@ghaskins +1  openblockchain/obc-dev-env/third-party/...

ghaskins:
@binhn https://github.com/ghaskins/obc-dev-env/blob/packer-baseimage/baseimage/PPA.md

ghaskins:
Closing for now to re-work patch without utilizing PPA

PR #768, opened 2016-03-02
Add behave test for range query. Resolves #767

### LINKED ISSUE #767: Behave test for shim range query functionality
Need to add a behave test for the shim range query functionality.
1. Adding an example "map" chaincode
2. Fixing a bug in the behave logic that caused the "invoke" and "query" to fail

srderson:
Also, fix issue with "invoke" and "query" actions in behave when functions had names other than "invoke" or "query"


srderson:

Issue #477, opened 2016-01-13
Chaincode shim table API


srderson:
Provide an API to emulate tables/rows/columns in the chaincode shim. This has been proposed as a way to hide low level state key/value store API from chaincode developers.

API
createTable(name, []columnDefinition)
insertRow(row)
getRow(key)
deleteRow(key)
replaceRow(row)

markparz:
Not required for Jan... true run-at

frankyclu:
can we add API to query by column value?

srderson:
Today the API has
`GetRow(tableName string, key []Column) (Row, error)`

and for partial keys
`GetRows(tableName string, key []Column) (<-chan Row, error)`
See comment at https://github.com/openblockchain/obc-peer/blob/master/openchain/chaincode/shim/chaincode.go#L459 for details.

are you asking about querying via non-key column values?

bmos299:
@srderson I'll work on this one with you.  Please let me know when you have some time.

PR #774, opened 2016-03-03
Initial table shim API for Go. Issue #477

srderson:
@binhn @jeffgarratt 

This is the initial pass a table API support for the Go shim.

The following types are supported for columns
string
int32
more will be added in the future.

The following API is available in the Go chaincode shim

```
CreateTable(name string, columnDefinitions []*ColumnDefinition) error
GetTable(tableName string) (*Table, error)
DeleteTable(tableName string) error
InsertRow(tableName string, row Row) (bool, error)
ReplaceRow(tableName string, row Row) (bool, error)
GetRow(tableName string, key []Column) (Row, error)
DeleteRow(tableName string, key []Column) error
```

The next step will be to add
`GetRows(tableName string, partialKey []Column) (Rows, error)`

There is a new behave test `chaincode shim table API, issue 477` that demonstrates usage.


srderson:

Issue #764, opened 2016-03-01
Crash When Running Test App


bmos299:
I am crashing when I try to deploy my chaincode (sample code 2).  I am attaching the chaincode logs and obc-peer log.  I also have security enabled so I am including the openchain.yaml and obcca.yaml.
It looks like chaincode actually gets deployed.  My application receives back a failure on the rest call.  the obc-peer than sits for a few minutes, cuts a message that the chaincode is deployed, and then crashes.  I am running at commit 02d50f9c7691871253966128031dbe23d77ff34f,

[chaincodelogs.txt](https://github.com/openblockchain/obc-peer/files/153688/chaincodelogs.txt)
[obcpeer.txt](https://github.com/openblockchain/obc-peer/files/153687/obcpeer.txt)

[openchain.yaml.txt](https://github.com/openblockchain/obc-peer/files/153700/openchain.yaml.txt)
[obcca.yaml.txt](https://github.com/openblockchain/obc-peer/files/153701/obcca.yaml.txt)

bmos299:
I can't run any new test cases without hitting the crash.   I ran my app against the bluemix service and it worked fine, so this points more towards a recent change.

jeffgarratt:
@bmos299 do the existing behave tests pass?

bmos299:
The crash occurs with the default openchain.yaml file.  The backtrace is pointing towards rocksdb so I am assigning to Sheehan and including the log. 
[rocksdb.txt](https://github.com/openblockchain/obc-peer/files/157694/rocksdb.txt)

bmos299:
@srderson please let me know what you need.
Thanks, Barry

srderson:
@bmos299 Could you run `go version` in your vagrant image and confirm that you're running 1.6? If so, this may be related to https://github.com/tecbot/gorocksdb/issues/42 and I need to update the gorocksdb package.

bmos299:
@srderson sure thing.

srderson:
@bmos299 I haven't reproduced this issue locally, so could you try PR #786 and see if it resolves this issue?

bmos299:
@srderson I will.

bmos299:
@srderson the crash is still occurring with PR #786 and Go 1.6.

bmos299:
...but is much less infrequent.

srderson:
@bmos299 The PR has been merged to the master branch now. Could you catch up with the master branch, reproduce, and attach the log?

[...]

PR #786, opened 2016-03-03
Update gorocksdb. Issue #764

srderson:
Updating gorocksdb to see if it resolves issue #764


corecode:
This seems to have fixed local crashes with rocksdb that I have been observing on my laptop.

srderson:
@bmos299 also confirmed this resolved the crash he was seeing.

PR #795, opened 2016-03-04
Support additional types in Go shim table. Issue #477

srderson:
This adds many more types to the table API available in the Go chaincode shim. Supported types are now

```
string
int32
int64
uint32
uint64
bytes
bool
```


srderson:

PR #805, opened 2016-03-08
CI integration

binhn:


pkg/config/config.go
```diff
@@ -50,10 +50,10 @@ type Resources struct {
...
 // Organization is an organization in the channel configuration.
 type Organization struct {
-	Name      string
-	ID        string
-	Policies  map[string]*Policy
-	MSPConfig *mb.FabricMSPConfig
+	Name            string
+	ID              string
+	Policies        map[string]*Policy
+	FabricMSPConfig *FabricMSPConfig
```
sykesm: Fabric is super-redundant. No need to include that in the name. (Now that it's a type in this package, the same could probably be said for the 'Config' suffix).

pkg/config/config_test.go
```diff
@@ -617,21 +556,21 @@ func TestNewOrgConfigGroup(t *testing.T) {
 		}
 	},
 	"values": {
-		"AnchorPeers": {
+		"Endpoints": {
 			"mod_policy": "Admins",
-			"value": "CgkKBWhvc3QxEHs=",
+			"value": "Cg1sb2NhbGhvc3Q6MTIz",
```
sykesm: We need to open a bug to find out why this isn't being decoded into an object.

pkg/config/config_test.go
```diff
@@ -617,21 +556,21 @@ func TestNewOrgConfigGroup(t *testing.T) {
...
 			"version": "0"
 		},
 		"MSP": {
 			"mod_policy": "Admins",
-			"value": "",
+			"value": "EuMCMjoKNi0tLS0tQkVHSU4gQ0VSVElGSUNBVEUtLS0tLQotLS0tLUVORCBDRVJUSUZJQ0FURS0tLS0tChIAOjgKNi0tLS0tQkVHSU4gQ0VSVElGSUNBVEUtLS0tLQotLS0tLUVORCBDRVJUSUZJQ0FURS0tLS0tCkIAWugBEjgKNi0tLS0tQkVHSU4gQ0VSVElGSUNBVEUtLS0tLQotLS0tLUVORCBDRVJUSUZJQ0FURS0tLS0tCho4CjYtLS0tLUJFR0lOIENFUlRJRklDQVRFLS0tLS0KLS0tLS1FTkQgQ0VSVElGSUNBVEUtLS0tLQoiOAo2LS0tLS1CRUdJTiBDRVJUSUZJQ0FURS0tLS0tCi0tLS0tRU5EIENFUlRJRklDQVRFLS0tLS0KKjgKNi0tLS0tQkVHSU4gQ0VSVElGSUNBVEUtLS0tLQotLS0tLUVORCBDRVJUSUZJQ0FURS0tLS0tCg==",
```
sykesm: We need to open a bug to find out why this isn't being decoded into an object.

pkg/config/consortiums_test.go
```diff
@@ -88,14 +86,16 @@ func TestAddOrgToConsortium(t *testing.T) {
...
 
 	orgToAdd := &Organization{
-		Name:      "Org1",
-		ID:        "Org1MSP",
-		Policies:  applicationOrgStandardPolicies(),
-		MSPConfig: &mb.FabricMSPConfig{},
+		Name:            "Org1",
+		ID:              "Org1MSP",
+		Policies:        orgStandardPolicies(),
+		FabricMSPConfig: baseFabricMSPConfig(),
```
sykesm: Really think this could be as simple as `MSP`.

pkg/config/msp.go
```diff
@@ -0,0 +1,264 @@
...
+
+// FabricMSPConfig is the configuration information for
+// a Fabric MSP.
+// Here we assume a default certificate validation policy, where
+// any certificate signed by any of the listed rootCA certs would
+// be considered as valid under this MSP.
+// This MSP may or may not come with a signing identity. If it does,
+// it can also issue signing identities. If it does not, it can only
+// be used to validate and verify certificates.
+type FabricMSPConfig struct {
```
sykesm: s/FabricMSPConfig/MSP/  ?

pkg/config/msp.go
```diff
@@ -0,0 +1,264 @@
...
+	// Identity revocation list
+	RevocationList []pkix.CertificateList
+	// SigningIdentity holds information on the signing identity
+	// this peer is to use, and which is to be imported by the
+	// MSP defined before
+	SigningIdentity *SigningIdentityInfo
+	// OrganizationalUnitIdentifiers holds one or more
+	// fabric organizational unit identifiers that belong to
+	// this MSP configuration
+	OrganizationalUnitIdentifiers []*FabricOUIdentifier
```
sykesm: Can we remove the Fabric prefix and shorten this? (... I see that's what the name is in the proto (sigh)...)

PR #806, opened 2016-03-08
Adding unix format for files and travis build result file

binhn:


binhn:

tuand27613:
.travis.yml doesn't set notifications for build results ?  

Will that be a separate commit ?  I think we should have the notifications sent to #github channel along with the pull request notices

genggjh:
how about the .proto files? will you consider keep these file into unix format?

Issue #811, opened 2016-03-09
performance degradation


dongmingh:
Commit level 42c6fe9ff5bceb4258eb9754fa33fef5f668ecd2 can only perform 25 invokes per second with security and privacy enabled and both peer and crypto set to error.  The test is performed with SoapUI with 100 threads.

After trace back, found that commit e3c27dcf853863bda1f962addf65e6db43232521 performs 152 Invokes per second but commit 329127cf14f2ce7637308e3c7693b5190abd8111 can only perform 25 Invokes per second.

commit e3c27dcf853863bda1f962addf65e6db43232521
Author: srderson sranderson@gmail.com
Date:   Thu Mar 3 12:14:42 2016 -0500

```
Initial table shim API for Go. Issue #477
```

commit 329127cf14f2ce7637308e3c7693b5190abd8111
Merge: 3c9ff3a 2cd84df
Author: Binh Q Nguyen binhn@us.ibm.com
Date:   Thu Mar 3 15:18:45 2016 -0500

```
Merge pull request #745 from adecaro/next_conf

Confidentiality 1.2
```

muralisrini:
@adecaro : after bracketing, issues seems to be with commit 329127c

adecaro:
@dongmingh @muralisrini I have just submitted a PR to address this issue. Please, let me know if it fixes the problem. Thanks :)

dongmingh:
With the commit from @adecaro , 
commit 085dcb505674780746bd65df05dafa878d77bc41
Author: Angelo De Caro adc@zurich.ibm.com
Date:   Thu Mar 10 17:15:54 2016 +0100

```
removed unused imports
```

With Tcert=200, threads=100, 5 min run, the throughput are as follow:
peer and crypto set to error, security and privacy enabled, tps=121 and 151 for invoke and query resp.
peer and crypto set to error, security and privacy disabled, tps=185 and 257 for invoke and query resp.
peer and crypto set to info, security and privacy enabled, tps=104 and 140 for invoke and query resp.
peer and crypto set to info, security and privacy disabled, tps=162 and 252 for invoke and query resp.

muralisrini:
that's closer to the original numbers. @adecaro : do you know what's causing the delta degradation (given that @dongmingh is able to reproduce the original values with the previous level of code) ?

dongmingh:
With peer and crypto set to error and security and privacy enabled, tps = 147/236 (invoke/query) about two weeks ago, see #684

[...]

PR #813, opened 2016-03-10
Fix: Performance Degradation #811

adecaro:
EC usage: ScalarMult (and related methods) must be invoked directly on Curve and not on Params. This is in order to take advantage of the speedup introduced by go1.6.


muralisrini:
unit and behave tests passed. Merging

Issue #751, opened 2016-02-29
Factor out and implement split `Orderer` and `Executor` services


jyellick:
The design of `obcpbft` attempts to separate the core pbft functions from the specific implementations, ie `obc-batch`, `obc-classic`, and `obc-sieve`.

However, a greater level of abstraction is required to isolate the ordering from the execution.
## API

This issue proposes a new `Executor` service with the following simple API:

```
ValidState(seqNo uint64, checkpointID []byte, sources []interface{})
Execute(seqNo uint64, request []byte, execInfo ExecuteInfo)
```

The `ValidState` call is invoked whenever the `Orderer` becomes aware of a state which the network has determined to be valid.  This call is not to be replied to.  In the PBFT standard case, this would be invoked when `f+1` checkpoint certificates have been received, but could also be invoked during a view change.  The `checkpointID` is discussed below, but is an opaque `[]byte` supplied by other `Executor`s.   The `sources` parameter indicates a list of locations which the `Executor` can retrieve the state from.  Note, for now this is labeled as `[]interface{}`, because it is not obvious whether this should be the replica ID, peer ID, or some other abstraction.

The `Execute` call is invoked for a request that should be executed with a particular sequence number.  The request is passed as an opaque `[]byte`.  The `ExecuteInfo` indicates what sort of response the `Executor` should deliver to the `Orderer` (presently only one response `Checkpoint` is proposed), and any other useful metadata, such as if the `Orderer` realizes there is going to be a gap in the sequence numbers (indicating a state transfer is necessary).

In the event that an execution request is tagged as requesting a `Checkpoint` response, after the request is invoked, then the `Executor` would invoke on the `Orderer`

```
Checkpoint(seqNo uint64, id []byte)
```

Where `id` is an opaque `[]byte`.  This should include at least, the a hash of the state corresponding to the sequence number, and optionally any metadata (such as in the case of `obc`, the block number). 
## State Transfer

It would be the executor's sole responsibility to perform state transfer.  The `Orderer` is frequently informing the `Executor` of known valid states, and so the `Executor` should never be stuck in an inconsistent state for any long amount of time.  Because it is the `Executor` which supplies the opaque `id []byte`, the executor can ensure that the checkpoint value contains all necessary information to recover the state.

Because execution requests contain sequence numbers, the `Executor` can detect if it is missing a transaction and optionally choose to initiate state transfer to a later point.

[...]

Issue #695, opened 2016-02-17
State transfer should retrieve defaultPeerIDs from Stack


jyellick:
State transfer currently expects a list of PeerIDs in its constructor.  This makes testing more difficult, and is especially problematic for the `N=1` PBFT case.

Instead, state transfer should interrogate the stack to discover the IDs of other validating peers that it can sync from.

@kchristidis Could you tag with statetransfer, and assign to me?

tuand27613:
all yours @jyellick

jyellick:
This was included in PR #833

Issue #756, opened 2016-03-01
Consensus should run with whitelisted VPs when security is enabled


kchristidis:
This is a follow-up to #462.

We currently require that VPs are named as `vp0`, `vp1`, ...`vpX` where `X+1` is the maximum number of validating nodes that we have set in the config file.  [As noted in #462](https://github.com/openblockchain/obc-peer/issues/462#issuecomment-173421403), this approach is flaky and needs to go.

The proper way to do this is: create a whitelist of hashed raw enrollment certificates that will be allowed to connect to the network. We dropped this because it was causing issues when automatically deploying a network through Swarm.

@adecaro: Can we modify the CA so that it pre-generates a list of N enrollment certificates. We could then use them to populate the network's whitelist. Later on, as each VP signs up, we'll have the CA return one of these pre-generated certificates. Are there any security implications we need to be aware of in this approach?

Another way to do this is: Wait for the first N validating nodes to connect to the network, and don't kick off consensus until this happens. These are your validators, this is your whitelist. (Chet is thinking along these lines as well IIRC.)

kchristidis:
@adecaro Can you have a look at the question above and give me your take? Let's see if we can maybe tackle this during Sprint 1.

binhn:
I would prefer to wait for N rather listing pre-generated certs, which would be harder to handle in configuration

kchristidis:
Got it. (@adecaro we're good then, I can work on this without being blocked.)

bmos299:
@RicHernandez2 can you talk to @kchristidis about this Issue.

kchristidis:
Consider the scenario where a VP disconnects and reconnects, i.e. it has a whitelist installed already.

This VP will:
1. Attempt to connect to the peers in the whitelist, and reject any connections from validators not in that list.
2. It'll wait until 2f such connections are formed before it starts consensus.

Sounds right? @corecode @jyellick @tuand27613

tuand27613:
Should that vp wait until 2f connections ?

kchristidis:
@tuand27613 That's right. 2f, so that you can have 2f+1 PREPARE messages, etc.

kchristidis:
There is an issue where `GetWhitelist()` doesn't produce the proper unmarshaled protobuf. See code [here](https://gist.github.com/kchristidis/232b6954b0d0857c4401); 87 bytes are read, but the unmarshaled protobuf comes up blank.

tuand27613:
[...]

PR #833, opened 2016-03-16
obcpbft Executor split

jyellick:
This changeset is intended to logically split the obc functions from the pbft ones.

It is specifically intended to address issue #751, but also closes issues #739 and #695.  Other consensus related bugs, in particular ones involving state transfer, are possibly fixed through this changeset as well.

There is a new file `obc-executor.go` which is responsible for all of the interactions with the ledger, both through executions and through state transfer.  All PBFT implementations (classic, batch, and sieve) now only depend on the network oriented pieces of OBC.  The core PBFT implementation in `pbft-core.go` no longer contains any references to obc, other than logging and a single crypto hash utility.

The test framework has been restructured extensively.  Previously, there was a `testnet` which contained a list of `instance`s, which implemented both the `consensus.Stack` and `innerStack` interfaces.  This conflation caused considerable problems in removing ledger dependence from the pbft tests.  To address this, `testnet` was moved to its own `mock_network_test.go` file and has been changed to refer only to `endpoint`s which have a limited interface.  Then, a `pbftTestnet` to be used to interact with the pbft-core and a `consumerTestnet` designed to interact with the pbft plugins were created in `pbft-core_mock_test.go` and `mock_consumer_test.go` respectively.

There is also a new `omniProto` type in the file `mock_utilities_test.go` which implements most interfaces used in obcpbft.  It contains default implementations which panic, but can be overridden setting a corresponding `<Function>Impl` variable.  This allows for tests to create one-off limited implementations simply.

This is unfortunately a rather large changeset.  Although all go tests, and behave tests pass, it definitely deserves a degree of scrutiny in code review.  I'd appreciate it if @kchristidis @corecode and @tuand27613 could all take a look.

Please do not merge until the people listed above have had a chance to review.


openchain/consensus/obcpbft/obc-executor.go
```diff
@@ -0,0 +1,472 @@
...
+func (obcex *obcExecutor) Execute(seqNo uint64, txs []*pb.Transaction, execInfo *ExecutionInfo) {
+	request := &transaction{
+		seqNo:    seqNo,
+		txs:      txs,
+		execInfo: execInfo,
+	}
+	select {
+	case obcex.executionQueue <- request:
+		logger.Debug("%v queued request for sequence number %d", obcex.id, seqNo)
+	default:
```
corecode, line 151: how does this work?

openchain/consensus/obcpbft/obc-sieve.go
```diff
@@ -36,29 +36,52 @@ type obcSieve struct {
...
+	logger.Debug("Replica %d obtaining startup information", id)
+	startupInfo := <-op.startup
+	close(op.startup)
+
+	op.pbft = newPbftCore(id, config, op, startupInfo)
 
 	return op
 }
 
+func (op *obcSieve) Startup(seqNo uint64, id []byte) {
```
corecode, line 90: Why doesn't NewOBCExecutor return the values passed to Startup()?  Why do all consumers have to implement this channel init blocking thing?

openchain/consensus/obcpbft/obc-sieve.go
```diff
@@ -317,6 +339,51 @@ func (op *obcSieve) processExecute() {
 		op.id, verify.BlockNumber)
 	op.broadcastMsg(&SieveMessage{&SieveMessage_Verify{verify}})
 	op.recvVerify(verify)
+
+	op.pbft.unlock()
+
+	logger.Debug("Sieve replica %d waiting for result decision on block %d", op.id, seqNo)
+	select {
+	case result := <-op.validResultChan:
```
corecode, line 361: Putting the logic here makes the flow very difficult to follow.

openchain/consensus/obcpbft/obc-sieve.go
```diff
@@ -535,35 +604,34 @@ func (op *obcSieve) executeVerifySet(vset *VerifySet) {
 
 	if !sync {
 		if !shouldCommit {
-			logger.Error("Execute vset: not deterministic")
-			op.rollback()
+			logger.Error("Seive replica %d execute vset: not deterministic", op.id)
```
corecode: "Seive"

openchain/consensus/obcpbft/pbft-core.go
```diff
@@ -218,51 +209,7 @@ func newPbftCore(id uint64, config *viper.Viper, consumer innerStack, ledger con
...
-	}
-	genesisHash, err := ledger.HashBlock(genesisBlock)
-	if err != nil {
-		panic(fmt.Errorf("Cannot hash genesis block: %s", err))
-	}
-	instance.chkpts[0] = &blockState{
-		blockNumber: 0,
-		blockHash:   base64.StdEncoding.EncodeToString(genesisHash),
-	}
+	instance.chkpts[0] = base64.StdEncoding.EncodeToString(startupID)
```
corecode, line 212: Does this handle a restart of a peer?

openchain/consensus/obcpbft/pbft-core.go
```diff
@@ -840,60 +762,25 @@ func (instance *pbftCore) executeOne(idx msgID) bool {
 	if digest == "" {
 		logger.Info("Replica %d executing/committing null request for view=%d/seqNo=%d",
 			instance.id, idx.v, idx.n)
+		// Note, that it is not possible for a null request to be a multiple of the checkpoint interval, so no checkpointing checking is needed in this path
```
corecode: Why can't a null request be a multiple of the checkpoint interval?

openchain/consensus/obcpbft/obc-executor.go
jyellick, line 151: The executionQueue is a buffered channel.  Its size is configurable, but for pbft, it should be at least the size of the pbft log.  Because there can not be more outstanding executions than the the log size (because a checkpoint is required to move the watermarks, and a checkpoint is generated from the execution result), it should never be possible for pbft to overflow this channel.

However, in the interest of generality, and to recover from any unforeseen errors, the write to the channel is done via a select in a non-blocking way.  If there is no room on the channel, then the channel is emptied, and a flag (a special request with nil txs)  is set in the channel indicating that there was a gap and that state transfer should be performed.

openchain/consensus/obcpbft/obc-sieve.go
jyellick, line 90: This is a little hacky, and I don't like the code duplication.  I considered having this value returned via the `NewOBCExecutor`, but chose not to for a few different reasons.

In design, I tried to think of `Exector` and the `Orderer` as being two different processes, talking via the network.  Because of this, it makes sense that the instantiation of the `OBCExecutor` should not require a connection to the `Orderer`, similarly, the fact that the `Orderer` is invoking the constructor of the `Executor` is a local implementation detail, so relying on output from its instantiation seemed wrong.  You can think of that `Startup` method more as a network `Handshake` method, logically.

If there are strong feelings that this is the wrong approach, I can rework the patch to provide the seqNo and ID via the constructor call.

openchain/consensus/obcpbft/obc-sieve.go
jyellick, line 361: Before this patch, the sieve flow worked roughly as:
1. Begin Sieve Request (Sieve Message thread)
2. BeginBatch (Sieve Message thread)
3. ExecTxs (Sieve Message thread)
4. PreviewBatch (Sieve Message thread)
5. Send Verify (Sieve Message thread)
6. Assign final ordering (PBFT Message thread)
7. Queue execution (PBFT Message thread)
8. Make decision (Execution thread)
9. CommitBatch (Execution thread)

Although the flow is certainly simpler, the problem here is that the Batch, Exec, and Preview operations will all happening on the Sieve Message Thread, while the Commit is happening via the Execution thread.  The new flow looks as follows:
1. Begin Sieve Request (Sieve Message thread)
2. BeginBatch (Execution thread)
3. ExecTxs (Execution thread)
4. PreviewBatch (Execution thread)
5. Send Verify (Execution thread)
6. Assign final ordering (PBFT Message thread)
7. Queue execution (PBFT Message thread)
8. Make decision (PBFT Message thread)
9. CommitBatch (Execution thread)

This is another effect of attempting to view the `Executor` as a network service, that executions are queued asynchronously, and the `Orderer` can only interact with their results.

openchain/consensus/obcpbft/pbft-core.go
jyellick, line 212: Partially, assuming that all VPs (or at least 2f+1 honest ones) were stopped while idle, so that their state, and chain were the same, then this should allow us to handle the restart.

However, this is no different than the previous behavior, but PBFT takes its initial checkpoint value via the constructor, rather than interrogating the ledger directly.

openchain/consensus/obcpbft/pbft-core.go
jyellick: As I think more about this, I actually think this is a false assertion.

When a primary is faulty, and a view change occurs.  The primary has assigned requests to sequence numbers, but has chosen to skip some sequence numbers.  At view change, the latest checkpoint quorum certificate is selected to work from, and the missing requests are filled in with nulls.

The erroneous logic was that if we had requests beyond the next checkpoint sequence number, then we would have simply chosen that checkpoint as the basis for our new view, but there is no reason to believe that the Primary ever assigned a request to that checkpoint sequence number.

I'd be open to suggestions as to how to address this.  I think the obvious, but incorrect solution is to add an API to retrieve a snapshot ID from the executor.  But this seems like the wrong solution to me.  The problem being that it introduces a dependency on the current state of execution, so it would also need to include a sequence number, and it might block for a very long time, might be missed because of state transfer, etc.

My suggestions would be that we could extend the notion of a 'Null Request' to the executor, so that it does not perform an execution but does send back a checkpoint ID.  In this way it gets into the execution queue normally, and otherwise follows the normal execution/checkpoint path.

jyellick:
@corecode I've pushed another commit to address the 'Seive' misspelling, as well as to handle the null checkpoint request.

@tuand27613 @kchristidis would still love to hear any feedback you might have.

kchristidis:
(@jyellick Not ignoring this, just tied up with #756.)

jyellick:
@tuand27613 @kchristidis @corecode I think this is just about ready to be pulled, but I'd appreciate any last minute looks it could get before doing so.

There was a bug that was fixed where the ledger was not setting the previous block hash which was causing the Sieve behave test to occasionally fail.

There was also a race in the new Sieve path which should now be fixed.

Additionally, the pbft logging was cleaned up a bit so that we wouldn't get all of the 'out of view window' warning messages in the log constantly.  Instead, we now only log at warning if it is outside of the view window and not on the lower boundary.

jyellick:
With the inclusion of a5fdf8e the troubles I was seeing with high volumes of requests seem to have gone away.  Was able to successfully commit 2000 transactions via behave with a checkpoint interval of 2 (which typically exacerbates the problem).  @tuand27613 If you could double check that the tests pass on a system that isn't mine, I think this should be in merge-able shape.

jyellick:
For posterity's sake, here is some quick documentation on the latest Sieve deadlock that was causing problems.

Execution was happening like this:
1. Sieve does speculative execution, sends out Verify, executor thread is now blocked until decision
2. Verify set arrives via PBFT, unblocks executor thread with decision which triggers state transfer
3. Executor thread requests state transfer, executor thread blocks
4. New speculative execution queued
5. Verify set arrives via PBFT, blocks waiting for second speculative execution to complete

The threads are now in a deadlocked condition.  The state transfer messages are stuck behind the PBFT message, and cannot be delivered, the execute thread is waiting on state transfer, and the PBFT message is waiting on the execution thread.

The solution was to introduce yet another thread to handle the PBFT execution's tasks, immediately unblocking so that the message thread can complete the state transfer.

Issue #836, opened 2016-03-16
peer is currently hardcoded to only accept GOLANG type chaincode


ghaskins:
Some minor re-work is required to allow the system to be extended to other chaincode types.

ghaskins:
@joequant Related to #807

srderson:
@ghaskins closing as I see your PR is merged. Please reopen if needed.

PR #838, opened 2016-03-16
Generalize the chaincode language type supported

ghaskins:
Things are somewhat hardcoded to only support GOLANG.  Lets generalize this to solve issue #836.  

This patch largely consists of restructuring the code so that golang specific code is now in openchain/container/golangcontainer, leaving room for other openchain/container/XXcontainer implementations in the future.  Also, the general peer has been modified so that it no longer either assumes golang or explicitly rejects non-golang types, but allows us to more easily extend the types accepted.  Lastly we move some logic under openchain/container/util so that it may be shared.  Leaving the code in the top-level openchain/container/\* would have created import cycles.

All unit/behave tests pass.


ghaskins:
@muralisrini

muralisrini:
@ghaskins : I think this is a good step to modularize language separation. Few observations 
- there is golang specific stuff in language neutral files
  openchain/container/util/write.go (this being a "util" file has the feel of being language neutral)
- there is golang type checking in language neutral files 
  openchain/devops.go
  openchan/container/vm.go

Seems there is room for some more language abstraction (say with a "lang" package) ... am I asking for too much :-) ?

Also, it might help with the above if we were to go back to having "golang" for the Dockerfile in yaml. This may even be necessary if we allow chaincodes of different languages deployed in the same peer ?

Comments please @jeffgarratt @srderson @binhn

ghaskins:
@muralisrini thank you for the review.

Regarding point 1 (util/write.go go-isms), note that it was meant to be a utility for container generation in general, not necessarily specific to golang as a chaincode platform.  The are things that use it are both in container (see obcca generation) as well as container/golangcontainer, so I wasn't sure where the best home for it would be.  I suppose it could go under the golangcontainer section, but I was trying to keep that more chaincode oriented.  Maybe that distinction isn't necessary.

Regarding point 2 (devops/vm golang infiltration), I will review and see if I can come up with further refinement.

Regarding point 3 (yaml change) I think I agree with you.  It might make sense to leave that as is so each platform could define its own.  My original rationale is they could all use it thus the generalization, but perhaps that isn't as flexible.

ghaskins:
@muralisrini regarding point 1, would you rather see more purity with golang isms being classified and grouped appropriately, or clarity on the intent of golangcontainer being golang-chaincode specific?  For the latter, perhaps I should s/golangcontainer/golangchaincode

muralisrini:
@ghaskins : hopefully this addresses the comment....

I was thinking along the lines of a language package with a "Factory" which will encapsulate coarse grained interfaces such as "createPackage". This level of abstraction might lead to a bit of duplication of code at a finer grain (such are reading, tarring , etc for different languages) but will clearly separate language implementations. This will also delegate language checks to the lang package thus avoiding check at different points in code.

Or put it another way, if I want to write support for a new language hopefully I could just look in one place (the language package or factory, etc) and copy the pattern.

I'm assuming this is just "more of the same" work that you have begun and not a radical surgery....

ghaskins:
@muralisrini I refactored some things according to your suggestions, as we discussed in DM earlier.  As before, branch passes unit/behave.  Let me know what you think.

muralisrini:
@ghaskins : a quick look shows golangism's gone from vm and devops using language encapsulation from "platforms".  That's a good litmus test of the attempt at language type generalization.

muralisrini:
Test and behave worked for me as well.

In summary
- platforms.go defines a minimal interface which uniformly uses ChaincodeSpec to encapsulate platform operations (currently  used by devops and vm)
- platform specific operations are all in platform files (currently just in golang package)... (Note - there are Gopath specific methods in container/util files outside of platform but these are for generating peer and obcca images which are part of the golang-fabric)

Net result is that chaincode/platforms package encapsulation should make it easier to add chaincode support for new languges.

PR #845, opened 2016-03-18
Access control at chaincode level

adecaro:
This PR shows how to do access control at chaincode level by using a naive asset management as driving example (asset_management.go)


stanliberman:
I have a few questions/comments:
- Seeing as how most clients would need functionality of isCaller(), would it make sense to make it part of the API?
- transfer() does DeleteRow() and then InsertRow(). As far as I understand, there are no atomic guarantees on partial failures. UpdateRow() may be more appropriate.
- Is it possible to attach additional metadata to certificates for entitlements? Currently each chaincode will need to manage more granular permissions internally. I think it would be very beneficial to have the ability to store those with the identity (likely ecert).

ghaskins:
@stanliberman One thing to keep in mind is that any chaincode invoke is always within a transaction context, so the operations should be implicitly atomic.  Or did I misunderstand the issue you are raising?

binhn:
@stanliberman the next rev of tcert may contain attributes for the purpose you mentioned

stanliberman:
@ghaskins I talking specifically about local state. If there was a failure on InsertRow() it would have already been deleted from the table. Do atomic transaction guarantees include local state store?

@binhn That's great! Looking forward to it.

ghaskins:
@stanliberman Chaincode should only have one "local state" and that is surfaced via the PutState/DeleteState ABI on which the table row mechanism is based.  Each chaincode invoke() is considered under an implicit transaction.  Therefore, if you mutate your state (such as DeleteRow) and then subsequently fail (such as InsertRow), just abort the entire invoke operation and the previously successful mutation will automatically roll back (or its a bug if it doesn't).  FWIW there is no way to express subordinate transactions (such as to wrap the Delete+Insert), so your bounds on atomicity are at the invoke level.  The corollary to this is that you should not be storing any "local state" outside of the PutState KV-store:  Your code won't properly participate in consensus and will probably exhibit non-deterministic behavior.

Issue #879, opened 2016-03-25
obc panic and grpc timeout with pbft consensus


dongmingh:
commit f08ccce33b13f23eb1af97a43bf654aa48129794
Merge: ab0e167 bfecc0f
Author: tuand27613 tdang@us.ibm.com
Date:   Thu Mar 24 17:30:24 2016 -0400

```
Merge pull request #833 from jyellick/executor-split

obcpbft Executor split
```

Running PBFT consensus, security and privacy enabled, 4 docker containers,  invoke requests with threads = 100.  

Steps to recreate:
1. bring up 4 docker containers
2. login as lukas
3. deploy example02
4. issue single query and invoke transactions and confirm that the responses are correct
5. use SoapUI to generate invoke transactions with 100 threads
Then vp0 and vp3 have panic and vp1 and vp2 have grpc timeout.  See logs below.

vp0 has panic:
20:03:26.325 [consensus/obcpbft] Execute -> ERRO 002 name:"vp0"  error queueing request (queue full) for sequence number 72
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xb code=0x1 addr=0x1 pc=0x9e7a82]

goroutine 51 [running]:
panic(0xcd2c00, 0xc820010090)
        /opt/go/src/runtime/panic.go:464 +0x3e6
github.com/openblockchain/obc-peer/openchain/consensus/obcpbft.(_obcExecutor).execute(0xc8203d5440, 0xc82861bbf0, 0x0, 0x0, 0x0, 0x0)
        /opt/gopath/src/github.com/openblockchain/obc-peer/openchain/consensus/obcpbft/obc-executor.go:477 +0x82
github.com/openblockchain/obc-peer/openchain/consensus/obcpbft.(_obcExecutor).queueThread(0xc8203d5440)
        /opt/gopath/src/github.com/openblockchain/obc-peer/openchain/consensus/obcpbft/obc-executor.go:301 +0x13ef
created by github.com/openblockchain/obc-peer/openchain/consensus/obcpbft.NewOBCExecutor
        /opt/gopath/src/github.com/openblockchain/obc-peer/openchain/consensus/obcpbft/obc-executor.go:108 +0x5d4

vp1 has gRPC timeout:
20:03:26.723 [peer] handleChat -> ERRO 003 Error during Chat, stopping handler: rpc error: code = 13 desc = "transport is closing"
[...]

PR #909, opened 2016-03-30
Nil execinfo fix for issue #879

jyellick:
This is a relatively small changeset.  It includes a test which reproduces the error seen in issue #879, as well as a fix, and a warning about a misconfiguration which would likely lead to this error.


jyellick:

All go tests pass
All behave tests pass

Issue #888, opened 2016-03-30
Incorporation of EnrollmentID into TCerts


dulcep:
Currently TCert structure does not include “EnrollmentID”. Encrypted EnrollmentID need to be added to the TCert. They are encrypted to support privacy and un-linkability.

binhn:
@diegomasini Let's separate out the encryption of the attributes into another issue. We need the attributes in first.

diegomasini:
@binhn ok, that sounds good, we will do that.

PR #911, opened 2016-03-30
Add the EnrollmentID into TCerts

diegomasini:
@binhn this is the initial implementation to incorporate the EnrollmentID into TCerts (see #888 created by @dulcep).


binhn:
Closed due to incomplete and we have to move the code to Hyperledger. Please rework on the new repo

Issue #788, opened 2016-03-04
inconsistent naming of chaincode transaction types across layers


muralisrini:
In Transaction proto
        deploy is CHAINCODE_NEW 
        invoke is CHAINCODE_EXECUTE
        query is CHAINCODE_QUERY

In OpenchainMessage proto
        deploy and invoke are CHAIN_TRANSACTION
        query is CHAIN_QUERY

In Chaincode interface
       deploy and invoke call Run 
        // Query is to be used for read-only access to chaincode state
       query calls Query

We need to provide meaningful and consistent names to transaction types.

binhn:
I would propose the following changes to Transaction:

```
Transaction
    enum Type {
        UNDEFINED = 0;

// deploy a chaincode to the network and call `init` function
        CHAINCODE_DEPLOY = 1;  

// call a chaincode `invoke` function as a transaction
        CHAINCODE_INVOKE = 2;  

// call a chaincode `query` function
        CHAINCODE_QUERY = 4;  

// terminate a chaincode; chaincode no longer accepts tx
        CHAINCODE_TERMINATE = 6;  
    }
```

binhn:
The `OpenchainMessage` should abstract the message (eg Transaction) such that handler may look at the payload to further discriminate.

```
message OpenchainMessage {
    enum Type {
        ...
        CHAIN_TRANSACTION = 6;
        ...
```

This means we need to change consensus framework event handler.

The chaincode interface should define functions according to the transaction types, including:
- Init (called as part of deploy)
- Invoke
- Query

muralisrini:
The original PR is closed but not the issue. Reopening the issue. Related new PR is #1012

binhn:
pr #1012 has been merged, so close this issue

Issue #581, opened 2016-02-01
deploy transaction with same parameters should return error


muralisrini:

corecode:
Why?  isn't it conceivable that you want to run two separate chaincode instances that are created the same way?

muralisrini:
We need to separate instance (not there today) and deploy (which is a one time operation). Semantics of deploy is "one time" (it always returns the same "transaction id" as stored on the ledger).

Also when chaincode instance is implemented, each instance would need to be identified uniquely as well - but that's a different discussion.

binhn:
@muralisrini I tried and succeeded in deploying a chaincode twice:

```
vagrant@vagrant-ubuntu-trusty-64:/opt/gopath/src/github.com/openblockchain/obc-peer$ OPENCHAIN_PEER_ADDRESS=10.0.2.15:30303 ./obc-peer chaincode deploy -p github.com/openblockchain/obc-peer/openchain/example/chaincode/chaincode_example02 -c '{"Function":"init", "Args":["a", "1", "b", "1"]}' --logging-level=info
15:01:51.623 [main] chaincodeDeploy -> INFO 001 Deploy result: type:GOLANG chaincodeID:<path:"github.com/openblockchain/obc-peer/openchain/example/chaincode/chaincode_example02" name:"bddbf5b909a9eacf97bfeaba9991b04a36dd5e437167f6bda6fbc0add4641d81b7eba987236acdf1a73d6742af1a0ada2b24b86123141708281504cd241fdbb8" > ctorMsg:<function:"init" args:"a" args:"1" args:"b" args:"1" > 
vagrant@vagrant-ubuntu-trusty-64:/opt/gopath/src/github.com/openblockchain/obc-peer$ OPENCHAIN_PEER_ADDRESS=10.0.2.15:30303 ./obc-peer chaincode deploy -p github.com/openblockchain/obc-peer/openchain/example/chaincode/chaincode_example02 -c '{"Function":"init", "Args":["a", "1", "b", "1"]}' --logging-level=info
[...]

PR #921, opened 2016-03-31
Code changes for the issue #788

ratnakar-asara:
Code changes  for issue #788 "inconsistent naming of chaincode transaction types across layers"

### DIFF
--- a/openchain/api.go
+++ b/openchain/api.go
@@ -104,7 +104,7 @@ func (s *ServerOpenchain) GetBlockByNumber(ctx context.Context, num *pb.BlockNum
 	// individual transaction.
 	blockTransactions := block.GetTransactions()
 	for _, transaction := range blockTransactions {
-		if transaction.Type == pb.Transaction_CHAINCODE_NEW {
+		if transaction.Type == pb.Transaction_CHAINCODE_DEPLOY {
 			deploymentSpec := &pb.ChaincodeDeploymentSpec{}
 			err := proto.Unmarshal(transaction.Payload, deploymentSpec)
 			if err != nil {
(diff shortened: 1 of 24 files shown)


ratnakar-asara:
Signed-off-by: rasara@us.ibm.com

srderson:
@muralisrini @ratnakar-asara First, thanks for this PR. This is a much needed change. I notice there are changes in openchain.proto in the transaction type such as `CHAINCODE_QUERY = 4;` to `CHAINCODE_QUERY = 3;`. This would normally break compatibility. Is this a problem or is it ok as long as you're also using an updated client?

@stanliberman This PR contains the Init function in a chaincode now so it's likely of interest to you.

I know one of Stan's requirements was a guarantee that the Init function is only run one. Is that the case here?

muralisrini:
@srderson : chaincode messages are just between the peer and chaincode (does not bleed into user apis)... so should be fine.

Currently we allow deploy to be called for the same chaincode any number of times (recorded in #581). When that gets fixed, init of a chaincode cannot be called multiple times.

muralisrini:
@srderson : as part of this work @ratnakar-asara had to make APIs such as the REST APIs compatible as well. So yes, as long as clients side are uptodate we should be fine.

@angrbrd : can you review to make sure please ?

As far calling init only once, there's issue #581 to prevent duplicate deploys. Once that's fixed, the init function can only be run once (at the first deploy).

muralisrini:
@ratnakar-asara : we will have to do this work on the Hyperledger codebase anyway. So please keep you branch around for us to continue there. Will close this PR on this repo so we can open a new one on Hyperledger.

muralisrini:
@stanliberman : we will continue with this PR under hyperledger github

PR #927, opened 2016-04-01
Replacing all instances of "server-host-override" with "serverhostoverride"

mrshah-at-work:
We cannot use hyphens in environment variables.

We can use OPENCHAIN_PEER_TLS_SERVERHOSTOVERRIDE instead of
OPENCHAIN_PEER_TLS_SERVER-HOST-OVERRIDE


mrshah-at-work:
Signed-off-by: Mihir Shah mrshah@us.ibm.com

srderson:
Reviewed and confirmed all tests are passing.

PR #942, opened 2016-04-03
move .sh files to scripts folder

binhn:


core/ledger/kvledger/kv_ledger_provider_test.go
```diff
@@ -525,13 +525,18 @@ func testConfig(t *testing.T) (conf *lgr.Config, cleanup func()) {
...
 
-func testutilNewProvider(conf *lgr.Config, t *testing.T) *Provider {
+func testutilNewProvider(conf *lgr.Config, t *testing.T, allowAnyCollection bool) *Provider {
 	cryptoProvider, err := sw.NewDefaultSecurityLevelWithKeystore(sw.NewDummyKeyStore())
 	assert.NoError(t, err)
 
+	ccInfoProvider := &mock.DeployedChaincodeInfoProvider{}
+	if allowAnyCollection {
+		ccInfoProvider.CollectionInfoReturns(&peer.StaticCollectionConfig{BlockToLive: 0}, nil)
+	}
```
manish-sethi: These hidden statements cause confusion later. I would prefer to set this behavior of mock in the specific tests.

core/ledger/kvledger/kv_ledger_provider.go
```diff
@@ -302,11 +334,12 @@ func (p *Provider) openInternal(ledgerID string) (ledger.PeerLedger, error) {
 		}
 	}
 
-	// Create a kvLedger for this chain/ledger, which encapsulates the underlying data stores
-	// (id store, blockstore, state database, history database)
+	// TODO: We have way more function arguments and it makes the code look ugly.
+	// Need to reduce it - FAB-17683
```
manish-sethi: I think that these type of generic TODO with in code does not add value. You can have the link to the code in the Jira instead. BTW, I don't see an easy way to reduce these without building further abstraction - which is exact opposite of what we are trying here.

core/ledger/kvledger/kv_ledger.go
```diff
@@ -37,20 +39,31 @@ var logger = flogging.MustGetLogger("kvledger")
...
+	txtmgmt                   txmgr.TxMgr
+	historyDB                 *history.DB
+	configHistoryRetriever    ledger.ConfigHistoryRetriever
+	blockAPIsRWLock           *sync.RWMutex
+	stats                     *ledgerStats
+	commitHash                []byte
+	isPvtstoreAheadOfBlkstore atomic.Value
+	// as the value is read during missing pvtData reconciliation
+	// and may be updated during a regular block commit, we use
+	// atomic value to ensure consistent read.
```
manish-sethi: move the comment up and start the standard way with `// isPvtstoreAheadOfBlkstore`

core/ledger/kvledger/kv_ledger.go
```diff
@@ -37,20 +39,31 @@ var logger = flogging.MustGetLogger("kvledger")
...
+}
+
+// TODO: we have used terms like blkstorage, blockstorage, blockstore. We need to use
+// a single term consistently. While typing code, it is getting harder to memorize what
+// term to be used when referring to the package or struct or a variable name - FAB-17683.
+
+// TODO: We have way more function arguments and it makes the code look ugly.
+// Need to reduce it - FAB-17683.
```
manish-sethi: Same as about the other TODO comments

core/ledger/kvledger/kv_ledger.go
```diff
@@ -149,8 +170,8 @@ func (l *kvLedger) initTxMgr(
 	return err
 }
 
-func (l *kvLedger) initBlockStore(btlPolicy pvtdatapolicy.BTLPolicy) {
-	l.blockStore.Init(btlPolicy)
+func (l *kvLedger) setBLTPolicyForPvtStore(btlPolicy pvtdatapolicy.BTLPolicy) {
+	l.pvtdataStore.Init(btlPolicy)
```
manish-sethi: You can remove this tiny function and call the actual function directly

core/ledger/kvledger/kv_ledger.go
```diff
@@ -63,11 +76,16 @@ func newKVLedger(
...
+	l := &kvLedger{
+		ledgerID:        ledgerID,
+		blockStore:      blockStore,
+		pvtdataStore:    pvtdataStore,
+		historyDB:       historyDB,
+		blockAPIsRWLock: &sync.RWMutex{},
+	}
 
 	btlPolicy := pvtdatapolicy.ConstructBTLPolicy(&collectionInfoRetriever{ledgerID, l, ccInfoProvider})
+	l.setBLTPolicyForPvtStore(btlPolicy)
```
manish-sethi: This is definately tempting to move this line before `l.initTxMgr`. But unfortunately, there is an implicit dependency here. The btlPolicy that you pass here internally uses queryexecuter and indirectly ends up using txmgr. This did not run into any error because, in the private data store function, this object is not accessed until at a later stage. But, since at this level, where we are wiring things together, its easier to understand and manage these initializing order than inside the individual components.

core/ledger/kvledger/kv_ledger.go
```diff
@@ -545,28 +619,51 @@ func (l *kvLedger) addBlockCommitHash(block *common.Block, updateBatchBytes []by
...
-	l.blockAPIsRWLock.RUnlock()
-	return blockAndPvtdata, err
+	defer l.blockAPIsRWLock.RUnlock()
+
+	var block *common.Block
+	var pvtdata []*ledger.TxPvtData
+	var err error
+
+	// TODO: Among retrieve, fetch, and get, we need to use a single
+	// term consistently - FAB-17684
```
manish-sethi: Same comment as I mentioned about other generic todos

core/ledger/kvledger/kv_ledger.go
```diff
@@ -545,28 +619,51 @@ func (l *kvLedger) addBlockCommitHash(block *common.Block, updateBatchBytes []by
 // GetPvtDataAndBlockByNum returns the block and the corresponding pvt data.
 // The pvt data is filtered by the list of 'collections' supplied
 func (l *kvLedger) GetPvtDataAndBlockByNum(blockNum uint64, filter ledger.PvtNsCollFilter) (*ledger.BlockAndPvtData, error) {
-	blockAndPvtdata, err := l.blockStore.GetPvtDataAndBlockByNum(blockNum, filter)
 	l.blockAPIsRWLock.RLock()
-	l.blockAPIsRWLock.RUnlock()
-	return blockAndPvtdata, err
+	defer l.blockAPIsRWLock.RUnlock()
```
manish-sethi, line 615: You don't need to do in this PR but just to take a note - In the existing code, the read lock pattern is different that may cause confusion. Better to convert the existing pattern to this one, which is more natural to understand. The current pattern was done keeping commit performance in mind, specially where block retrieving queries are more frequent... but we can always revert if we face some degradation.

core/ledger/kvledger/kv_ledger_test.go
```diff
@@ -567,6 +573,363 @@ func TestLedgerWithCouchDbEnabledWithBinaryAndJSONData(t *testing.T) {
...
+	}
+	blockNumAtCrash := dataAtCrash.Block.Header.Number
+	var pvtdataAtCrash []*ledger.TxPvtData
+	for _, p := range dataAtCrash.PvtData {
+		pvtdataAtCrash = append(pvtdataAtCrash, p)
+	}
+	// call Commit on pvt data store and mimic a crash before committing the block to block store
+	lgr.(*kvLedger).pvtdataStore.Commit(blockNumAtCrash, pvtdataAtCrash, nil)
+
+	// Now, assume that peer fails here before committing the transaction to the statedb and historydb
```
manish-sethi: "before committing the block to blockstore"

core/ledger/kvledger/kv_ledger_test.go
```diff
@@ -567,6 +573,363 @@ func TestLedgerWithCouchDbEnabledWithBinaryAndJSONData(t *testing.T) {
...
+	sampleData := sampleDataWithPvtdataForSelectiveTx(t, bg)
+	for _, d := range sampleData[0:9] { // commit block number 0 to 8
+		assert.NoError(t, kvlgr.commitToPvtAndBlockStore(d))
+	}
+
+	isPvtStoreAhead, err = kvlgr.isPvtDataStoreAheadOfBlockStore()
+	assert.NoError(t, err)
+	assert.False(t, isPvtStoreAhead)
+
+	// close and reopen. It mimics peer crash.
```
manish-sethi: It mimics peer crash -> this confused me for a moment. Though, it is correct from the point of view of the data not getting written to statedb but, nothing of that sort of thing is getting tested here. The tests that are moved here still have the flavor of testing of commits between the two store only. After some time, we may forget the scope of these tests and I guess that it would make sense to either have them in a separate test file or renaming the tests to makes the scope clear.

PR #958, opened 2016-04-04
update devenv to remove open blockchain

binhn:


pkg/config/policies.go
```diff
@@ -43,29 +43,13 @@ func (c *ConfigTx) UpdateConsortiumChannelCreationPolicy(consortiumName string,
...
-}
-
-// GetPoliciesForConsortium returns a map of policies for a specific consortium.
-func (c *ConfigTx) GetPoliciesForConsortium(consortiumName string) (map[string]Policy, error) {
-	consortium, ok := c.base.ChannelGroup.Groups[ConsortiumsGroupKey].Groups[consortiumName]
+// GetPoliciesForConsortiumOrg returns a map of policies for a specific consortium org.
+func (c *ConfigTx) GetPoliciesForConsortiumOrg(consortiumName, orgName string) (map[string]Policy, error) {
+	consortium, ok := c.updated.ChannelGroup.Groups[ConsortiumsGroupKey].Groups[consortiumName]
 	if !ok {
 		return nil, fmt.Errorf("consortium %s does not exist in channel config", consortiumName)
```
cendhu, line 50: Why do you return an `error` instead of empty map? Many databases return empty results when it does not find a matching result (some doesn't though). Even our `GetState()` chaincode API returns `nil` when the key does not exist. If `error` is must here, can't we define error type so that it is easy for the consumer to check what kind of error it is and ignore as needed? Suppose if I use the library to check whether a `consortiumName` exists, do I need to perform string matching on the error string?

pkg/config/policies.go
```diff
@@ -43,29 +43,13 @@ func (c *ConfigTx) UpdateConsortiumChannelCreationPolicy(consortiumName string,
...
-
-	return getPolicies(consortiums.Policies)
-}
-
-// GetPoliciesForConsortium returns a map of policies for a specific consortium.
-func (c *ConfigTx) GetPoliciesForConsortium(consortiumName string) (map[string]Policy, error) {
-	consortium, ok := c.base.ChannelGroup.Groups[ConsortiumsGroupKey].Groups[consortiumName]
+// GetPoliciesForConsortiumOrg returns a map of policies for a specific consortium org.
+func (c *ConfigTx) GetPoliciesForConsortiumOrg(consortiumName, orgName string) (map[string]Policy, error) {
+	consortium, ok := c.updated.ChannelGroup.Groups[ConsortiumsGroupKey].Groups[consortiumName]
```
DereckLuo: get policies for consortium org should probably retrieve policies from the base configuration instead of updated configuration.

PR #972, opened 2016-04-04
Update name of listenAddress property

cp68:
The listenAddress property in core.yaml was not being recognized because the code was looking for listenaddress with a lower case "a". This caused problems because I need to manually set the listen address to be a specific interface on my host and the "address" property to a different interface. There is a firewall in between two of my peers and they need to know the external address of each other, not the internal address.

### DIFF
--- a/main.go
+++ b/main.go
@@ -311,7 +311,7 @@ func serve(args []string) error {
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


cp68:
Signed-off by Chris Price (cp68@ntrs.com)

PR #977, opened 2016-04-05
move docker-compose files from root dir

tuand27613:
Move docker-compose config files from root dir to subdir of consensus/


tuand27613:

PR #988, opened 2016-04-05
Make config environment variable naming consistent

jyellick:
Per comments from @jeffgarratt all environment variables should be prefixed with `CORE_` not `HYPERLEDGER_`. 

The obcpbft and noops consensus plugins had their config prefix set to `HYPERLEDGER_` which was breaking the state transfer behave test.

All go tests except for 'github.com/hyperledger/fabric/core/chaincode' pass, which periodically fails, but also periodically fails in master.

All behave tests pass.


binhn:
@jyellick got conflick -- created and merged #997

PR #995, opened 2016-04-05
change core.yaml to asset.yaml in chaincode examples (try 2)

binhn:


binhn:

PR #1010, opened 2016-04-07
Remove and replace openchain

binhn:


binhn:

PR #1012, opened 2016-04-07
Chaincode Interface changes

ratnakar-asara:
- Chaincode interface has Init and Invoke methods instead of RUN
- Enum changes across layers

Signed-off-by: Ratnakar Asara rasara@us.ibm.com


binhn:
@ratnakar-asara pending document for the changes

PR #1017, opened 2016-04-07
Update protocol-spec and devnet to hyperledger

binhn:


binhn:

PR #1042, opened 2016-04-10
create a hook to allow mounting external directories

joequant:
This creates an environment hook to allow the user to mount an
external directory.  In the OBC workflow, the user could mount
an arbitrary directory into /openblockchain, however this is now
different as devenv has been merged with the fabric tree.  This
commit allows the user to add an external directory which
could have extra dev tools or test chaincode.

Signed-off-by: Joseph C Wang joequant@gmail.com

### DIFF
--- a/devenv/Vagrantfile
+++ b/devenv/Vagrantfile
@@ -15,6 +15,7 @@
 
 # This is the mount point for the sync_folders of the source
 SRCMOUNT = "/hyperledger"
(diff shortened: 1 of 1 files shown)


ghaskins:
+1  I have been meaning to submit a similar patch for a while now.  Thank you.

srderson:
LGTM. Thanks!

Issue #680, opened 2016-02-15
Unmarshaling error during state transfer SYNC_STATE_SNAPSHOT processing


tuand27613:
getting this message while initiating a state transfer request to other validating peers:
`
[33m20:19:56.769 [consensus/statetransfer] tryOverPeers -> WARN 719[0m name:"vp3"  in tryOverPeers loop trying name:"vp2"  : name:"vp3"  received a corrupt delta from name:"vp2"  : Error unmarashaling size: unexpected EOF
`

state transfer code will keep retrying, send out SYNC_STATE_GET_SNAPSHOT, get the same error and loop forever

tuand27613:
@jeffgarratt  taking a look.

tuand27613:
not sure if it's related but look at #686

tuand27613:
@jeffgarratt  @jyellick ,   not seeing the error when running with security.enabled=false

jyellick:
@tuand27613 Great work, I think that narrows down the problem considerably.  This means that state transfer without security completes successfully?

@srderson It seems that something in the state delta is encrypted before marshaling when security is enabled, causing unmarshaling to fail?  Do you have any thoughts or insight into what that might be?

tuand27613:
@jyellick  @jeffgarratt   I take it back.  got the latest code from obc-peer  including PR 677,681, 690 , run again with security.enabled=false  and I _still_ get the `unmarshalling unexpected EOF` error

jeffgarratt:
@jyellick @tuand27613 @srderson It appears to be working as designed, the last message will have an empty slice for the Delta field, indicating end of snapshot.  Be sure to check the length prior to attempting to Unmarshal with state_deltas.Unmarshal.  I will update the comment to be more explicit about the mechanism of indicating termination.

jyellick:
@jeffgarratt Yes, thanks for figuring this out, this was not how our mock implementation functioned which is why we did not catch it in earlier testing.  I'll fix that up now and put a PR together.

jyellick:
@tuand27613 You should be able to verify that this issue is fixed now.  Simply execute the new behave tests without decomposing.  If you grep the vp3 log for `obcpbft\|statetransfer` you will see state transfer be initiated, transfer blocks, then state deltas, then finish after transfering to block 8, and pbft will execute the outstanding 9th request.

tuand27613:
state transfer working for batch/classic/sieve

[...]

PR #1044, opened 2016-04-10
DEVENV: Rebrand vagrant environment to hyperledger

ghaskins:
This rebrands the vagrant environment from "obc" to "hyperledger".

We have published the base image as hyperledger/fabric-baseimage.  Note that the docker environment still remains as "openblockchain" for now because we lack access to the dockerhub account that controls this namespace.  Docker will be updated in a future patch once this is resolved.

Passes all unit-tests.  Behave tests all pass except #680.  However, this test has been intermittently failing for some time and not believed to be related to this commit.


ghaskins:
@srderson @binhn FYI

@jyellick #680 still consistently fails for me (with or without this patch).  Let me know if there is some information I can gather for you.

srderson:
LGTM. Thanks!

As an additional data point, I ran the behave tests with this PR and #680 passed for me, but it sounds as though investigation is needed.

Issue #610, opened 2016-02-05
The peer.gomaxprocs property in yaml not used


binhn:
In openchain.yaml, we define `peer.gomaxprocs: 2`, but in the code, we are not using it. This is important for performance analysis to find the best profile for obc-peer.

```
func main() {
    runtime.GOMAXPROCS(2)
...

    viper.BindPFlag("peer_gomaxprocs", flags.Lookup("peer-gomaxprocs"))
```

binhn:
@adecaro FYI, we need the same config for obc-ca

adecaro:
@binhn Addressed in obc-ca with PR #628

bcbrock:
I would request that when this support is added, the default should follow the current GOLANG defaults, that is, unless otherwise overridden by the user the code should not set GOMAXPROCS, but simply allow the system to use its default value. Currently this means that the obc-peer will be able to use as many threads as there are CPUs on the system.

I'm also saying please don't override this in the default YAML file (but maybe put a comment demonstrating how to do it).

adecaro:
@binhn, I checked and I can confirm that the property is used by the membership service.

binhn:
@bcbrock @adecaro I set the config to -1 to use the system default setting, which is the same as the number of available cores.

PR #1058, opened 2016-04-11
change gomaxprocs to system default -- issue #610

binhn:


binhn:

binhn:
@srderson @hyperledger/fabric-maintainers please review and merge

srderson:
Do we need to have this option in the yaml? Maybe I misunderstood, but I thought that @bcbrock was suggesting that we just remove it completely and let the user use the standard GOMAXPROCS environmental variable.

bcbrock:
@srderson @binhn - Sheehan, you're right, I think we should remove every reference to 'gomaxprocs' from the code, docs. and config files, since this setting can also be controlled directly using the GOMAXPROCS environment variable.

I think Binh's approach would be my second choice. However, note that in order to standardize around a way of command processing we would need for the code that calls runtime.GOMAXPROCS() to first check to see if the user had actually defined the env. variable GOMAXPROCS and use that in preference to the YAML setting, as is the case for config. options processed by VIPER. This would need to be done here and in the 'membersrvc' server.

So to avoid confusion I think it is simplest to eliminate gomaxprocs everywhere and let anyone who needs to modify GOMAXPROCS simply set the environment variable. 

I don't think it was discussed in the original issue, but I _think_ the reason that this crept into the codebase is that when you guys started, Go 1.4 was defaulting to using a single thread (GOMAXPROCS = 1), so you probably added this as a necessary way to enable multithreading. I would be surprised if anyone could make the case now for why they _needed_ to control this in a config file.

binhn:
@srderson @bcbrock I thought about removing it, and in fact @jeffgarratt suggested the same thing, but I found env vars are cumbersome to express on 1 line as how we do CLI today, especially on some OS platform. This approach would allow controlling GOMAXPROCS from the cmd line without having to set env vars. And env var technique is still available if the user wants to since the negative value in the yaml wouldn't effect any other settings.

bcbrock:
@binhn When you talk about expressing GOMAXPROCS on the command line, are you assuming that the options found in YAML files automatically become command-line flags? This isn't true (See #180).

In any event I think your solution is OK.

srderson:
LGTM. Based on discussion sounds like there's agreement to adopt this approach.

joshhus:
Can we adopt some method for flagging resolved "issues" that need a doc fix also. Or is there one. I seem to be unable to assign an issue or to add a label to an issue.

christo4ferris:
@joshhus seems to me that simple fixes, such as this to change the default value for a property, should be done by the developer. Same for tests, if you make a code change, the tests should be changed accordingly, and if none exists, tests should be added by the OP.

joshhus:
Agree. Don't want to lose track of any larger issues that a customer could run into / i.e. can't be fixed internally.

PR #1068, opened 2016-04-12
Update utxo's readme

noamher:


noamher:
Signed-off-by: noam@digitalasset.com

srderson:
LGTM. Thanks!

cmd/configtxgen/main_test.go
```diff
@@ -77,7 +78,20 @@ func TestMissingConsortiumValue(t *testing.T) {
...
+
+func TestUnsuccessfulChannelTxFileCreation(t *testing.T) {
+	configTxDest := filepath.Join(tmpDir, "configtx")
+
+	config := genesisconfig.Load(genesisconfig.SampleSingleMSPChannelProfile, configtest.GetDevConfigDir())
+	ioutil.WriteFile(configTxDest, []byte{}, 0440)
+
+	err := doOutputChannelCreateTx(config, nil, "foo", configTxDest).Error()
+	assert.Error(t, doOutputChannelCreateTx(config, nil, "foo", configTxDest), "Not able to write to file")
+	assert.Regexp(t, regexp.MustCompile("Error writing channel create tx: open /tmp/configtxgen.+/configtx: permission denied"), err)
```
wlahti: This works in CI but won't work locally depending on the developer's OS (i.e. it fails for me on mac).

Also a nit: you can leave off regexp.MustCompile

cmd/configtxgen/main_test.go
```diff
@@ -77,7 +78,20 @@ func TestMissingConsortiumValue(t *testing.T) {
...
 
-	assert.Error(t, doOutputChannelCreateTx(config, nil, "foo", configTxDest), "Missing Consortium value in Application Profile definition")
+	assert.EqualError(t, doOutputChannelCreateTx(config, nil, "foo", configTxDest), "config update generation failure: cannot define a new channel with no Consortium value")
+}
+
+func TestUnsuccessfulChannelTxFileCreation(t *testing.T) {
+	configTxDest := filepath.Join(tmpDir, "configtx")
+
+	config := genesisconfig.Load(genesisconfig.SampleSingleMSPChannelProfile, configtest.GetDevConfigDir())
+	ioutil.WriteFile(configTxDest, []byte{}, 0440)
```
wlahti: Just to be safe, I think we should check the error here.

cmd/configtxgen/main_test.go
```diff
@@ -77,7 +78,20 @@ func TestMissingConsortiumValue(t *testing.T) {
...
+
+	err := doOutputChannelCreateTx(config, nil, "foo", configTxDest).Error()
+	assert.Error(t, doOutputChannelCreateTx(config, nil, "foo", configTxDest), "Not able to write to file")
+	assert.Regexp(t, regexp.MustCompile("Error writing channel create tx: open /tmp/configtxgen.+/configtx: permission denied"), err)
+
+	os.Remove(configTxDest)
```
wlahti: I think we should use a defer for this remove call near where we create the file. That's the pattern we typically use in fabric.

PR #1082, opened 2016-04-13
Fix reply of SYNC_STATE_GET_SNAPSHOT

yeasy:
The reply of SYNC_STATE_GET_SNAPSHOT should be SYNC_STATE_SNAPSHOT.


srderson:
LGTM. Thanks.

PR #1089, opened 2016-04-13
Fix the GetRows API. Resolves #1088

### LINKED ISSUE #1088: Chaincode API GetRows fails when table contains only a single column
1. Create a table with a single column that is the key.
2. From a chaincode, call the `GetRows` API and include a valid key at the parameter.

Expected: A single row should be returned

Actual: No rows are returned

srderson:
Fixes GetRows API for tables with a single column. Resolves issue #1088

### DIFF
--- a/bddtests/peer_basic.feature
+++ b/bddtests/peer_basic.feature
@@ -424,7 +424,7 @@ Feature: lanching 3 peers
 			|arg1|arg2|arg3|
 			| a  | b  | 20 |
 	    Then I should have received a transactionID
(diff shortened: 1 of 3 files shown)


lehors:
@srderson line 408 of table.go reads:
            return nil, errors.New("getRowTableTwo failed. Must include 1 key")

I think this should say getRowTableFour instead of getRowTableTwo.

srderson:
@lehors Thanks, I've fixed the error message.

Issue #617, opened 2016-02-06
Can't script peer login


bcbrock:
Ideally a script like this should work:

```
echo -n NPKYL39uKbkj | OPENCHAIN_PEER_ADDRESS=localhost:30303 ./obc-peer login lukas
```

Instead it crashes (both with and without the -n option to echo)

```
Enter password for user 'lukas': panic: inappropriate ioctl for device

goroutine 1 [running]:
github.com/openblockchain/obc-peer/vendor/github.com/howeyc/gopass.getch(0x0)
    /gsa/ausgsa/home/b/c/bcbrock/obc/code/src/github.com/openblockchain/obc-peer/vendor/github.com/howeyc/gopass/nix.go:13 +0x73
github.com/openblockchain/obc-peer/vendor/github.com/howeyc/gopass.getPasswd(0x0, 0x0, 0x0, 0x0)
    /gsa/ausgsa/home/b/c/bcbrock/obc/code/src/github.com/openblockchain/obc-peer/vendor/github.com/howeyc/gopass/pass.go:18 +0x12e
github.com/openblockchain/obc-peer/vendor/github.com/howeyc/gopass.GetPasswd(0x0, 0x0, 0x0)
    /gsa/ausgsa/home/b/c/bcbrock/obc/code/src/github.com/openblockchain/obc-peer/vendor/github.com/howeyc/gopass/pass.go:37 +0x2d
main.login(0xc8202744c0, 0x1, 0x1, 0x7f8f0dde9ac8, 0xc82024f860)
    /gsa/ausgsa/home/b/c/bcbrock/obc/code/src/github.com/openblockchain/obc-peer/main.go:492 +0x4e1
main.glob.func8(0x135da60, 0xc8202744c0, 0x1, 0x1, 0x0, 0x0)
    /gsa/ausgsa/home/b/c/bcbrock/obc/code/src/github.com/openblockchain/obc-peer/main.go:115 +0x41
github.com/openblockchain/obc-peer/vendor/github.com/spf13/cobra.(*Command).execute(0x135da60, 0xc820274420, 0x1, 0x1, 0x0, 0x0)
    /gsa/ausgsa/home/b/c/bcbrock/obc/code/src/github.com/openblockchain/obc-peer/vendor/github.com/spf13/cobra/command.go:497 +0x63b
github.com/openblockchain/obc-peer/vendor/github.com/spf13/cobra.(*Command).Execute(0x135d360, 0x0, 0x0)
[...]

PR #1090, opened 2016-04-13
Fixes issue #617

genggjh:


genggjh:
First patch is using to update vender code to fix panic issue.
Second patch is using to add a new flag '--password' when run login command.

bcbrock:
@genggjh Hi David - I'm not sure why you needed to pass the command object into the login routine, none of the other routines do this even though they handle flags.

Also a typo:
`
The password for user. You will be requested to enter the password if this flag does not specified.
`
should be
`
The password for user. You will be requested to enter the password if this flag IS not specified.
`

genggjh:
@bcbrock , thanks for your comment, I will fix the typo and refine the code to get the password from global variable.

bcbrock:
@genggjh Also, please add some brief documentation of the feature in Section 6.1.3.2 of docs/protocol-spec.md.

genggjh:
@bcbrock I tried to find out where to doc this new argument, now you did it, thanks.

bcbrock:
This looks OK to me, assuming it works. I see that there are no tests for _any_ of the CLI functions in this repo (@angrbrd @bmos299 ?) so I don't think we need to require @genggjh to provide one.

srderson:
LGTM. Confirmed tests are passing.

@angrbrd @bmos299 - Let me know if I should open an issue for CLI tests or if we already have one.

Issue #920, opened 2016-03-31
PBFT: queue full


dongmingh:
code level:
commit 6ef721d1ad93c786db37ae6ee44cd2623dfca1d8
Merge: 0b3f7b1 ba5fffe
Author: Sheehan Anderson sranderson@gmail.com
Date:   Wed Mar 30 18:28:33 2016 -0400

```
Merge pull request #914 from muralisrini/convert_lang_to_upper_/issue#912

convert chaincode lang arg toupper
```

setup:
1. 4 docker-composer with classic PBFT, security enabled
2. login lukas
3. deploy example02
4. use SoapUI to generate invokes transaction with threads=100
5. the transaction halts with the following message on vp0:

20:13:42.216 [consensus/obcpbft] Execute -> ERRO 002 name:"vp0"  error queueing request (queue full) for sequence number 78
20:13:42.223 [consensus/obcpbft] Execute -> ERRO 003 name:"vp0"  error queueing request (queue full) for sequence number 79
20:13:51.093 [consensus/statetransfer] stateThread -> ERRO 004 name:"vp0"  could not retrieve blocks as recent as 80 as the block hash advertised

jyellick:
@dongmingh Could you provide logs at DEBUG level for all replicas?

jyellick:
This message has also been seen by @bcbrock in issue #879, what I believe is occurring here, is that a PBFT replica falls behind, seeing that the rest of the network has moved on to processing requests outside of its sliding watermark window, and initiates state transfer.  Before state transfer can complete, the network attempts to queue more transactions than the queue can accommodate, this is not a bug, but rather a side effect of the network being so busy. That error message should probably be reduced to a warning message.

The real problem here, is is likely related to `ERRO 004 name:"vp0" could not retrieve blocks as recent as 80 as the block hash advertised`.  Once PBFT advertises a block hash, it should be available on the network, and this will be the issue that needs investigating.

dongmingh:
This issue seems to be intermittent.  I am unable to recreate it this morning when attempting to collect logs for @jyellick 
I did not see the issue yesterday morning when verifying the fix of #879.  Then the issue occurred last afternoon in my test.
If I see the issue again, I will capture logs and attach here.

tuand27613:
@dongmingh  any luck ?

bcbrock:
@tuand27613 @jyellick I am seeing this in this release:

```
commit 92cc78029d04736928c3c07b7c3336023de2d747
[...]

Issue #793, opened 2016-03-04
Travis CI only: raise Exception("Max time exceeded waiting for transactions with current response map = {0}"


rameshthoomu:
**Failing scenarios:**

peer_basic.feature:391  chaincode example02 with 4 peers and 1 obcca, issue #567 -- @1.1 Consensus Options
peer_basic.feature:477  chaincode example02 with 4 peers and 1 obcca, issue #680 (State transfer) -- @1.3 Consensus Options

0 features passed, 1 failed, 0 skipped
20 scenarios passed, 2 failed, 1 skipped
470 steps passed, 2 failed, 36 skipped, 0 undefined
Took 10m5.521s
The command "behave" exited with 1.

Above mentioned two behave test scenarios are failing in outside of vagrant environment (Using Travis) and observed @1.1 Consensus Options test scenario failed in inside the vagrant environment. 

Attached are the log files:

[peer_basic feature 477.txt](https://github.com/openblockchain/obc-peer/files/159442/peer_basic.feature.477.txt)
[peer_basic features 391.txt](https://github.com/openblockchain/obc-peer/files/159443/peer_basic.features.391.txt)

Full log file from Travis:
[Behave script_Failed.txt](https://github.com/openblockchain/obc-peer/files/159445/Behave.script_Failed.txt)

Full log file from Inside vagrant env:
[inside_vagrant_log_file.txt](https://github.com/openblockchain/obc-peer/files/159453/inside_vagrant_log_file.txt)

rameshthoomu:
Updated bug description

tuand27613:
reading through the log files, I'm seeing things such as
in 477.txt
- line 269 `ERRO[1108] HTTP Error                                    err=no such id: dev-vp0-bb540edfc1ee2ac0f5e2ec6000677f4cd1c6728046d5e32dede7fea11a42f86a6943b76a8f9154f4792032551ed320871ff7b7076047e4184292e01e3421889c statusCode=404`
- line 509 `Exception: Max time exceeded waiting for transactions with current response map = {'bddtests_vp0_1': 200, 'bddtests_vp1_1': 200, 'bddtests_vp2_1': 200, 'bddtests_vp3_1': 404}`

in 391.txt
[...]

Issue #699, opened 2016-02-18
Classic PBFT hangs under stress


bcbrock:
Observed in

```
commit 7abd041f5b9e5ce346d1608f26b6880f6e235a77
Merge: 5077256 4fd8ae5
Author: Binh Q Nguyen <binhn@us.ibm.com>
Date:   Wed Feb 17 13:29:30 2016 -0500

    Merge pull request #682 from adecaro/crash_issue669

    Fix for issue #669
```

Same setup and irritation as in #687. I have a 4-peer docker-compose network running classic PBFT using all default parameters. I send 1000 transactions as fast as possible. The good news is that the network usually does not crash (#696 is rare) and eventually arrives at the correct answer. The issue is that it takes 8 or 9 minutes after I send in the last transaction for the database to correctly record the last transaction. 

You can see in this trace that the network is only processing 20 or 40 transactions every 30 seconds after the last transaction has been sent over the REST API. Note that security is _not_ enabled in this network (#684).

```
[bcbrock@arlab178]$ driveEx02 172.17.0.1 -rest -n 1000 -wait 30 -reCheck
Waiting 30 seconds for chain state to stabilize
 A: Got -638, expected -1000 B: Got 638, expected 1000
Waiting 30 seconds for chain state to stabilize
 A: Got -658, expected -1000 B: Got 658, expected 1000
Waiting 30 seconds for chain state to stabilize
 A: Got -678, expected -1000 B: Got 678, expected 1000
Waiting 30 seconds for chain state to stabilize
 A: Got -698, expected -1000 B: Got 698, expected 1000
Waiting 30 seconds for chain state to stabilize
 A: Got -738, expected -1000 B: Got 738, expected 1000
Waiting 30 seconds for chain state to stabilize
 A: Got -758, expected -1000 B: Got 758, expected 1000
Waiting 30 seconds for chain state to stabilize
 A: Got -778, expected -1000 B: Got 778, expected 1000
Waiting 30 seconds for chain state to stabilize
 A: Got -798, expected -1000 B: Got 798, expected 1000
Waiting 30 seconds for chain state to stabilize
 A: Got -838, expected -1000 B: Got 838, expected 1000
Waiting 30 seconds for chain state to stabilize
 A: Got -858, expected -1000 B: Got 858, expected 1000
Waiting 30 seconds for chain state to stabilize
 A: Got -878, expected -1000 B: Got 878, expected 1000
Waiting 30 seconds for chain state to stabilize
[...]

PR #1094, opened 2016-04-13
Fix for issue #920

### LINKED ISSUE #917: PBFT Classic Consensus seems to keep changing views
Description:

PBFT Classic seems to keep changing views, it went from view 4 to view 17 in a span of 10 minutes

Steps to repro:
1.  Setup pbft classic on a laptop
2.  Run Deploy, Query
3.  Run GO SDK tests

Results:

Primary logs have gone through 13 view changes in that time

Expected Results:

Primary and VPs should not be constantly changing views, as they are unable to process transactions when voting on a new primary (view change)

[vp0 View Mania.txt](https://github.com/openblockchain/obc-peer/files/197552/vp0.View.Mania.txt)

jyellick:
This is designed to address issue #920, but likely also fixes #917, #793, and #699 

There was a bug that was introduced when splitting out the executor service.  PBFT would send its checkpoints based on the last executed sequence number.  This assumption used to be correct when executions were synchronous, but no longer.  Under low load environments, typically this problem would never be observed, but when message processing begins to occur faster than executions, the problem occurs.

This changeset causes checkpoints to use the sequence number provided by the executor, not the one stored internally by pbft and fixes some sequence number accounting that was incorrect in the executor.

This changeset also reduces the severity of the queue full message to warning.

Passes all go tests.
Passes all behave tests.

### DIFF
--- a/consensus/obcpbft/obc-executor.go
+++ b/consensus/obcpbft/obc-executor.go
@@ -279,8 +279,9 @@ func (obcex *obcExecutor) queueThread() {
 
 		var err error
 		var id *BlockInfo
-		var seqNo uint64
 		var sync bool
+
+		seqNo := transaction.seqNo
 
 		if nil == transaction.execInfo {
 			transaction.execInfo = &ExecutionInfo{}
@@ -293,7 +294,6 @@ func (obcex *obcExecutor) queueThread() {
 			logger.Info("%v executor queue apparently has a gap in it, initiating state transfer", obcex.id)
 			sync = true
 		case execInfo.Null:
-			seqNo = transaction.seqNo
 			id, err = obcex.getCurrentInfo()
 			if nil != err {
 				logger.Critical("Requested to send a checkpoint for a Null request, but could not create one: %v", err)
--- a/consensus/obcpbft/obc-executor.go
+++ b/consensus/obcpbft/obc-executor.go
@@ -149,7 +149,7 @@ func (obcex *obcExecutor) Execute(seqNo uint64, txs []*pb.Transaction, execInfo 
 	case obcex.executionQueue <- request:
 		logger.Debug("%v queued request for sequence number %d", obcex.id, seqNo)
 	default:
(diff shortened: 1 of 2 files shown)


srderson:
@jyellick Is it possible to add a test to check for this issue or is it difficult to reproduce?

srderson:
I can also confirm that all unit and behave tests are passing with this PR.

jyellick:
@srderson Ordinarily I would have included a test, but this is essentially a race between an executor counter and a pbft counter, which stay in sync unless the system is under stress.  The bug is using the pbft counter when it is the executor counter which is correct.  To write a test case, it would essentially need to rely on a long sleep or similar, or require more extensive modifications to allow the unit tests to hook into the execution path more directly.  Because this is a simple logical error in picking the wrong counter (or rather, not picking the new counter after it was split into two), it seems like a regression is unlikely.

If you feel a unit test is still required for this changeset, I can write one up, just let me know.

joshhus:
Are we recording our troubleshooting finds and fixes for eventual publication / customers? The ones that could be potentially exposed ...

jyellick:
@joshhus These are documented via the github issue / PR system, did you have something else in mind?

joshhus:
Not really. Are the Troubleshooting issues labeled as "Documentation" or "Troubleshooting" so ID can find them / publish them as necessary.

jyellick:
@joshhus When I say these are documented in github, I mean that the problem that was encountered in test  is documented in the issue (for instance, for this PR, issue #920).  Then, once the cause of the problem is discovered and fixed, it documented in the accompanying PR.

Although I don't think anyone is out there running the fabric with PBFT in any sort of capacity in which they are not routinely updating the code, I suspect you could look for issues tagged as [bug](https://github.com/hyperledger/fabric/issues?q=is%3Aopen+is%3Aissue+label%3Abug) that you could use these for that purpose.

srderson:
@jyellick understand the lack of a test. LGTM

Issue #1080, opened 2016-04-13
more behave tests failure with latest hyperledger code


gongsu832:
commit 8580de69c9371e26342068746e6340169711a5dc

```
Failing scenarios:
  peer_basic.feature:420  chaincode example02 with 4 peers and 1 membersrvc, issue #567 -- @1.1 Consensus Options
  peer_basic.feature:421  chaincode example02 with 4 peers and 1 membersrvc, issue #567 -- @1.2 Consensus Options
  peer_basic.feature:422  chaincode example02 with 4 peers and 1 membersrvc, issue #567 -- @1.3 Consensus Options
  peer_basic.feature:423  chaincode example02 with 4 peers and 1 membersrvc, issue #567 -- @1.4 Consensus Options
  peer_basic.feature:504  chaincode example02 with 4 peers and 1 membersrvc, issue #680 (State transfer) -- @1.1 Consensus Options
  peer_basic.feature:505  chaincode example02 with 4 peers and 1 membersrvc, issue #680 (State transfer) -- @1.2 Consensus Options
  peer_basic.feature:506  chaincode example02 with 4 peers and 1 membersrvc, issue #680 (State transfer) -- @1.3 Consensus Options
  peer_basic.feature:555  chaincode example02 with 4 peers and 1 membersrvc, issue #724 -- @1.1 Consensus Options
  peer_basic.feature:633  4 peers and 1 membersrvc, consensus still works if one backup replica fails -- @1.1 Consensus Options
  peer_basic.feature:634  4 peers and 1 membersrvc, consensus still works if one backup replica fails -- @1.2 Consensus Options
  peer_basic.feature:635  4 peers and 1 membersrvc, consensus still works if one backup replica fails -- @1.3 Consensus Options
  peer_basic.feature:695  4 peers and 1 membersrvc, consensus fails if 2 backup replicas fail -- @1.1 Consensus Options
  peer_basic.feature:696  4 peers and 1 membersrvc, consensus fails if 2 backup replicas fail -- @1.2 Consensus Options
  peer_basic.feature:697  4 peers and 1 membersrvc, consensus fails if 2 backup replicas fail -- @1.3 Consensus Options
  peer_basic.feature:740  4 peers and 1 membersrvc, consensus still works if 1 peer (vp3) is byzantine -- @1.1 Consensus Options
[...]

PR #1095, opened 2016-04-13
Increase bddtest query timeout to 30; Fixes issue 1080

bcbrock:
@gongsu832 and I were seeing lots of timeouts while running queries during bddtests. See #1080. Simply increasing the timeout from "3" to "30" fixes these errors (I tried "10" - didn't work). This change does not increase the time required to run the bddtest suite from inside Vagrant.

Speaking for myself, I was seeing these issues in a standalone (non-Vagrant) system. I'm running on a beefy X86 server, and the bddtest suits takes about 4x the time required within Vagrant for some reason. So there may be some other issue at play, but this change seems to do no harm inside Vagrant, and seems to be required outside.


bcbrock:

srderson:
LGTM

Issue #1074, opened 2016-04-12
Chaincode install path is hard coded


harrijk:
The default chaincode `installpath` value is set to `/opt/gopath/bin` in the core.yaml file.  If one decides to change the `installpath`parameter to something other than the default value, the following error occurs when deploying chaincode:

`12:14:29.626 [container] start -> ERRO 04b start-could not start container API error (500): Cannot start container dev-jdoe-3a46d4d629d8c1f931bb29c1b52ca9698cba5d75487771882068975b15f0440fb73bab315f5dad72eaaaeaead5ea6a211ac013cfb8d55c64547d348d51504e7d: [8] System error: exec: /opt/gopath/bin/3a46d4d629d8c1f931bb29c1b52ca9698cba5d75487771882068975b15f0440fb73bab315f5dad72eaaaeaead5ea6a211ac013cfb8d55c64547d348d51504e7d: stat /opt/gopath/bin/3a46d4d629d8c1f931bb29c1b52ca9698cba5d75487771882068975b15f0440fb73bab315f5dad72eaaaeaead5ea6a211ac013cfb8d55c64547d348d51504e7d : no such file or directory`

When you change your default chaincode installpath within core.yaml, this also assumes that one has included the path within the Docker image used for chaincode operations.  When I built my base Docker image for chaincode use, I updated my PATH environment variable to include the chaincode installpath (/go/bin) set in my core.yaml file.

I took a look at the fabric core file `core/chaincode/chaincode_support.go` and noticed that on line 51 the constant `chaincodeInstallPathDefault` is set to `/opt/gopath/bin`.  Within the `NewChaincodeSupport` method on line 116 of chaincode_support.go, the `s.chaincodeInstallPath` is assigned the constant `chaincodeInstallPathDefault` (/opt/gopath/bin) rather than using the `installpath` value within the core.yaml file.

Just as a test, I went ahead and rebuilt the peer binary after updating  the `chaindeInstallPathDefault` constant in `chaincode_support.go` to match the `installpath` value used in my core.yaml file.  I then deployed chaincode and the error that I was experiencing previously did not occur.

Perhaps I'm missing something here... very possible. :-)

srderson:
@muralisrini can you comment on this issue?

muralisrini:
[...]

PR #1103, opened 2016-04-15
Let chaincodeInstallPath read from config

### LINKED ISSUE #1102: Chaincode config should be able to read from sys variable
Seems the chaincode config is hard coded in core/chaincode/chaincode_support.go (no find where it reads from the core.yaml yet).

``` sh
const (
    // DefaultChain is the name of the default chain.
    DefaultChain ChainName = "default"
    // DevModeUserRunsChaincode property allows user to run chaincode in development environment
    DevModeUserRunsChaincode       string = "dev"
    chaincodeStartupTimeoutDefault int    = 5000
    chaincodeInstallPathDefault    string = "/opt/gopath/bin/"
    peerAddressDefault             string = "0.0.0.0:30303"
)
s.chaincodeInstallPath = chaincodeInstallPathDefault
```

When using system variable like CORE_CHAINCODE_INSTALLPATH to setup new path, the chaincode module still use the default one.

I suggest we can solve this, and more optimization to let it read from relative sys path like $GOPATH/bin/ as default value.

yeasy:
Currently the chaincodeInstallPath directly uses the default hard-coded value. 

Add code to try reading it from config first.

This closes #1102.

### DIFF
--- a/core/chaincode/chaincode_support.go
+++ b/core/chaincode/chaincode_support.go
@@ -113,7 +113,10 @@ func NewChaincodeSupport(chainname ChainName, getPeerEndpoint func() (*pb.PeerEn
 	s.ccStartupTimeout = ccstartuptimeout * time.Millisecond
 
 	//TODO I'm not sure if this needs to be on a per chain basis... too lowel and just needs to be a global default ?
-	s.chaincodeInstallPath = chaincodeInstallPathDefault
+	s.chaincodeInstallPath = viper.GetString("chaincode.installpath")
+	if s.chaincodeInstallPath == "" {
+		s.chaincodeInstallPath = chaincodeInstallPathDefault
+	}
 
 	return s
 }


srderson:
Does this also resolve issue #1074? It looks like the same problem.

yeasy:
@srderson 
Yes, seems the same problem. 
Could you help add this patch to close #1074 too?
Thanks a lot!

muralisrini:
@yeasy : can you take a look at recent comments in #1074 and provide your view please ? If you agree that the right thing to do would be to remove the "installpath" option, perhaps we can modify your PR to do that ?

srderson:
@yeasy See https://github.com/hyperledger/fabric/issues/1074#issuecomment-211934650

yeasy:
@srderson
Sure, I will add those suggested comments in another patch.

srderson:
Confirmed all tests are passing. LGTM.

PR #1104, opened 2016-04-15
Fix deprecated DOCKER_OPTS

yeasy:
The option `--api-enable-cors=true` was deprecated at Docker 1.9+.

Use `--api-cors-header='*'` instead as recommended style.


srderson:
LGTM. Thanks

PR #1111, opened 2016-04-15
Fix view change deadlock, caused by confusion of S and S prime

jyellick:
In the view change code, the matching checkpoint sets S' was being used instead of the view change set S.

Because of this, view change was requiring 2f+1 low watermarks at or below the sequence number within the checkpoint set, when this needs to be considered over the whole view change set.  This could lead to a scenario where view change never completes, and the network locks up, view changing indefinitely.

This changeset fixes this behavior, and includes a test which verifies its correctness.

It also includes a fix for another observed deadlock in which the executor thread attempts to deliver a checkpoint while the pbft message thread attempts to initiate state transfer.

Finally, this changeset also causes the 'fall behind' recovery to mark the current view as active.  The correctness fo this should be verified with PBFT protocol experts, and it is flagged as such in the code, but otherwise could cause a replica to fail to rejoin a network if its view change timer expires before realizing it is behind.

All of these fixes have been verified to help address the stress related problems observed in #920, #917, #793, and #699 

Passes all go tests.
Passes all behave tests.

### DIFF
--- a/consensus/obcpbft/obc-executor.go
+++ b/consensus/obcpbft/obc-executor.go
@@ -320,7 +320,9 @@ func (obcex *obcExecutor) queueThread() {
 		if execInfo.Checkpoint {
 			idAsBytes, _ := obcex.createID(id)
 			logger.Debug("%v sending checkpoint: %x", obcex.id, idAsBytes)
-			obcex.orderer.Checkpoint(seqNo, idAsBytes)
+			go func() {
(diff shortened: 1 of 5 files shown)


srderson:
LGTM. Also confirmed unit and behave tests are passing.

kchristidis:
FWIW, I also confirm that the view-change modification makes sense.

PR #1126, opened 2016-04-17
change the slack url according to info from LF

binhn:
We are using the slack self-register URL


binhn:

srderson:
It looks like the space in `[Hyperledger Slack]( https://slack.hyperledger.org/)` is breaking the link. See preview at https://github.com/binhn/fabric/blob/bn-readme/README.md

binhn:
@srderson I removed the space. Thanks.

srderson:
LGTM. Thanks

Issue #1002, opened 2016-04-06
Incorporation of attributes into TCerts


diegomasini:
Currently TCert structure does not include “attributes”. Encrypted attributes need to be added to the TCert. They are encrypted to support privacy and un-linkability.

diegomasini:
@binhn, as discussed with you issue #888 is separated in two, the previous one for the inclusion of the EnrollmendID into TCerts and this one for the inclusion of attributes into TCerts.

Issue #889, opened 2016-03-30
Generation of keys used to encrypt EnrollmentID and Attributes


dulcep:
Generation of TCerts needs to include the generation of keys to encrypt EnrollmentID and attributes, such keys must be unique per TCert. Related to #888

ghaskins:
Is the design of the key agreement protocol available anywhere for review?

PR #1130, opened 2016-04-18
Incorporation of attributes (issue #1002) and EnrollmentID (issue #888) into TCerts

diegomasini:
@binhn these are the changes to address issues #1002 and #888. They also covers the generation of the keys needed to encrypt the attributes and the enrollment ID (issue #889) although the encryption can be disabled with a flag in the configuration file.

The unit tests and the manual tests are working. There is a problem with the behave tests and we couldn't find why some behave tests are failing. The number of tests failing is not deterministic across executions. Please don't merge until we find the reason why the behave tests are failing. I already contacted @muralisrini about this, I will talk to @jeffgarratt and probably he can help with this.

Mentioning @raullaprida and @andresgaragiola since they are working on this too trying to find out the problem with the tests.


diegomasini:
@binhn,@jeffgarratt Behave and Unit tests are working with the latest changes. @binhn the PR is ready to be merged.

Signed-off-by: Diego Masini diego.masini@gmail.com

PR #1136, opened 2016-04-18
Viper usage

manish-sethi:
@binh, @srderson  though in the ledger code, the viper configs were getting loaded once only except in the Genesis code (which anyway does not come in the request path). Still, I have separated this explicitly so it's clear to others packages.

### DIFF
--- a/core/ledger/genesis/config.go
+++ b/core/ledger/genesis/config.go
@@ -1,0 +1,66 @@
+/*
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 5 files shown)


manish-sethi:

binhn:
@manish-sethi I wonder if [init() should be used](https://golang.org/doc/effective_go.html#init) to initialize the constants. It sounds like it might be applicable.

manish-sethi:
@binhn I had thought about init() but the issue that I ran into was that init() is invoked before main() i.e., by the time init() is invoked, viper properties are not set. 
One of the solutions to this problem could have been to introduce a method (say initViper()) in a separate package (say config) and let all the init() methods call config.initViper() as a first thing. However, this would create issues in tests that want to load their own yaml files (This may be solved by passing yaml file name, path etc. as environment variables but that may lead to broken tests for many new users).
An alternate solution (that we discussed earlier) could be that each package can expose a method initConfig() and let main() invoke these methods (after setting the viper properties) but that would lead to main file being a frequently changing file and so I am not sure about this too as of now.

binhn:
@manish-sethi that's a good point on timing. I thought main.go would be loaded first prior to other files.

srderson:
LGTM. Thanks

PR #1138, opened 2016-04-18
Added clarifications

mcr222:
Added explanation on how to start peers with CA.

Added warning to avoid deleting all the docker images when setting up peers.

Signed-off-by: Marc Cayuela Rafols markitus222@hotmail.com


srderson:
LGTM

Issue #886, opened 2016-03-29
obc-peer stop command is not working


rameshthoomu:
I tried to stop obc-peer applying `obc-peer stop` command. After execute the command obc-peer status is showing as stopped when I execute

`obc-peer status` peer status is showing as `UNKNOWN` and 
`fuser -n tcp 5000` command resulting me that peer is still running. 

Attached snapshot of the error message.
![obc-peer status_stop](https://cloud.githubusercontent.com/assets/17034502/14110470/914a6782-f594-11e5-9729-26933848b9ba.PNG)

srderson:
Resolved via PR #1141

PR #1141, opened 2016-04-18
Fixing peer stop command - issue #886

gennadylaventman:
On running peer side:
1. Peer store its pid into peer.pid file in "peer.fileSystemPath"
2. Once ServerAdmin (admin.go) get StopServer, it removes pid file and executes os.Exit(0)
3. In case of GetStatus it return STARTED, based of that while peer process is still running and able to response to grpc - it seems OK

On another side:
1. Process try to connect to peer using grpc, as it was before. But one it expect to disconnect, as other side is exited.
2. If grpc doesn't work, it look for pid file and try to send SIGTERM to process.


gennadylaventman:
Signed-off-by: Gennady Laventman gennady@il.ibm.com

core/admin.go
```diff
@@ -59,15 +59,15 @@ func worker(id int, die chan struct{}) {
...
-	die := make(chan struct{})
-	log.Debug("Creating %d workers", viper.GetInt("peer.workers"))
-	for i := 0; i < viper.GetInt("peer.workers"); i++ {
-		go worker(i, die)
-	}
-	runtime.Gosched()
-	<-time.After(1 * time.Millisecond)
-	close(die)
+	status := &pb.ServerStatus{Status: pb.ServerStatus_STARTED}
+	//die := make(chan struct{})
```
srderson: Can this block of commented out code be removed?

gennadylaventman:
@srderson Yes, sure. I will remove it now and commit.

gennadylaventman:
@srderson Done.

srderson:
LGTM. Thanks

PR #1142, opened 2016-04-18
An id field in JSON-RPC request: allow to be both string or numeric

uzzz:
As for JSON-RPC spec **id** field in a request may be a string:

```
An identifier established by the Client that MUST contain a String, Number, or NULL value if included.
```

and some clients actually do send strings as IDs. This patch allows to unmarshal both int64 and string IDs. What do you think about it?

Signed-off-by: Maksim Zhylinski maksim.zhylinski@altoros.com

### DIFF
--- a/core/rest/rest_api.go
+++ b/core/rest/rest_api.go
@@ -21,6 +21,7 @@ package rest
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"google/protobuf"
 	"io"
@@ -73,7 +74,37 @@ type rpcRequest struct {
 	Jsonrpc *string           `json:"jsonrpc,omitempty"`
 	Method  *string           `json:"method,omitempty"`
 	Params  *pb.ChaincodeSpec `json:"params,omitempty"`
-	ID      *int64            `json:"id,omitempty"`
+	ID      *rpcID            `json:"id,omitempty"`
+}
+
+type rpcID struct {
+	StringValue *string
+	IntValue    *int64
+}
+
+func (id *rpcID) UnmarshalJSON(b []byte) error {
+	var err error
+	s, n := "", int64(0)
+
+	if err = json.Unmarshal(b, &s); err == nil {
+		id.StringValue = &s
+		return nil
+	}
+	if err = json.Unmarshal(b, &n); err == nil {
+		id.IntValue = &n
+		return nil
+	}
+	return fmt.Errorf("cannot unmarshal %s into Go value of type int64 or string", string(b))
+}
+
+func (id *rpcID) MarshalJSON() ([]byte, error) {
+	if id.StringValue != nil {
+		return json.Marshal(id.StringValue)
+	}
+	if id.IntValue != nil {
+		return json.Marshal(id.IntValue)
+	}
+	return nil, errors.New("cannot marshal rpcID")
 }
 
 // rpcResponse defines the JSON RPC 2.0 response payload for the /chaincode endpoint.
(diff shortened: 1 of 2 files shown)


angrbrd:
I will test this out. Thanks.

angrbrd:
Hi @uzzz. I think this looks good. Thank you for taking a look. The only thing I would add is that you should also update the rest_api.json for the Swagger to update the specification of the "id" field. Please do that and I will merge. Thanks.

"id": {
   "type": "integer",
   "format": "int64",
   "description": "An integer number used to correlate the request and response objects. If it is not included, the request is assumed to be a notification and the server will not generate a response."
}

uzzz:
It looks like swagger supports multiple types per property, but it does not support multiple formats & multiple default values. So I can't define something like this:

```
"id": {
  "type": ["integer", "string"],
  "format": ["int64", "string"],
  "default": [123, "123"]
}
```

angrbrd:
@uzzz, I know Swagger is not always perfect... that's OK with me :-]. But what if we do it like this? I think this is the best option we have:

https://github.com/angrbrd/fabric/commit/04f94dcba47fcdf226dc14e609eb2cb8e746316c

If you agree, we can take your code change and this Swagger change to complete this fix. Let me know what you think.

uzzz:
But it's not what API actually looks like. Here's a screenshot from the swagger editor:

screenshot 2016-04-22 12 13 29

You can see it generates incorrect request while correct would be

```
{
  ...
  "id": 123,
  ...
}
```

or

```
{
  ...
  "id": "one-two-three",
  ...
}
```

I would suggest keeping old documentation, at least it's correct. And user will get string id as a bonus :)

My motivation for making this PR was that I actually met JSON-RPC library implementation (and tried to use it with hyperledger) that was using string ids. So I checked with JSON-RPC spec and figured out that strings are valid too. So this patch won't break such client libraries, but for those who want to write their own requests and reading swagger docs it's OK if they will see only int parameter, I think.

srderson:
Sounds reasonable and LGTM. Confirmed that all tests are passing.

pkg/configtx/example_test.go
```diff
@@ -155,159 +123,123 @@ func Example_systemChannel() {
...
-		panic(err)
+		panic(nil)
 	}
 
-	// Signature and nonce is different on every example run
-	data.Signatures = nil
+}
+
+// This example shows the addition and removal of ACLs.
+func Example_aCLs() {
```
caod123, line 155: this looks odd here but I'm assuming it looks fine in the generated GoDocs

pkg/configtx/example_test.go
```diff
@@ -407,29 +322,6 @@ func Example_organization() {
...
-		panic(err)
-	}
-
-	err = c.RemoveOrdererEndpoint("OrdererOrg", configtx.Address{Host: "127.0.0.1", Port: 9050})
-	if err != nil {
-		panic(err)
-	}
 }
 
 func ExampleNewSystemChannelGenesisBlock() {
```
caod123, line 327: does this need to be Example_newSystemChannelGenesisBlock?

pkg/configtx/example_test.go
```diff
@@ -540,629 +432,10 @@ func ExampleNewSystemChannelGenesisBlock() {
...
-	//			},
-	//			"version": "0"
-	//		},
-	//		"sequence": "0"
-	//	},
-	//	"last_update": null
-	//}
 }
 
 func ExampleNewCreateChannelTx() {
```
caod123, line 441: does this need to be Example_newCreateChannelTx?

Issue #1119, opened 2016-04-15
Fix PBFT consensus to avoid normal case view changes.


jyellick:
I discussed this in slack with @corecode but I am observing behavior where PBFT is being forced into a view change when it should not be.

Consider the scenario when vp2 is the primary and vp3 is offline, K=2, L=4

vp0 - send checkpoints 2, 4
vp0 - receive vp0 checkpoint 2, 4
vp1 - send checkpoints 2, 4
vp1 - receive vp1 checkpoints 2, 4
vp2 - send checkpoints 2, 4
vp2 - receive vp0,vp1,vp2 checkpoints 2, and 4
vp0,1 - receive vp2 checkpoints 2, and 4 (do not move watermarks, only f+1)
vp2 - move watermarks and send pre-prepare 5
vp0,1 - receive and ignore vp2 pre-prepare 5
vp0 receive vp1 checkpoints 2, and 4, move watermarks
vp1 receive vp0 checkpoints 2, and 4, move watermarks

We are now in a scenario where vp0 and vp1 have discarded a pre-prepare message from vp2, so the network will cease to make progress until the view change timer expires, causing an (unnecessary) view change when the primary was not behaving in a byzantine fashion.

Although this problem seems like it would be mitigated by the log, it is still seen frequently in stress testing.
1. Primary orders L requests
2. Network reaches consensus on order before first request is executed
3. All requests are executed, triggering the transmission of the checkpoints in quick succession
4. Sometimes, primary is first to get the checkpoint quorum, it broadcasts the next pre-prepare
5. The backups receive this pre-prepare before the checkpoint quorum, and initiate view change

My suspicion is this is especially likely in docker-ized testing like we do with behave, because all replicas are competing for the same CPU resources, and one replica's threads being serviced necessarily causes other replicas to be idle, but even though this is likely less of a concern in production, it still seems like a scenario worth at least considering addressing.

In an experimental branch, I was able to decrease the frequency of this occurrence by increasing the log size from 2_K to 4_K, and modifying the behavior of the primary to only submit pre-prepares if the sequence number was going to be less than or equal to L-K.   In this way, the primary had some additional buffer on the other side to give the other replicas an opportunity to receive eachother's checkpoints.  Still, there is nothing preventing the failure from still occurring, as if checkpoint messages are sufficiently delayed between peers, the primary could still advance its watermarks twice and send a pre-prepare to a replica which had not yet advanced its watermarks once.

[...]

PR #1143, opened 2016-04-18
Implement issue #1119

jyellick:
This is a simple implementation of the enhancement suggested in #1119 

In short, view changes were unnecessarily occurring because of network ordering guarantees that are assumed in the PBFT paper, but which are not present in reality.  Although the absence of these guarantees does not affect the 'correctness' it has been observed to decrease the throughput.

This changeset increases the default log multiplier from 2, to 4.  Simultaneously, it reduces the sequence number at which the primary will pause sending requests from the high water mark `H = h + logMultiplier * K` to `H = h + logMultiplier * K / 2`.  This has the effect of in a sense keeping our 'send buffer' the same size, but doubling our 'receive buffer'.  In high volume testing, this has been shown to reduce the incidence of unnecessary view changes and increase the throughput of the network.

A test case is included to confirm the new primary request sending behavior.

If there are objections from @kchristidis @corecode or @tuand27613 I'd be open to making this configurable, but there was not an obvious way to make the configuration expressive (Is it a multiple of the checkpoint interval? Is it an absolute value?) and the potential for mis-configuring things in a way that would break seemed high.

This changeset also forces the `logMultiplier` to be at least 2, otherwise things would break.

All behave tests pass.
All go tests pass.

### DIFF
--- a/consensus/obcpbft/config.yaml
+++ b/consensus/obcpbft/config.yaml
@@ -25,9 +25,11 @@ general:
     # overall throughput in normal case operation.
     K: 10
 
-   # Affects the log size which is K * logmultiplier
-   # For high volume/high latency environments, a higher log size may increase throughput
-    logmultiplier: 2
+    # Affects the receive log size which is K * logmultiplier
+    # The primary will only send sequence numbers which fall within K * logmultiplier/2 of
+    # its high watermark, so this cannot be set to less than 2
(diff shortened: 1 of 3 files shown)


kchristidis:
Code reviewed, all looks good here.

srderson:
LGTM. Thanks

Issue #908, opened 2016-03-30
Allowing configuration to use SHA2


binhn:
Currently we only support SHA3 with configurable length. We need to also support SHA2, which can be configured as:

```
security:
    level: 256
        sha: 2
```

andresgaragiola:
@binhn I'm working in this issue

andresgaragiola:
@binhn After pull request #1148 this issue should be closed.

binhn:
closed by PR 1148

PR #1148, opened 2016-04-19
Fixed Issue #908

andresgaragiola:
Fixed issue #908.


andresgaragiola:
DCO1.1 Signed-off-by: Andres Garagiola andresgaragiola@gmail.com

core.yaml
```diff
@@ -389,7 +389,11 @@ security:
     # Can be 256 or 384. If you change here, you have to change also
     # the same property in membersrvc.yaml to the same value
     level: 256
-
+ 
+    # Can be SHA2 or SHA3. If you change here, you have to change also
+    # the same property in membersrvc.yaml to the same value
+    hashAlgorithm: SHA3
```
srderson, line 395: Any thoughts on whether https://github.com/hyperledger/fabric/blob/master/core/util/utils.go#L34 should read this same configuration setting or whether a new one should be created? @manish-sethi

srderson:
LGTM. Confirmed all tests are passing. Thanks

core.yaml
manish-sethi, line 395: @srderson, I think that it's better make the util method take these properties as parameters and then define properties separately for different usages (e.g., one we can define for state).

PR #1152, opened 2016-04-19
Add docker based cluster quick start method

yeasy:
Besides the vagrant and bare metal setup, this will give new users a quick try.

Many teams show interests with this setup, might be good idea to put here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,6 +128,24 @@ CGO_CFLAGS=" " CGO_LDFLAGS="-lrocksdb -lstdc++ -lm -lz -lbz2 -lsnappy" go instal
 up at the IP address currently assumed by the test environment
 (`172.17.0.1`). Use `ifconfig` or `ip addr` to find the docker bridge.
 
+## Quick Start with Docker <a name="docker"></a>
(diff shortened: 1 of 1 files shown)


srderson:
I've seen a couple people creating docker images which is awesome. As these are not official or maintained by the Hyperledger Project, I think it would be best to keep a list in the wiki. Thoughts?

I've created a page at https://github.com/hyperledger/fabric/wiki/Docker-Images. I think it makes sense to move most of your proposed text to this page and add a link in the README. This will allow others to post their own docker images without filling up the README.

christo4ferris:
@srderson that's a good thought, but I think we want to provide for their development and evolution under the terms of the hyperledger project governance (apache licensing etc). Further, I think that this could also lead to different deployment options. I guess we could treat like StackForge or cloudfoundry-community, but again those are separate and not under the formal governance of the main project.

srderson:
@christo4ferris Agreed, I'd like to see these images, tools, etc. have a path to be included officially included in Hyperledger. The wiki feels less formal to me so I thought it may be a good starting point for people to link to contributions currently residing in their personal repositories. I'm open to linking to these in the README or wiki though, whatever makes sense for Hyperledger.

yeasy:
@srderson @christo4ferris 
Thanks for the comments!
That makes sense to me, I will close this pr first.
Thanks!

Issue #596, opened 2016-02-03
Request for Profiling Services


bcbrock:
I would like to add an HTTP service to obc-peer, obc-ca and others (?) to support profiling. This is very easy to set up and use as documented here: https://golang.org/pkg/net/http/pprof/
I would propose to add a command line flag to enable this service, e.g.,

```
./obc-peer peer --pprof-service=localhost:6060
```

The above would spawn an HTTP server thread listening on localist:6060, using the default 'request router'. Since the current REST API services use a non-default router this shouldn't interfere with REST APIs.

Here's an example of a profile I collected this morning of the obc-peer running with security enabled:

[pprof.pdf](https://github.com/openblockchain/obc-peer/files/116610/pprof.pdf)

muralisrini:
+1
We used pprof while debugging some memory issues.

jeffgarratt:
@bcbrock #1179 was the re-submission of #1157.  It has been merged so this function is now present.  You can refer to https://github.com/hyperledger/fabric/blob/master/bddtests/docker-compose-4-consensus-batch.yml#L14-L18 as the mechanism for configuring through docker compose.  Perhaps this issue can now be closed?

bcbrock:
Thanks @jeffgarratt

PR #1157, opened 2016-04-19
Issues/1001/devops to peer

jeffgarratt:
This PR is still in progress.  Wanted to allow others to review.  @srderson @jyellick


bcbrock:
@jeffgarratt - The profiling stuff will satisfy/close my issue #596. I think this new profiling feature should be included in user-level documentation, and I would also also suggest that an error log message be printed if the call to `http.ListenAndServe(profileListenAddress, nil)` fails.

Issue #1158, opened 2016-04-19
CLI "peer" command would be better called "start"


lehors:
The CLI commands are:
      peer        Runs the peer.
      status      Returns status of the peer.
      stop        Stops the running peer.
This means that to start a peer one has to type: "./peer peer" and we have the pair "peer, stop" to start and stop the peer. This is not intuitive.

I suggest instead we have the commands:
      start        Starts the peer.
      status      Returns status of the peer.
      stop        Stops the running peer.
Which would be much more intuitive.

lehors:
I'd be happy to submit a PR to make the change if others agree with the change.

christo4ferris:
SGTM

jeffgarratt:
The executable has the dual purpose of acting as the Peer, as well as the CLI.  Since we have now renamed the executable 'peer', it may be prudent to rename the 'peer' command to 'server', with several subcommands of 'start','stop', etc.  In this way the now 'server' related commands (or CLI) may feel more segregated.

for example, starting the peer would be...

> peer server start
>   And the other commands would be readily identifiable as CLI based.

Thoughts?  @lehors @binhn

christo4ferris:
@jeffgarratt other than 'server' what other targets would be invoked in that manner?

jeffgarratt:
@christo4ferris there is also 'chaincode', which supports several sub-commands ('deploy', 'invoke', 'query').

for example, to deploy a chaincode...

> peer chaincode deploy

binhn:
The list of commands:

```
Available Commands: 
  peer        Runs the peer.
  status      Returns status of the peer.
  stop        Stops the running peer.
  login       Logs in a user on CLI.
  network     Lists all network peers.
  chaincode   chaincode specific commands.
  help        Help about any command
```

I ageed with @jeffgarratt that `start` and `stop` impact the target peer, but the remaining ones are CLI and some with subcommands or parameters. The online help is not very complete, which we should also cover in this issue that @lehors can take care of.

jeffgarratt:
@lehors @binhn @christo4ferris 

I would think something like the following would be inutitive:

```
peer
   help
   server
      start
      stop
      status
   chaincode
      deploy
      invoke
      query
   network
      login
      list
```

So...
to start server:  peer server start

to login:  peer network login

Seems to read well as you type the command.  Could also consider changing 'server' to 'node' if the community is more use to that nomenclature.

lehors:
[...]

Issue #1131, opened 2016-04-18
Remove need for -o flag when compiling peer


srderson:
The documentation currently gives the following instructions for compiling the peer.
`go build -o peer`

This generates a binary named `peer`. Without the `-o` flag, the binary would be named `fabric`.

We want to remove the use of the `-o` flag as it causes confusion and ideally keep the binary named `peer`.

The proposal is to create a top level folder named `peer` and move `main.go` and `core.yaml` into this folder.

```
cd $GOPATH/src/github.com/hyperledger/fabric/peer
go build
```

would generate a binary named `peer` in this case.

genggjh:
Why can't we just use the fabric as the binary file name or give another name? Instead of peer, after all you know './peer peer' is really a little bit confusing.
Maybe './peer -d' is another better choice. -d means run peer as daemon process.

srderson:
The reason for not using 'fabric' as the binary name is because this encompasses other parts of the system that are not the 'peer' such as 'membership services', 'sdk', 'api', etc. I imagine the list will grow.

Agree that `./peer peer` is a bit confusing and the `./peer -d` may be a better option.

RicHernandez2:
Please keep the documentation up do date on what you decide.  For now I'm calling ./peer peer, till documentation says otherwise.

srderson:
@RicHernandez2 Yes, I will also update docs. For the first pass, I'm just moving peer and core.yaml. If we choose to deprecate `./peer peer` in favor of `./peer -d` we can likely keep both for a while for backwards compatibility.

genggjh:
+1

srderson:
Closing as main and core.yaml have been moved. @genggjh It seems there's discussion on the `peer` command taking place in issue #1158

PR #1168, opened 2016-04-20
Fix main cmd name to fabric

yeasy:
The old name peer does not match current executable name.


srderson:
See issue #1158 for discussion. The term 'Fabric' refers to the peer, membership services, sdk, etc. The consensus for now seems to be that the server should be referred to as the 'peer', not 'fabric'.

genggjh:
@yeasy We have discussion related this command in https://github.com/hyperledger/fabric/issues/1131.

PR #1169, opened 2016-04-20
Added missing Chaincode name for the deploy api

cx0der:
Without the `name` attribute deploy API fails with https://github.com/hyperledger/fabric/blob/95d2b1ac230bc5abffa12929315053a5e1df04fa/core/rest/rest_api.go#L695

### DIFF
--- a/docs/API/SandboxSetup.md
+++ b/docs/API/SandboxSetup.md
@@ -201,7 +201,7 @@ POST host:port/chaincode
   "params": {
       "type": 1,
       "chaincodeID":{
-          "name":"52b0d803fc395b5e34d8d4a7cd69fb6aa00099b8fabed83504ac1c5d61a425aca5b3ad3bf96643ea4fdaac132c417c37b00f88fa800de7ece387d008a76d3586"
+          "name":"mycc"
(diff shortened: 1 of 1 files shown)


cx0der:
Signed-off-by: Shiva t.shivakumar@gmail.com

muralisrini:
@cx0der : the changes look ok. You shouldn't need "-path" in the "deploy" option in "--peer-chaincodedev" mode (I think... I know that to be true in the "cli" mode). Can you check and make that change too please ?

angrbrd:
@cx0der 

@muralisrini is correct in that the "path" is not necessary for a deploy when the peer is running in development mode. Please remove the "path" from the payload in the deploy request with and without security enabled. Otherwise, looks good. Thanks for catching my batch replace during the repo move :-].

cx0der:
Thanks for the comments, I have removed the path attribute. Also corrected the expected responses.

angrbrd:
@cx0der looks great. Thanks.

@srderson @binhn , please merge this change.

srderson:
LGTM. Thanks

PR #1174, opened 2016-04-20
Fix viewchange watermark movement

jyellick:
The previous code was checking the low watermark against the selected checkpoint, and if they differed, would move the watermarks, reset its `lastExec` and kick off state transfer.

If the replica had already processed beyond this point, but had not yet seen a stable checkpoint to move its watermarks, then its `lastExec` was lowered, causing it to re-execute transactions which had already occurred, and state transfer was requested, even though it was unnecessary.

This changeset corrects this behavior by moving the watermarks based on the checkpoint, but not necessarily performing state transfer or resetting `lastExec`.  Then, based on `lastExec` the view change determines whether state transfer is needed, and only then modified `lastExec`.

Included is a changeset which tests for this behavior.  It also tests for the inverse bad behavior (of moving the watermarks based on the `lastExec` rather than the checkpoint).

@kchristidis Could you review this logic?

All go tests pass.
All behave tests pass.

### DIFF
--- a/consensus/obcpbft/pbft-core_test.go
+++ b/consensus/obcpbft/pbft-core_test.go
@@ -375,6 +375,77 @@ func TestInconsistentPrePrepare(t *testing.T) {
 }
 
 // This test is designed to detect a conflation of S and S' from the paper in the view change
+func TestViewChangeWatermarksMovement(t *testing.T) {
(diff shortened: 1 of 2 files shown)


kchristidis:
Code reviewed, looks good to go.

:+1:

PR #1187, opened 2016-04-20
Merging in client_sdk branch

angrbrd:
Merging in initial Node.js client sdk code and tape UT.


srderson:
@angrbrd @smithbk Do the proto files have to be duplicated? Is it possible to share the files in the protos folder?

binhn:
@angrbrd how can we test this prior to merge?

angrbrd:
@srderson, I guess it is possible for us to pull in the .proto files directly from the /fabric/protos directory inside our .js files. But I guess the question is how do we do that when we package this as an npm package? Does it make sense to reference them from /fabric/protos when we are packaging the npm? Will that still work? My impression was that we have to have them inside the directory containing the Node project to package them. Let me know what you think.

angrbrd:
@binhn, if you are worried that it will break any of the existing code that is not a concern, as this submission does not interact with any of the existing code at all. As far as UT, I did start a UT file together with the submission that includes two tests covering current functionality. The test file is under /sdk/node/test/unit/chain-tests.js.

binhn:
@angrbrd I know it doesn't break existing code, but want to make sure this code has test-cases that I can run prior to merge.  Thanks for the info.

binhn:
Regarding protos, we can either ask the developers to copy the protos from the peer over, or we copy as done during packaging. However, I think we need to introduce version somehow to make sure the nodejs sdk is at the same level as the peer.

binhn:
I scanned through the code and noticed that we also need to support SHA2, issue #908, for both 256 and 384

angrbrd:
@binhn , issue #908 should already be closed as the associated PR #1148 has been merged.

core/ledger/kvledger/upgrade_dbs.go
```diff
@@ -24,29 +32,39 @@ func UpgradeDBs(rootFSPath string) error {
 	defer fileLock.Unlock()
```
cendhu, line 29: Not related to this PR. We have this exact `fileLock` code in four places (updatedbs, rollback, reset, ledger initializing). If possible, it will be addressed as part of [FAB-17609](https://jira.hyperledger.org/browse/FAB-17609)

core/ledger/kvledger/kv_ledger_provider.go
```diff
@@ -498,6 +498,10 @@ func openIDStore(path string) (s *idStore, e error) {
 	return &idStore{db, path}, nil
 }
 
+func (s *idStore) getFormat() ([]byte, error) {
```
cendhu: Let's see whether @manish-sethi like this tiny function 😄 
Anyway, in one of the comment, I have mentioned a way to remove this function.

core/ledger/kvledger/kv_ledger_provider_test.go
```diff
@@ -504,7 +504,8 @@ func constructTestLedgerID(i int) string {
 }
 
 func testConfig(t *testing.T) (conf *lgr.Config, cleanup func()) {
-	path, err := ioutil.TempDir("", "kvledger")
+	// use /tmp instead of os.TempDir so that we can mount dirs for couchdb
```
cendhu: The default folders shared between mac osx and docker are 
- /Users/
- /Volumes/
- /private/  
- /tmp

The `os.TempDir` creates a temp folder in `/var/folders` and hence the mount would be denied. Given that the test passes in CI (linux environment), I assume that `/tmp` is also shared between the linux host and docker. 

We can update this comment to add more information as `instead of os.TempDir` would confuse us when we look at the code after a month.

core/ledger/kvledger/upgrade_dbs.go
```diff
@@ -24,29 +32,39 @@ func UpgradeDBs(rootFSPath string) error {
...
-	logger.Debugf("Attempting to upgrade idStore data format to current format %s", dataformat.Version20)
+	logger.Info("Updating the ledger provider database format ...")
+	return idStore.upgradeFormat()
+}
 
-	dbPath := LedgerProviderPath(rootFSPath)
-	db := leveldbhelper.CreateDB(&leveldbhelper.Conf{DBPath: dbPath})
-	db.Open()
-	defer db.Close()
+func dropCouchAndLevelDBs(config *ledger.Config) error {
```
cendhu: This function name is a bit confusing as we use either CouchDB or leveldb as statedb but not both. Though we use leveldb for all internal states, we don't explicitly name it leveldb instead call it `configHistoryDB`, `HistoryDB`, `BookkeeperDB`.

Further, the dropDBs() also try to drop the `stateleveldb`. Assume that couchDB is used as the stateDB. With that assumption, the code drops `statecouchdb` and also try to drop `stateleveldb`. Hence, It looks hacky.

As we can drop CouchDB during rollback/reset too, we can move this logic to `dropDBs()` itself and either drop whole statecouchdb or whole stateleveldb. However, this require changes to both rollback and reset code (need to pass `ledger.config` rather than the `ledger.config.rootfspath`). Hence, I am fine with doing it in a separate PR. When we implement snapshot processing, we can avoid dropping the whole DB and drop only the dbs associated with the channel.  At that time, we could infact pass a config struct to dropDBs with info such as statedb used, channels to be dropped (one or all), rootfspath, etc...). 

Wdyt @manish-sethi ?

core/ledger/kvledger/txmgmt/statedb/statecouchdb/couchdb_test_export.go
```diff
@@ -24,15 +24,17 @@ func StartCouchDB(t *testing.T, binds []string) (addr string, stopCouchDBFunc fu
 
 // DeleteApplicationDBs deletes all the databases other than fabric internal database
 func DeleteApplicationDBs(t testing.TB, config *ledger.CouchDBConfig) {
```
cendhu: Given that `DropApplicationDBs()` is exported from the `couchdbutil.go`, I don't think we need `DeleteApplicationDBs()` in the `test_export.go`. Wherever we have called `DeleteApplicationDBs()`, now we can call `DropApplicationDBs()` and assert on the returned error..

core/ledger/kvledger/txmgmt/statedb/statecouchdb/couchdbutil.go
```diff
@@ -297,3 +298,28 @@ func escapeUpperCase(dbName string) string {
...
+	for _, dbName := range dbNames {
+		db := &couchDatabase{
+			couchInstance: couchInstance,
+			dbName:        dbName,
+		}
+		_, err := db.dropDatabase()
+		if err != nil {
+			logger.Errorf("Error dropping CouchDB database %s", dbName)
+			return err
+		}
```
cendhu: Earlier, we called `dropDB()` present in the `test_export.go`  instead of these lines. Now, it would look wrong to keep `dropDB()` in the `test_export.go` as it is used only by the local test file. Hence, I would suggest replacing these lines with `dropDB()` and move the `dropDB()` to this file & remove assertions.

core/ledger/kvledger/txmgmt/statedb/statecouchdb/couchdbutil_test.go
```diff
@@ -184,3 +186,35 @@ func TestConstructedNamespaceDBName(t *testing.T) {
...
+
+func TestDropApplicationDBs(t *testing.T) {
+	config := testConfig()
+	couchDBEnv.startCouchDB(t)
+	config.Address = couchDBEnv.couchAddress
+	defer couchDBEnv.cleanup(config)
+	database := "testdropapplicationdbs"
+
+	couchInstance, err := createCouchInstance(config, &disabled.Provider{})
+	assert.NoError(t, err, "Error when trying to create couch instance")
```
cendhu, line 197: I understand that this whole test file follows a similar assertion statement with an error message. I don't know how this will help us. `assert.NoError(t, err)` is a much cleaner way and more readable. We get the error only when we add new features or change existing code. At that time, this error message does not help much in debugging. We look at the line number and directly go to the test file to see why the error has occurred. We also refer to debug logs. For this PR, it is fine as existing UT func also uses similar statements.  Going forward, we can avoid using an error message within the assertion when it is not necessary as it affects the readability (need to read some extra words which does add much value).

core/ledger/kvledger/txmgmt/statedb/statecouchdb/couchdbutil_test.go
```diff
@@ -184,3 +186,35 @@ func TestConstructedNamespaceDBName(t *testing.T) {
...
+	numCouchdbs := 10
+	for i := 0; i < numCouchdbs; i++ {
+		db := couchDatabase{couchInstance: couchInstance, dbName: fmt.Sprintf("%s_%d", database, i)}
+		err = db.createDatabaseIfNotExist()
+		assert.NoErrorf(t, err, "Error when trying to create database %s", db.dbName)
+	}
+
+	dbs, err := couchInstance.retrieveApplicationDBNames()
+	assert.NoError(t, err, "Error when retrieving application db names")
+	assert.Equal(t, 1+numCouchdbs, len(dbs), "Number of databases created")
```
cendhu: the error message seems incomplete.

core/ledger/kvledger/upgrade_dbs.go
```diff
@@ -24,29 +32,39 @@ func UpgradeDBs(rootFSPath string) error {
...
-	if err := UpgradeIDStoreFormat(rootFSPath); err != nil {
+	idStore := &idStore{db, dbPath}
+	format, err := idStore.getFormat()
+	if err != nil {
 		return err
 	}
+	if bytes.Equal(format, []byte(dataformat.Version20)) {
+		logger.Info("Ledger data format is current, do nothing.")
+		return nil
+	}
```
cendhu: This is checked twice. Once here and again at line 59. This is because `upgradeFormat()` has side-effect. As per the name of the function, it should upgrade the format. However, it is also checking whether the store has the latest format (which has been checked already here). I would suggest splitting `idStore.upgradeFormat()` into `idStore.hasLastestFormat()` and `idStore.upgradeFormat()`. We can also use `idStore.hasVersion20Format()` to be explicit. Then, replace these lines with the newly introduced function.  We can also replace a few lines in `openIDStore()` with the newly introduced function but it is not related to this PR. 

With this change, we can get rid of the newly introduced function `idStore.getFormat()`.

core/ledger/kvledger/txmgmt/statedb/statecouchdb/couchdbutil_test.go
```diff
@@ -184,3 +186,35 @@ func TestConstructedNamespaceDBName(t *testing.T) {
...
+	numCouchdbs := 10
+	for i := 0; i < numCouchdbs; i++ {
+		db := couchDatabase{couchInstance: couchInstance, dbName: fmt.Sprintf("%s_%d", database, i)}
+		err = db.createDatabaseIfNotExist()
+		assert.NoErrorf(t, err, "Error when trying to create database %s", db.dbName)
+	}
+
+	dbs, err := couchInstance.retrieveApplicationDBNames()
+	assert.NoError(t, err, "Error when retrieving application db names")
+	assert.Equal(t, 1+numCouchdbs, len(dbs), "Expected number of databases are created")
```
cendhu:
```suggestion
	assert.Equal(t, 1+numCouchdbs, len(dbs), "Expected number of databases are not created")
```

core/ledger/kvledger/upgrade_dbs.go
```diff
@@ -24,29 +33,44 @@ func UpgradeDBs(rootFSPath string) error {
...
+	dbPath := LedgerProviderPath(rootFSPath)
+	db := leveldbhelper.CreateDB(&leveldbhelper.Conf{DBPath: dbPath})
+	db.Open()
+	defer db.Close()
 
-	// upgrades idStore (ledgerProvider) and drop databases
-	if err := UpgradeIDStoreFormat(rootFSPath); err != nil {
-		return err
+	if emptydb, _ := db.IsEmpty(); emptydb {
+		return errors.Errorf("ledger database %s is empty. Make sure core.yaml has the correct ledger path.", dbPath)
```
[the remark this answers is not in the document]
manish-sethi: I don't have a strong feeling either way to keep or remove this check. However, to take a note, another problem with this check (extending the no channel use case) is that it will change behavior if you start the peer once and then execute the upgrade command again. Because, then the version entry would have been added and this condition won't be true. 

From the user point of view, even then there won't be any data present  but this message will no longer be shown.

core/ledger/kvledger/upgrade_dbs.go
```diff
@@ -24,29 +33,44 @@ func UpgradeDBs(rootFSPath string) error {
...
+	if bytes.Equal(format, []byte(dataformat.Version20)) {
+		logger.Info("Ledger data format is current, do nothing.")
+		return nil
+	}
+	if format != nil {
+		err = &dataformat.ErrVersionMismatch{
+			ExpectedVersion: "",
+			Version:         string(format),
+			DBInfo:          fmt.Sprintf("leveldb for channel-IDs at [%s]", dbPath),
+		}
```
manish-sethi: We had this check deliberately  in the function `idStore.upgradeFormat`. So that we do not accidentally end-up using that function without a precondition check somewhere.  I suggest to keep it there for sure. I am fine, if you want to check here again.

Also, like `dataformat.Version20`, it's better to use `dataformat.Version1x` explicitly instead of `nil`, or `""`. For byte comparison, you may have to define `Version1xBytes` explicitly though.

Not for this PR but I think that we discussed previously that we can introduce two generic constants `CurrentFormat` and `PreviousFormat` instead of `Version20` and `Version1x`.

core/ledger/kvledger/txmgmt/statedb/statecouchdb/couchdbutil_test.go
```diff
@@ -184,3 +186,35 @@ func TestConstructedNamespaceDBName(t *testing.T) {
...
+	couchDBEnv.startCouchDB(t)
+	config.Address = couchDBEnv.couchAddress
+	defer couchDBEnv.cleanup(config)
+	database := "testdropapplicationdbs"
+
+	couchInstance, err := createCouchInstance(config, &disabled.Provider{})
+	assert.NoError(t, err, "Error when trying to create couch instance")
+
+	err = writeDataFormatVersion(couchInstance, dataformat.Version20)
+	assert.NoErrorf(t, err, "Error when trying to write data format version to %s", fabricInternalDBName)
```
manish-sethi: what is the significance of writing this explicitly for this test?

PR #1201, opened 2016-04-21
Fix up a few remaining fabric references. Issue #1131

srderson:
Cleans up a few remaining references to `fabric` that needs to changed to `peer` as that is the default binary name.


srderson:

Issue #1202, opened 2016-04-21
behave fails if peer is running


lehors:
Several behave scenarios fail when run while a peer is running.
Unfortunately that's exactly the situation you find yourself in when following the current README which instructs you to run the unit tests, with a peer running in a different window, and then goes into describing how to run behave tests.

lehors:
This has been rendered obsolete by issue #1223

PR #1164, opened 2016-04-19
move out install instructions

joequant:
This commit moves the install instructions out of the README.md which makes the repository more friendly to newbies.

Signed-off-by: Joseph Wang joequant@gmail.com

PR #1203, opened 2016-04-21
Update instructions on how to run behave tests

lehors:
This change to the README mitigates issue #1202 by instructing the user to kill the unit tests peer that makes behave fail.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,7 +86,7 @@ To run a specific test use the `-run RE` flag where RE is a regular expression t
     go test -v -run=TestGetFoo
 
 #### 3.2 Behave Tests
-[Behave](http://pythonhosted.org/behave/) tests will setup networks of peers with different security and consensus configurations and verify that transactions run properly. To run these tests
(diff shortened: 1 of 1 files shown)


lehors:

srderson:
Waiting to merge until we finish discussion on PR #1164. If we merge #1164 this PR will need to be rebased.

christo4ferris:
this should be automated with a bash script that detects a running peer(s) and tears down an environment before launching the behave tests. No?

binhn:
I agreed with @christo4ferris -- we need to look at the script to see if we can detect a local running peer; issue #1223. @lehors could you take a look at that.

lehors:
@binhn I did look at it and have a fix but in doing so I uncovered another problem with the CLI: It can't run from outside its source directory, because the paths to the config files are relative. I will open a separate issue for that and submit a fix, and then submit the fix to stop any running peer when behave is launched.
I'll close this PR as it will become irrelevant.

PR #1210, opened 2016-04-22
Get-Set Multiple keys and Copy chaincode state

manish-sethi:
This PR include changes for 
1) get and set multiple keys to amortize the cost of grpc communication between shim and peer
2) copying state from one chaincode to another for supporting the scenario of upgrading of a chaincode

### DIFF
--- a/core/ledger/ledger.go
+++ b/core/ledger/ledger.go
@@ -216,6 +216,23 @@ func (ledger *Ledger) DeleteState(chaincodeID string, key string) error {
 	return ledger.state.Delete(chaincodeID, key)
 }
 
+// CopyState copies all the key-values from sourceChaincodeID to destChaincodeID
+func (ledger *Ledger) CopyState(sourceChaincodeID string, destChaincodeID string) error {
+	return ledger.state.CopyState(sourceChaincodeID, destChaincodeID)
+}
+
+// GetStateMultipleKeys returns the values for the multiple keys.
+// This method is mainly to amortize the cost of grpc communication between chaincode shim peer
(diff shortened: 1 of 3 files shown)


manish-sethi:
@srderson can you have a look and merge. @muralisrini - FYI

srderson:
LGTM. Confirmed all tests are passing.

core/ledger/kvledger/txmgmt/txmgr/lockbasedtxmgr/lockbased_txmgr.go
```diff
@@ -59,36 +59,46 @@ func (c *current) maxTxNumber() uint64 {
...
-	bookkeepingProvider bookkeeping.Provider,
-	ccInfoProvider ledger.DeployedChaincodeInfoProvider,
-	customTxProcessors map[common.HeaderType]ledger.CustomTxProcessor,
-	hasher ledger.Hasher,
-) (*LockBasedTxMgr, error) {
-
-	if hasher == nil {
+func NewLockBasedTxMgr(initializer *Initializer) (*LockBasedTxMgr, error) {
+
+	if initializer.Hasher == nil {
```
manish-sethi, line 76: In isolation, without checking `initializer` for nil, this check seems a half job. Also, we check here only for Hasher and not for other parameters, seem odd. Though, the difference is that the hasher is an external ledger dependency that is passed from the peer. So, I guess that to make it clear, we should move this way up (directly when the ledger is initialized from outside). Not introduced in this PR, so can be done separately.

PR #1211, opened 2016-04-22
increase behave timeout for issue793

rameshthoomu:


christo4ferris:
LGTM

srderson:
LGTM

common/ledger/blkstorage/blockindex.go
```diff
@@ -248,6 +251,41 @@ func (index *blockIndex) getTXLocByBlockNumTranNum(blockNum uint64, tranNum uint
...
+	}
+
+	writer.Write([]byte{exportFileFormatVersion})
+
+	dbItr := index.db.GetIterator([]byte{txIDIdxKeyPrefix}, []byte{txIDIdxKeyPrefix + 1})
+	defer dbItr.Release()
+	var previousTxID []byte
+	if err := dbItr.Error(); err != nil {
+		return errors.Wrap(err, "internal leveldb error while obtaining db iterator")
+	}
```
cendhu, line 264: It looks disconnected from the code. It can be placed after line 261?

common/ledger/blkstorage/blockindex.go
```diff
@@ -248,6 +251,41 @@ func (index *blockIndex) getTXLocByBlockNumTranNum(blockNum uint64, tranNum uint
...
 }
 
+func (index *blockIndex) exportUniqueTxIDs(writer io.Writer) error {
+	if !index.isAttributeIndexed(IndexableAttrTxID) {
+		return ErrAttrNotIndexed
+	}
+
+	writer.Write([]byte{exportFileFormatVersion})
+
+	dbItr := index.db.GetIterator([]byte{txIDIdxKeyPrefix}, []byte{txIDIdxKeyPrefix + 1})
```
cendhu, line 260: I think this works only for the `rune` based prefix. For a `string` prefix such as the one used for [private data](https://github.com/hyperledger/fabric/blob/master/core/ledger/kvledger/txmgmt/privacyenabledstate/common_storage_db.go#L30), this would not work. As I don't see a need to change `t` to string prefix in the future, this looks good.

common/ledger/blkstorage/blockindex.go
```diff
@@ -248,6 +251,41 @@ func (index *blockIndex) getTXLocByBlockNumTranNum(blockNum uint64, tranNum uint
...
+	defer dbItr.Release()
+	var previousTxID []byte
+	if err := dbItr.Error(); err != nil {
+		return errors.Wrap(err, "internal leveldb error while obtaining db iterator")
+	}
+	for dbItr.Next() {
+		if err := dbItr.Error(); err != nil {
+			return errors.Wrap(err, "internal leveldb error while iterating for txids")
+		}
+		txID, err := decodeTxIDKey(dbItr.Key())
```
cendhu: I was expecting `decodeTxIDKey()` to give the complete decoded key. Instead, it returns a part of the encoded key. Most of [the similarly named functions](https://github.com/hyperledger/fabric/blob/master/core/ledger/pvtdatastorage/kv_encoding.go) in `pvtdatastorage` pkg decode and returns the complete key. Here, the `decodeTxIDKey()` seems to return only the part of the encoded txID key. I understand that we need to write entry length to the snapshot file so that we know how much to read. Here, the encoded `len(txID)` would be the same for all txIDs as long as we use SHA256. Having the length is good as changing of hash to SHA512 in the future would not create a problem. 

Anyway, I don't have complete clarity on what name to suggest here. `retrieveTxIDFromKey(dbitr.Key())` may look okay.

common/ledger/blkstorage/blockindex.go
```diff
@@ -248,6 +251,41 @@ func (index *blockIndex) getTXLocByBlockNumTranNum(blockNum uint64, tranNum uint
...
+			return err
+		}
+
+		// duplicate TxID may be present in the index
+		if bytes.Equal(previousTxID, txID) {
+			continue
+		}
+		previousTxID = make([]byte, len(txID))
+		copy(previousTxID, txID)
+		if _, err := writer.Write(txID); err != nil {
```
cendhu: We do write the order-preserving encoded length of `txID` that is hidden here. If we make `decodeTxIDKey()` to return only `txID` and then explicitly compute the length, it would make the code more explicit (shouldn't impact the performance significantly). 

Do we really need order-preserving txIDLength in the snapshot file? or just encoded length?

common/ledger/blkstorage/blockindex.go
```diff
@@ -248,6 +251,41 @@ func (index *blockIndex) getTXLocByBlockNumTranNum(blockNum uint64, tranNum uint
 	return txFLP, nil
 }
 
+func (index *blockIndex) exportUniqueTxIDs(writer io.Writer) error {
```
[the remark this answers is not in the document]
manish-sethi: TL;DR - No, we cannot keep all the content in memory. 

I'll shortly push a PR that invokes this function. That should make it clear how we are going to used this function. Then, we can follow the similar pattern for exporting other data items.

common/ledger/blkstorage/blockindex.go
manish-sethi: Yes, with string, this won't let you proceed. This will result in a compilation error.

common/ledger/blkstorage/blockindex.go
manish-sethi: Sure, I can rearrange these lines in next commit.

common/ledger/blkstorage/blockindex.go
```diff
@@ -248,6 +252,44 @@ func (index *blockIndex) getTXLocByBlockNumTranNum(blockNum uint64, tranNum uint
...
+	}
+	writer.Write([]byte{exportFileFormatVersion})
+	dbItr := index.db.GetIterator([]byte{txIDIdxKeyPrefix}, []byte{txIDIdxKeyPrefix + 1})
+	defer dbItr.Release()
+	if err := dbItr.Error(); err != nil {
+		return errors.Wrap(err, "internal leveldb error while obtaining db iterator")
+	}
+
+	var previousTxID string
+	reusableBuf := make([]byte, binary.MaxVarintLen64)
```
cendhu, line 267: Nice. I like this. Going forward, we will keep following this approach to reduce garbage collector work. Otherwise, as we add more features, I am sure that GC would start to steal a lot of CPU periodically from the peer.

Issue #1207, opened 2016-04-22
Example of a chaincode implementing RBAC 


elli-androulaki:
Labels: security, application layer
To be assigned to: @adecaro

jeffgarratt:
Will review #1212 and look to remove go unit test and replace with behave test and simply take the chaincode only and not the yaml or test file.  @adecaro @srderson

frankyclu:
@jeffgarratt for the record, the go test of the asset management chaincode sample only works in the first run, would require restarting image after each run

christo4ferris:
@elli-androulaki closing this as I believe we have this now. Please reopen if you disagree.

Issue #1139, opened 2016-04-18
Chaincode access control using Tcert attributes


binhn:
Using Tcert attributes to determine access control on chaincode. This may be done as part of the uber chaincode #1133 on each chaincode-function (deploy, invoke, query, upgrade, terminate)

Issue #961, opened 2016-04-04
Access control for chaincode design


binhn:
We need ability to control who can issue specific transactions on specific chaincode

jeffgarratt:
Will document the work on https://github.com/hyperledger/fabric/wiki/Access-Control-Specification

frankyclu:
Add some clarification:

This function does not limit to invocation transactions, For example, some has mentioned that they would like to filter the information in the query response so that a user can only see what he/she is entitled to see.

jeffgarratt:
This is in progress.  The current focus will be on user defined ACLs for their specific application.

jeffgarratt:
@binhn here is s breakdown of the work for this issue, will track through item list below.  I want to use this work as an archetype of how new functionality should be added to system through a test first approach.
- [ ] define behave tests for ACL demonstration leveraging asset control chaincode
- [ ] extend behave implementation to support new constructs (eg. multi chaincode access from scenario)
- [ ] document the use case and guidelines within the ACL spec wiki page

jeffgarratt:
Classification of access control mechanisms: ordered by complexity
- access control based on TCerts (No attributes, not role based) see -> asset_mgt.go (doable now with behave and should pass)
- Role based access control using TCerts w/o using attributes -> @adecaro will provide simple example (doable soon, will coordinate with @adecaro )
- Attribute based access control using Tcerts with Attributes -> ( @adecaro @diegomasini to determine)
- attribute access control with User Defined membership services (use their own PKI infastructure) TBD

PR #1212, opened 2016-04-22
Example of a chaincode using role-based access control (RBAC)

adecaro:
The PR contains a chaincode showing how to implement role-based access control using TCerts.
This addresses Issue #1207.


srderson:
Why does rbac.yaml contain a complete copy of core.yaml? My concern is that we will not be able to keep these in sync.

Also, is it possible to include a README with this chaincode describing its functionality? I'd preferably like to do that for all example chaincodes.

adecaro:
Hi @srderson, this chaincode is part of a joint effort with @jeffgarratt to write behave tests for the access control mechanism supported by the fabric.

A copy of core.yaml is there for test purposes. It will not stay there forever. In fact, once the client sdk will be ready I will write an app to show how to invoke the chaincode leveraging the client sdk drectly. Actually, having a test that shows how to invoke such a chaincode will help us driving the design of the client sdk to ensure that all the required API are exposed as expected. What do you think?

For the README, I will write it soon :)

srderson:
@adecaro @jeffgarratt So it looks like this chaincode is deployed in the `rbac_test.go` test case? Is it possible to use a behave test instead? The unit tests where chaincodes are deployed are generally fragile. Ideally I think they should all be moved over to behave tests.

adecaro:
@srderson, Indeed, rbac_test simulates a minimal fabric environment, deploys the chaincode and invokes it in order to test its behavior. Actually, the final goal is to have behave tests. I'm just concerned that without the client sdk, it would be difficult to run successfully any behave test because the CLI (nor the REST API) does not provide the required API to drive the chaincode.

Anyway, I will try to understand if we can have something cleaner without waiting the client sdk before merging :)

jeffgarratt:
@adecaro @srderson The behave test could be implemented as part of the effort I am working on now.  Either #1139 or #961.

binhn:
@srderson  should we wait for the behave tests to merge or just merge it since it is a example chaincode?

srderson:
@binhn I'd like to avoid merging `rbac.yaml` and `rbac_test.go`. The current effort is to remove these types of tests, not add more. Also, I'd really like to see a README with all example chaincodes, at least documenting their purpose and API.

adecaro:
@binhn @srderson  I also think we can wait the behave tests to merge. I'm working with @jeffgarratt to have them ready asap.

christo4ferris:
@adecaro could you please fix the license in these new files? It should be the text that I have in PR #1288 thanks

jeffgarratt:
@adecaro @srderson I need to touch base with @adecaro and work this workflow.  Have come to a block point based upon current flow (which is already a bit precarious).

adecaro:
Hi @christo4ferris, I have updated the license and also added a readme file to describe the chaincode. 

@christo4ferris, @jeffgarratt, @srderson I can also add a demo app like I did for the asset management chaincode to show how to invoke it. What do you think?

examples/chaincode/go/rbac/README.md
```diff
@@ -0,0 +1,82 @@
...
+the DER (Distinguished Encoding Rules) certificate encoding of one of the administrator ECert/TCert. 
+
+For simplicity, there is only one administrator.
+
+A possible work-flow could be the following:
+
+1. Alice is the deployer of the chaincode;
+2. Alice wants to assign the administrator role to Bob;
+3. Alice obtains, via an out-of-band channel, a TCert of Bob, let us call this certificate *BobCert*;
+4. Alice constructs a deploy transaction, as described in *application-ACL.md*,  setting the transaction metadata to *DER(BobCert)*.
```
BiscuitBomber: Please can we make this a hyperlink/url too ?

examples/chaincode/go/rbac/rbac.go
```diff
@@ -0,0 +1,305 @@
...
+
+	myLogger.Debug("Assign 'admin' role...")
+
+	// Give to the deployer the role 'admin'
+	deployer, err := stub.GetCallerMetadata()
+	if err != nil {
+		return nil, errors.New("Failed getting metadata.")
+	}
+	if len(deployer) == 0 {
+		return nil, errors.New("Invalid admin certificate. Empty.")
```
BiscuitBomber: "Empty.": Maybe "Missing caller metadata in admin certificate", "Deployer information is required in admin certificate", or so?

(trying to be a bit more informative in reporting the error.)

examples/chaincode/go/rbac/README.md
christo4ferris: could you please clarify... you want a link to this README from where? (I am assuming that is what you're asking for here)

BiscuitBomber: Sorry the suggestion wasn't clear. I meant to suggest: "as described in [Application ACLs](https://github.com/hyperledger/fabric/blob/master/docs/tech/application-ACL.md)"
 instead of "as described in ***application-ACL.md***" [in line 33]

That is:  [Application ACLs] (https://github.com/hyperledger/fabric/blob/master/docs/tech/application-ACL.md)

examples/chaincode/go/rbac/rbac.go
adecaro: Hi @BiscuitBomber, what do you think of the following: "Missing metadata. Admin certificate expected in the metadata field."?

examples/chaincode/go/rbac/README.md
adecaro: @BiscuitBomber @christo4ferris Fixed. Thanks for the feedback :)

examples/chaincode/go/rbac/rbac.go
JonathanLevi: @adecaro: this is marvelous! Thank you.

JonathanLevi: @adecaro I think that this is marvelous! Thank you.

JonathanLevi: @adecaro I think that this is a marvelous error message!

BiscuitBomber: Oh, I see - I have learned something new from your proposed change already.

How about "An admin certificate must be provided in the metadata field."? Although I don't have a strong preference either way, so long as the message is instructive.

adecaro: Deal :)

jeffgarratt:
@adecaro @srderson @binhn Want to hold this PR until it is reconciled with TCert with attrs example.

ghaskins:
Can this be closed and then reopened when its ready?

christo4ferris:
this needs some unit/function tests

Issue #1128, opened 2016-04-17
Cache viper config properties in each module


binhn:
Calling viper to get config properties is slow. We want to cache the frequent use properties outside of critical loops. This should be implemented in all modules on the main transaction path.

bcbrock:
@binhn I would like to volunteer to work on this - I have time this week and next.

My suspicion is that the large majority of parameters don't need to be cached. My goal would be to cache parameters only until the word "viper" stops appearing in the graphical performance profiles. I would propose to add a configuration cache data structure to each package, initialized through a call from the main fabric routine.

binhn:
@bcbrock That's great. Thanks. I assign this to you.

Manish @manish-sethi did the Ledger as a sample to follow https://github.com/hyperledger/fabric/pull/1136

bcbrock:
@binhn My current thought is that there is no need to cache these variables in Go language test files. In many cases this would require the test programs to include the `peer` package which they don't do at the moment. Also they are not performance critical. On the other hand it would make the codebase more consistent if we did cache them. What do you think?

PR #1215, opened 2016-04-22
Phase 1 of solving issue 1128

bcbrock:
I need to stage the commits for solving #1128 into several phases, that
is caching commonly used configuratioin parameters. If I try to do a single
monolithic change I've already seen that I'll never be able to simultaneously
resolve all of the conflicts that will arise.

Phase 1 sets up the infrastructure, and redefines the peer-package functions
GetLocalAddress(), GetValidatorStreamAddress() and GetpeerEndpint() as
caching versions.

Once this is accepted, Phase2 will replace all config parameters that appear
frequenctly in single-peer runs with NOOPS with the function calls defined here. Later phases will fix up any remaining issues. I've already seen that when we're done it will make
significant improvements in performance.

Note, the Go tests all run, but some Behave tests fail intermittently. I Don't think this changeset is the cause.

Also note that a few of the changes here come from running 'gofmt' on files I changed.

### DIFF
--- a/core/config.go
+++ b/core/config.go
@@ -62,11 +62,11 @@ func SetupTestConfig() {
 	viper.AutomaticEnv()
 	replacer := strings.NewReplacer(".", "_")
 	viper.SetEnvKeyReplacer(replacer)
-	viper.SetConfigName("core") // name of config file (without extension)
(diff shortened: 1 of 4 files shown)


bcbrock:

PR #1219, opened 2016-04-22
Non deterministic version of example02 for testing pbft - sieve

RicHernandez2:
Random number generator for testing sieve's ability to sieve out non deterministic results.
[chaincode_nondeter_example02.zip](https://github.com/hyperledger/fabric/files/232462/chaincode_nondeter_example02.zip)

### DIFF
--- a/examples/chaincode/go/chaincode_nondeter_example02.go
+++ b/examples/chaincode/go/chaincode_nondeter_example02.go
@@ -1,0 +1,198 @@
+/*
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+*/
+
+package main
+
+import (
+	"errors"
+	"fmt"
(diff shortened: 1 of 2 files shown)


RicHernandez2:
Signed-off-by: RicHernandez2 rhernan@us.ibm.com

examples/chaincode/go/chaincode_nondeter_example02.go
```diff
@@ -0,0 +1,198 @@
...
+		return nil, errors.New("Incorrect number of arguments. Expecting 4")
+	}
+
+	// Initialize the chaincode
+	A = args[0]
+	Aval, err = strconv.Atoi(args[1])
+	if err != nil {
+		return nil, errors.New("Expecting integer value for asset holding")
+	}
+	B = rand.Intn(100)
```
tuand27613, line 51: I didn't see a seed() function being called which means that rand.Intn() will be deterministic and always return the same sequence of numbers across all the peers.

I think that every time you want to generate a random number, you need to do something like rand.Seed(time now) then rand.Intn().  See this [example](https://gobyexample.com/random-numbers) 

Also, I think it will be better if you do this in the invoke() method  and use an random number for the X value rather than making the Bval random.

srderson:
This PR contains changes to an old version of `docs/dev-setup/devnet-setup.md` which is why it cannot merged. Could you remove that file from this PR?

Regarding `examples/chaincode/go/chaincode_nondeter_example02.go`, I'd like to keep "test" chaincodes out of the `examples` folder. It's valid for a test case to have a "here's a chaincode you should never write" test, but I think they will confuse users in the `examples` folder. In PR #1221 I've created a folder for test chaincodes under the bddtests folder. Could you move the chaincode there once #1221 is merged?

Finally, could you write a behave test for this chaincode? It would be a great test of sieve consensus.

examples/chaincode/go/chaincode_nondeter_example02.go
RicHernandez2, line 51: I looked into that, I'll put in some logging to see what the output is, because it crashed 1 to 4 vps pretty hardcore.

PR #1225, opened 2016-04-23
docker-compose-files: fixed paths

adecaro:
This PR fixes a compilation error caused by an incorrect configuration in Dockerfile. It also adds back the infinitloop.sh require by the cli container.

@tuand27613, may you review this?


consensus/docker-compose-files/scripts/infiniteloop.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
```
tuand27613: Hi @adecaro,  we already have hyperledger/fabric/scripts/infiniteloop.sh  .

Are you replacing that copy of infiniteloop.sh ? or are you adding a new .sh script ?  
In any case, I think we're better off keeping hyperledger/fabric/scripts/infiniteloop.sh  and updating compose-consensus-4.yaml to put in the right path for the cli container ( `command: ../scripts/infiniteloop.sh` I think but verify that )

adecaro: Hi @tuand27613, sorry I didn't know infiniteloop.sh has been moved.

Anyway, I added back that script in that position because the cli container was failing saying the script was not found. How should we set that path?

tuand27613: ah, so the cli container working dir moved to `./fabric/peer`  so the invoke of the shell script becomes `../scripts/infiniteloop.sh`

adecaro:
@tuand27613 I have removed the infiniteloop.sh and updates the command in compose-consensus-4.xml. Now, it works :)

tuand27613:
👍 @adecaro

consensus/docker-compose-files/Dockerfile
```diff
@@ -5,6 +5,8 @@ RUN mkdir -p /var/hyperledger/db
 RUN mkdir -p /var/hyperledger/production
 WORKDIR $GOPATH/src/github.com/hyperledger/fabric/
 COPY . .
+WORKDIR $GOPATH/src/github.com/hyperledger/fabric/peer
+RUN pwd
```
christo4ferris: what is this for? debugging? Seems like if we want to have debugging, we should make it conditional, and probably not the default.

adecaro: Hi @christo4ferris, thanks for pointing it out. The 'pwd' command is actually not needed. I will remove it.

srderson:
LGTM. Confirmed tests are passing

Issue #1222, opened 2016-04-22
README.md link to main.go needs new target


joshhus:
https://github.com/hyperledger/fabric
Under 2.0 Run: 
Encoded link for "main.go" text results in "page not found"
The target to the following URL is "page not found" - https://github.com/hyperledger/fabric/blob/master/main.go

genggjh:
it was moved into peer subfolder, I will fix it.

angrbrd:
I just tried this link and it seems that this has already been fixed. Closing.

srderson:
PR #1226

PR #1226, opened 2016-04-23
Update the directory of main.go - issue #1222

genggjh:

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ cd $GOPATH/src/github.com/hyperledger/fabric/peer
 ./peer
 ```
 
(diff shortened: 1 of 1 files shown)


binhn:
@genggjh @christo4ferris  -  I don't think this is right.  We are in `$GOPATH/src/github.com/hyperledger/fabric/peer`, so the current directory contains `main.go`, NOT `peer/main.go`

christo4ferris:
@binhn @genggjh actually, no... the navigation is relative to the root of the repo for markdown files such as this README.md - I tested the PR and it is working as expected. The current README.md main.go link returns a 404.

binhn:
@christo4ferris thanks for testing it out.

genggjh:
@christo4ferris thanks for your clarification.

Issue #1224, opened 2016-04-23
Peer CLI fails to run when invoked outside its directory


lehors:
When invoked from anywhere else than the fabric/peer directory the peer command fails because it can't find the configuration files that are all searched for using relative paths.

```
vagrant@hyperledger-devenv:v0.0.9-4ba4f96:/opt/gopath/src/github.com/hyperledger/fabric$ peer/peer
panic: Error reading CORE_PBFT plugin config: Unsupported Config Type ""

goroutine 1 [running]:
panic(0xcac6a0, 0xc8201a6f30)
        /opt/go/src/runtime/panic.go:464 +0x3e6
github.com/hyperledger/fabric/consensus/obcpbft.loadConfig(0xc820156a90)
        /opt/gopath/src/github.com/hyperledger/fabric/consensus/obcpbft/obc-pbft.go:83 +0x601
...
```

lehors:
Fixed with PR  #1229

PR #1227, opened 2016-04-23
Fix Issue #1224

lehors:
This PR changes the peer command so that it doesn't have to be invoked from its directory (fabric/peer).
The configuration files are searched in $GOPATH/src/hyperledger/fabric/... allowing the command to be invoked from anywhere.


lehors:
Sorry about the mode change, that's a mistake. I'll resubmit a clean PR.

PR #1228, opened 2016-04-23
add custom options for start docker controller

ckeyer:
add `network mode`,`nds` option for start docker container. to use customize ipam&DNS

### DIFF
--- a/core/container/dockercontroller/dockercontroller.go
+++ b/core/container/dockercontroller/dockercontroller.go
@@ -29,6 +29,7 @@ import (
 	"github.com/hyperledger/fabric/core/container/ccintf"
 	cutil "github.com/hyperledger/fabric/core/container/util"
 	"github.com/op/go-logging"
+	"github.com/spf13/viper"
 	"golang.org/x/net/context"
 )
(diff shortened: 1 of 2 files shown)


binhn:
@jeffgarratt could you review this

orderer/common/server/main.go
```diff
@@ -241,7 +241,9 @@ func Main() {
 	}
 	ab.RegisterAtomicBroadcastServer(grpcServer.Server(), server)
 	logger.Info("Beginning to serve requests")
-	grpcServer.Start()
+	if err := grpcServer.Start(); err != nil {
```
sykesm, line 255: The error that is returned isn't actually from `Start` but from `grpc.Server#Serve(net.Listener)`. (The method is poorly named.) This means the error message is misleading as it implies the server didn't start yet, in reality, it very likely did start but exited.

Also the message is lacking context about *which* gRPC server failed (there's more than one) and does not use a format or context string so the message and the error will not be separated by a space. We typically would use the `Fatalf` form with a format specifier for these kinds of messages.

PR #1229, opened 2016-04-23
Fix Issue 1224

lehors:
This PR changes the peer command so that it doesn't have to be invoked from its directory (fabric/peer).
The configuration files are searched in $GOPATH/src/hyperledger/fabric/... allowing the command to be invoked from anywhere.


lehors:

binhn:
@lehors would it work if the $GOPATH contains a list of paths?

lehors:
@binhn Ah, good question. I haven't tested that. Let me have a look.

lehors:
@binhn I committed another change to handle lists in GOPATH.

binhn:
@lehors I think we also need the same code in [noops](https://github.com/hyperledger/fabric/blob/master/consensus/noops/configutil.go)

Also just curious if it'd be any faster if we change the order to have the absolute path first in the `AddConfigPath`.

lehors:
@binhn I made the change to noops.go. Sorry for missing this. I didn't realize we had those config files peppered around. The other ones were easy to find because it caused the peer command to fail at launch.

The reason I put the added config path after the others was to be conservative. Apparently someone had similar issues and added various paths. This is especially visible in [obc-pbft.go](https://github.com/hyperledger/fabric/blob/master/consensus/obcpbft/obc-pbft.go) where we have:
    config.AddConfigPath("./")
    config.AddConfigPath("../consensus/obcpbft/")
    config.AddConfigPath("../../consensus/obcpbft")

Not knowing the history behind this I thought it was more prudent to append to it so it wouldn't risk breaking anything. But I'd be happy to put it before if you prefer.

binhn:
@lehors thanks.  I will accept it as-is, and we can experiment with the order of the paths later when we run performance suite

Issue #915, opened 2016-03-31
Sieve with 4 validating peers - One peer never executes a single transaction


bcbrock:
I have set up a network of 4 validating peers running Sieve. I have deployed 4 chain codes to the network. I send several hundred transactions to each peer. One of the peers, which happens to be the last one created, was sent 1300 transactions which the logs show it has received and then forwarded on to "the validator".  However at the end of the run, it is clear from the chaincode logs that this peer has never executed a single transaction via any of its 4 chaincodes.

Is this expected? My expectation was that all peers should be independently executing every transaction, since they are all working correctly and the chain codes are deterministic.

bcbrock:
@corecode @vukolic Could one of you please comment on this issue?

corecode:
could you please provide DEBUG level logs for all 4 peers?

bcbrock:
Thanks @corecode - Sorry for the delay, I have reduced the problem to reduce the log file size. Here there are 4 peers each managing 2 chain codes. In the tarball, vp[0-3]_logs/stderr are the peer logs, and dev-vp*_logs/stderr are the chaincode logs.

In this example, peer 1 (vp1) is the one that never executed a transaction. The run "fails" due to this fact.

[vplogs.tar.gz](https://github.com/openblockchain/obc-peer/files/198346/vplogs.tar.gz)

corecode:
@jyellick I think this is because sieve now does not stop the pbft timer anymore; probably got lost in changes #833.  Could you verify my thoughts?

jyellick:
@corecode This actually appears to be a race.  In order to not block waiting to acquire the pbft lock, which can cause a deadlock, you'll see in `Validate` that the timer is reset on a goroutine.  Invariably, the verify set arrives before this goroutine executes, and so the timer is not stopped as it should be.  If this happens to two nodes, then the view advances, and state transfer occurs, which is why the error is never observed in more than 2 replicas.

@bcbrock As I implement #919 I will do my best to eliminate much of the Sieve hacky cruft which accumulated in the #833 changeset.  Unless there are objections, I think this issue would be best addressed in the #919 work.

corecode:
The goroutine starts the timer, and doesn't stop it.  I think once a decision comes out of `validResultChan`, the timer must be stopped.

jyellick:
[...]

PR #1231, opened 2016-04-24
Issue1128 phase2

bcbrock:
In the second phase of solving #1128, occurrences of the variables "security.enabled",
"peer.validator.enabled" and "peer.tls.enabled" in mainline code are replaced
with the cached versions.

Note: Based on one benchmark, the cumlative performance improvement over the baseline
is 26% for the first phase (already committed), and is now 29% after the second phase (this PR).

Once this PR is processed I'll finish by looking to see if there are commonly used variables that appear in other benchmarks, and by modifying the Go language tests for consistency.

### DIFF
--- a/consensus/helper/helper.go
+++ b/consensus/helper/helper.go
@@ -22,7 +22,6 @@ package helper
 import (
 	"fmt"
 
-	"github.com/spf13/viper"
 	"golang.org/x/net/context"
 
 	"github.com/hyperledger/fabric/core/chaincode"
@@ -45,7 +44,7 @@ type Helper struct {
 // NewHelper constructs the consensus helper object
 func NewHelper(mhc peer.MessageHandlerCoordinator) consensus.Stack {
 	return &Helper{coordinator: mhc,
-		secOn:     viper.GetBool("security.enabled"),
+		secOn:     peer.SecurityEnabled(),
 		secHelper: mhc.GetSecHelper()}
 }
 
--- a/consensus/helper/helper.go
+++ b/consensus/helper/helper.go
@@ -24,6 +24,7 @@ import (
 
 	"golang.org/x/net/context"
 
+	"github.com/hyperledger/fabric/core"
 	"github.com/hyperledger/fabric/core/chaincode"
 	"github.com/hyperledger/fabric/consensus"
 	crypto "github.com/hyperledger/fabric/core/crypto"
@@ -44,7 +45,7 @@ type Helper struct {
 // NewHelper constructs the consensus helper object
 func NewHelper(mhc peer.MessageHandlerCoordinator) consensus.Stack {
 	return &Helper{coordinator: mhc,
-		secOn:     peer.SecurityEnabled(),
+		secOn:     core.SecurityEnabled(),
 		secHelper: mhc.GetSecHelper()}
 }
 
--- a/core/devops.go
+++ b/core/devops.go
@@ -125,7 +125,7 @@ func (d *Devops) Deploy(ctx context.Context, spec *pb.ChaincodeSpec) (*pb.Chainc
 	var tx *pb.Transaction
 	var sec crypto.Client
 
-	if viper.GetBool("security.enabled") {
+	if peer.SecurityEnabled() {
 		if devopsLogger.IsEnabledFor(logging.DEBUG) {
 			devopsLogger.Debug("Initializing secure devops using context %s", spec.SecureContext)
 		}
@@ -178,7 +178,7 @@ func (d *Devops) invokeOrQuery(ctx context.Context, chaincodeInvocationSpec *pb.
 	var transaction *pb.Transaction
 	var err error
 	var sec crypto.Client
-	if viper.GetBool("security.enabled") {
+	if peer.SecurityEnabled() {
 		if devopsLogger.IsEnabledFor(logging.DEBUG) {
 			devopsLogger.Debug("Initializing secure devops using context %s", chaincodeInvocationSpec.ChaincodeSpec.SecureContext)
 		}
--- a/core/peer/handler.go
+++ b/core/peer/handler.go
@@ -158,7 +158,7 @@ func (d *Handler) beforeHello(e *fsm.Event) {
 	peerLogger.Debug("Received %s from endpoint=%s", e.Event, helloMessage)
 
 	// If security enabled, need to verify the signature on the hello message
-	if viper.GetBool("security.enabled") {
+	if SecurityEnabled() {
 		if err := d.Coordinator.GetSecHelper().Verify(helloMessage.PeerEndpoint.PkiID, msg.Signature, msg.Payload); err != nil {
 			e.Cancel(fmt.Errorf("Error Verifying signature for received HelloMessage: %s", err))
 			return
--- a/core/peer/handler_sync_state.go
+++ b/core/peer/handler_sync_state.go
@@ -22,8 +22,6 @@ package peer
 import (
 	"sync"
 
-	"github.com/spf13/viper"
-
 	pb "github.com/hyperledger/fabric/protos"
 )
 
@@ -54,7 +52,7 @@ func (srh *syncStateSnapshotRequestHandler) createRequest() *pb.SyncStateSnapsho
 }
 
 func makeStateSnapshotChannel() chan *pb.SyncStateSnapshot {
-	return make(chan *pb.SyncStateSnapshot, viper.GetInt("peer.sync.state.snapshot.channelSize"))
+	return make(chan *pb.SyncStateSnapshot, SyncStateSnapshotChannelSize())
 }
 
 func newSyncStateSnapshotRequestHandler() *syncStateSnapshotRequestHandler {
@@ -82,7 +80,7 @@ func (ssdh *syncStateDeltasHandler) createRequest(syncBlockRange *pb.SyncBlockRa
 }
 
 func makeSyncStateDeltasChannel() chan *pb.SyncStateDeltas {
-	return make(chan *pb.SyncStateDeltas, viper.GetInt("peer.sync.state.deltas.channelSize"))
+	return make(chan *pb.SyncStateDeltas, SyncStateDeltasChannelSize())
 }
 
 func newSyncStateDeltasHandler() *syncStateDeltasHandler {
--- a/core/peer/peer.go
+++ b/core/peer/peer.go
@@ -144,7 +144,7 @@ func GetLocalIP() string {
 // NewPeerClientConnectionWithAddress Returns a new grpc.ClientConn to the configured local PEER.
 func NewPeerClientConnectionWithAddress(peerAddress string) (*grpc.ClientConn, error) {
 	var opts []grpc.DialOption
-	if viper.GetBool("peer.tls.enabled") {
+	if TlsEnabled() {
 		var sn string
 		if viper.GetString("peer.tls.serverhostoverride") != "" {
 			sn = viper.GetString("peer.tls.serverhostoverride")
@@ -222,7 +222,7 @@ func NewPeerWithHandler(secHelperFunc func() crypto.Peer, handlerFact HandlerFac
 	peer.secHelper = secHelperFunc()
 
 	// Install security object for peer
-	if viper.GetBool("security.enabled") {
+	if SecurityEnabled() {
 		if peer.secHelper == nil {
 			return nil, fmt.Errorf("Security helper not provided")
 		}
@@ -242,11 +242,11 @@ func NewPeerWithEngine(secHelperFunc func() crypto.Peer, engFactory EngineFactor
 	peer = new(PeerImpl)
 	peer.handlerMap = &handlerMap{m: make(map[pb.PeerID]MessageHandler)}
 
-	peer.isValidator = viper.GetBool("peer.validator.enabled")
+	peer.isValidator = ValidatorEnabled()
 	peer.secHelper = secHelperFunc()
 
 	// Install security object for peer
-	if viper.GetBool("security.enabled") {
+	if SecurityEnabled() {
 		if peer.secHelper == nil {
 			return nil, fmt.Errorf("Security helper not provided")
 		}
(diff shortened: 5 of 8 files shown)


bcbrock:

binhn:
@bcbrock thanks - that sounds great on performance improvements

christo4ferris:
@bcbrock this could really use some unit tests, IMO. @binhn @srderson would you agree?

binhn:
@bcbrock  I got the following failures when running behave. I haven't dug deeper; did it work for you?

```
Failing scenarios:
  peer_basic.feature:437  chaincode example02 with 4 peers and 1 membersrvc, issue #567 -- @1.1 Consensus Options
  peer_basic.feature:439  chaincode example02 with 4 peers and 1 membersrvc, issue #567 -- @1.3 Consensus Options
  peer_basic.feature:522  chaincode example02 with 4 peers and 1 membersrvc, issue #680 (State transfer) -- @1.2 Consensus Options
  peer_basic.feature:651  4 peers and 1 membersrvc, consensus still works if one backup replica fails -- @1.2 Consensus Options
  peer_basic.feature:713  4 peers and 1 membersrvc, consensus fails if 2 backup replicas fail -- @1.2 Consensus Options
  peer_basic.feature:792  chaincode example02 with 4 peers,1 membersrvc, and 1 non-validating peer. -- @1.1 Consensus Options

0 features passed, 1 failed, 1 skipped
17 scenarios passed, 6 failed, 4 skipped
260 steps passed, 6 failed, 287 skipped, 0 undefined
Took 7m46.741s
```

bcbrock:
@binhn - I am sorry to hear that you are having problems with this. On my system, the Behave test suite does not work unless PR #1235 is installed - and even then it is unreliable. I just ran Behave with this patch and #1235 applied and it failed. I then re-ran Behave with the collect-logs option to try to diagnose the problem, and it worked perfectly. These simple optimizations may be exciting some timing-related issues. Is there a button you can push to rerun the CI tests?

binhn:
@bcbrock So it looks like timing.  I ran it again this morning after cleaning up docker images, etc. and it completed successfully:

```
1 feature passed, 0 failed, 1 skipped
23 scenarios passed, 0 failed, 4 skipped
369 steps passed, 0 failed, 184 skipped, 0 undefined
Took 11m47.679s
```

bcbrock:
@binhn Thanks for merging this. 

I also re-ran just now with logs and also got failures.  All of the failures in my case involved Sieve (maybe coincidence, maybe not). Looking at the logs, in one case I saw that one of the peers never received one of the transactions. I have previously reported similar, as-yet unresolved errors with Sieve - #915.

bcbrock:
@christo4ferris - I can add some unit tests, but I'm not sure what you had in mind? To me this is a straightforward replacement of a computed constant by the cached version of the constant. The functionality is being tested every time these branch conditions are evaluated. I do see a coverage question regarding whether every branch is ever in fact reached, but that seems like a system-level test issue to me.

christo4ferris:
it's moot at this point, but I was specifically thinking unit tests for the new function.

common/ledger/snapshot/file.go
```diff
@@ -0,0 +1,200 @@
...
+	"hash"
+	"io"
+	"os"
+
+	"github.com/golang/protobuf/proto"
+	"github.com/pkg/errors"
+)
+
+// FileCreator creates a new file for ledger snapshot. This is expected to be used by various
+// components of ledger, such as blockstorage and statedb for exporting the relevent snapshot data
```
cendhu: We could think of moving this package to `fabric/internal/pkg/snapshot`.  Otherwise, when we refactor `common/ledger` package, we can also move this package.

common/ledger/snapshot/file.go
```diff
@@ -0,0 +1,200 @@
...
+type FileCreator struct {
```
cendhu:
```suggestion
type FileWriter struct {
```
`FileReader` and `FileWriter` seem to compliment each other well. I assume we have prefixed `File` so that we can introduce `snapshot.GRPCReader` and `snapshot.GRPCWriter` later. If not, we could just rename it `Reader` and `Writer`.

common/ledger/snapshot/file.go
```diff
@@ -0,0 +1,200 @@
...
+	file              *os.File
+	dataformat        byte
+	hasher            hash.Hash
+	bufWriter         *bufio.Writer
+	multiWriter       io.Writer
+	varintReusableBuf []byte
+}
+
+// CreateFile creates a new file for exporting the ledger snapshot data
+// This function returns an error if the file already exists. The `dataformat` is the first byte
```
cendhu, line 32: Correct me if my assumptions about the future code (i.e., to be added code) are wrong:

1. The `filePath` includes the block height. Hence, if the file already exists, we throw an error. 
2. If the last execution of `peer node snapshot` fails, our recovery logic would remove incomplete data files before restarting the work (if metadata file exists but incomplete, we don't delete data files). These logics will be part of each sub-ledger package.
3. When we add recovery logic, the file deleter would be part of this package.

common/ledger/snapshot/file.go
```diff
@@ -0,0 +1,200 @@
...
+// written to the file. The hasher is used for computing the hash of the data stream
+func CreateFile(filePath string, dataformat byte, hasher hash.Hash) (*FileCreator, error) {
```
cendhu:
```suggestion
func NewFileWriter(filePath string, dataformat byte, hasher hash.Hash) (*FileWriter, error) {
```

common/ledger/snapshot/file.go
```diff
@@ -0,0 +1,200 @@
...
+		file:              file,
+		bufWriter:         bufWriter,
+		multiWriter:       multiWriter,
+		hasher:            hasher,
+		varintReusableBuf: make([]byte, binary.MaxVarintLen64),
+	}, nil
+}
+
+// EncodeString encodes and appends the string to the data stream
+func (c *FileCreator) EncodeString(str string) error {
```
cendhu:
```suggestion
func (c *FileWriter) WriteString(str string) error {
```
`Writer.WriteXXX` seems like a correct usage as per `bufio` and other golang io packages. 

By using `WriteString()`, we aren't really hiding some important info from the consumer. In fact, `encoding/decoding` and `length` have now become internal detail, and consumers need not be aware of it. 

The abstraction that we expose to the consumer is reading or writing the following from/to a file -- `string`, `proto`, `[]byte`, `uint64`

1. `EndoceString()` -> `WriteString()`
1. `EncodeProtoMessage()` -> `WriteProtoMessage()`
2. `EncodeBytes()` -> `WriteBytes()`
3. `EncodeUVarint()` -> `WriteUint64()`

common/ledger/snapshot/file.go
```diff
@@ -0,0 +1,200 @@
...
+		return nil, errors.New(fmt.Sprintf("unexpected data format: %x", dataFormat))
+	}
+	return &FileReader{
+		file:      file,
+		bufReader: bufReader,
+	}, nil
+}
+
+// DecodeString reads and decodes a string
+func (r *FileReader) DecodeString() (string, error) {
```
cendhu, line 146:
```suggestion
func (r *FileReader) ReadString() (string, error) {
```
Similarly, we can change 

1. `DecodeBytes()` -> `ReadBytes()`
2. `DecodeProtoMessage()` -> `ReadProtoMessage()`
3. `DecodeUVarInt()` -> `ReadUint64()`

common/ledger/snapshot/file.go
```diff
@@ -0,0 +1,200 @@
...
+// FileReader reads from a ledger snapshot file. This is expected to be used for loading the ledger snapshot data
+// during bootstrapping a channel from snapshot. The data should be read, using the functions `DecodeXXX`,
+// in the same sequence in which the data was written by the functions `EncodeXXX` in the `FileCreator`.
+// Note that the FileReader does not verifies the hash of stream and it is expected that the hash has been verified
+// by the consumer. Later, if we decide to perform this, on-the-side, while loading the snapshot data, the FileRedear,
+// like the FileCreator, would take a `hasher` as an input
+type FileReader struct {
+	file              *os.File
+	bufReader         *bufio.Reader
+	reusableByteSlice []byte
```
cendhu, line 119:
```suggestion
	varintReusableBuf []byte
```
to maintain consistency with what we have defined in the `FileCreator`

common/ledger/snapshot/file.go
```diff
@@ -0,0 +1,200 @@
...
+}
+
+// OpenFile constructs a FileReader. This function returns an error if the format of the file, stored in the
+// first byte, does not match with the expectedDataFormat
+func OpenFile(filePath string, expectDataformat byte) (*FileReader, error) {
```
cendhu, line 124:
```suggestion
func NewFileReader(filePath string, expectDataformat byte) (*FileReader, error) {
```

common/ledger/snapshot/file.go
```diff
@@ -0,0 +1,200 @@
...
+
+	"github.com/golang/protobuf/proto"
+	"github.com/pkg/errors"
+)
+
+// FileCreator creates a new file for ledger snapshot. This is expected to be used by various
+// components of ledger, such as blockstorage and statedb for exporting the relevent snapshot data
+type FileCreator struct {
+	file              *os.File
+	dataformat        byte
```
cendhu: Unused variables. Can be removed as it is not needed.

common/ledger/snapshot/file.go
```diff
@@ -0,0 +1,200 @@
...
+	bufWriter         *bufio.Writer
+	multiWriter       io.Writer
+	varintReusableBuf []byte
+}
+
+// CreateFile creates a new file for exporting the ledger snapshot data
+// This function returns an error if the file already exists. The `dataformat` is the first byte
+// written to the file. The hasher is used for computing the hash of the data stream
+func CreateFile(filePath string, dataformat byte, hasher hash.Hash) (*FileCreator, error) {
+	file, err := os.OpenFile(filePath, os.O_RDWR|os.O_CREATE|os.O_EXCL, 0444)
```
cendhu, line 37: I assume `-r--r--r--` permission is used so that admin is forced to use the snapshot API to delete existing snapshots. In the `blkstorage` pkg, we use `drw-rw-rw-` for the block files. A comment to explain this reasoning may help us in the future.

common/ledger/snapshot/file.go
```diff
@@ -0,0 +1,200 @@
...
+		return nil, errors.Wrapf(err, "error while flushing to the snapshot file: %s ", c.file.Name())
+	}
+	if err := c.file.Close(); err != nil {
+		return nil, errors.Wrapf(err, "error while closing the snapshot file: %s ", c.file.Name())
+	}
+	return c.hasher.Sum(nil), nil
+}
+
+// Close closes the underlying file, if not already done. A consumer can invoke this function if the consumer
+// encountered some error and simply wants to abandon the snapshot file creation (typically, intended to be used in a defer statement)
```
cendhu, line 102: Shouldn't we name it `Cleanup()` and remove the created file after checking whether `c.Done` is set to `false`?

common/ledger/snapshot/file.go
```diff
@@ -0,0 +1,200 @@
...
+
+// Done closes the snapshot file and returns the final hash of the data stream
+func (c *FileCreator) Done() ([]byte, error) {
+	if err := c.bufWriter.Flush(); err != nil {
+		return nil, errors.Wrapf(err, "error while flushing to the snapshot file: %s ", c.file.Name())
+	}
+	if err := c.file.Close(); err != nil {
+		return nil, errors.Wrapf(err, "error while closing the snapshot file: %s ", c.file.Name())
+	}
+	return c.hasher.Sum(nil), nil
```
cendhu, line 98: so we aren't computing hash per chunk of data instead for the full data in memory?

common/ledger/snapshot/file.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+	return errors.Wrapf(c.file.Close(), "error while closing the snapshot file: %s", c.file.Name())
+}
+
+// FileReader reads from a ledger snapshot file. This is expected to be used for loading the ledger snapshot data
+// during bootstrapping a channel from snapshot. The data should be read, using the functions `DecodeXXX`,
+// in the same sequence in which the data was written by the functions `EncodeXXX` in the `FileCreator`.
+// Note that the FileReader does not verifies the hash of stream and it is expected that the hash has been verified
+// by the consumer. Later, if we decide to perform this, on-the-side, while loading the snapshot data, the FileRedear,
+// like the FileCreator, would take a `hasher` as an input
```
cendhu, line 115: Given that the hash computation logic is present in this package, shouldn't the verification logic also be in the same package? Otherwise, are we thinking that the admin would verify the hash of the file using sha256sum cli command?

common/ledger/snapshot/file.go
```diff
@@ -0,0 +1,200 @@
...
+type FileReader struct {
```
cendhu, line 116: Wouldn't batch reads increase sequential IO and hence the performance? Anyway, we can update the code when we process a snapshot file.

common/ledger/snapshot/file.go
```diff
@@ -0,0 +1,200 @@
...
+func (c *FileCreator) EncodeProtoMessage(m proto.Message) error {
+	b, err := proto.Marshal(m)
+	if err != nil {
+		return errors.Wrapf(err, "error marshalling proto message to write to the snapshot file: %s", c.file.Name())
+	}
+	return c.EncodeBytes(b)
+}
+
+// EncodeBytes encodes and appends bytes to the data stream
+func (c *FileCreator) EncodeBytes(b []byte) error {
```
cendhu: Do we need to export this? I think consumer can mostly create snapshot of a given store using `WriteProtoMessage()`, `WriteString()` and `WriteUint64()` itself.

common/ledger/blkstorage/blockindex.go
```diff
@@ -252,42 +256,67 @@ func (index *blockIndex) getTXLocByBlockNumTranNum(blockNum uint64, tranNum uint
...
 		}
 		previousTxID = txID
-
-		n := binary.PutUvarint(reusableBuf, uint64(len(txID)))
-		if _, err := writer.Write(reusableBuf[:n]); err != nil {
-			return err
-		}
-		if _, err := writer.Write([]byte(txID)); err != nil {
-			return err
+		if err := dataFile.EncodeString(txID); err != nil {
```
cendhu, line 292:
```suggestion
		if err := dataFileWriter.WriteString(txID); err != nil {
```
might read better.

common/ledger/snapshot/file_test.go
```diff
@@ -0,0 +1,194 @@
...
+	"testing"
+
+	"github.com/golang/protobuf/proto"
+	"github.com/hyperledger/fabric-protos-go/common"
+	"github.com/stretchr/testify/require"
+)
+
+func TestFileCreateAndRead(t *testing.T) {
+	testDir := testPath(t)
+	//defer os.RemoveAll(testDir)
```
cendhu: forgot to uncomment?

PR #1233, opened 2016-04-24
Fix issue 1232

### LINKED ISSUE #1232: Go language tests need to be serialized, or modified to eliminate TCP port conflicts
When the Go language test suite is run as documented:

```
go test -timeout=20m $(go list github.com/hyperledger/fabric/... | grep -v /vendor/ | grep -v /examples/)
```

Go is running the tests in parallel. This is nice because the test suite runs pretty quickly.

The problem is that this parallelization creates conflicts. Tests that create the same TCP server ports will fail if run in parallel.  Here's what I'm seeing now:

```
--- FAIL: TestExecuteDeployTransaction (0.00s)
    exectransaction_test.go:162: Error starting peer listener listen tcp 0.0.0.0:40303: bind: address already in use
--- FAIL: TestExecuteInvokeTransaction (0.00s)
    exectransaction_test.go:293: Error starting peer listener listen tcp 0.0.0.0:40303: bind: address already in use
--- FAIL: TestExecuteQuery (0.00s)
    exectransaction_test.go:391: Error starting peer listener listen tcp 0.0.0.0:40303: bind: address already in use
--- FAIL: TestExecuteInvokeInvalidTransaction (0.00s)
    exectransaction_test.go:474: Error starting peer listener listen tcp 0.0.0.0:40303: bind: address already in use
--- FAIL: TestExecuteInvalidQuery (0.00s)
    exectransaction_test.go:535: Error starting peer listener listen tcp 0.0.0.0:40303: bind: address already in use
--- FAIL: TestChaincodeInvokeChaincode (0.00s)
    exectransaction_test.go:606: Error starting peer listener listen tcp 0.0.0.0:40303: bind: address already in use
--- FAIL: TestChaincodeQueryChaincode (0.00s)
    exectransaction_test.go:718: Error starting peer listener listen tcp 0.0.0.0:40303: bind: address already in use
--- FAIL: TestExecuteDeploySysChaincode (0.00s)
    exectransaction_test.go:861: Error starting peer listener listen tcp 0.0.0.0:40303: bind: address already in use
FAIL
FAIL    github.com/hyperledger/fabric/core/chaincode    0.032s
```

Note that the problem is not with this test, it is the fact that _another_ test that is running concurrently using port 40303.

The big-hammer approach to solve this problem is to serialize the Go tests by adding `-p 1` to the test command:

```
go test -timeout=20m -p 1 $(go list github.com/hyperledger/fabric/... | grep -v /vendor/ | grep -v /examples/)
```

This significantly slows things down, though.

I am about to submit a PR that solves the problem by modifying the port number that is used by the conflicting test. This works, but seems like a band-aid on a bigger problem. If this continue to happen in the future as more tests are added we might need to, for example, document a list of server ports for tests to use and assign each test permanently to a fixed port.

bcbrock:
Fixes #1232 by changing the port number used by the events_test.go so that it
does not conflict with other tests being run in parallel.

### DIFF
--- a/events/events_test.go
+++ b/events/events_test.go
@@ -148,7 +148,7 @@ func TestMain(m *testing.M) {
 
 	//use a different address than what we usually use for "peer"
 	//we override the peerAddress set in chaincode_support.go
-	peerAddress = "0.0.0.0:40303"
+	peerAddress = "0.0.0.0:50303"
 
 	lis, err := net.Listen("tcp", peerAddress)
 	if err != nil {


bcbrock:

core/container/externalbuilder/copy.go
```diff
@@ -10,47 +10,15 @@ import (
...
-	}
-
-	return errors.WithMessagef(cpErr, "failed to copy %s to %s", srcroot, destroot)
-}
-
 // CopyDir creates a copy of a dir
-func CopyDir(srcroot, destroot string) error {
-	return filepath.Walk(srcroot, func(path string, info os.FileInfo, err error) error {
+func CopyDir(logger *flogging.FabricLogger, srcroot, destroot string) error {
+	cpErr := filepath.Walk(srcroot, func(path string, info os.FileInfo, err error) error {
```
sykesm: nit: please use the idiomatic name `err`

core/container/externalbuilder/copy.go
```diff
@@ -68,14 +36,63 @@ func CopyDir(srcroot, destroot string) error {
...
+		// its something else that we don't support copying
+		return errors.Errorf("Refusing to copy special file %s", path)
 	})
+
+	// If an error occurred, clean up any created files.
+	if cpErr != nil {
+		rmErr := os.RemoveAll(destroot)
+		if rmErr != nil {
+			logger.Errorf("Failed to clean targeting dir %s: %s", destroot, rmErr)
+		}
```
sykesm: nit: this can use an if with assign to capture the err to `err` and log it.

core/container/externalbuilder/copy.go
```diff
@@ -68,14 +36,63 @@ func CopyDir(srcroot, destroot string) error {
...
+		return errors.WithMessagef(cpErr, "failed to copy %s to %s", srcroot, destroot)
+	}
+	return nil
+}
+
+func copySymlink(srcroot, srcpath, destpath string) error {
+
```
sykesm: nit: unnecessary blank line; please remove to be consistent with the rest of the file.

core/container/externalbuilder/copy.go
```diff
@@ -68,14 +36,63 @@ func CopyDir(srcroot, destroot string) error {
...
+	}
+
+	// Determine where the symlink points to. If it points outside
+	// of the source root, then we do not want to copy it.
+	symlinkDir := filepath.Dir(srcpath)
+	symlinkTarget := filepath.Clean(filepath.Join(symlinkDir, symlinkDest))
+	relativeTarget, err := filepath.Rel(srcroot, symlinkTarget)
+	if err != nil {
+		return err
+	} else if strings.HasPrefix(relativeTarget, "..") {
```
sykesm: we don't use 'else if' after error paths in idiomatic go when it's not associated with the if/assign. In this case, there are two things going on:

1. Check for error from filepath.Rel with fast return
2. Return an error as the destination is not relative

https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

This test is also incorrect. Imagine a file called `..bob`; you're not testing for a path element but a prefix.

core/container/externalbuilder/copy.go
```diff
@@ -68,14 +36,63 @@ func CopyDir(srcroot, destroot string) error {
...
+}
+
+func copySymlink(srcroot, srcpath, destpath string) error {
+
+	// If the symlink is absolute, then we do not want to copy it.
+	symlinkDest, err := os.Readlink(srcpath)
+	if err != nil {
+		return err
+	} else if filepath.IsAbs(symlinkDest) {
+		return errors.Errorf("Refusing to copy absolute symlink %s -> %s", srcpath, symlinkDest)
```
sykesm: error messages are not capitalized.

https://github.com/golang/go/wiki/CodeReviewComments#error-strings

core/container/externalbuilder/copy.go
```diff
@@ -68,14 +36,63 @@ func CopyDir(srcroot, destroot string) error {
...
+
+		// its something else that we don't support copying
+		return errors.Errorf("Refusing to copy special file %s", path)
 	})
+
+	// If an error occurred, clean up any created files.
+	if cpErr != nil {
+		rmErr := os.RemoveAll(destroot)
+		if rmErr != nil {
+			logger.Errorf("Failed to clean targeting dir %s: %s", destroot, rmErr)
```
sykesm: What is a "targeting" dir? I think the message could be improved a bit.

core/container/externalbuilder/copy.go
```diff
@@ -68,14 +36,63 @@ func CopyDir(srcroot, destroot string) error {
...
+		if info.Mode().IsRegular() {
+			if err = copyFile(path, destpath); err != nil {
+				return err
+			}
+			return nil
 		}
-		return nil
+
+		// its something else that we don't support copying
+		return errors.Errorf("Refusing to copy special file %s", path)
```
sykesm: Errors are not capitalized.

https://github.com/golang/go/wiki/CodeReviewComments#error-strings

core/container/externalbuilder/copy.go
```diff
@@ -68,14 +36,63 @@ func CopyDir(srcroot, destroot string) error {
...
-			return err
+		if info.Mode().IsRegular() {
+			if err = copyFile(path, destpath); err != nil {
+				return err
+			}
+			return nil
 		}
-		return nil
+
+		// its something else that we don't support copying
```
sykesm: nit: it's instead of its, right?

Maybe fix the two others in here that were not introduced with this commit?

core/container/externalbuilder/copy_test.go
```diff
@@ -45,51 +58,77 @@ var _ = Describe("copy", func() {
...
+
+			Expect(destRootDir).NotTo(BeAnExistingFile())
 		})
+	})
 
-		It("fails to move and try copy", func() {
-			err = MoveOrCopyDir(logger, srcRootDir, destRootDir)
+	When("Source contains a file other than a regular file, directory, or symlink", func() {
+		It("Returns an error and removes the destination directory", func() {
+			socket, err := net.Listen("unix", filepath.Join(srcRootDir, "uds-in-root.txt"))
```
sykesm, line 153: This is going to make people on windows upset. We have a few choices:

- Ignore the people that develop on Windows
- Move this test into something that logically has the unix build tag
- Skip this test when `runtime.GOOS == "windows"`
- Remove the test entirely

core/container/externalbuilder/copy_test.go
```diff
@@ -45,51 +58,77 @@ var _ = Describe("copy", func() {
 		os.RemoveAll(destRootDir)
 	})
 
-	When("dest dir does not exist", func() {
-		BeforeEach(func() {
-			destRootDir, err = ioutil.TempDir("", "dest-")
-			Expect(err).NotTo(HaveOccurred())
-			err = os.RemoveAll(destRootDir)
+	It("Copies all files and subdirectories", func() {
```
sykesm: The strings inside of context, when, describe, and it are generally *not* capitalized unless they are referring to an identifier in the code. If you look at the other tests in this package you'll see you've not adhered to that convention.

core/container/externalbuilder/copy_test.go
```diff
@@ -19,23 +20,35 @@ import (
...
-		err = os.Chmod(srcSubFile.Name(), os.ModePerm)
+
+		err = os.Symlink("file-in-subdir.txt", filepath.Join(srcRootDir, "subdir", "symlink-in-subdir.txt"))
+		Expect(err).NotTo(HaveOccurred())
+		err = os.Symlink("../file-in-root.txt", filepath.Join(srcRootDir, "subdir", "symlink-to-root.txt"))
+		Expect(err).NotTo(HaveOccurred())
+
+		destRootDir, err = ioutil.TempDir("", "dest-")
 		Expect(err).NotTo(HaveOccurred())
+		err = os.RemoveAll(destRootDir)
```
sykesm: I don't understand why we would create and immediately remove a directory.

The way I'd expect this to go is something like this:

1. Create temporary directory for the test. Let's call this "tempdir".
2. Populate "tempdir/src" with the contents.
3. Reference "tempdir/dest" as the non-existent destination directory.

In the `AfterEach`, cleanup should be as simple as an `os.RemoveAll(tempdir)`.

core/container/externalbuilder/copy.go
sykesm: we don't use 'else if' after error paths in idiomatic go when it's not associated with the if/assign. In this case, there are two things going on:

1. Check for error from os.Readlink with fast return
2. Return an error as the destination is not allowed.

https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

core/container/externalbuilder/copy.go
```diff
@@ -68,14 +36,63 @@ func CopyDir(srcroot, destroot string) error {
 			return nil
 		}
 
+		// filepath.Walk does not follow symbolic links; we need to copy
+		// symbolic links as-is because some chaincode types (Node.js) rely
+		// on the use of symbolic links.
+		if info.Mode()&os.ModeSymlink == os.ModeSymlink {
+			if err = copySymlink(srcroot, path, destpath); err != nil {
```
sykesm: If you're going to use an if/else, please use := to get a new var for the scope of the block. Otherwise, just assign err check it after the call.

That said, the whole check is unnecessary as you're simply returning the result of copySymlink. A simple `return copySymlink(...)` should suffice.

core/container/externalbuilder/copy.go
```diff
@@ -68,14 +36,63 @@ func CopyDir(srcroot, destroot string) error {
...
 		// its a file, copy to corresponding path in the dest.
 		// Intermediate directories are ensured to exist because parent
 		// node is always visited before children in `filepath.Walk`.
-		if err = copyFile(path, destpath); err != nil {
-			return err
+		if info.Mode().IsRegular() {
+			if err = copyFile(path, destpath); err != nil {
+				return err
+			}
+			return nil
```
sykesm: With the introduction of the `IsRegular` check, this should be as simple as `return copyFile(...)`

core/container/externalbuilder/copy.go
```diff
@@ -68,14 +36,63 @@ func CopyDir(srcroot, destroot string) error {
 			return nil
```
sykesm: You didn't write this but we're just returning the result of `os.MkdirAll` so there's no need for the check. A simple `return os.MkdirAll(...)` should suffice.

core/container/externalbuilder/copy.go
sykesm: The entire back-half of this walk function is a switch:

```go
switch {
case info.IsDir():
    return os.MkdirAll(...)
case info.Mode()&os.ModeSymlink == os.ModeSymlink:
    return copySymlink(...)
case info.IsRegular():
    return copyFile(...)
default:
    return errors.New("refusing to copy file %s with mode %s", path, info.Mode())
}
```

There's no need to spread the error handling all over. This would address many of the comments below.

core/container/externalbuilder/copy.go
```diff
@@ -68,14 +36,63 @@ func CopyDir(srcroot, destroot string) error {
...
+
+	// Determine where the symlink points to. If it points outside
+	// of the source root, then we do not want to copy it.
+	symlinkDir := filepath.Dir(srcpath)
+	symlinkTarget := filepath.Clean(filepath.Join(symlinkDir, symlinkDest))
+	relativeTarget, err := filepath.Rel(srcroot, symlinkTarget)
+	if err != nil {
+		return err
+	} else if strings.HasPrefix(relativeTarget, "..") {
+		return errors.Errorf("Refusing to copy symlink %s -> %s pointing outside of source root", srcpath, symlinkDest)
```
sykesm: We don't capitalize error messages.

https://github.com/golang/go/wiki/CodeReviewComments#error-strings

Issue #1237, opened 2016-04-25
go vet yields errors in obcpbft/obc-batch.go


christo4ferris:
go vet yields errors in obcpbft/obc-batch.go

obcpbft/obc-batch.go:64: possible formatting directive in Error call

PR incoming

christo4ferris:
the problem is far more extensive because fabric is using a vendored version of logging package taken before the introduction of separate Errorf etc methods.

PR #1238, opened 2016-04-25
Issue 1237

christo4ferris:
changed logger.Error to logger.Errorf


christo4ferris:

fixes #1237

bcbrock:
@christo4ferris - ~~I think Go vet wanted you to remove the redundant parameter `op.pbft.L` from the call, not change the function being called.~~ I'm sorry I don't know what I'm talking about - I don't know what it was complaining about.

christo4ferris:
@bcbrock neither do I... AFAIK logging.Logger.Errorf is a very real and valid method https://github.com/op/go-logging/blob/master/logger.go line 213. I am baffled.

bcbrock:
@christo4ferris - We are using the "vendor" package: https://github.com/hyperledger/fabric/blob/master/vendor/github.com/op/go-logging/logger.go
This package is not up-to-date. I believe that if we did update it, we would have to change every call of Debug(), Info() etc. in our code to Debugf(), Infof() etc. Maybe @srderson can comment?

In terms of the Go vet error, I'm wondering now if maybe this is a bug where Go vet is not implementing "vendoring" correctly in some cases. The error message from Go vet makes sense in that context.

christo4ferris:
@bcbrock thanks for the explanation... now to hunt down why we are doing this

christo4ferris:
closing this because the correct fix needs much more investment.

PR #1250, opened 2016-04-25
Add basic support for CHAINTOOL

ghaskins:
This PR adds support for chaintool (https://github.com/ghaskins/chaintool) for discussion/proposed-inclusion

In this patch:
- Add support for a new chaincodeType "CAR" (Chaincode Archive) platform
  - Supports the deployment of chaincode emitted from the chaintool tooling
- Adds ./devenv/tools which is intended to be added to $PATH for fabric development environments
- Adds ./devenv/tools to the $PATH under vagrant
- Adds chaintool v0.5.1 java bytecode based binary to ./devenv/tools
- Adds a reference port of "example02" to the chaintool environment
- Adds a unit-test for deploying CAR based chaincode

Todo:
- Add support for querying/browsing the peer for the available interfaces a specific chaincodeID supports.
- Include the source code as part of (or along side of) fabric
- Modify behave to accept chaincodeType during deploy
- Include behave tests for CAR type chaincode

Add tests/behave pass


ghaskins:
@muralisrini

ghaskins:
Hmm, it seems that I cannot rely on the project being available in the same $GOPATH between vagrant and travis.  Will need to rethink how the unit test gets its data.

christo4ferris:
@ghaskins I restarted the build. FlyFi FTW

ghaskins:
@christo4ferris thank you!

ghaskins:
@jeffgarratt one item to discuss is something I had brought up a few weeks ago:  how to best integrate with behave.

Background: my original intention was that the workflow that I am proposing would make chaintool-style deployments the _only_ supported method.  This would mean all deployments are created with chaintool and we no longer have chaincodeType in the deploy method (since its encapsulated in the chaincode archive (CAR)).  However, @muralisrini made the very good point that this may be a bit overzealous in this early stage of the concept and make the patches harder to accept.  So we agreed that a good intermediate (and even long term) option was to simply parameterize the deploy to optionally accept chaintool style deployments while still supporting the current style.  This was implemented by adding a "CAR" enum, as you can see in the diff.

Anyway, the implication of this current approach is that it would require some considerations at the behave level because the current behave grammar assumes chaincodeType=GOLANG.  There are several different approaches we could take to mutate the behave ecosystem to support other types and I was wondering if you have an opinion on how you'd prefer it to be done?

jeffgarratt:
@srderson   @ghaskins Great job of extension without intrusion.  This is a testament your comprehension of the mechanisms of the system.  I look forward to seeing the interest this sparks within the community.  Kudos.

srderson:
@ghaskins I ran `go test -timeout=20m $(go list github.com/hyperledger/fabric/... | grep -v /vendor/ | grep -v /examples/)` and hit the following error

```
2016/05/02 16:12:56 setting Number of procs to -1, was 2
--- FAIL: TestCar_BuildImage (0.56s)
    car_test.go:36: Error getting chaincode package bytes: Error determining chaintool path dynamically
FAIL
FAIL    github.com/hyperledger/fabric/core/chaincode/platforms/car/test 0.590s
```

Could you take a look?

ghaskins:
Already discussed with @srderson in slack, but for anyone following the thread here:  I believe the issue is that this PR requires the vagrant to be destroy/up so that the JRE/$PATH mods are picked up.  Failure to rebuild vagrant would result in the chaintool application appearing unavailable as the error above indicates.  This is the current theory.  Sheehan is going to rebuild his environment and try again.

srderson:
Can confirm that the unit tests pass for me after running `vagrant destory` `vagrant up`. Thanks @ghaskins

srderson:
LGTM. Confirmed all tests are passing. Thanks!

Issue #1223, opened 2016-04-22
Behave tests should detect and shutdown local running peer


binhn:
Often we forget to shutdown the local running peer as part of the unit-tests, and behave tests result in errors. We should try to detect and shutdown the peer prior to running the behave tests.

srderson:
I'd prefer that the unit tests not require a running peer. Possibly the tests that require such functionality should be moved into the behave tests.

binhn:
@srderson ok, we need to investigate which unit-tests require peer and will report back here

christo4ferris:
Agree that unit tests should not require a running peer.

PR #1251, opened 2016-04-25
Fix Issue 1223

lehors:
This simple fix to issue #1223 stops the local running peer if any before starting the behave tests so that it doesn't get in the way.
This depends on the fix to issue-1224 so that "peer stop" can be invoked successfully from the bddtests directory.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,7 +49,6 @@ before_script:
 script:
 
  - echo "Executing Behave test scripts"
- - ps -ef | grep peer | grep -v grep | kill -9 `awk '{print $2}'`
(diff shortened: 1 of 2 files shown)


lehors:
Hmm, the whole build and testing is kind of screwy in my opinion because it is based on running programs from specific directories. The Travis check apparently fails because it doesn't have the peer command in its $PATH.

lehors:
Note that with this fix the clunky command to kill the running peer in .travis.yml can be removed.

christo4ferris:
LGTM

srderson:
Confirmed all tests are passing. LGTM. Thanks!

Issue #996, opened 2016-04-05
Security mode throughput is periodic in security.cert.batch.size


bcbrock:
I was asking questions on Slack about the poor system-level throughput with security enabled. @muralisrini suggested setting security.cert.batch.size to 1000 and ignoring the throughput of the initial transactions that "bring down the tcerts". In doing these experiments on a Docker-compose network consisting of a single validating peer running NOOPS + OBC-CA, I observed the behavior shown in the following graph. 

Transactions are carefully grouped by the driver into groups of 200 (including invoke, deploy and query). Groups that contain the 1st, 1001st, 2001st etc. transactions have the lowest throughput. Surprisingly, the throughput increases dramatically as the transaction indices increase within a group of 1000.  I am not familiar with the algorithms involved but this looks like some kind of O(N) algorithm is being run based on the transaction index. Looking at `top` it also appears that the utilization decreases in the obc-peer process as the transaction index modulo 1000 increases. 

Unfortunately the throughput is only better with higher indexes, and is still never very good. I've opened this issue in the hope that it might prompt someone to figure out and fix part of the performance shortfall.

Note: I'm still using recent OBC code; I haven't converted my drivers to HyperLedger yet.

![txrate](https://cloud.githubusercontent.com/assets/11968347/14300478/3b343bd0-fb56-11e5-9a7a-1e4d8a7a7913.png)

bcbrock:
More details: There is a single simple chaincode, similar to chaincode_example02 (but slightly different).  The first block of 200 TX is 1 deploy, 198 invokes, 1 query.  All but the first block of 200 TX consist of 199 invokes plus 1 query. [The queries are correctness checks at the end]. Throughput is measured for each complete block of  200 TX - each block of 200 TX is considered as a separate test against a network that remains deployed between tests.

IBMers can recreate by
1. git clone https://github.ibm.com/bcbrock/busywork
2. Install as directed in the README.md
3. cd ... bcbrock/busywork/counters
4. make issue996

adecaro:
@bcbrock, which golang version are you using?

bcbrock:
@adecaro @muralisrini I am using golang version 1.6rc2 (built from source before Go 1.6 was officially released; I should upgrade this) on a 4-socket X86 server. This test is essentially the only work running on the system.

[...]

PR #1256, opened 2016-04-26
Close clients

adecaro:
We now distinguish between long-living and short-living client instances. 

A short-living client instance is one that is initialized (InitShortLivingClient), used (for a limited amount of time) and discarded. For this kind of instances there is no need to invoke CloseClient. An internal garbage collector will reclaim them if the time elapsed from the last initialization is more than a certain lower-bound (that can be specified using the property 'security.client.refreshTimePeriod' – an integer that represents the number of minutes).

On the other hand, a long-living client instance (InitClient) is one that can be used indefinitely and therefore CloseClient needs to be invoked explicitly when the instance is not required anymore and the 
resources can be released.

This PR introduces the differentiation described above and addresses issues #811, #996

### DIFF
--- a/core/crypto/client.go
+++ b/core/crypto/client.go
@@ -227,7 +227,7 @@ func initClientLayer() {
 		done = make(chan struct{})
 
 		// Start client instances cleaner
-		go clientInstancesCleaner()
+		//go clientInstancesCleaner()
 
 		initialized = true
 		log.Debug("Initilize client layer...done!")
--- a/core/crypto/client.go
+++ b/core/crypto/client.go
@@ -22,16 +22,22 @@ package crypto
 import (
 	"github.com/hyperledger/fabric/core/crypto/utils"
 	"sync"
+	"time"
 )
 
 // Private Variables
 
 type clientEntry struct {
-	client  Client
-	counter int64
+	client    Client
+	counter   int64
+	timestamp time.Time
 }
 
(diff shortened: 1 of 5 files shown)


christo4ferris:
restarted build

christo4ferris:
re-running TravisCI build

bcbrock:
@adecaro - Do you think this is really working in spite of the Travis failure? If so I can try it probably Thursday afternoon or Friday.

adecaro:
@christo4ferris @bcbrock It was okay on my machine. Let me check more deeply.

adecaro:
@christo4ferris @bcbrock, actually the only error I see is the following:

'Max time exceeded waiting for transactions with current response map....'

Moreover, the number of behave tests that fail is always the same. 
I don't see yet how this can be linked to the modification I did. 

Does the error tell you anything?

christo4ferris:
re-running Travis one more time.

adecaro:
I just had the master failing the behave tests. Looking at the logs I'm not really able to figure it out why. I can only say that I haven't see any timeout issue.

christo4ferris:
@adecaro third time is not a charm... same failure fingerprint as others. It will need to be run locally and logs inspected.

adecaro:
I close this PR to better investigate what is wrong with the tests!

Issue #904, opened 2016-03-30
On error, need to stop peer rather than return with error


tuand27613:
As noted in https://github.com/openblockchain/obc-peer/commit/246d4311ef50f1e845fea15fb3b9914669b74995#commitcomment-16896253,

there are cases where a caller won't know how to recover given a error return, see
https://github.com/openblockchain/obc-peer/blob/master/openchain/ledger/ledger.go#L70
or 
https://github.com/openblockchain/obc-peer/blob/master/openchain/ledger/ledger.go#L62

A caller will not know how to recover at these points. Ledger ( or any other component ) needs
to recover internally or stop

PR #1258, opened 2016-04-26
added check for error

### LINKED ISSUE #1257: test fails to catch nil ledger
core/chaincode/exectransaction_test.go fails to catch possible nil ledger

christo4ferris:
added check for error in ledger.Getledger()

Fixes #1257

### DIFF
--- a/core/chaincode/exectransaction_test.go
+++ b/core/chaincode/exectransaction_test.go
@@ -73,6 +73,9 @@ func deploy(ctx context.Context, spec *pb.ChaincodeSpec) ([]byte, error) {
 	}
 
 	ledger, err := ledger.GetLedger()
+	if err != nil {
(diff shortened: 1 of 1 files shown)


christo4ferris:
re-ran build - original failure was due to known occasional pbft test failure.

corecode:
I don't think this additional conditional is needed.  The test will fail either way.  Under which condition could `GetLedger` even fail?

christo4ferris:
when/if someone changes it. this is cleaner, IMO

srderson:
Ideally I guess we would have used `_` instead of `err` if we wanted to ignore it. I personally agree with @christo4ferris that it's more readable with the check, but I see how opinions could differ here.

It's a moot point because we have issue #904 so eventually `ledger.GetLedger()` will not return an error. So LGTM until #904 is resolved and then this will be removed as part of that change.

PR #1261, opened 2016-04-26
Added StackOverflow communication channel

mcr222:
To avoid having slack filled with the same questions over and over, a new tag for StackOverflow has been created.

The more the people uses it less answers will have to be repeated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,7 +137,9 @@ up at the IP address currently assumed by the test environment
 We welcome contributions to the Hyperledger Project in many forms. There's always plenty to do! Full details of how to contribute to this project are documented in the [CONTRIBUTING.md](CONTRIBUTING.md) file.
 
 ## Communication <a name="communication"></a>
-We use [Hyperledger Slack](https://slack.hyperledger.org/)  for communication and Google Hangouts&trade; for screen sharing between developers.
+We use [Hyperledger Slack](https://slack.hyperledger.org/) for communication and Google Hangouts&trade; for screen sharing between developers.
+
+For general purpose questions, use [StackOverflow](http://stackoverflow.com/questions/tagged/hyperledger). Even if questions are answered through Hyperledger Slack, please post them to StackOverflow so answers don't have to be repeated continuously.
 
 ## Coding Golang <a name="coding"></a>
 - We require a file [header](docs/dev-setup/headers.txt) in all source code files. Simply copy and paste the header when you create a new file.


mcr222:
Signed-off-by: Marc Cayuela Rafols markitus222@hotmail.com

christo4ferris:
As this is a README change, will forgo re-running the TravisCI build.

Thanks for adding the tag @mcr222! subscribed;-) 
This change LGTM

srderson:
LGTM

core/ledger/kvledger/txmgmt/statedb/commontests/test_common.go
```diff
@@ -980,3 +982,112 @@ func TestApplyUpdatesWithNilHeight(t *testing.T, dbProvider statedb.VersionedDBP
...
+	t *testing.T,
+	dbProvider statedb.VersionedDBProvider,
+	valueFormat byte,
+	dbValueDeserializer func(b []byte) (*statedb.VersionedValue, error)) {
+
+	db, err := dbProvider.GetDBHandle("test-full-scan-iterator")
+	assert.NoError(t, err)
+
+	// generateSampleData returns a slice of KVs. The returned value contains five KVs for each of the namespaces
+	generateSampleData := func(namespaces ...string) []*statedb.VersionedKV {
```
cendhu, line 996: In other tests, we have made similar utils as a separate function. We could do the same here. At the first read, it was really a big distraction when I just wanted to understand the test logic at high level before diving into details.

core/ledger/kvledger/txmgmt/statedb/statedb.go
```diff
@@ -65,6 +65,12 @@ type VersionedDB interface {
 	// BytesKeySupported returns true if the implementation (underlying db) supports the any bytes to be used as key.
 	// For instance, leveldb supports any bytes for the key while the couchdb supports only valid utf-8 string
 	BytesKeySupported() bool
+	// GetFullScanIterator returns a FullScanIterator that can be used to iterate over entire data in the statedb.
+	// `skipNamespace` parameter can be used to control if the consumer wants the FullScanIterator
+	// to skip one or more namespaces from the returned results. The second parameter returns the format information
+	// about the value bytes returned by the Next function in the returned FullScanIterator.
+	// The intended use of this iterator is to generate the snapshot files for the statedb.
+	GetFullScanIterator(skipNamespace func(string) bool) (FullScanIterator, byte, error)
```
cendhu, line 73: Clarification questions: 

1. When do we use `skipNamespace`? 
   - If the goal is to use it when we resume snapshot processing after a failure to avoid duplication in effort, shouldn't we track at the key level? I understand that `common/ledger/snapshot` pkg requires update to find out which key was written successfully last. Further, there are some challenges in finding that by flushing writes periodically, tracking the offset of last written key, etc... 
   - Is it introduced to skip private data, i.e., `namespace = chaincodeName + pvtdataprefix + collectionName`? 

2. Why do we pass `skipNamespace func(string) bool` instead of `map[string]bool`? Assuming that `skipNamespace` is to skip pvtdata. 
   - The stateDB does not know whether a given namespace belong to a private data or hashed data or public data. Only the privacyenabledstate pkg would know. Hence, we make the statedb to rely on the logic present in the privacyenabledstate pkg by passing a function as an argument. Am I correct?
   - Using the collection config store, can't we retrieve all collections and pass the namespaces instead of a function? I assume that passing 10s/100s/1000s of collection names in `map[string]bool` is not preferred and hence, we go with `func (string) bool`. Is my understanding correct?
   - Another option is to pass all chaincode namespaces retrieved from deployedCCProvider. However, this might create additional dependencies and again the list could be large.

core/ledger/kvledger/txmgmt/statedb/statedb.go
```diff
@@ -86,6 +92,16 @@ type IndexCapable interface {
 	ProcessIndexesForChaincodeDeploy(namespace string, indexFilesData map[string][]byte) error
 }
 
+// FullScanIterator provides a mean to iterate over entire statedb. The intended use of this iterator
+// is to generate the snapshot files for the statedb
+type FullScanIterator interface {
+	// Next returns the key-values in the lexical order of <Namespace, key>
+	// A particular statedb implementation is free to chose any deterministic bytes representation for the <version, value, metadata>
+	Next() (*CompositeKey, []byte, error)
```
cendhu, line 100: Why do we return `CompositeKey, []byte` instead of `VersionedKV`?

1. In leveldb, we store `version`, `value`, and `metadata` using a proto message called `DBValue`. We do know that we need to store marshalled proto message in the snapshot file. As you responded in one of the comment #1231, is this the optimization and also the reason for exposing `EncodeBytes()` from `common/ledger/snapshot` pkg?   
2. For couchdb, it might be bit complicated as `couchDoc.jsonvalue []byte` stores `_id`,` _rev`, `valueBytes`, `~version`, `_deleted`.  Further, it has `couchDoc.attachments` too. While the proposed optimisation might be applicable for leveldb, need to think a bit more to analyse the benefits with the couchdb. 

Returning of `VersionedKV` and using a proto at `privacyenabledstate` pkg looks cleaner to me but it nullifies the optimization for the leveldb. Also, I am not sure how much improvement we get from this optimization. Anyway, I can revisit when exporting CouchDB data.

core/ledger/kvledger/txmgmt/statedb/stateleveldb/stateleveldb.go
```diff
@@ -21,10 +21,12 @@ import (
...
-	dataKeyPrefix    = []byte{'d'}
-	nsKeySep         = []byte{0x00}
-	lastKeyIndicator = byte(0x01)
-	savePointKey     = []byte{'s'}
+	dataKeyPrefix               = []byte{'d'}
+	dataKeyStopper              = []byte{'e'}
+	nsKeySep                    = []byte{0x00}
+	lastKeyIndicator            = byte(0x01)
+	savePointKey                = []byte{'s'}
+	fullScanIteratorValueFormat = byte(1)
```
cendhu, line 29: Clarification Question:

We have the dataFormat defined at the individual `statedb` pkg instead of `privacyenabledstate` pkg such  that each `statedb` can have independent format changes. Is that correct?

core/ledger/kvledger/txmgmt/statedb/stateleveldb/stateleveldb.go
```diff
@@ -268,3 +284,64 @@ func (scanner *kvScanner) GetBookmarkAndClose() string {
...
+	dbItr := db.GetIterator(dataKeyPrefix, dataKeyStopper)
+	if err := dbItr.Error(); err != nil {
+		return nil, byte(0), errors.Wrap(err, "internal leveldb error while obtaining db iterator")
+	}
+	return &fullDBScanner{
+			db:            db,
+			dbItr:         dbItr,
+			skipNamespace: skipNamespace,
+		},
+		fullScanIteratorValueFormat,
```
cendhu, line 305: We could move this variable to `fullDBScanner` right? because it is the format returned by the scanner.

core/ledger/kvledger/txmgmt/statedb/stateleveldb/stateleveldb.go
```diff
@@ -268,3 +284,64 @@ func (scanner *kvScanner) GetBookmarkAndClose() string {
...
+}
+
+// Next returns the key-values in the lexical order of <Namespace, key>
+// The bytes returned for the <version, value, metadata> are the same as they are stored in the leveldb.
+// Since, the primary intended use of this function is to generate the snapshot files for the statedb, the same
+// bytes can be consumed back as is. Hence, we do not decode or transform these bytes for the efficiency
+func (s *fullDBScanner) Next() (*statedb.CompositeKey, []byte, error) {
+	for s.dbItr.Next() {
+		dbKey := s.dbItr.Key()
+		logger.Debugw("", "dbKey", dbKey)
```
cendhu: forgot to remove?

core/ledger/kvledger/txmgmt/statedb/stateleveldb/stateleveldb.go
```diff
@@ -268,3 +284,64 @@ func (scanner *kvScanner) GetBookmarkAndClose() string {
...
+		ns, key := decodeDataKey(dbKey)
+		compositeKey := &statedb.CompositeKey{
+			Namespace: ns,
+			Key:       key,
+		}
+
+		switch {
+		case ns == s.currentNamespace:
+			return compositeKey, dbVal, nil
+
```
cendhu: Nit. Not sure whether we need an empty line between cases.

core/ledger/kvledger/txmgmt/statedb/stateleveldb/stateleveldb.go
```diff
@@ -268,3 +284,64 @@ func (scanner *kvScanner) GetBookmarkAndClose() string {
...
+		// new namespace begins
+		case !s.skipNamespace(ns):
```
cendhu: `s.skipNamespace(ns)` sounds like asking the `fullDBScanner` to skip the `ns` namespace. To avoid confusion, it can be renamed to `s.isNamespaceToBeSkipped(ns)` or `s.isNamespaceRequested()` or something similar.

core/ledger/kvledger/txmgmt/statedb/commontests/test_common.go
```diff
@@ -980,3 +982,112 @@ func TestApplyUpdatesWithNilHeight(t *testing.T, dbProvider statedb.VersionedDBP
 	assert.Equal(t, savePoint, ht) // savepoint should still be what was set with batch1
 	// (because batch2 calls ApplyUpdates with savepoint as nil)
 }
+
+func TestFullScanIterator(
+	t *testing.T,
+	dbProvider statedb.VersionedDBProvider,
+	valueFormat byte,
+	dbValueDeserializer func(b []byte) (*statedb.VersionedValue, error)) {
```
cendhu, line 990: When we export couchDB, we might need to create a new proto for returning version, metadata, and value. Again, we might need to create a proto including the key and versioned value at the privacyenabledstate. Given that metadata and value are []byte, I am not really sure about the performance gain that we get with reusing the serialized dbValue. May be some extra memory copy might happen for raw bytes -- https://github.com/golang/protobuf/blob/master/proto/buffer.go#L168 but only when we have the complete code, we could understand the benefits.

core/ledger/kvledger/txmgmt/statedb/commontests/test_common.go
```diff
@@ -980,3 +982,112 @@ func TestApplyUpdatesWithNilHeight(t *testing.T, dbProvider statedb.VersionedDBP
...
+			}
+			versionedVal, err := dbValueDeserializer(serializedVersionedValue)
+			require.NoError(t, err)
+			results = append(results, &statedb.VersionedKV{
+				CompositeKey:   *compositeKV,
+				VersionedValue: *versionedVal,
+			})
+		}
+
+		expectedNamespacesInResults := stringset{"ns1", "ns2", "ns3", "ns4"}.minus(skipNamespaces)
```
cendhu, line 1048: Similar to other tests in ledger package, we can move this logic to a local `testFullScanIterator()` and pass `skipNamespaces` and `expectedResults` as a parameters along with others. As a result, we can get rid of `minus()` too. One drawback is that we might pass 6 parameters. 

After going over the details, I find that these sub-functions are very dependent on other parameters used in the main function to reduce the number of lines of code. I find this test a bit more complex than required.

core/ledger/kvledger/txmgmt/statedb/commontests/test_common.go
```diff
@@ -980,3 +982,112 @@ func TestApplyUpdatesWithNilHeight(t *testing.T, dbProvider statedb.VersionedDBP
...
+		expectedResults := []*statedb.VersionedKV{}
+		for _, ns := range expectedNamespacesInResults {
+			expectedResults = append(expectedResults, generateSampleData(ns)...)
```
cendhu, line 1051: Clarification Question:

so the `generateSampleData()` needs to be deterministic and hence not it is not created as a separate function?

core/ledger/kvledger/txmgmt/statedb/stateleveldb/stateleveldb.go
```diff
@@ -217,6 +228,11 @@ func decodeDataKey(encodedDataKey []byte) (string, string) {
 	return string(split[0][1:]), string(split[1])
 }
 
+func dataKeyStarterForNextNamespace(ns string) []byte {
```
cendhu, line 231: Nit. 

```suggestion
func dataKeyStarterForNextNamespace(currentNamespace string) []byte {
```

core/ledger/kvledger/txmgmt/statedb/stateleveldb/stateleveldb_test.go
```diff
@@ -125,3 +126,88 @@ func TestApplyUpdatesWithNilHeight(t *testing.T) {
...
+	batch.Put("ns3", "key3", []byte("value3"), version.NewHeight(1, 1))
+	vdb.ApplyUpdates(batch, version.NewHeight(2, 2))
+
+	itr, _, err = vdb.GetFullScanIterator(
+		func(ns string) bool {
+			return ns == "ns2"
+		},
+	)
+
+	compositeKey, _, err := itr.Next()
```
cendhu, line 203: missed `required.Nil(t, err)`?

core/ledger/kvledger/txmgmt/statedb/stateleveldb/stateleveldb_test.go
```diff
@@ -125,3 +126,88 @@ func TestApplyUpdatesWithNilHeight(t *testing.T) {
...
+
+	// error from function Next when switching to new iterator for skipping a namespace
+	reInitEnv()
+	batch := statedb.NewUpdateBatch()
+	batch.Put("ns1", "key1", []byte("value1"), version.NewHeight(1, 1))
+	batch.Put("ns2", "key2", []byte("value2"), version.NewHeight(1, 1))
+	batch.Put("ns3", "key3", []byte("value3"), version.NewHeight(1, 1))
+	vdb.ApplyUpdates(batch, version.NewHeight(2, 2))
+
+	itr, _, err = vdb.GetFullScanIterator(
```
cendhu, line 197: missed `required.Nil(t, err)`?

core/ledger/kvledger/txmgmt/statedb/stateleveldb/stateleveldb_test.go
```diff
@@ -125,3 +126,88 @@ func TestApplyUpdatesWithNilHeight(t *testing.T) {
...
+	_, _, err := vdb.GetFullScanIterator(
+		func(string) bool {
+			return false
+		},
+	)
+	require.Contains(t, err.Error(), "internal leveldb error while obtaining db iterator:")
+
+	// error from function Next
+	reInitEnv()
+	itr, _, err := vdb.GetFullScanIterator(
```
cendhu, line 179: missed `required.Nil(t, err)`?

core/ledger/kvledger/txmgmt/statedb/stateleveldb/stateleveldb.go
```diff
@@ -268,3 +284,64 @@ func (scanner *kvScanner) GetBookmarkAndClose() string {
...
+		// skip the new namespace
+		default:
+			s.dbItr.Release()
+			s.dbItr = s.db.GetIterator(dataKeyStarterForNextNamespace(ns), dataKeyStopper)
+			if err := s.dbItr.Error(); err != nil {
+				return nil, nil, errors.Wrapf(err, "internal leveldb error while obtaining db iterator for skipping a namespace [%s]", ns)
+			}
+		}
+	}
+	return nil, nil, errors.Wrap(s.dbItr.Error(), "internal leveldb error while retrieving data from db iterator")
```
cendhu, line 339: At the first read, it looked as if we are returning an error when the iter is exhausted. Then, I looked deeply into the `errors.Wrap()` and understood. I think, a similar statements were used in #1231. Explicitly checking for errors and returning one might avoid this confusion. Anyway, at least I will not get confused going forward.

PR #1268, opened 2016-04-27
Issue 1266

### LINKED ISSUE #1266: golint warnings in peer/main.go
should fix golint warnings in main.go

christo4ferris:
fix golint errors

Fixes #1266

main.go:578:9: if block ends with a return statement, so drop this else
and outdent its block
main.go:586:10: if block ends with a return statement, so drop this
else and outdent its block
main.go:630:10: if block ends with a return statement, so drop this
else and outdent its block
main.go:736:10: should omit 2nd value from range; this loop is
equivalent to `for k := range 
…`

### DIFF
--- a/bddtests/out.txt
+++ b/bddtests/out.txt
@@ -1,0 +1,1057 @@
+Feature: lanching 3 peers # peer_basic.feature:11
+  As a HyperLedger developer
+  I want to be able to launch a 3 peers
+  Scenario: Peers list test, single peer issue #827                            # peer_basic.feature:16
+    Given we compose "docker-compose-1.yml"                                    # steps/peer_basic_impl.py:95
+    And I wait "1" seconds                                                     # steps/peer_basic_impl.py:132
+    When requesting "/network/peers" from "vp0"                                # steps/peer_basic_impl.py:109
+    Then I should get a JSON response with array "peers" contains "1" elements # steps/peer_basic_impl.py:125
+
+  Scenario: Peers list test,3 peers issue #827                                 # peer_basic.feature:23
+    Given we compose "docker-compose-3.yml"                                    # steps/peer_basic_impl.py:95
+    And I wait "1" seconds                                                     # steps/peer_basic_impl.py:132
+    When requesting "/network/peers" from "vp0"                                # steps/peer_basic_impl.py:109
+    Then I should get a JSON response with array "peers" contains "3" elements # steps/peer_basic_impl.py:125
+
+  @wip @issue_767
+  Scenario: Range query test, single peer, issue #767                                                           # peer_basic.feature:32
+    Given we compose "docker-compose-1.yml"                                                                     # None
+    And I wait "1" seconds                                                                                      # None
+    When requesting "/chain" from "vp0"                                                                         # None
+    Then I should get a JSON response with "height" = "1"                                                       # None
+    When I deploy chaincode "github.com/hyperledger/fabric/examples/chaincode/go/map" with ctor "init" to "vp0" # None
+      |  |
+      |  |
+    Then I should have received a chaincode name                                                                # None
+    Then I wait up to "60" seconds for transaction to be committed to all peers                                 # None
+    When requesting "/chain" from "vp0"                                                                         # None
+    Then I should get a JSON response with "height" = "2"                                                       # None
+    When I invoke chaincode "map" function name "put" on "vp0"                                                  # None
+      | arg1 | arg2   |
+      | key1 | value1 |
+    Then I should have received a transactionID                                                                 # None
+    Then I wait up to "25" seconds for transaction to be committed to all peers                                 # None
+    When requesting "/chain" from "vp0"                                                                         # None
+    Then I should get a JSON response with "height" = "3"                                                       # None
+    When I query chaincode "map" function name "get" on "vp0"                                                   # None
+      | arg1 |
+      | key1 |
+    Then I should get a JSON response with "OK" = "value1"                                                      # None
+    When I invoke chaincode "map" function name "put" on "vp0"                                                  # None
+      | arg1 | arg2   |
+      | key2 | value2 |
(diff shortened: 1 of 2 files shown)


christo4ferris:
closing in deference to #1281

core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb.go
```diff
@@ -207,6 +219,14 @@ func newVersionedDB(couchInstance *couchInstance, redoLogger *redoLogger, dbName
 		return nil, err
 	}
 
+	dbsInfo, err := vdb.getLatestNamespaceDBsInfo()
+	if err != nil {
+		return nil, err
+	}
+	for _, dbInfo := range dbsInfo {
+		vdb.namespaceDBsInfo[dbInfo.Namespace] = dbInfo
+	}
```
cendhu: When `dbsInfo` is nil while executing `newVersionedDB()`, shouldn't we create `channelMetadata` with either the channelName or metadataDBName to channelName mapping and store an entry?

**Reasons**
1. Without maintaining the mapping between dbName and full namespace, we can find the per channel databases by enhancing `DeployedChaincodeInfoProvider`. However, to avoid the extra dependencies, reduce complexity in peer rollback/reset, and make the `statecouchdb` pkg self-sufficient, we store the mapping explicitly. Similarly, we can store the full channelName associated with the metadataDB as well to avoid any future problems. 
2. For example, say that we want to find all channels and namespaces maintained by the couchDB provider without depending on any other pkgs. This can be done by
   -  issuing `curl -X GET http://127.0.0.1:5984/_all_dbs`
   -  reading channelMetadata from each DB. If channelMetadata does not exist, it might either be a namespace db or orphan database (as a result of failure) or other internal databases.
   -  finding the channel name and associated namespaces by looking at the channelMetadata.

core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb.go
```diff
@@ -207,6 +219,14 @@ func newVersionedDB(couchInstance *couchInstance, redoLogger *redoLogger, dbName
...
+
```
cendhu: This code appears suddenly inbetween redo log reply logic and break the existing flow. Can you move these line to 201?

core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb.go
```diff
@@ -238,6 +258,9 @@ func (vdb *VersionedDB) getNamespaceDBHandle(namespace string) (*couchDatabase,
 	namespaceDBName := constructNamespaceDBName(vdb.chainName, namespace)
 	vdb.mux.Lock()
 	defer vdb.mux.Unlock()
+	if err := vdb.updateAndSaveNamespaceDBsInfo(namespace, namespaceDBName); err != nil {
+		return nil, err
+	}
```
manish-sethi: It will be clearer if you move this inside the check ` if db == nil` at line 265 below.

core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb_test.go
```diff
@@ -1365,3 +1365,109 @@ func TestMissingRevisionRetrievalFromCache(t *testing.T) {
...
+			DBName:    constructNamespaceDBName(channelName, ns),
+		}
+	}
+	savedChannelMetadata, err = vdb.getSavedChannelMetadata()
+	require.NoError(t, err)
+	require.Equal(t, expectedChannelMetadata, savedChannelMetadata)
+	require.Equal(t, expectedChannelMetadata, vdb.channelMetadata)
+}
+
+func testChannelMetadataNegativecases(t *testing.T) {
```
manish-sethi: Noop - `TestChannelMeta...`

core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb_test.go
```diff
@@ -1365,3 +1365,109 @@ func TestMissingRevisionRetrievalFromCache(t *testing.T) {
...
+	namepsaces := make([]string, 10)
+	for i := 0; i < 10; i++ {
+		namepsaces[i] = fmt.Sprintf("nsname_%d", i)
+		_, err := vdb.getNamespaceDBHandle(namepsaces[i])
+		require.NoError(t, err)
+		savedChannelMetadata, err := vdb.getSavedChannelMetadata()
+		require.NoError(t, err)
+		require.Equal(t, i+2, len(savedChannelMetadata.NamespaceDBsInfo))
+		require.Contains(t, savedChannelMetadata.NamespaceDBsInfo, namepsaces[i])
+		require.Equal(t, vdb.channelMetadata, savedChannelMetadata)
```
manish-sethi: Do these assertions add any value, given what you have line 1411 onward?

core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb_test.go
```diff
@@ -1365,3 +1365,109 @@ func TestMissingRevisionRetrievalFromCache(t *testing.T) {
...
+	}
+
+	// call getNamespaceDBHandle for existing dbs, verify that no new db names are added to dbMetadataMapping
+	for _, ns := range namepsaces {
+		_, err := vdb.getNamespaceDBHandle(ns)
+		require.NoError(t, err)
+		require.Equal(t, 11, len(vdb.channelMetadata.NamespaceDBsInfo))
+		savedChannelMetadata, err := vdb.getSavedChannelMetadata()
+		require.NoError(t, err)
+		require.Equal(t, vdb.channelMetadata, savedChannelMetadata)
```
manish-sethi: Same comment as above.

PR #1269, opened 2016-04-27
add custom options for start docker controller

ckeyer:
#1228


lehors:
Hi, LGTM but it would be nice to have that documented somewhere. Not to put the blame on you because I think this project is quite deficient when it comes to documentation.

lehors:
Thanks!

muralisrini:
@ckeyer, @lehors :  One suggestion.

I think it would be better if we can take the viper lookups out off dockercontroller.go
- other files/packages that need these properties would have to duplicate property lookup code
- we are spreading usage of viper access when we should really be consolidating them via a config package or something

An approach could be to wrap docker.Client returned by NewDockerClient in dockerutil.go in a new object with all the global properties read from core.yaml for the docker.Client. These properties will then be available to all users via the wrapped object.  NewDockerClient already reads some of these properties today but does not have to return them (they are directly set on docker.Client). So this really is more of the same thing. 

NewDockerClient would return a DockerClient object from the util package instead of returning docker.Client directly where DockerClient would be something like
type DockerClient struct {
      Client docker.Client
      Dns string
      NetworkMode string
}

What do you think ?

mastersingh24:
@ckeyer - first, thanks for doing this.  But I worry that there will be other options which people want to set.  I agree with @muralisrini about unifying access to config properties as well.  Specific to this change, if we are going to allow people to set custom HostConfig properties, then I would suggest that we actually externalize the entire set of options for HostConfig into a single config parameter.  I would also suggest that we potentially externalize that to a separate config file referenced by core.yaml.  The go-dockerclient which we are using actually supports JSON and YAML bindings for the various structs that it uses - e.g. https://godoc.org/github.com/fsouza/go-dockerclient#HostConfig

We could have a config file specific to docker and cover all properties we might want to make configurable.   Of course at some point, we'll need to have policies for conformity across config, but I hate to see us one-off ever new property we want to make configurable

mastersingh24:
@ckeyer - realized that perhaps my comment seemed like we were trying to do too much here in terms of the centralizing of config part.  I think that probably deserves a different (perhaps larger) story / set of stories.

In this case, I am suggesting that you modify _core.yaml_ as follows:

```
vm:

    # Endpoint of the vm management system.  For docker can be one of the following in general
    # unix:///var/run/docker.sock
    # http://localhost:2375
    # https://localhost:2376
    endpoint: unix:///var/run/docker.sock

    # settings for docker vms
    docker:
        hostConfig:
            Dns:
                - 192.168.1.2
                - 192.168.1.3
            NetworkMode: bridge    
        tls:
            enabled: false
            cert:
                file: /path/to/server.pem
            ca:
                file: /path/to/ca.pem
            key:
                file: /path/to/server-key.pem
```

(basically creating a _hostConfig_ section so we can support other hostConfig properties) and then in your code you can do something like

```
    var hostConfig *docker.HostConfig

    err := viper.UnmarshalKey("vm.docker.hostConfig", &hostConfig)
    if err != nil {
        fmt.Println("error:\n", err)
    } else {
        fmt.Printf("hostConfig:\n%+v", hostConfig)
    }
```

mastersingh24:
@ckeyer  - glad you found my suggestions useful and thanks for making the change
LGTM

peer/core.yaml
```diff
@@ -222,7 +222,7 @@ peer:
         enabled:    true
 
         # number of workers that
-        # test the peers for being
+        # tastes the peers for being
```
ghaskins: typo

peer/core.yaml
```diff
@@ -41,7 +41,7 @@ logging:
     # defined at runtime and are not checked for validity during option
     # processing.
 
-    # Default logging levels are specified here for each of the peer
+    # Default logging levels are specified here for each of the obc-peer
```
srderson: 'obc-peer' should just be 'peer'

core/container/dockercontroller/dockercontroller.go
```diff
@@ -26,16 +26,32 @@ import (
...
+	if err != nil {
+		dockerLogger.Fatalf("Load docker HostConfig wrong, error: %s", err.Error())
+	}
+
+	if hostConfig.NetworkMode == "" {
+		hostConfig.NetworkMode = "host"
+	}
+	dockerLogger.Debug("Load docker HostConfig: %+v", hostConfig)
+}
+
```
muralisrini, line 58: Looks like Unmarshall and pass the docker.HostConfig directly to docker will allow users to add any of its 48 properties directly in core.yaml to control container process in a SUPPORTED fashion.  ("Privileged" is one such property.)

My suggestion would be to use only the specific properties such as NetworkMode, Dns that are declared in the yaml to construct hostConfig (unmarshalling is ok for convenience but filter in only supported properties).

ckeyer, line 58: #1228 like that?

muralisrini, line 58: Yes like that but combined with this work to read into a HostConfig in init(). For example something like

```
func init() {
    networkMode := viper.GetString("vm.docker.networkMode")
    if networkMode == "" {
        networkMode = "host"
    }
    dns := viper.GetStringSlice("vm.docker.dns")        //use only supported properties
    hostConfig = docker.HostConfig{}
    hostConfig.NetworkMode = networkMode
    hostConfig.Dns = dns

    dockerLogger.Debug("Load docker HostConfig: %+v", hostConfig)
}
```

Will note that in my first comment, I was going for wrapping the docker.Client in another struct so we can embed the HostConfig in the wrapped struct.

But even referring to a global HostConfig constructed in init() as above is a step in the right direction.

muralisrini:
@ckeyer any update on my suggestion ? basically add @mastersingh24 idea to setup config in init along with yours of pulling only host and dns.

mastersingh24:
@ckeyer - to make things easier, I created a pull request against your fork rebasing on later version of fabric -- https://github.com/ckeyer/fabric/pull/2

mastersingh24:
travis failing due to #1887

mastersingh24:
restart travis

binhn:
@ckeyer @muralisrini @mastersingh24 We should ignore the failure on `Failing scenarios:
  peer_basic.feature:967  chaincode example02 with 4 peers, two stopped` since it is a new test for PBFT.

yeasy:
Has anyone verified this?

I set "CORE_VM_DOCKER_HOSTCONFIG_NETWORKMODE=bridge" in the peer node.

However, the started chaincode container still got the networkmode="host".

Am I missing something?

Thanks!

ckeyer:
@yeasy  A bug #2140

yeasy:
@ckeyer
Still not works, let's keep updating this important feature.
Thanks!

ckeyer:
@yeasy  because `viper.UnmarshalKey` not support the Environment Variables, I have modified.

PR #1275, opened 2016-04-27
modify docs about ledger.state.dataStructure.name

ckeyer:

### DIFF
--- a/peer/core.yaml
+++ b/peer/core.yaml
@@ -344,9 +344,11 @@ ledger:
     deltaHistorySize: 500
 
     # The data structure in which the state will be stored. Different data
-    # structures may offer different performance characteristics. Options are
-    # 'buckettree' and 'trie'. If not set, the default data structure is the
-    # 'buckettree'. This CANNOT be changed after the DB has been created.
+    # structures may offer different performance characteristics. 
+    # Options are 'buckettree', 'trie' and 'raw'.
+    # ( Note:'raw' is experimental and incomplete. )
+    # If not set, the default data structure is the 'buckettree'.
+    # This CANNOT be changed after the DB has been created.
     dataStructure:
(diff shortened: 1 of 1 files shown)


ckeyer:
This is not match https://github.com/hyperledger/fabric/blob/master/core/ledger/statemgmt/state/state.go#L58-L67

christo4ferris:
re-running Travis even though this is just a doc change. Otherwise, LGTM thanks @ckeyer

srderson:
Does the default of 'bucketTree' not work? I notice there are still variables in the code indicating that 'bucketTree' is the default. For example https://github.com/hyperledger/fabric/blob/2b2ae0cd337fde98fa0cdfaf9b0865dba72dfd49/core/ledger/statemgmt/state/state.go#L37

If we don't have a default, these should probably be cleaned up.

@manish-sethi

ckeyer:
@manish-sethi  this variables is used there https://github.com/hyperledger/fabric/blob/2b2ae0cd337fde98fa0cdfaf9b0865dba72dfd49/core/ledger/statemgmt/state/config.go#L48

manish-sethi:
Default of 'bucketTree' should work. The relevant code is what @ckeyer pointed out in his last comment.
The 'raw' should not be exposed in the core.yaml as it does not implements iterator functions yet. Also, I am not sure how should we compute the state-hash in 'raw' implementation (e.g., return something dummy?)
This is just to benchmark the max throughput that we can get so we can see how other implementations fair.

christo4ferris:
We should document that 'raw' is experimental and incomplete, then... not just omit it. We should be endeavoring to keep docs as consistent with the code as possible.

christo4ferris:
LGTM thanks @ckeyer

srderson:
LGTM, thanks!

PR #1265, opened 2016-04-26
Persistence and more; #1000

corecode:
This addresses #1000 and many other small bug fixes.

Signed-off-by: Simon Schubert sis@zurich.ibm.com

Issue #796, opened 2016-03-05
Deal with a byzantine leader that will censor a specific transaction


kchristidis:
The problem: VP1 submits a transaction to the leader but the leader won't add it to a new block. We need to figure out a way to initiate a view-change. (VP1's vote alone cannot bring about a view-change.)

The solution will involve VP1 broadcasting the transaction to the other VPs, and all the VPs keeping timers. But we need to be careful in how we kick off these timers (per-VP, instead of per-request timers will probably do the job). Had discussed this with @jyellick earlier this week, let's outline the solution during scrum.

Also, @corecode noted that we can't keep all pending requests in memory, cause we can be flooded with requests and run out of memory.

vukolic:
A high level implementation could go like this
1) when a replica becomes aware of a request - for the first time - it broadcasts it to all other replicas (incl. the leader) and triggers a timer (for this particular request)
2) if that timer expires - with out request being committed - replica triggers view change 

Per-request timer could be implemented using a sort of an ordered hashmap ordered by insertion time
1) When a replica receives the request it puts it in the ordered hashmap (ordered by insertion time, where keys are request hashes)
2) When you commit request you remove the request from the hashmap
3) periodically check the first entry (per insertion order) 
   3a) if its insertion time + TIMEOUT < current time trigger view change

PS: this sketch solution assumes that per request timers are the same for all requests which is not necessarily the case for deploy transactions but should/could be for invoke transactions. If different timeouts are needed, the sketch above needs to be tweaked a bit 

PS2: DoS against rate limiting of requests (to address issue broght by @corecode) needs to be done, but perhaps as a separate mechanism/issue

PS3: It might be interesting to incentivize this mechanism (in future)

corecode:
The whole crux is doing this with O(1) memory use.  We did have a mechanism in place previously, but I disabled it in favor of not running out of memory.

corecode:
How about this: every consenter can only challenge the primary with one single request.

However, a byzantine primary then may only process this one challenged request.  This means that we should have another mechanism to change the primary when consensus throughput is slow.

@vukolic could you point us to academic literature that deals with these kinds of fairness?

vukolic:
@corecode " changing the primary when consensus throughput is low" is another issue which is not the same as fairness. Namely having "every consenter challenge the primary with one single request" does not solve the fairness issue. Some pointers for addressing "performance attacks" are at the end of this post.

[...]

PR #1277, opened 2016-04-27
pbft: allow replicas to complain about primary

corecode:
This addresses #796.  Applies on top of #1265.

### DIFF
--- a/consensus/obcpbft/custodian/custodian.go
+++ b/consensus/obcpbft/custodian/custodian.go
@@ -1,0 +1,181 @@
+/*
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+*/
+
+package custodian
+
+import (
+	"sync"
+	"time"
+)
+
+type custody struct {
+	id       string
+	data     interface{}
+	deadline time.Time
+	canceled bool
+}
+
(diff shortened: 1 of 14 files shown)


corecode:

corecode:
this behave test passes locally

jyellick:
@corecode Could you write a brief overview of how the complaints system functions?

My (admittedly rough, and probably incorrect) understanding, is that a peer keeps track of the requests it receives (and then broadcasts to the network), and will broadcast a complaint to the network when it believes the primary is censoring its request.  When a peer receives a complaint, it re-submits the request to the primary, and then if the primary censors the request, sends a view change.  Is this correct / how is it wrong?

So a few immediate questions.  Why do I see complaints at the plugin specific level instead of in a common way for all plugins? I see this being done in Sieve and batch, but not classic, I assume this is intentional, but why?

I've admittedly not been the best about this, but @kchristidis pointed out to me that some of my checkins were missing function documentation and were generally not meeting go lint/vet standards.  Could you make sure they pass for your new files? (We'll need to do a wholesale linting of the other files, but do not want to create merge headaches).  Some documentation might also help make the review easier.

corecode:
Yes, your description is correct.  The reason why it is not in the pbft core is that single requests get forwarded to the primary, and the pbft core only sees a batch and not separate requests.  I did not implement the functionality for classic pbft yet, because batch with size set to 1 will provide the same performance and functionality, so immediate support for complaints did not seem necessary.  More broadly, I think we could remove classic in favor of batch with size=1.

I'll add function documentation in a followup commit.

srderson:
@corecode Could you rebase this PR? There is a merge conflict. Thanks

corecode:
@srderson Rebased, ready to merge.

consensus/obcpbft/custodian/custodian.go
```diff
@@ -0,0 +1,164 @@
...
+)
+
+type custody struct {
+	id       string
+	data     interface{}
+	deadline time.Time
+	canceled bool
+}
+
+type CustodyPair struct {
```
srderson: Could you run golint on this file please. You will see warnings for missing comments on public types and functions. Thanks

srderson:
Thanks. LGTM

corecode:
As indicated in #1445, this PR contains a race condition in one of the unit tests (`TestBatchCustody`).

I believe the cause for the failed test is that as complaints get submitted and trigger a view change, the request order is changed, which leads to the older request being discarded as stale.  However, replica 0 (previous leader) still has a PBFT request filed, but due to the test configuration, cannot communicate this (can't send messages).  Therefore PBFT stays in view change.

Issue #1091, opened 2016-04-13
REST API /chain/blocks/<n> returns empty blocks (other than block 0)


bcbrock:
This issue is observed in conjunction with another error case, see #920. When a test program makes the REST API call /chain/blocks/<n>, it gets a normal HTTP response but the data is NULL, i.e., the block has no transactions. I don't think this makes sense except for the genesis block. This issue is observed in

```
commit 92cc78029d04736928c3c07b7c3336023de2d747
Merge: c4b31f9 96210ef
Author: Sheehan Anderson <sranderson@gmail.com>
Date:   Sun Apr 10 21:17:43 2016 -0400

    Merge pull request #1044 from ghaskins/rebrand
```

and is easy to reproduce if required.

@jyellick offered the following opinion in #920:

  My suspicion is that the peer is currently performing state transfer and so his block height does not    necessarily match the number of blocks he has. This is probably fooling the error checking that exists at that layer which assumes the chain is always unbroken.

  I think we need to think about a policy for more clearly handling requests when a replica knows its state is not complete. Do we make a best effort to return a correct result? Do we reject all queries? This seems like it should be a well studied problem, maybe <at>chetmurthy could weigh in?

bcbrock:
I can confirm that I am seeing this, see #920. What is being returned is null (not sure is this is really the empty string, or something that prints as "null")

```
vagrant@hyperledger-devenv:v0.0.9-0c353b1:/opt/gopath/src/github.ibm.com/bcbrock/busywork/scripts$ curl http://172.17.0.2:5000/chain/blocks/71
null
```

jyellick:
@bcbrock I wouldn't rule out that this is state transfer related, it's possible that in this scenario, state transfer is being invoked so many times, it is producing some gaps that are for whatever reason not being filled in, I'll investigate this possibility. (Though I'll agree, this should reply with an error, rather than an empty block if it is missing)

bcbrock:
@jyellick I was a bit optimistic when I said in #920 that the runs were going to completion. I'm seeing now that as often as not they actually fail in various ways. I assume it is best to wait until this issue (1091) is fixed before posting the other issues, would you agree?

I can also add here that the evidence suggests that if peer 0 has empty blocks in its blockchain, peer 0  also appears not to have executed every transaction.

bcbrock:
[...]

PR #1279, opened 2016-04-27
Consensus threading and queues

jyellick:
This PR is built upon the changeset which @corecode did for persistence.

The first purpose of this changeset is to switch the `pbft-core.go` and each of the respective plugins to operate on a single thread.  This eliminates the need to acquire and drop locks in the code, and eliminates several conditions which existed in the code.

The second purpose is to enable message queueing for consensus so that other message types (in particular state transfer peer messages) are not blocked because the handler is waiting to deliver messages to consensus as this blocking can create deadlocks.  This is accomplished through the standard go fan in pattern, where each handler creates a buffered channel, and these buffered channels fan into a single buffered channel, which is then sent to consensus for processing.  The message fan is currently only employed for consensus requests.  In the future, the same fan code/structure can be utilized for the queueing of queries and potentially peer messages.

With this changeset, (admittedly before rebasing onto Simon's latest), using the classic pbft plugin, my local vagrant environment was able to process 10k requests, being delivered as rapidly as they could be accepted to a single peer, by 20 threads (at 500 requests each).  This processing did not trigger a view change, and all requests were processed successfully.  This is an improvement over last sprint, where deadlocks were often seen beyond about 2k requests.  @bcbrock you might be interested to try your stress tests again against this changset.  I will re-run my local stress test against this PR, assuming it passes the CI tests and will post the results here.

Some code review from @corecode @tuand27613 @kchristidis or others would be appreciated if cycles allow.


bcbrock:
@jyellick - Thanks, I'll try it Thursday afternoon or Friday.

jyellick:
@bcbrock Just retested with the 10k request flood.  The current branch still passes it.

bcbrock:
@jyellick - In this case: 4 peers, PBFT classic, 10K back-to-back TX sent to peer 0 - it seems to work most of the time. At least Issue #1091 seems to be gone. It's hard to characterize/repeat the non-working attempts. Other scenarios definitely fail.

My 2 cents would be to go ahead and merge assuming the reviewers agree that the architecture is good, then we can start to tackle the next set of issues.

srderson:
@jyellick There's a merge conflict on this PR. Assume this is because of the requirements on @corecode's PR which also have conflicts.

jyellick:
Closing, because this was rebased into #1325

PR #1285, opened 2016-04-27
Issue 1278

### LINKED ISSUE #1278: golint error in handler.go
golint: handler.go:84:11: if block ends with a return statement, so drop this else and outdent its block

christo4ferris:
fix lint warning in handler.go

Fixes #1278

handler.go:84:11: if block ends with a return statement, so drop this
else and outdent its block

### DIFF
--- a/consensus/helper/handler.go
+++ b/consensus/helper/handler.go
@@ -81,9 +81,8 @@ func (handler *ConsensusHandler) HandleMessage(msg *pb.Message) error {
 		if err == nil {
 			if tx.Type == pb.Transaction_CHAINCODE_QUERY {
 				return handler.doChainQuery(tx)
-			} else {
-				return handler.doChainTransaction(msg, tx)
 			}
+		  return handler.doChainTransaction(msg, tx)
 		}
 	}
 


srderson:
LGTM. Thanks

msp/cache/second_chance_test.go
```diff
@@ -14,7 +14,7 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-func TestSecondChanceCache(t *testing.T) {
+func TestSecondChanceCacheCase1(t *testing.T) {
```
sykesm: Some questions based on very quick glance at the change:

What do `Case1` and `Case2` mean?
Can we use more descriptive names for the scenarios?
Are the scenarios different enough as to require separate tests?

Looks to me like the new test can simply replace the old one.

msp/cache/second_chance.go
```diff
@@ -82,7 +82,7 @@ func (cache *secondChanceCache) add(key string, value interface{}) {
 	var item cacheItem
 	item.key = key
 	item.value = value
-	atomic.StoreInt32(&item.referenced, 1)
+	atomic.StoreInt32(&item.referenced, 0)
```
sykesm: fwiw, should be no need for the atomic store as the zero value is already 0.

PR #1289, opened 2016-04-28
make sure dbBucketNode is not nil before printing it

danielhtshih:
It is possible to get nil, nil from stateImpl.bucketCache.get(_bucketNode.bucketKey), if it calls fetchBucketNodeFromDB(bucketKey *bucketKey) (_bucketNode, error) and gets nil, nil returned.
It gets crash if we want to print nil string via logger.Debug().


danielhtshih:
Signed-off-by: Daniel HT Shih hotingwow@gmail.com

manish-sethi:
@danielhtshih - Thanks for pointing it out. However, this variable being 'nil' should not cause this behavior. logger gracefully prints nil values (as <nil>).
I looked at the code and came across one issue that could cause this. I have submitted a PR - https://github.com/hyperledger/fabric/pull/1320

srderson:
Closing as this is now resolved fully in PR #1320. @danielhtshih Thanks for catching this!

PR #1297, opened 2016-04-28
fix for 1290

christo4ferris:
partial fix for 1290

another batch will follow in separate PR


christo4ferris:
rebased - note, there are no code changes - just license header replacement to correct an incorrect license header statement. If there's a better way of handling this change, I am all ears

christo4ferris:
restarted build, but there are no code changes, so hard to fathom how the tests are failing other than randomly

christo4ferris:
same two errors - I am baffled unless the tests themselves are broken.

christo4ferris:
closing this and submitting another PR

core/ledger/kvledger/txmgmt/privacyenabledstate/sanpshot_test.go
```diff
@@ -0,0 +1,272 @@
...
+	}
+	samplePublicState := generateSampleData(
+		"",
+		"ns1",
+		"ns2",
+		"ns4",
+	)
+
+	samplePvtStateHashes := generateSampleData(
+		deriveHashedDataNs("", "coll1"),
```
cendhu: Nit. As per chaincode naming [rule](https://github.com/hyperledger/fabric/blob/6764026031f1f39362620473d09dde54a76d0922/core/scc/lscc/lscc.go#L56), we may never have this scenario but it does not hurt if we have it in the test. For pubState, I understand that the empty namespace denotes the ns where we store the configblock. Just worried that it shouldn't give a wrong message when we read the code in the future.

core/ledger/kvledger/txmgmt/privacyenabledstate/sanpshot_test.go
```diff
@@ -0,0 +1,272 @@
...
+		coll := nsColl[1]
+		updateBatch.HashUpdates.PutValHashAndMetadata(ns, coll, []byte(s.Key), s.Value, s.Metadata, s.Version)
+	}
+	for _, s := range pvtState {
+		nsColl := strings.Split(s.Namespace, nsJoiner+pvtDataPrefix)
+		ns := nsColl[0]
+		coll := nsColl[1]
+		updateBatch.PvtUpdates.Put(ns, coll, s.Key, s.Value, s.Version)
+	}
+	err := db.ApplyPrivacyAwareUpdates(updateBatch, version.NewHeight(2, 2))
```
cendhu: For the pvtdata, we can do a small check to see whether the db has the pvtupdates. For public and pvt state hashes, we check this through snapshot files. I am asking this because in the `ExportPubStateAndPvtStateHashes()`, we don't have a `case isPvtDataNs(compositeKey.Namespace): return nil, err` to ensure that the iterator does not return pvtdata. Either we can add that case or have an extra check in the test for additional robustness. If you think this is not mandatory, I am fine with the current test.

core/ledger/kvledger/txmgmt/privacyenabledstate/snapshot.go
```diff
@@ -0,0 +1,184 @@
...
+			dataFile:             dataFile,
+			metadataFile:         metadataFile,
+			kvCountsPerNamespace: map[string]uint64{},
+		},
+		nil
+}
+
+func (w *snapshotWriter) addData(ck *statedb.CompositeKey, dbValue []byte) error {
+	counts := w.kvCountsPerNamespace[ck.Namespace]
+	if counts == 0 {
```
cendhu: Is there any specific reason for using counts?

If we have a chaincode or collection but no data associated with them, the iterator would not return anything related to those namespaces as there would be no entry on the db. For couchdb, the respective namespace's database might be an empty one with indexes or no databases at all. Hence, I assume that you have used `if count == 0` as an alternative to the standard `if !ok`.

common/ledger/snapshot/file.go
```diff
@@ -191,6 +191,9 @@ func (r *FileReader) decodeBytes() ([]byte, error) {
 		return nil, err
 	}
 	size := int(sizeUint)
+	if size == 0 {
+		return []byte{}, nil
```
cendhu, line 195: It can also be `return nil, nil`.

PR #1298, opened 2016-04-28
Updated documentation to reflect newer Chaincode interface with Init method.

stanliberman:
Updated documentation to reflect newer Chaincode interface with Init method.

Signed-off-by: Stanislav Liberman stanislav.liberman@cmegroup.com


muralisrini:
@stanliberman :  

 _`Init` is a constructor function, which will only be invoked by Deploy transaction. The parameters `function` and `args` point to the function implemented by the chaincode that the `Invoke` calls and passes the `args`. Similar behavior for the `Query` function. The `Query` function is not allowed to modify the state of the chaincode; it can only read and calculate the return value as a byte array._

Perhaps seperate the `function` and `args` part from the methods names ? So something like 

The `Init`, `Invoke` and `Query` functions take `function` and `args` as parameters to be used by those methods to support a variety of transactions.  `Init` is a constructor function, which will only be invoked by the Deploy transaction.  The `Query` function is not allowed to modify the state of the chaincode; it can only read and calculate the return value as a byte array.

stanliberman:
@muralisrini Thanks! I've updated the wording per your suggestion.

muralisrini:
@stanliberman : LGTM, thanks!

srderson:
LGTM. Thanks!

Issue #618, opened 2016-02-06
We need to be able to control logging levels in chaincodes


bcbrock:
There is currently no way for the writer of a chaincode to control the amount of logging done by the shim. The shim currently defaults to DEBUG logging, which creates large log files and wastes time. Also, logging in the shim is not set up as "prettily" as in the obc-peer command - there is less information printed and no color coding for example.

The simplest solution for now may simply be to document how to use the `github.com/op/go-logging` package to control what is logged, e.g., by adding the import

```
github.com/op/go-logging
```

and making the first line of the main() routine

```
logger.SetLevel(logger.ERROR, "")
```

In the future I think there is a broader question of how to control logging behavior and myriad other behaviors of chaincodes without have to modify their source code, and without requiring every chaincode to implement an ad-hoc command parser for configuration options passed in by a normal chaincode invocation.

PR #1306, opened 2016-04-29
Added the `shim.SetLoggingLevel` API

### LINKED ISSUE #1303: There are 3 loggers logging as the "chaincode" module

bcbrock:
This is a solution to #618. Go language chaincodes can now call
`shim.SetLoggingLevel` to control logging in the shim. There is no requirement
that the chaincode have included the op/go-logging package, or know anything
about how logging is implemented in the shim. The appropriate documentation
has also been updated.

Since I was mucking around with chaincode logging I went ahead and also fixed #1303 by deleting the redundant logger from the chaincode package.

### DIFF
--- a/core/chaincode/chaincode_support.go
+++ b/core/chaincode/chaincode_support.go
@@ -27,7 +27,6 @@ import (
 	"time"
 
 	"github.com/golang/protobuf/proto"
-	"github.com/op/go-logging"
 	"github.com/spf13/viper"
 	"golang.org/x/net/context"
 
@@ -37,8 +36,6 @@ import (
 	"github.com/hyperledger/fabric/core/ledger"
 	pb "github.com/hyperledger/fabric/protos"
 )
-
-var chaincodeLog = logging.MustGetLogger("chaincode")
 
 // ChainName is the name of the chain to which this chaincode support belongs to.
 type ChainName string
@@ -107,12 +104,12 @@ func NewChaincodeSupport(chainname ChainName, getPeerEndpoint func() (*pb.PeerEn
 
 	peerEndpoint, err := getPeerEndpoint()
 	if err != nil {
-		chaincodeLog.Error(fmt.Sprintf("Error getting PeerEndpoint, using peer.address: %s", err))
+		chaincodeLogger.Error(fmt.Sprintf("Error getting PeerEndpoint, using peer.address: %s", err))
 		s.peerAddress = viper.GetString("peer.address")
 	} else {
 		s.peerAddress = peerEndpoint.Address
 	}
-	chaincodeLog.Info("Chaincode support using peerAddress: %s\n", s.peerAddress)
+	chaincodeLogger.Info("Chaincode support using peerAddress: %s\n", s.peerAddress)
 	//peerAddress = viper.GetString("peer.address")
 	if s.peerAddress == "" {
 		s.peerAddress = peerAddressDefault
@@ -226,7 +223,7 @@ func (chaincodeSupport *ChaincodeSupport) sendInitOrReady(context context.Contex
 	var ok bool
 	if chrte, ok = chaincodeSupport.chaincodeHasBeenLaunched(chaincode); !ok {
 		chaincodeSupport.runningChaincodes.Unlock()
-		chaincodeLog.Debug("handler not found for chaincode %s", chaincode)
+		chaincodeLogger.Debug("handler not found for chaincode %s", chaincode)
 		return fmt.Errorf("handler not found for chaincode %s", chaincode)
 	}
 	chaincodeSupport.runningChaincodes.Unlock()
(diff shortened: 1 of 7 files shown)


bcbrock:
Why does the DCO-bot not recognize my Signed-off-by? No matter how I struggle and strive, the DCO-bot I can't satisfy.

christo4ferris:
weird... let me look at the code

bcbrock:
Thanks, I can't tell that my signature looks any different than yours.

srderson:
You need to put < > around your email address if you're not doing that.

bcbrock:
Why does it work for Chris not to use < > ?

christo4ferris:
I do;-) that explains... I can update regex

muralisrini:
@bcbrock :  a suggestion if it makes sense 
_Go language chaincodes can also control the logging level of the chaincode
 shim interface through the `shim.SetLoggingLevel` API documented
 [here](../API/ChaincodeAPI.md)._

Perhaps modify that to 
Go language chaincodes can also control the logging level of the chaincode
 _shim_ interface through the `shim.SetLoggingLevel` API documented
 [here](../API/ChaincodeAPI.md). Having chaincode log to the shim logger simplifies log collection and would help in trouble shooting and debugging.

bcbrock:
@muralisrini - I'm not sure I understand your suggestion. The API proposed here allows the chaincode to control how much logging the shim does. The shim is set to DEBUG by default, so it creates a lot of noise. 

Are you suggesting that there should be APIs that allow the chaincode to use the same logger object that the shim is using? We could do that as well - but I would still prefer to be able to tell the shim to "shut up" independently of the amount of logging being done by the chaincode.

muralisrini:
@bcbrock : I might have put the extra line "Having chaincode log to the shim logger simplifies log collection and would help in trouble shooting and debugging." in the wrong place... I apologize.

What I meant say was about this section

 _As independently executed programs, user-provided chaincodes can use any
 appropriate technique to create their private logs - from simple print
 statements to annotated and level-controlled logs.  For example, one way to
 create basic levelled logs is to first add..._

If the chaincode developer puts the chaincode to a different log then we would lose interleaving of chaincode and shim logs. Which is OK and good for many puroposes. It might also be good to provide way by which the chaincode can write to the same log as shim (if only for trouble shooting/ debug purposes). It might mean we expose another API on the shim for logging from chaincode ? What do you think ?

bcbrock:
@muralisrini - I think your idea to have the shim provide logging services that are consistent with and interleaved with the system logging is a great idea.

One way we could do that would be to have the shim export shim.Debugf(), shim.Infof(), etc. functions for use by the chaincode, plus an interface to set the logging level for the chaincode independent of the shim logging level. [Note, we really should be naming using the 'f' forms of logging functions for consistency with Golang base package conventions.] This seems a bit odd though because "shim _isa_ logger" is not really true, and would probably limit a chaincode to only 1 logger.

So I think it might be nicer to have the shim provide a shim.NewLogger() interface, and then define Debugf() etc. on the new object returned by the call. In a way this would be redundant because we would just be passing through the APIs to op/go-logging objects. However I think there are practical advantages and abstraction advantages to not requiring the chaincodes to import op/go-logging,  or really have to know anything about how logging is implemented.

Thoughts?

srderson:
@bcbrock There's a merge conflict with this PR and it needs to be rebased. Looks like it's still under discussion?

bcbrock:
@srderson Yes I'm waiting for @muralisrini to weigh in again. I'll let you know when we make a decision. Thanks for the heads up.

muralisrini:
I like just exposing interfaces to use the one defined logger in the shim. Simple and implies that shim's logger will be the one used. If the chaincode developer wants other logging mechanism, they could implement it themselves.  This should be good for most purposes. What do you think ?

bcbrock:
@muralisrini I don't like the idea of using the same logging object used by the shim, because the chaincode logs will not be unique - they will be marked as coming from the shim - and we should not have to require the chaincode writers to put "coming from chaincode" into every log message. Both of the two methods I suggested would correctly interleave the chanicode's logs with the shims logs and use the same formatting. The chaincode logs would be identified with a user supplied string (supplied once) and the short form of the UUID to make them unique.

bcbrock:
Also, what started this is the need to control shim logging level separately from chaincode logging level. So they can't be the same logging object unless we want to wrap it in a front-end that provides this functionality by making the single logging object  _look like_ a different object when called from the chaincode.

muralisrini:
I see. The first of your suggestions creates a new logger for the chaincode - separate from the shim's - in the shim layer while the second allows many loggers to be created by the shim (with NewLogger()) ? Did I understand that correctly ?

I can think of good reasons for both... first one is simple and good enough for simple cases. The second might be useful for large chaincodes composed of many packages. Perhaps we should err on the side of caution and go with the second.

bcbrock:
Closing for now. Will reopen once I have implemented what Murali and I discussed privately.

PR #1197, opened 2016-04-21
Confidentiality 2.0

adecaro:
This PR implements confidentiality as described in Section 4.3.2.1 (Confidentiality against users) of the protocol spec document. The previous versions are still available. The default version of confidentiality will remain 1.2. The new version can be activated by setting the field "ConfidentialityProtocolVersion" in "ChaincodeSpec" to "2.0".

Signed-off-by: Angelo De Caro adc@zurich.ibm.com

PR #1307, opened 2016-04-29
[Confidentiality 2.0] Introducing the crypto/primitives package

adecaro:
This PR introduces the crypto/primitives package that collects all the basic crypto algorithms used by the crypto layer. 

The PR is the first step toward deploying the new version of confidentiality as described in Section 4.3.2.1 (Confidentiality against users) of the protocol spec document.

Notes: golint, go vet and gofmt applied.

@BiscuitBomber, please, may you give a look at this PR? Does it go in the right direction? 
The spaces and blank lines you were mentioning in the other PR are actually introduced by gofmt  (https://golang.org/cmd/gofmt/) that I use, in addition to go vet and golint, to formats Go files.


srderson:
@adecaro thanks for your patience. LGTM and I can confirm that all unit and behave tests are passing. This PR has been discussed in PR #1197.

As this project is still new, we're all learning as we go, so obviously there will be a few bumps as we strive to be as transparent as possible. I think the key takeaways from and for the community are to create issues that clearly and fully describe plans prior to implementation, and make every effort to keep PRs small so they're easily reviewable by everyone.

PR #1314, opened 2016-05-01
Changed Broadcast from sequential to parallel

yacovm:
Each message is now sent on its own goroutine, and the method aggregates the results
of the message sending and returns them to the caller

This improves the throughput of the system because it shortens the invocation time of Broadcast, since sending to different peers is done in parallel.

I tested it by simulating network latencies with the tc [1] command on 4 peers running in docker containers with pbft and it counted the transactions per second, and it increased throughput:

Delay | Speedup:
1ms | 80%
2ms | 85%
5ms | 130%
10ms | 140%
15ms | 150%
20ms | 160%

[1]
http://linux.die.net/man/8/tc

### DIFF
--- a/core/peer/peer.go
+++ b/core/peer/peer.go
@@ -411,15 +411,38 @@ func (p *PeerImpl) cloneHandlerMap(typ pb.PeerEndpoint_Type) map[pb.PeerID]Messa
 // Broadcast will broadcast to all registered PeerEndpoints if the type is PeerEndpoint_UNDEFINED
 func (p *PeerImpl) Broadcast(msg *pb.Message, typ pb.PeerEndpoint_Type) []error {
 	cloneMap := p.cloneHandlerMap(typ)
-	var errorsFromHandlers []error
+	errorsFromHandlers := make(chan error,len(cloneMap))
+	var bcWG sync.WaitGroup
+
+	start := time.Now()
+
 	for _, msgHandler := range cloneMap {
-		err := msgHandler.SendMessage(msg)
-		if err != nil {
-			toPeerEndpoint, _ := msgHandler.To()
-			errorsFromHandlers = append(errorsFromHandlers, fmt.Errorf("Error broadcasting msg (%s) to PeerEndpoint (%s): %s", msg.Type, toPeerEndpoint, err))
-		}
+		bcWG.Add(1)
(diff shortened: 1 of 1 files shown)


yacovm:

muralisrini:
@yacovm : thanks! This is a good change IMO... 
- there is no inherent dependency on serialization or ordering in message reception, so parallel is logically equivalent to current serial. Plus, in the end, network determines order even in serial anyway
- it is a local change and does not bleed over outside the Broadcast method
- "Broadcast" suggests parallel sends
- help performance

(Not suggesting we do this in this PR... but wonder if it also makes sense to randomize the sends ?)

Given the nature of this change will be good to get more eyes, especially @jeffgarratt @binhn

yacovm:
Do you mean randomizing the order of the sends? Why?

muralisrini:
now that we are not serializing we have the opportunity to add bit more entropy ... more unpredictability than we'd have with just TCP/IP. Again, just a thought for brainstorming. Not suggesting we do this for this PR.

yacovm:
hmm, but isn't the time it takes for all goroutines to reach the system call layer (a few hundred micro seconds) negligible in comparison to the time it takes for the network to transmit them (milliseconds)?

Plus it's done in Parallel... I'm asking this because if we gain something from it I can do another PR.

muralisrini:
So basically randomization will get lost in the much larger network latency ? That's probably true.

Let's ignore the randomization for now ? It is here if we want to revisit.

yacovm:
OK, So we're waiting for @jeffgarratt  || @binhn  to approve the PR I guess?

muralisrini:
yes.

jeffgarratt:
@muralisrini @yacovm Looks good to me!! NOTE: Keep look out for future ramifications to consensus as this may alter sequencing wrt to consensus interactions and expose pre-existing but unidentified issues.

FYI @jyellick @corecode

jeffgarratt:
@muralisrini @yacovm  Looks good to me!!  NOTE:  Keep look out for future ramifications to consensus as this may alter sequencing wrt to consensus interactions and expose pre-existing but unidentified issues.

srderson:
This is a nice contribution! Thanks @yacovm! LGTM and I confirmed all unit and behave tests are passing, including consensus tests.

Issue #875, opened 2016-03-24
we need the shim to provide the transaction timestamp to the contract


kletkeman:
In our domain, timestamps at the nano level are critical to sorting out the history of events relating to an asset. This in fact is equally important in any domain, even finance. You must know the time at which something was signed.

Currently, our contract generates timestamps when the event does not carry one, but on today's interlock call with @mastersingh24, we confirmed that we will _never_ achieve consensus in a network with actual consensus running because the state can not converge on one timestamp unless it comes in in the transaction. We absolutely cannot assume that we can force every application and device in the entire world to send timestamps, so we are suddenly very stuck.

After some discussion of messy front ends to inject timestamps, Gari suggested that the transaction already has a timestamp that is generated at the moment of initial creation and then forever propagated, and that contracts should be asking the shim for that timestamp in order to have something to define the time accurately. Since the transaction is propagated, every peer will get the same txn time if this is implemented.

This issue is as high a priority as we can imagine. We cannot run on pbft with timestamps generated inside the contract, but we must have timestamps. 

Since the transaction UUID is already available in the stub:

// ChaincodeStub for shim side handling.
type ChaincodeStub struct {
    UUID string
}

May we please get the transaction timestamp in there as well? 

Edit: Note that the timestamp we need is defined in [3.1.2.1 Transaction Data Structure](https://github.com/hyperledger/fabric/blob/master/docs/protocol-spec.md#3121-transaction-data-structure).

corecode:
Timestamp service is important, and we're currently working on a principled design.

Using a timestamp provided by the transaction is possible, but do keep in mind that the timestamp supplied is as unreliably and untrustworthy as the party submitting the transaction is.

As a quick workaround, you could add a timestamp parameter to your `invoke` method.  Whoever submits this transaction may cheat, however, and provide an incorrect time.

kletkeman:
We already have a timestamp parameter on our invokes. But we cannot force people to use it properly, which mitigates the value almost to the point of being useless. 

I'm not sure what you mean when you say "timestamp service" and how that would be different from the invoke's understanding of its invoke time, which must exist somewhere in the peer and be both propagated to the fabric and accessible to the shim (perhaps with a bit of work.) 

Could you comment on what I am missing with my presumptions? Thx.

corecode:
[...]

PR #1317, opened 2016-05-01
Implement issue #875: Passing transaction timestamp to the chaincode

binhn:
Adding shim stub GetTxTimestamp(). See chaincode_example01 for unit test case


srderson:
@binhn Do we need test cases or documentation associated with this change?

Could you provide more detail in the comment in chaincode.go?

```
// GetTxTimestamp returns transaction timestamp
 +func (stub *ChaincodeStub) GetTxTimestamp() (*gp.Timestamp, error) {
```

Based on that comment it's not clear to me who generates the timestamp or when it is generated.

binhn:
@srderson ok, will add more comment and doc -- the timestamp is currently the transaction created timestamp

binhn:
@srderson I am working on resolving the conflicts

binhn:
@srderson I push a new PR to resolve the conflicts https://github.com/hyperledger/fabric/pull/1433
We should close this one.

PR #1315, opened 2016-05-01
Issue892

andresgaragiola:
Fix of issue892.

PR #1310, opened 2016-04-30
Implemented Attribute Certificate Authority.

andresgaragiola:
This pull request include the ACA (Attribute Certificate Authority) implementation.

PR #1319, opened 2016-05-02
Added support to create deploy, invoke and query transactions using user specified attributes

diegomasini:
Modified CLI and rest API to allow a user to deploy, query or invoke specifying attributes for ABAC. The given attributes are validated against an Attribute Authority. A local implementation of an Attribute Authority is used for testing purposes.


srderson:
@diegomasini I haven't reviewed this change yet, but there are merge conflicts so you will need to rebase on master when you have a chance. Thanks.

srderson:
@diegomasini This change is too big to be easily reviewed. Can you break it down into smaller PRs? I also see no documentation.

Can you suggest someone you can review this code? Maybe @smithbk?

diegomasini:
Hello @srderson, sure, I think the problem was that the PR was processed before another PRs from @andresgaragiola that contains part of this changes, that's why this is to big. I merged with it locally because I needed part of the functionality there. I will closed this PR and submit it again after PR #1315 and #1310 are accepted. Thanks and sorry for the inconvenience.

PR #1320, opened 2016-05-02
Fixed a bug in bucket-tree caching

manish-sethi:
Making a copy of bytes received from rocks db before freeing the memory.
To me it appears that this is what causes the behavior described in the following PR.
https://github.com/hyperledger/fabric/pull/1289


manish-sethi:
@danielhtshih Can you please take the changes from this PR and confirm that it solves the issue in your test scenario?

srderson:
Confirmed tests are passing. LGTM. Thanks!

Issue #1326, opened 2016-05-02
Travis failing with unit-test platforms/car


ghaskins:
A discrepancy between travis/vagrant provisioners coupled with a bug providing a false positive against unit-test failures allowed PR #1250 to be merged without noticing that the unit-tests were failing under the Travis environment.  The unit-test masking bug was later fixed, which caused any subsequent runs to be (correctly) reported as failing.  To fix for real, we will need to align the travis provisioner more closely with vagrant to reduce the likelihood of reoccurance

ghaskins:
@christo4ferris @rameshthoomu

bcbrock:
@ghaskins - This platforms/car test is also failing in my non-vagrant environment. Maybe #1329 should be generalized to never run the test unless specifically requested, or this directory should be removed altogether from the Go tests.

ghaskins:
@bcbrock I am open to that, but I suspect we'd be papering over a broader issue: environmental prerequisites not being met within your chosen environment.  Without seeing your error logs, Im guessing you are either seeing it complaining about protoc-gen-go or chaintool binaries not being found? We probably need to find a more general solution or I fear this will keep happening over and over as more chaincode environments and other dependency driving forces are added in the future.  For now, perhaps the best solution is to invert the logic:  rather than exclude travis, I will only run the test under vagrant.

bcbrock:
Yes, first I has to install protoc-gen-go, then figure out what "chaintool" was and get it on the path, then install java. Now it looks like you are building a Docker container and assuming that the system java executable exists in /bin?  (Mine doesn't) I would agree with your "papering over" comment if chaintool was accepted as a permanent part of the hyperledger/fabric, but it seems to be an experiment - e.g., you checked in the binary (Java JAR) instead of the source code plus build scripts. Is java only required inside the chaincode container or in general for this to work?

I agree we need to have some guidelines on how many external dependencies we want to have in the core verification suite. Java seems reasonable to me if it can be brought in in a portable way.

The current fail:

```
2016/05/03 14:36:55 Failed Chaincode docker build:
Step 1 : FROM openblockchain/baseimage
 ---> 13359bc4b3b5
Step 2 : COPY bin/* /usr/local/bin/
 ---> 002e2d97f48d
Removing intermediate container d62fa566ac39
Step 3 : COPY package.car /tmp/package.car
 ---> bbad0879f0ad
Removing intermediate container 818d7ecf8df9
[...]

PR #1327, opened 2016-05-02
Add a common "host.sh" for sharing provisioning logic

ghaskins:
Travis is missing the protoc compiler, so this starts to incorporate a shared facility to keep travis
updated along with vagrant.  It can be exploited by other hosting environments too.

(possible) Fix for #1326.  We won't know until travis run completes.  Still running local test/behave in parallel and will update with final result. 

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,12 +31,11 @@ install:
 
 before_script:
 
- - echo " CREATING BASE IMAGE "
- - echo " Calling docker.sh to build Openblockchain/baseimage "
- - cd $HOME/gopath/src/github.com/$USER_NAME/$REPO_NAME/scripts/provision/ && chmod +x docker.sh && ./docker.sh 0.0.9
- - cd $HOME/gopath/src/github.com/$USER_NAME/$REPO_NAME/scripts/provision/ && chmod +x host.sh && sudo ./host.sh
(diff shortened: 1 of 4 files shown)


christo4ferris:
@ghaskins thanks for catching this... looks like there's still an issue.

ghaskins:
@christo4ferris yeah, still trying to figure it out.  Worst case, I will add a patch that neutralizes the failing unit-test under Travis, such as 

```
diff --git a/core/chaincode/platforms/car/test/car_test.go b/core/chaincode/platforms/car/test/car_test.go
index 371d0d3..f150bab 100644
--- a/core/chaincode/platforms/car/test/car_test.go
+++ b/core/chaincode/platforms/car/test/car_test.go
@@ -15,6 +15,10 @@ func TestMain(m *testing.M) {
 }

 func TestCar_BuildImage(t *testing.T) {
+       if os.Getenv("TRAVIS") != "" {
+               t.Skip("skipping test; TravisCI not supported")
+       }
+
        vm, err := container.NewVM()
        if err != nil {
                t.Fail()
```

But for now, we are still trying to figure out why the Travis environment doesnt seem to see the protoc compiler

ghaskins:
@christo4ferris we ended up merging #1329 for now so at least people can make forward progress.  This issue with the travis configuration is still under investigation.

ghaskins:
Deferring this work for now

PR #1329, opened 2016-05-02
Disable the platforms/car unit test when run under vagrant

ghaskins:
This is an experiment to see if I can work-around some travis wonkiness related to Issue #1326 

Preference is to fix the travis environment.  However, if this works we can use it as a stop-gap until the travis issue is resolved and then revert this one.


ghaskins:
Pardon the typo:  this should read "when run under Travis"

ghaskins:
I have confirmed that the test still runs when not under Travis, and successfully bypasses the test under Travis

srderson:
LGTM. Thanks for the quick fix!

PR #1330, opened 2016-05-03
Increase timeouts on some behave tests

srderson:
Simply increases the timeouts on some behave tests that intermittently fail because the timeout is too low.

### DIFF
--- a/bddtests/peer_basic.feature
+++ b/bddtests/peer_basic.feature
@@ -465,7 +465,7 @@ Feature: lanching 3 peers
                     |arg1|arg2|arg3|
                     | b  | a  | 1  |
 	        Then I should have received a transactionID
-	        Then I wait up to "30" seconds for transaction to be committed to peers:
+	        Then I wait up to "60" seconds for transaction to be committed to peers:
                     | vp0  | vp1 | vp2 | vp3 |
 
             When I query chaincode "example2" function name "query" with value "a" on peers:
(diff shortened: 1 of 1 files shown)


srderson:

binhn:
@srderson I don't think this is the ultimate fix. It seems to me if their server is busy, the timing becomes unpredictable. We should come up with better strategy or just ignore these things until we switch over to jenkins; otherwise the behave keeps taking longer and longer to complete.

srderson:
@binhn These are timeouts needed to make behave tests pass on my local system, a mid 2014 Macbook Pro. I figure if I see these timeouts, other users may encounter them when running behave tests locally. Feel free to close and we can explore a better solution.

binhn:
@srderson since these timeouts are "waiting up to", so I will go ahead and merge. Thanks

orderer/common/channelparticipation/restapi.go
```diff
@@ -121,8 +125,8 @@ func (h *HTTPHandler) serveListOne(resp http.ResponseWriter, req *http.Request)
 		return
 	}
 
-	channelID, ok := h.extractChannelID(req, resp)
-	if !ok {
+	channelID, err := h.extractChannelID(req, resp)
+	if err != nil {
```
C0rWin, line 129: I seem missed it previous review, but why did not you return an error here?

orderer/common/channelparticipation/restapi.go
```diff
@@ -143,8 +147,8 @@ func (h *HTTPHandler) serveJoin(resp http.ResponseWriter, req *http.Request) {
...
```
C0rWin, line 151: Same as above?

PR #1333, opened 2016-05-03
[ci skip] update glossary

frankyclu:
add new key terms and definitions to the glossary list

DCO1.1 Signed-off-by Frank Lu fylu@us.ibm.com

### DIFF
--- a/docs/glossary.md
+++ b/docs/glossary.md
@@ -7,6 +7,12 @@
 <col>
 <col>
 <tr>
+<td width="20%"><b>Chain Member</b></td>
+<td>
+Entities that do not participate in the validation process of a blockchain network, but help to maintain the integrity of a network. Unlike Chain transactors, chain members maintain a local copy of the ledger.
+</td>
+</tr>
+<tr>
 <td width="20%"><b>Chain Transactor</b></td>
 <td>
 Entities that have permission to create transactions and query network data.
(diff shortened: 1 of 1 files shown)


srderson:
@frankyclu Please add [ci skip] to your comment in future doc only changes. This will prevent the travis build from running. Thanks

joshhus:
A few quick examples would be helpful adds to the definitions. e.g. Chain Member examples: NVN? etc. (... What are some example "entities" that fit the description.)

joshhus:
For Network owners, maybe avoid "full member" ... confusing compared to just a "member"? ... perhaps define a Network owner simply as a VN owner.

joshhus:
Please tweak: "Approval from network owner is required when adding or deleting a participant with ownership right" to: "Approval from a network owner is required when adding or deleting a participant with ownership rights." (I guess a participant with ownership rights = an owner.) ... Note this is kind of interesting; could a nefarious owner, on his own, keep adding nefarious owners to the point of influencing consensus?

joshhus:
So Owner and Member registration also involves "inviting" whereas User reg. does not involve "inviting." ... Sounds reasonable. Does the inviting come after the registration, as presented?

joshhus:
Finally, for the transaction defs. - so a "Public Transaction" is not a chaincode transaction? What kind of code is it? .... (as compared to a Confidential Chaincode Transaction). Thanks!

christo4ferris:
Network Owner seems an odd term, since there can and likely will be multiples. I would think Stakeholder would be more suitable term. At the end of the day, they are Validators. Also, should there be a heirarchy? e.g. a Network Stakeholder is also a Network Member, etc.

srderson:
@frankyclu You also need to sign off on this PR. Thanks.

christo4ferris:
@frankyclu there are also a few vestigial references to "OBC" in the glossary that should be changed to "Hyperledger Fabric". Also, there is a parenthetical reference to "the Dev team" which should be removed.

frankyclu:
@joshhus 
1) updated the sentence, I modified a few words to reflect the fact that "for a new owner to be added to co-own a network, existing members must agree"
2) yes, inviting come after the registration (someone has to setup the network first before new members can be added)
3) I added a new paragraph "confidential transaction" to differentiate the difference between that and "confidential chaincode transaction"

@christo4ferris 
1) I updated a few words and hopefully it will make it more clear (to point out there could be multiple "owners". 

I think the word "owner"  is better because for some would all members that transact on a network are technically "stakeholders" of that network (e.g. bitcoin owners have a stake in the bitcoin network although they don't run full nodes

2) done, replaced OBC with hyperledger fabric

joshhus:
Do you folks think we should move the "terminology" section of prot-spec to the "glossary". I see prot-spec being split up eventually also anyway: e.g. into a separate security doc. We could open Issues for these for the community. 

I agree that Network Owner is sort of, well, is it a necessary concept; can we do better with another term. e.g. what do we really mean; isn't it just an owner of a VN/VP.

joshhus:
If we keep Network Owner, let's add a clear definition of that to glossary. ty.

PR #1335, opened 2016-05-03
Invert our platforms/car unit-test exclusion logic

ghaskins:
We previously made the test exclude TravisCI.  However, there are
other environments besides travis that do not have the prerequisites
installed and thus fail this test as well.  Therefore, we invert the
logic so that the test _only_ runs under vagrant and is disabled
everywhere else.  When the dependency issue is resolved, we can
remove this exclusion.

### DIFF
--- a/core/chaincode/platforms/car/test/car_test.go
+++ b/core/chaincode/platforms/car/test/car_test.go
@@ -15,8 +15,8 @@ func TestMain(m *testing.M) {
 }
 
 func TestCar_BuildImage(t *testing.T) {
-	if os.Getenv("TRAVIS") != "" {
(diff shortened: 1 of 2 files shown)


ghaskins:
@bcbrock #1326

srderson:
@ghaskins saw you and @bcbrock chatting on the #fabric-dev-env channel, but didn't follow the entire conversation. Is this ready to merge or are you planning additional changes?

ghaskins:
@srderson This can go in, I believe it should be an effective stop-gap for @bcbrock

srderson:
LGTM

PR #1336, opened 2016-05-03
Add behave test for crash fault tolerance

jyellick:
Added a behave test to explicitly test crash fault tolerance.

In this behave test the following occurs:

```
vp{0,1,2,3} up
execute some transactions
vp{1,2,3} down
vp{1,2} up
execute some transactions
verify result
```

If vp1 and vp2 do not recover their state, there are not 2f+1 functioning nodes who can come to a consensus of the next sequence number, and the final execute will not occur.

Discovered a bug had been introduced in the persistence, likely during one of its many rebases, where the checkpointing code had bifurcated, and one method became dead code.  This dead code contained the call to persist the checkpoint, while the active code did not.  This changeset also fixes this problem

This is likely of particular interest to @corecode but it's correctness should be easily verifiable by anyone familiar with the consensus code (looking at @tuand27613 and or @kchristidis)

### DIFF
--- a/bddtests/peer_basic.feature
+++ b/bddtests/peer_basic.feature
@@ -777,3 +777,70 @@ Feature: lanching 3 peers
     Examples: Consensus Options
         |          ComposeFile                                                       |   WaitTime   |
         |   docker-compose-4-consensus-batch.yml docker-compose-4-consensus-nvp0.yml |      60      |
+
+  #@doNotDecompose
+  @issue_1000
+	Scenario Outline: chaincode example02 with 4 peers and 1 membersrvc, test crash fault
+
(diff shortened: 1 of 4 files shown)


jyellick:
Preliminary review is fine, but hold on integration, see a problem.

jyellick:
The above commit fixes a couple of the things revealed by the failing Sieve test, but the test still fails.

This is not a regression from this PR, but existed prior, so my vote would be for merging this now and uncommenting the Sieve test case, once the corresponding bug in the Sieve crash fault tolerance is fixed .

corecode:
nice, looking good.

srderson:
@jyellick sounds like a good approach, LGTM

PR #1345, opened 2016-05-04
[ci skip] Adding the Hyperledger Fabric

JonathanLevi:
Adding an H1 level project title
[ci skip]

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
 [![Build Status](https://travis-ci.org/hyperledger/fabric.svg?branch=master)](https://travis-ci.org/hyperledger/fabric)
+# Hyperledger Fabric
 
 ## Incubation Notice
 This is a Hyperledger project in _Incubation_. It was proposed to the community and documented [here](https://goo.gl/RYQZ5N). Information on what _Incubation_ entails can be found in the [Hyperledger Project Lifecycle document](https://goo.gl/4edNRc).


common/ledger/blkstorage/blockindex.go
```diff
@@ -288,19 +283,30 @@ func (index *blockIndex) exportUniqueTxIDs(dir string, newHashFunc snapshot.NewH
 			continue
 		}
 		previousTxID = txID
+		if numTxIDs == 0 { // first iteration, create the data file
+			dataFile, err = snapshot.CreateFile(filepath.Join(dir, snapshotDataFileName), snapshotFileFormat, newHashFunc)
```
cendhu, line 287: Is it possible to have no TxId in the store? Even the genesis block has a txID right?  https://github.com/hyperledger/fabric/commit/4eff62460e450c8ab349437ad2070a4b1e6d939b

core/ledger/confighistory/mgr.go
```diff
@@ -181,23 +181,27 @@ func (r *Retriever) CollectionConfigAt(blockNum uint64, chaincodeName string) (*
...
+
 	var numCollectionConfigs uint64 = 0
+	var dataFileWriter *snapshot.FileWriter
+	var err error
 	for nsItr.Next() {
 		if err := nsItr.Error(); err != nil {
 			return nil, errors.Wrap(err, "internal leveldb error while iterating for collection config history")
 		}
+		if numCollectionConfigs == 0 { // first iteration, create the data file
+			dataFileWriter, err = snapshot.CreateFile(filepath.Join(dir, snapshotDataFileName), snapshotFileFormat, newHashFunc)
```
cendhu, line 199: **Clarification Question**

- When the snapshot files are processed by the new peer, how would the new peer know all required files are present in a given directory? In other words, an admin can forget to copy a required file. I assume that in the snapshot.metadata file, there would be a mapping between file names (all non-empty) and its hash along with many other data such as height, hash. Then, the kvledger at the new peer would ensure all files are present as per the snapshot.metadata file before processing the snapshots. Am I correct?

**Comment**

1. I agree that the store would be empty when there is no explicit collection config. I am worried that admin might say that a file is missing when it is missing intentionally. I assume snapshot.metadata is not in human readable format to find the list of files (a counter argument would be that admin can still look at each file and complain about empty content). A consistent set of files for all scenario looks natural to me and avoids confusion.
2. If we allow empty files, the code path for both empty and non-empty files would be the same (does not look like a hack to me when we allow empty files for code simplicity). Further, the kvledger needs not to have logic like which files do not exist and which components should not be called, etc... Even if we pass this logic to individual components, it adds unnecessary complexity. Earlier code looked neat to me.

common/ledger/blkstorage/blockindex.go
manish-sethi: As far as possible, it's not good to drag the assumptions in the lower level code that how the consumer uses it. At this level of the code, you can open a store and invoke the `Export` function.

PR #1346, opened 2016-05-04
[ci skip] Updating project title

JonathanLevi:
README >> "Hyperledger Fabric"


JonathanLevi:
Can't tell how come changing the title breaks the build... to save time, will re-submit.

srderson:
The behave tests do not always pass currently. Seems that some of the consensus tests have timing issues.

JonathanLevi:
Yeah, noted the various threads regarding the tests' behavior! [thank you @srderson]

We can just leave/ignore this proposed top level title change though (we now have the incubation notice and the project name instead of 'README'), in the restored README.md, so there's no need for this one any more.

Issue #1347, opened 2016-05-04
Broken links in README.md


lehors:
The landing of PR #1341 moving install content to new file broke several links in the top README.md file.

joshhus:
We updated the README.md also. Will investigate, thanks.

bmos299:
I noticed in the README.md file that both the Installation: and Protocol Specification: link are broken (i.e. result in 404 Response codes).

lehors:
@bmos299 You're right I added that to my PR.
Thanks.

joshhus:
Thanks for looking at this, folks! (Still new to the git CLI / cloning / syncing process.)

PR #1348, opened 2016-05-04
Fix broken links in README

lehors:
Addresses Issue #1347

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,4 +24,4 @@ If you are new to the project, you can begin by reviewing the following document
 
   - [Whitepaper WG](https://github.com/hyperledger/hyperledger/wiki/Whitepaper-WG)
   - [Requirements WG](https://github.com/hyperledger/hyperledger/wiki/Requirements-WG)
-  - [Protocol Specification](https://github.com/joshhus/fabric/blob/TOCReal/docs/protocol-spec.md)
+  - [Protocol Specification](docs/protocol-spec.md)
--- a/README.md
+++ b/README.md
@@ -7,14 +7,14 @@ Use the following topics to navigate through the project directories and documen
 [Apache License Version 2.0](LICENSE): The Hyperledger Project software license.
 
 ## Installing the fabric
-[Installation](INSTALL.md): Describes how to install the blockchain fabric and use project tools.
+[Installation](docs/dev-setup/install.md): Describes how to install the blockchain fabric and use project tools.
 
 ## Contributing to the project
 [Contributing](CONTRIBUTING.md): Describes the rules and procedures for contributing to the Hyperledger Project.<br>
 [Issues](https://github.com/hyperledger/fabric/issues): Lists the current work items, which all community members are eligible to work on.
 
 ## Maintaining the project
-[Maintaining](MAINTAINING.md): Lists the project maintainers and contact information.
+[Maintainers](MAINTAINERS.txt): Lists the project maintainers and contact information.
 
 ## Documentation 
 [README](docs/README.md): Project documentation table of contents (linked).


joshhus:
Yes, LGTM. Thanks.

srderson:
LGTM

core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb.go
```diff
@@ -833,8 +848,41 @@ func (vdb *VersionedDB) readChannelMetadata() (*channelMetadata, error) {
...
+	// if namespacesToScan is empty, we can return early with a nil FullScanIterator. However,
+	// the implementation of this method needs be consistent with the same method implemented in
+	// the stateleveldb pkg. Hence, we don't return a nil FullScanIterator by checking the length
+	// of the namespacesToScan.
+
+	dbsToScan := []*namespaceDB{}
+	for _, ns := range namespacesToScan {
+		db, err := vdb.getNamespaceDBHandle(ns)
+		if err != nil {
+			return nil, byte(0), errors.Wrapf(err, "failed to get database handle for the namespace %s", ns)
```
manish-sethi: Here, and at other other places, on the usage of `errors` package, I have been following a pattern - https://github.com/hyperledger/fabric/pull/1363#discussion_r434172541.

core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb.go
```diff
@@ -833,8 +848,41 @@ func (vdb *VersionedDB) readChannelMetadata() (*channelMetadata, error) {
...
+		dbsToScan = append(dbsToScan, &namespaceDB{ns, db})
+	}
+
+	// some of the databases might contains internal keys.
+	// The scanner must skip these keys.
+	toSkipKeys := map[string]bool{
+		savepointDocID:       true,
+		channelMetadataDocID: true,
+	}
+	return newDBsScanner(dbsToScan, vdb.couchInstance.internalQueryLimit(), toSkipKeys)
```
manish-sethi: We cannot use `toSkipKeys` blindly for each db. This is to be filtered only from the db that is used for the `empty` namespace. Because, that db could have a name clash with the metadata db under the current naming scheme. The DBs for user namespaces can have these keys legitimately.

core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb.go
```diff
@@ -1000,3 +1044,115 @@ func constructVersionedValue(cv *CacheValue) (*statedb.VersionedValue, error) {
...
+			continue
+		}
+		if s.toSkipKeys[couchDoc.id] {
+			continue
+		}
+		// TODO: for config block, either we need to perform a deserialization and then
+		// a deterministic serialization either here or we need to always serialize the config
+		// block deterministically after the validateAndPrepareBatch in the txmgr before
+		// committing the config block. As we anyway need to rebuild DBs to add channelMetadata,
+		// we can use the opportunity to employ the later approach which is more cleaner.
```
manish-sethi: As discussed offline, this is already fixed in 2.0. So, remove this comment

core/ledger/kvledger/txmgmt/statedb/statecouchdb/value.proto
```diff
@@ -10,6 +10,11 @@ option go_package = "github.com/hyperledger/fabric/core/ledger/kvledger/txmgmt/s
```
manish-sethi: Can you change this file name for this to be more specific, perhaps `dbvalue.proto`

core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb.go
```diff
@@ -1000,3 +1044,115 @@ func constructVersionedValue(cv *CacheValue) (*statedb.VersionedValue, error) {
...
+		fields, err := validateAndRetrieveFields(couchDoc)
+		if err != nil {
+			return nil, nil, errors.Wrapf(
+				err,
+				"failed to validate and retrieve fields from couch doc with id %s",
+				couchDoc.id,
+			)
+		}
+		val, err := encodeValueVersionMetadata(fields.value, fields.versionAndMetadata)
```
manish-sethi: want to call this as `dbVal` to distinguish?

core/ledger/kvledger/txmgmt/statedb/statecouchdb/couchdb.go
```diff
@@ -152,6 +152,7 @@ type fileDetails struct {
 
 //couchDoc defines the structure for a JSON document value
 type couchDoc struct {
+	id          string
```
manish-sethi: Maintaining same info in the two different manners causes confusion. Especially, when one would have to remember that the the actual data that affects the interaction with couchdb is `_id` entry in the `jsonValue` field below and this is just a proxy of that. Can you instead define a function on this type that returns you the `_id` field from the `jsonValue` field below.

core/ledger/kvledger/txmgmt/statedb/statecouchdb/couchdoc_conv.go
```diff
@@ -66,58 +67,66 @@ func (v jsonValue) toBytes() ([]byte, error) {
...
-		if returnValue, err = json.Marshal(jsonResult); err != nil {
-			return nil, err
+	delete(jsonDoc, idField)
+	delete(jsonDoc, revField)
+	delete(jsonDoc, versionField)
+
+	if jsonDoc[deletedField] != nil {
+		docFields.value = nil
+		return docFields, nil
+	}
```
manish-sethi: Is this ever possible for a retrieved couchdoc to have this field?

core/ledger/kvledger/txmgmt/statedb/statecouchdb/couchdoc_conv.go
```diff
@@ -66,58 +67,66 @@ func (v jsonValue) toBytes() ([]byte, error) {
...
-		revision = jsonResult[revField].(string)
+
+	docFields := &couchDocFields{}
+	if jsonDoc[revField] != nil {
+		docFields.revision = jsonDoc[revField].(string)
+	}
+	if jsonDoc[versionField] == nil {
+		return nil, fmt.Errorf("version field %s was not found", versionField)
 	}
+	docFields.versionAndMetadata = []byte(jsonDoc[versionField].(string))
```
manish-sethi: This line and the line 74 above can be simplified, if you declare the `docFields.versionAndMetadata` as string type

core/ledger/kvledger/txmgmt/statedb/statecouchdb/couchdoc_conv_test.go
```diff
@@ -0,0 +1,145 @@
...
+
+func testKVAndDocConversion(t *testing.T, kv *keyValue) {
+	doc, err := keyValToCouchDoc(kv)
+	require.NoError(t, err)
+	actualKV, err := couchDocToKeyValue(doc)
+	require.NoError(t, err)
+	require.Equal(t, kv, actualKV)
+}
+
+func testKVAndDocConversionUnsortedJSON(t *testing.T, kv *keyValue, sortedJSON string) {
```
manish-sethi: I don't find much value in this test, given that you already have `testKVAndDocConversion` that covers everything for all the fields and below `TestSortJSON` that covers more comprehensive content.

core/ledger/kvledger/txmgmt/statedb/statecouchdb/couchdoc_conv_test.go
```diff
@@ -0,0 +1,145 @@
...
+func TestKVAndDocConversion(t *testing.T) {
+	keyDelete := &keyValue{
+		"key0", "rev0",
+		&statedb.VersionedValue{
+			Value:    nil,
+			Version:  version.NewHeight(1, 0),
+			Metadata: nil,
+		},
+	}
+	testKVAndDocConversion(t, keyDelete)
```
manish-sethi: As these calls are pretty much independent, better to make them as subtests (using `t.Run`)

core/ledger/kvledger/txmgmt/statedb/statecouchdb/couchdb.go
manish-sethi: > We can pass `PUT db/"id" -d {_id, _rev}` to CouchDB. Hence, I do not see an issue having `id` as a string in struct and also as `_id` in the json bytes. Not a hack to me.
> 
That adds to the reason not to have the `id` field as it may give a wrong impression that this is how we are passing the id while interacting with couchdb.

> If you prefer not having `id string` in the couchdoc struct, I can add the following function and use it only when the namespace is empty (but the jsonValue would be marshalled twice once at `getKey` and another and `validateAndRetrieveFields` for valid keys in an empty namespace and that's fine as we don not expect more than 3 keys for now).

As an alternate, can you enhance the struct `couchDocFields` and function `validateAndRetrieveFields` to include the `id` field as well.

PR #1349, opened 2016-05-04
Use make to build the fabric

ghaskins:
RFC @bcbrock @corecode @srderson 

Specifically withholding DCO for now until testing and discussion are complete.

---

This patch adds a top-level makefile with various build targets:

*) peer (default) - builds the fabric ./peer/peer binary
*) membersrvc - builds the ./membersrvc/membersrvc binary
*) unit-test - runs the go-test based unit tests
*) behave - runs the behave test
*) behave-deps - ensures pre-requisites are availble for running behave manually
*) clean - cleans the build area
*) dist-clean - superset of 'clean' that also removes persistent state

Why?

golang is just one part.

of our build system.

```
 git pull && make clean peer
```

```
 a) automatically managed

  OR

```

current process.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,10 +33,10 @@ before_script:
 
  - echo " CREATING BASE IMAGE "
  - cd $HOME/gopath/src/github.com/$USER_NAME/$REPO_NAME/scripts && chmod +x foldercopy.sh && ./foldercopy.sh $TR_PULL_REQUEST $USER_NAME
- - cd /$HOME/gopath/src/github.com/hyperledger/$REPO_NAME/peer
  - sudo rm -rf /var/hyperledger/ && sudo mkdir /var/hyperledger/ && sudo chown $USER:$USER /var/hyperledger
+ - cd /$HOME/gopath/src/github.com/hyperledger/fabric
  - make peer && echo " STARTING PEER PROCESS "
- - ./peer peer &
+ - (cd ./peer; ./peer peer) &
  - make unit-test
 
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,14 +32,12 @@ install:
 before_script:
 
  - echo " CREATING BASE IMAGE "
- - echo " Calling docker.sh to build Openblockchain/baseimage "
- - cd $HOME/gopath/src/github.com/$USER_NAME/$REPO_NAME/scripts/provision/ && chmod +x docker.sh && ./docker.sh 0.0.9
  - cd $HOME/gopath/src/github.com/$USER_NAME/$REPO_NAME/scripts && chmod +x foldercopy.sh && ./foldercopy.sh $TR_PULL_REQUEST $USER_NAME
  - cd /$HOME/gopath/src/github.com/hyperledger/$REPO_NAME/peer
  - sudo rm -rf /var/hyperledger/ && sudo mkdir /var/hyperledger/ && sudo chown $USER:$USER /var/hyperledger
- - go build && echo " STARTING PEER PROCESS "
+ - make peer && echo " STARTING PEER PROCESS "
  - ./peer peer &
- - go test -timeout=20m $(go list github.com/hyperledger/fabric/... | grep -v /vendor/ | grep -v /examples/)
+ - make unit-test
 
 script:
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,9 +25,12 @@ install:
    echo " Installing Rocks DB, g++ compilers & Dependencies "
    sudo add-apt-repository -y ppa:ubuntu-toolchain-r/test && sudo apt-get -qq update && sudo apt-get -qq install g++-4.8 && sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-4.8 90
    sudo pip install --upgrade pip && sudo pip install behave && sudo apt-get install build-essential -y
-   cd /opt  && sudo git clone --branch v4.1 --single-branch --depth 1 https://github.com/facebook/rocksdb.git && cd rocksdb && sudo make shared_lib
-   sudo cp /opt/rocksdb/*.so* /usr/lib/ && export LD_LIBRARY_PATH=/opt/rocksdb:$LD_LIBRARY_PATH && sudo apt-get update && sudo apt-get install -y libsnappy-dev zlib1g-dev libbz2-dev
-   export CGO_CFLAGS="-I/opt/rocksdb/include" && export CGO_LDFLAGS="-L/opt/rocksdb -lrocksdb -lstdc++ -lm -lz -lbz2 -lsnappy"
+   sudo apt-get install -y libsnappy-dev zlib1g-dev libbz2-dev
+   cd /tmp
+   git clone --branch v4.1 --single-branch --depth 1 https://github.com/facebook/rocksdb.git
+   cd rocksdb
+   make shared_lib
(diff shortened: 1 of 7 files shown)


.travis.yml
```diff
@@ -32,14 +32,12 @@ install:
 before_script:
 
  - echo " CREATING BASE IMAGE "
- - echo " Calling docker.sh to build Openblockchain/baseimage "
- - cd $HOME/gopath/src/github.com/$USER_NAME/$REPO_NAME/scripts/provision/ && chmod +x docker.sh && ./docker.sh 0.0.9
  - cd $HOME/gopath/src/github.com/$USER_NAME/$REPO_NAME/scripts && chmod +x foldercopy.sh && ./foldercopy.sh $TR_PULL_REQUEST $USER_NAME
  - cd /$HOME/gopath/src/github.com/hyperledger/$REPO_NAME/peer
  - sudo rm -rf /var/hyperledger/ && sudo mkdir /var/hyperledger/ && sudo chown $USER:$USER /var/hyperledger
- - go build && echo " STARTING PEER PROCESS "
+ - make peer && echo " STARTING PEER PROCESS "
```
ghaskins, line 42: I just noticed this isn't going to work.  I need to cd to the TLD, do the make, and then cd into peer.

docs/API/SandboxSetup.md
```diff
@@ -23,9 +23,9 @@ From your command line terminal, move to the `devenv` subdirectory of your works
 
 To set up the local development environment with security enabled, you must first build and run the <b>Certificate Authority (CA)</b> server:
 
-    cd $GOPATH/src/github.com/hyperledger/fabric/membersrvc
-    go build
-    ./membersrvc
+    cd $GOPATH/src/github.com/hyperledger/fabric
+    make membersrvc
+    (cd menbersrvc; ./membersrvc)
```
ghaskins: typo in cd command

jyellick:
@ghaskins This looks like a great idea to me.  You said you chose make because it is ubiquitous and well understood.  Did you consider any other tools like Gradle (a tool I know @jeffgarratt is fond of)? I'm by no means an expert on build tools, just curious.

ghaskins:
@jyellick I didn't look at Gradle primarily b/c I am not familiar with it.  Of all the build tools, I am by far the most experienced with Make and to a lesser extent Maven and some of its derivatives like Leiningen. I figured there was probably some interesting projects out there, but my inclination was keep-it-simple so I went with the tried and true.  I am open to other ideas if there is a strong opinion that the assumed lower ubiquity/familiarity are worth the trade offs.

bcbrock:
I like the idea of having a Makefile - I was about to suggest this myself.  I have a similar Makefile in my busywork project which I will contrast with this proposed one.

First, I don't think the dependencies are working correctly. I would expect that if I say `make peer` that the peer would be rebuilt if any of the Go files changed. This does not happen for me  - I'm not sure what it would take to get the peer to be rebuilt here.

I have a target specifically to create the peer and membersrvc Docker images. In yours this is only a side-effect of running behave. (busywork also needs these images to run its Docker-compose networks).

The unit tests need to have a local peer running, don't they? That should be part of the make target I think. And the behave tests _can't_ have a peer running, which should also be part of the make target.

For consistency in testing I have defined a lot of system-wide cleanup targets:

```
############################################################################
# "Clean" routines

.PHONY: docker_chaincode_clean docker_ips \
    docker_fabric_clean docker_ps_clean docker_reset \
    fabric_clean fabric_reset

docker_chaincode_clean: docker_ps_clean
    -docker images | egrep '^dev-|^<none>' | \
        awk '{print $$3}' | xargs docker rmi -f

docker_ips:
    docker ps -a -q | xargs -n 1 docker_ip  

docker_fabric_clean: docker_chaincode_clean
    -docker rmi -f hyperledger-peer membersrvc

docker_ps_clean:
    -docker ps -a -q | xargs docker rm -f

docker_reset:
    sudo service docker stop
    sudo rm -rf /var/lib/docker
    sudo service docker start

fabric_clean:
    -pkill peer
    -pkill membersrvc

fabric_reset: fabric_clean docker_chaincode_clean
    rm -rf /var/hyperledger/*
```

Before running a test I typically do a `fabric_reset` that kills all peers and membersrvc, wipes out the blockchain database, kills all Docker processes and deletes all of the chaincode containers.

One of the reasons I hesitated suggesting my own Makefile was because testing consistency requires these Draconian measures, which then requires that  a user be running in a private VM (like Vagrant) or on a private system. I was thinking about how to guarantee that a user understood what havoc they might be wreaking if anyone else was using their machine at the same time.

ghaskins:
@bcbrock All excellent points.  Perhaps trying to capture execution environments (such as the unit-test and behave) were a bit overzealous at least in this early pass and I should stick with just dependency/pre-requisite management and let the user run these by hand?  Or should we go the other direction and try to automate the entire thing?  If the former, that simple.  If the latter, perhaps we really should look at more modern tooling (e.g. Gradle) as I know they are often more advanced in this regard.

Regarding the peer target not detecting code changes, this is a good point.  My initial inclination was that could be managed by the user (via a make clean cycle) because detecting all the .go code impacted might be expensive to do each time the user invokes make.  But now that you bring it up, I wonder if we can cache an index so that its only expensive the first time.

ghaskins:
@bcbrock another idea: if we switch to "go install" and make the target phony, go can figure it out and detect its own stale objects internally

smithbk:
Going forward, we also need to incorporate building and running tests (unit and behave) for each client SDK.   Anya will be doing the node client SDK portion.

ghaskins:
@smithbk yeah, i fully expect whatever this mechanism ends up looking like to be extended for cases such as the one you bring up.

ghaskins:
@bcbrock You had said: "In yours this is only a side-effect of running behave".  Note that its slightly more granular that this.  There is a "make behave-deps" which only generates the artifacts (if they don't already exist either from previously running behave-deps or the unit-test target).  Aside from the name implying a relationship to behave (which we can generalize), it seems that prereq could be utilized by your workflow too, couldn't it?

bcbrock:
I move for automating the whole thing with make targets. I have not seen anything complicated enough yet that cries out for another build system. To me the Makefile is just a repository for simple named scripts.

The .PHONY target should work with Go dependency checking. `go build` is pretty fast when nothing changed.

Yes, I could use behave-deps. I also have  a target `make behave_with_logs`.

If it were me I would also put some brief documentation at the top of the Makefile, like the bulleted list you included at the very top of this note, just so it's all in one place.

ghaskins:
@bcbrock seems reasonable to me.  I would prefer to stick with Make as well for the reasons already mentioned (ubiquity, simplicity, and its what I know).

Regarding your comments about phony+go-build and the documentation, those are good, ill add them to the PR.

Regarding "behave-deps" name, any suggestion for a more general name?

Regarding the feature automation, my inclination would be to leave them approximately what I have now and we can enhance their function over time with follow up PRs.

Any objections?

bcbrock:
@ghaskins I say `make images`, with options to build only 1 or the other since they take quite a while

jeffgarratt:
@ghaskins @jyellick @srderson  If we could get a java JDK installed then we could add Gradle at any point.  I think someone mentioned doing this, not sure if it was in plan or already there.

ghaskins:
@jeffgarratt FYI, JRE is available in vagrant/docker as of PR #1250.   Not sure that helps of if you need the full JDK

ghaskins:
Replaced by PR #1374

Issue #1350, opened 2016-05-04
Content of new docs/dev-setup/install.md file is backleveled and has broken links.


lehors:
This new file is supposed to hold the installation instructions that were in the top README.md file but this is from an older version than what we had before the move and the relative links were not updated to reflect the new location.

lehors:
I'll submit a PR shortly.

PR #1352, opened 2016-05-04
Fix new Install file - issue-1350

lehors:
[ci-skip]
This patch addresses issue  #1350 .
Content is brought back to latest version and broken links are fixed.

### DIFF
--- a/docs/dev-setup/install.md
+++ b/docs/dev-setup/install.md
@@ -11,11 +11,14 @@ Install the blockchain fabric by completing the following tasks:
 * [Working with CLI, REST, and Node.js](#working-with-cli-rest-and-nodejs-)
 * [Configuration](#configuration-)
 * [Logging](#logging-)
-* [Generating grpc code](#generating-grpc-code-)
+* [Generating gRPC code](#generating-gRPC-code-)
 * [Adding or updating Go packages](#adding-or-updating-go-packages-)
 
+## License <a name="license"></a>
+This software is made available under the [Apache License Version 2.0](LICENSE).
+
 ## Building the fabric core <a name="build"></a>
-The following instructions assume that you have followed the [development environment getting started instructions](docs/dev-setup/devenv.md).
+The following instructions assume that you have followed the [development environment getting started instructions](devenv.md).
 
 To access your VM, run
 ```
@@ -26,18 +29,19 @@ From within the VM, you can build, run, and test your environment.
 
 #### 1. Go build
 ```
-cd $GOPATH/src/github.com/hyperledger/fabric
-go build -o peer
+cd $GOPATH/src/github.com/hyperledger/fabric/peer
+go build
 ```
 
-#### 2. Run
+#### 2. Run/Execute
 
-To see what commands are available, simply execute the following command:
+To see what commands are available, simply execute the following commands:
+```
+cd $GOPATH/src/github.com/hyperledger/fabric/peer
+./peer
+```
 
-    cd $GOPATH/src/github.com/hyperledger/fabric
-    ./peer
-
-You should see some output similar to below (**NOTE**: The root command below is hardcoded in the [main.go](./main.go). Current build will actually create a *peer* executable file).
+You should see some output similar to below (**NOTE**: The root command below is hardcoded in the [main.go](../../peer/main.go) and the build creates the `peer` executable).
 
 ```
     Usage:
@@ -61,7 +65,7 @@ You should see some output similar to below (**NOTE**: The root command below is
 
 ```
 
-The **peer** command will run the peer process. You can then use the other commands to interact with the peer process. For example, the `status` command will return the status of the running peer.
+The `peer` command will initiate a peer process, with which one can interact by executing other commands. For example, the `status` command will return the status of the running peer.
 
 #### 3. Test
 New code must be accompanied by test cases both in unit and Behave tests.
@@ -74,7 +78,7 @@ To run all unit tests, in one window, run `./peer peer`. In a second window
 
 Note that the first time the tests are run, they can take some time due to the need to download a docker image that is about 1GB in size. This is why the timeout flag is added to the above command.
 
-To run a specific test use the `-run RE` flag where RE is a regular expression that matches the test case name. To run tests with verbose output use the `-v` flag. For example, to run the `TestGetFoo` test case, change to the directory containing the `foo_test.go` and enter:
+To run a specific test use the `-run RE` flag where RE is a regular expression that matches the test case name. To run tests with verbose output use the `-v` flag. For example, to run the `TestGetFoo` test case, change to the directory containing the `foo_test.go` and call/excecute
 
     go test -v -run=TestGetFoo
 
@@ -90,19 +94,19 @@ Some of the Behave tests run inside Docker containers. If a test fails and you w
 behave -D logs=Y
 ```
 
-Note, you must run the unit tests first to build the necessary Peer and Member Services docker images. These images can also be individually built using the commands
+Note, you must run the unit tests first to build the necessary `peer` and `member services` docker images. These images can also be individually built when `go test` is called with the following parameters:
 ```
 go test github.com/hyperledger/fabric/core/container -run=BuildImage_Peer
 go test github.com/hyperledger/fabric/core/container -run=BuildImage_Obcca
 ```
 
 ## Building outside of Vagrant <a name="vagrant"></a>
-This is not recommended, however some users may wish to build outside of Vagrant if they use an editor with built in Go tooling. The instructions are below.
+While not recommended, it is possible to build the project outside of Vagrant (e.g., for using an editor with built-in Go toolking). In such cases:
 
 - Follow all steps required to setup and run a Vagrant image:
-  - Make sure you you have [Go 1.6](https://golang.org/) or later installed
+  - Make sure you you have [Go 1.6](https://golang.org/) installed
   - Set the maximum number of open files to 10000 or greater for your OS
-  - Install [RocksDB](https://github.com/facebook/rocksdb/blob/master/INSTALL.md) version 4.1 and it's dependencies:
+  - Install [RocksDB](https://github.com/facebook/rocksdb/blob/master/INSTALL.md) version 4.1 and its dependencies
 ```
 apt-get install -y libsnappy-dev zlib1g-dev libbz2-dev
 cd /tmp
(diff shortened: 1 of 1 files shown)


lehors:

ghaskins:
LGTM

This should be merged ahead of 1349 since they both touch the markdown.

core/ledger/ledger_interface.go
```diff
@@ -700,6 +700,14 @@ type HashProvider interface {
 	GetHash(opts bccsp.HashOpts) (hash.Hash, error)
 }
 
+// ChannelInfoProvider interface enables ledger components to retrieve channel config data in all the config
+// blocks (current and previous config blocks).
+type ChannelInfoProvider interface {
```
cendhu: I would like to understand where this interface would be used. I assume that we would introduce a new API in the `DeployedChaincodeInfoProvider` that would receive this interface to get all mspIDs to construct the name of implicit collections. Is that correct?

core/ledger/ledger_interface.go
```diff
@@ -709,3 +717,4 @@ type HashProvider interface {
 //go:generate counterfeiter -o mock/cc_event_listener.go -fake-name ChaincodeLifecycleEventListener . ChaincodeLifecycleEventListener
 //go:generate counterfeiter -o mock/custom_tx_processor.go -fake-name CustomTxProcessor . CustomTxProcessor
 //go:generate counterfeiter -o mock/cc_event_provider.go -fake-name ChaincodeLifecycleEventProvider . ChaincodeLifecycleEventProvider
+//go:generate counterfeiter -o mock/channelinfo_provider.go -fake-name ChannelInfoProvider . ChannelInfoProvider
```
cendhu: The `ChannelInfoProvider` interface and the respective mock look disjoint from this PR. As there is no clarity on how this interface and mock would be used, it isn't clear to me why we have these two. Is the mock to be used by lifecycle pkg?

core/ledger/kvledger/channelinfo_provider.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/hyperledger/fabric/bccsp/factory"
+	"github.com/hyperledger/fabric/common/channelconfig"
+	"github.com/hyperledger/fabric/common/ledger/blkstorage"
+	"github.com/hyperledger/fabric/protoutil"
+)
+
+// ChannelInfoProvider implements ledger.ChannelInfoProvider interface
+type ChannelInfoProvider struct {
+	BlkStoreProvider *blkstorage.BlockStoreProvider
+}
```
cendhu: When will this be used? Will the `kv_ledger_provider.go` create a new `ChannelInfoProvider` and pass it to privacyenabledstate.`GetDBHandle()` while opening the ledger for a given channel so that it can call the `DeployedInforCCProvider`? 

If this struct is to be created by the kvledger pkg, shouldn't we have the following?
```suggestion
type ChannelInfoProvider struct {
	blockStore *blkstorage.BlockStore
}

func NewChannelInfoProvider(channelName string) (*ChannelInfoProvider, error) {
        blockStore, err := p.BlkStoreProvider.Open(channelName)
	if err != nil {
		return nil, err
	}
	return &ChannelInfoProvider{blockStore}, err
}
```
**OR**
```suggestion
type ChannelInfoProvider struct {
	blockStore *blkstorage.BlockStore
}

func NewChannelInfoProvider(blockStore blkstorage.BlockStore) (*ChannelInfoProvider, error) {
	return &ChannelInfoProvider{blockStore}, err
}
```

core/ledger/kvledger/channelinfo_provider.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+	blockchainInfo, err := blockStore.GetBlockchainInfo()
+	if err != nil {
+		return nil, err
+	}
+
+	// Iterate over the config blocks to get all the channel configs, extract MSPIDs and add to mspidsMap
+	mspidsMap := map[string]string{}
+	blockNum := blockchainInfo.Height
+	for blockNum > 0 {
```
cendhu: When I read this first, it added some confusion. As the `blockNum > 0` condition is evaluated first, it provided misinformation that we skip the genesis block. I think the `blockNum` denotes the height of block store in the first run. After that, it denotes the blockNum. To avoid this misinformation, confusion, and improve readability, I have a list of suggestions (S1, S2, S3, S4) to address this readability issue. If there is a better way to avoid this confusion, I am fine with that.

core/ledger/kvledger/channelinfo_provider.go
```diff
@@ -0,0 +1,79 @@
...
+	mspids := make([]string, 0, len(mspidsMap))
+	for mspid := range mspidsMap {
+		mspids = append(mspids, mspid)
+	}
+	return mspids, nil
+}
+
+// mostRecentConfigBlockBelow fetches the most recent config block below the blockNum
+func mostRecentConfigBlockBelow(blockNum uint64, blockStore *blkstorage.BlockStore) (*cb.Block, uint64, error) {
+	block, err := blockStore.RetrieveBlockByNumber(blockNum - 1)
```
cendhu: These two lines also add some confusion. When the block number is 0, i.e., the genesis block, the code reads like we try to fetch block with number -1 (that should get rounded to `Math.MaxUint64`). I understand that when the channel has only the genesis block, the argument to the function would be the height of the block store, i.e., 1 as per line 36. It is good to improve readability and avoid this confusion. The list of suggestions - S1, S2, S3, S4 fix this confusin.

core/ledger/kvledger/channelinfo_provider.go
```diff
@@ -0,0 +1,79 @@
...
+func (p *ChannelInfoProvider) GetAllMSPIDs(channelName string) ([]string, error) {
+	blockStore, err := p.BlkStoreProvider.Open(channelName)
+	if err != nil {
+		return nil, err
+	}
+	blockchainInfo, err := blockStore.GetBlockchainInfo()
+	if err != nil {
+		return nil, err
+	}
+
```
cendhu, line 32: **S1**
```suggestion
	if blockchainInfo.Height == 0 {
		return nil, nil	
	}
```

core/ledger/kvledger/channelinfo_provider.go
cendhu: **S2**
```suggestion
	blockNum := blockchainInfo.Height - 1
	for {
```

core/ledger/kvledger/channelinfo_provider.go
```diff
@@ -0,0 +1,79 @@
...
+	// Iterate over the config blocks to get all the channel configs, extract MSPIDs and add to mspidsMap
+	mspidsMap := map[string]string{}
+	blockNum := blockchainInfo.Height
+	for blockNum > 0 {
+		configBlock, configBlockNum, err := mostRecentConfigBlockBelow(blockNum, blockStore)
```
cendhu: **S3**
```suggestion
		configBlock, err := mostRecentConfigBlockAsOf(blockNum, blockStore)
```

- The block number is already present in the configBlock. Hence, we don't need to explicitly return it. 
- We can make the `mostRecentConfigBlockAsOf()` as an local method of `channelInfoProvider` if we decide to have blockStore in the ChannelInfoProvider instead of the blockStoreProvider so that we can avoid the 2nd argument too.

core/ledger/kvledger/channelinfo_provider.go
```diff
@@ -0,0 +1,79 @@
...
+		}
+
+		orgs := channelConfig.ApplicationConfig().Organizations()
+		for orgName, appOrg := range orgs {
+			mspid := appOrg.MSPID()
+			if _, ok := mspidsMap[mspid]; !ok {
+				mspidsMap[mspid] = orgName
+			}
+		}
+		blockNum = configBlockNum
```
cendhu: **S4**
```suggestion
		if configBlock.Header.BlockNumber == 0 {
		        break
		}
		blockNum = configBlock.Header.BlockNumber - 1
```

core/ledger/kvledger/channelinfo_provider.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	mspids := make([]string, 0, len(mspidsMap))
```
cendhu, line 58:
```suggestion
	mspids := make([]string, len(mspidsMap))
```

core/ledger/kvledger/channelinfo_provider.go
```diff
@@ -0,0 +1,79 @@
...
+			return nil, err
+		}
+
+		orgs := channelConfig.ApplicationConfig().Organizations()
+		for orgName, appOrg := range orgs {
+			mspid := appOrg.MSPID()
+			if _, ok := mspidsMap[mspid]; !ok {
+				mspidsMap[mspid] = orgName
+			}
+		}
```
cendhu: `orgName` isn't used right?

common/channelconfig/application.go
```diff
@@ -66,6 +66,16 @@ func (ac *ApplicationConfig) Organizations() map[string]ApplicationOrg {
 	return ac.applicationOrgs
 }
 
+// MSPIDs returns MSP IDs for the organizations
+func (ac *ApplicationConfig) MSPIDs() []string {
```
manish-sethi: Missed a UT for this function?

common/channelconfig/util.go
```diff
@@ -241,51 +241,56 @@ func ACLValues(acls map[string]string) *StandardConfigValue {
 
 // ValidateCapabilities validates whether the peer can meet the capabilities requirement in the given config block
 func ValidateCapabilities(block *cb.Block, bccsp bccsp.BCCSP) error {
+	cc, err := ExtractChannelConfig(block, bccsp)
+	if err != nil {
+		return err
+
```
manish-sethi: Extra line

core/ledger/kvledger/channelinfo_provider.go
```diff
@@ -0,0 +1,81 @@
...
+		if channelConfig.ApplicationConfig() == nil {
+			return nil, errors.Errorf("could not get application config for channel '%s'", p.channelName)
+		}
+		mspids := channelConfig.ApplicationConfig().MSPIDs()
+		for _, mspid := range mspids {
+			if _, ok := mspidsMap[mspid]; !ok {
+				mspidsMap[mspid] = ""
+			}
+		}
+
```
manish-sethi: Missed replacing this by the newly added function?

core/ledger/kvledger/channelinfo_provider.go
```diff
@@ -0,0 +1,81 @@
...
+	blockchainInfo, err := p.blockStore.GetBlockchainInfo()
+	if err != nil {
+		return nil, err
+	}
+	if blockchainInfo.Height == 0 {
+		return nil, nil
+	}
+
+	// Iterate over the config blocks to get all the channel configs, extract MSPIDs and add to mspidsMap
+	mspidsMap := map[string]string{}
```
manish-sethi: use rather `map[string]struct{}` type

Issue #470, opened 2016-01-12
repo: correct file mode (execute) introduced in a287fb1b


corecode:
a287fb1b introduced a repo-wide change in file modes, adding a +x.  We should remove these +x again.

christo4ferris:
@corecode @prjayach - this is odd... any idea why the permissions were changed? Agree they should be changed back

prjayach:
Not sure what happened here. I agree +x should not be there.

lehors:
I'll put a PR together for this.

PR #1357, opened 2016-05-05
Issue 1158 - Peer command usage

lehors:
This is a significant PR which changes the peer command usage per @jeffgarratt's suggestion to the following:

```
  node
    start      Starts the node.
    status     Returns status of the node.
    stop       Stops the running node.
  network
    login       Logs in user to CLI.
    list        Lists all network peers.
  chaincode
    deploy      Deploy the specified chaincode to the network.
    invoke      Invoke the specified chaincode.
    query       Query using the specified chaincode.
    help        Help about any command
```


srderson:
@lehors I haven't completed review yet, but this PR has a merge conflict and will need to be rebased on master. I can complete a first pass review without rebase so no rush.

ghaskins:
Overall, I think this is an improvement.  I am still a little bothered by the nomenclature for "peer chaincode|network".  The chaincode|network sub-function strikes me as a client to a peer, not a peer function itself.  Assuming I am not off in the weeds, it strikes me that "peer [start|status|stop]" and some other binary (e.g. "client") with functions for [chaincode|network] might make more sense.  This shouldn't be construed as a criticism of this patch: the subfunction relationship was preexisting and this patch moves us in a positive direction.  I am not just sure it's far enough.  I can be convinced otherwise.

christo4ferris:
@ghaskins I tend to agree, though I could also see the peer being a simple daemon process that is simply started, and a separate commandline binary that interacts with peers across the landscape - not just the one running locally.

@srderson @binhn has any design work been done around this? eg. what if my peer is running in the cloud - wouldn't I interact with that by logging in eg. (hlf short name for hyperledger fabric)
    hlf -login <some credential(s)> <network identifier>
    hlf -list
    hlf -status <node>
    hlf -deploy <chaincode> <args>
    hlf -invoke <chaincode> <args>

Also, it seems that starting and stopping a node should be a function of some sort of orchestration layer (k8s, swarm, openstack, ECS, Azure, whatever) that launches and then monitors and keeps an instance(s) running, etc.

ghaskins:
@christo4ferris I see your point, and this makes sense to me.  I think my hangup is that a single binary implements both the client and server functions more than the command structure itself.  It is a minor nitpick, though, so if others are ok with this change, I have no objections to merging.

srderson:
@christo4ferris @ghaskins @jeffgarratt - Originally it made sense (to me at least) that the client and server functions were a single binary when the vision was that every it was a full p2p network where there is little distinction between client and server, similar to Bitcoin, Ethereum, etc. Now it seems that requirements have led many to believe that there will be a much smaller number of servers and a larger number of client only nodes, similar to a full node vs lite wallet.

I think the [client SDK](https://github.com/hyperledger/fabric/wiki/Client-SDK-Specification) is an attempt to separate out some of this functionality. Eventually, someone may write a Go CLI without any peer functionality, something similar to what @christo4ferris is describing, though I'm not aware that it has been discussed.

I agree that the `peer` binary name makes the `node|chaincode|network` commands a bit wonky. Perhaps `hlf` would have been a better name for the binary?

lehors:
I don't mind taking another pass to rename the peer command to something else if you guys prefer. Maybe it's just a matter of getting used to but I find hlf kind of hard to read though. Putting aside all your points without implying they aren't valid, peer is easy to say, type, and read. hlf is easy to say, type and not so to read.
Another possibility would be hlpf (hyperledger project fabric). Easier to read than hlf but harder to type. One can't have it all. :-)

binhn:
I like `peer` but would not mind `hf`, quite easy to type and read

christo4ferris:
I could see hf. I also know we have caused endless rebases for @lehors and we really need to refactor to have a peer (or peerd) and a separate cli so maybe we do this later? I'm torn because it will be somewhat painful for users the longer we wait to change this.

christo4ferris:
LGTM

bddtests/peer_basic.feature
```diff
@@ -10,7 +10,7 @@
 #@chaincodeImagesUpToDate
```
ghaskins, line 10: Why the +x change?

bddtests/peer_basic.feature
```diff
@@ -10,7 +10,7 @@
...
 Feature: lanching 3 peers
     As a HyperLedger developer
-    I want to be able to launch a 3 peers
+    I want to be able to launch 3 peers
```
ghaskins, line 13: This is the only change making me nervous.  I am by no means a behave expert, but my understanding is this is a grammar change but I don't see any other hunks that correspond to this one when I would have expected this.  It's possible that I am either not seeing the corresponding update, or misunderstanding something fundamental.  Probably best if @jeffgarratt can comment on this particular hunk.

bddtests/compose-defaults.yml
```diff
@@ -6,7 +6,7 @@ vp:
     # TODO:  This is currently required due to BUG in variant logic based upon log level.
```
ghaskins, line 6: Another rogue +x?

bddtests/peer_basic.feature
christo4ferris, line 13: @ghaskins that's part of the Feature description (just optional text), not part of the DSL. However, kudos for being wary.

ghaskins, line 13: Ah, ok.  Thank you for the explanation!

ghaskins:
Aside from my note about a few suspicious looking +x changes, LGTM.  I'll leave it to @christo4ferris and others to decide if they want those fixed before merging, but otherwise feel comfortable merging this at any time.  Users will need to be notified of the change so the merge should be timed accordingly.

christo4ferris:
@ghaskins @srderson @binhn agree we want to clearly communicate this. we need notices in slack and email. as for the +x we should clean that up, IMO. However, we have a very small number of PRs now that a major overhaul would not be that painful. Let's rip the bandaid off quickly. 

SO, I would say let's merge this if we are ready, clearly communicate and then plan to address the +x issue.

lehors:
Sorry for the bogus +x, my emacs for some reason was changing the file permission on save. I have now fixed that and will revert the changes.

lehors:
File modes fix would probably have qualified as a "ci skip" commit but I decided it wasn't time to let something break everything so I let the tests run.
Looking at the file permissions I see a whole bunch throughout the project that would deserve to be cleaned up (that have nothing to do with this PR). This is something that probably ought to be addressed in a specific PR later on if we care.

lehors:
Regarding file modes, I just saw we have a long standing issue on this: issue #470.

christo4ferris:
correct... we'll need to clean that up.

srderson:
I'll likely merge this tomorrow morning. Will post a message in #fabric channel and send a note on the mailing list.

srderson:
LGTM. Thanks @lehors

PR #1363, opened 2016-05-05
add performance instrumentation and perfutil_readme.md

dongmingh:
Signed-off-by: Dongming Hwang dming@us.ibm.com


binhn:
@dongmingh Thanks for submitting the pull request.

I would like to improve a bit on the following:
1. perfutil.go should go into /core/util/ directory
2. perfutil_readme.md should go into /docs/tech and rename to perf_trace.md
3. fix misspellings in perfutil_readme.md, include capitalizing Hyperledger Fabric and remove IBM
4. for env var and REST endpoint, it would be better to call it "perftrace" rather than "perfutil"
5. why would REST endpoint /perfutil has the body of a "chaincodeSpec" ?
6. better names for the "actions" and should be defined as constants
7. call to perfutil.PerfTraceHandler() should be guarded as it is being called multiple times in the critical path
8. we can enhance later to provide summary average stats 
9. devops is not going to be in the main path when we complete the client SDK, so I would suggest the look into instrumenting core/peer.ProcessTransaction() where all transactions go through
10.

christo4ferris:
This PR is using the incorrect license header. Use [this one](https://github.com/hyperledger/fabric/blob/master/docs/dev-setup/headers.txt), please. Substitute "IBM 2016" for the "[COPYRIGHT OWNER]".

I'm also a bit unclear why we are developing code to handle profiling when there are so many libraries available that we might use.

christo4ferris:
If this PR is to be merged, it needs tests for perfutil.go at the very least. @bmos299 @binhn we should not be merging code without tests.

dongmingh:
@christo4ferris 
We will create a perftrace_test.go for test and evaluate Go-Metrics.

@binhn concerning your comments:
1.perfutil.go should go into /core/util/ directory
2.perfutil_readme.md should go into /docs/tech and rename to perf_trace.md

For above two comments, can we create a subdirectory, core/util/perftrace, to contain all related go files and readme,md (rename to README.md)? Then the user can read the README.md without clicking the file when they switch to that directory

3.fix misspellings in perfutil_readme.md, include capitalizing Hyperledger Fabric and remove IBM

We will take care of this.

4.for env var and REST endpoint, it would be better to call it "perftrace" rather than "perfutil"

We will rename perfutil to perftrace.

5.why would REST endpoint /perfutil has the body of a "chaincodeSpec" ?

We will change this not to use chaincodeSpec.

6.better names for the "actions" and should be defined as constants

We will rename action to reqType

7.call to perfutil.PerfTraceHandler() should be guarded as it is being called multiple times in the critical path

We will guard this API with lock()/unlock() and logic to prevent invalid uuid (nil or a different uuid from PerfUuid)

8.we can enhance later to provide summary average stats

We will add more functions, such as average and concurrence, to the perftrace later.

9.devops is not going to be in the main path when we complete the client SDK, so I would suggest the look into instrumenting core/peer.ProcessTransaction() where all transactions go through

We will move starting/ending tracepoints from devops to core/peer.ProcessTransaction().  Moreover, when tracepoints (neither starting nor ending) are deleted or added, the perftrace will calculate the elapsed time based on the new segments, i.e., there will be no harm to the perftrace.

binhn:
@christo4ferris A number of folks chased various tools but none gave us the info we need, especially related to gRPC and breaking the transaction path into segments.

binhn:
@dongmingh let's talk -- i will google hangout with you

core/perfutil/perfutil.go
```diff
@@ -0,0 +1,252 @@
...
+//            Create - create hash table entry and start a link list entry with the first entry having the name provided
+//			do not care: res
+//            CreatePTOP - close current link list entry and create a new link list entry with the name provided
+//			do not care: res
+//            EndPTOP - close current link list entry and end current link list
+//			do not care: name and now
+//            CreatePTOPFrom - close current link list entry, create a link list entry with the name and timestamps (now) provided,
+//                             and start a new one from Now() with the same trace name as the newly created entry with preface "Since_"
+//			do not care: res
+func PerfTraceHandler(uuid string, name string, now int64, res bool, action string) {
```
corecode, line 169: I think action should be an enum (`const`), instead of a string.

core/chaincode/handler.go
```diff
@@ -565,7 +577,18 @@ func (handler *Handler) handleGetState(msg *pb.ChaincodeMessage) {
...
 
 		readCommittedState := !handler.getIsTransaction(msg.Uuid)
+
+		//calculates time in ledger
+		perfutil.PerfTraceHandler(msg.Uuid, "GetLedger", 0, true, "CreatePTOP")
+
 		res, err := ledgerObj.GetState(chaincodeID, key, readCommittedState)
+
+		if err != nil {
+			perfutil.PerfTraceHandler(msg.Uuid, "GetLedger_done", 0, false, "CreatePTOP")
```
corecode, line 587: can pass `err == nil` as argument, instead of using a conditional.

core/perfutil/perfutil.go
```diff
@@ -0,0 +1,252 @@
...
+	case "CreatePTOP":
+		tt := GetPT(uuid)
+		EndCurrPTOP(tt, true)
+		tt = GetPT(uuid)
+		CreatePTOP(tt, name, now)
+	
+	case "EndPTOP":
+		tt := GetPT(uuid)
+		EndCurrPTOP(tt, res)
+		fmt.Printf("[PerfTraceHandler] ... action[%s]\n", action)
```
corecode, line 200: is this a leftover debug printf?

core/perfutil/perfutil.go
```diff
@@ -0,0 +1,252 @@
...
+
+//get the pointer to the end of the link list
+func GetPT(uuid string) *traceOperation {
+	if enableTrace {
+		to := allTransactionTraces[uuid]
+		if to == nil {
+			fmt.Printf("[GetPT] Transaction trace nil. \n")
+			return to
+		}
+		for {
```
corecode, line 75: Why is this a linked list with O(N) behavior, instead of using either front insertion (O(1)) or simply using a slice with `append`?

core/perfutil/perfutil.go
corecode, line 169: I think this API entry is too complex.  I suggest providing multiple functions, maybe even one per action type.

Most of the time `PerfTraceHandler` is being invoked with the same set of arguments - `now = 0, res = true, action = "CreatePTOP"` (often also `uuid = GetPerfUuid()`).

Lastly, `perfutil.PerfTraceHandler()` stutters.  I suggest dropping `Perf` from all function names, and shortening them.  E.g. `perfutil.Trace("foofunc")`.

core/chaincode/shim/handler.go
```diff
@@ -213,7 +214,7 @@ func (handler *Handler) handleInit(msg *pb.ChaincodeMessage) {
 			payload := []byte(unmarshalErr.Error())
 			// Send ERROR message to chaincode support and change state
 			chaincodeLogger.Debug("[%s]Incorrect payload format. Sending %s", shortuuid(msg.Uuid), pb.ChaincodeMessage_ERROR)
-			nextStateMsg = &pb.ChaincodeMessage{Type: pb.ChaincodeMessage_ERROR, Payload: payload, Uuid: msg.Uuid}
```
corecode, line 216: Why do you store `TraceName` and `Start` in the message, instead of tracing directly here?

core/perfutil/perfutil.go
muralisrini, line 169: Agree perfutil.Trace(...).

I like there is just one method. Keeps it simple -  good for search and "tracing the trace" which I suspect we'll be doing a lot of.

dongmingh, line 169: We can split this API into many simply APIs.  Each API only does one thing. Then the "action" will not be needed.

core/perfutil/perfutil.go
dongmingh, line 75: We want to keep the linked list in the order of tracepoints so that the trace can be printed in the order they are added.  I will look into slice with append to have O*1) behavior.

core/chaincode/shim/handler.go
dongmingh, line 216: This is to measure the latency of gRPC.  The receiver can add a tracepoint based on the TraceName and Start contained in this message.

core/perfutil/perfutil.go
christo4ferris, line 169: I agree the API is too complex.

christo4ferris, line 169: Am I right in understanding that only one uuid can be traced at a time? That seems ill advised. Why not simply time functions? I still maintain that go-metrics would be a better choice and would give us far more capability to monitor what's going on than simply recording the time along the path of a transaction's life.

christo4ferris:
@dongmingh please transfer this to Gerrit or close. Thanks!

dongmingh:
close.

PR #1367, opened 2016-05-05
Added support for multiple NVP rootnodes random selection

yacovm:
The root node is selected by looking at peer.discovery.rootnode with viper.
Non validating peers can only have 1 root node, so they can't do load balancing between
different validating peers.
Now the peer.discovery.rootnode can be a comma separated list and if the peer is non validating,
it will randomly select one of them.
If the peer is a validating peer and it's still given a comma separated list, it selectes the first peer (this is on purpose, because random selection with validating peers may reach a state of 2 mutually exclusive membership views (e.g {vp0, vp1}, {vp2, vp3})).

I tested it by running some validating peers and also spawning a non validating peer with CORE_PEER_DISCOVERY_ROOTNODE=172.17.0.2:30303,172.17.0.3:30303,172.17.0.4:30303,172.17.0.5:30303 and ran deployment + invocations through the non validating peer.

### DIFF
--- a/core/discovery_test.go
+++ b/core/discovery_test.go
@@ -19,14 +19,91 @@ under the License.
 
 package core
 
-import "testing"
+import (
 
-func TestDiscovery_GetRootNode(t *testing.T) {
-	rootNode, err := GetRootNode()
+	"github.com/hyperledger/fabric/discovery"
+
+	"testing"
+	"github.com/spf13/viper"
+)
+
+func TestDiscovery_GetEmptyRootNode(t *testing.T) {
+	discovery.SetDiscoveryService(NewStaticDiscovery(true))
+	assertRootNode(t,"")
+
+}
+
+func TestDiscovery_GetSingleValidatingPeer(t *testing.T) {
+	viper.Set("peer.discovery.rootnode","someHost")
+	discovery.SetDiscoveryService(NewStaticDiscovery(true))
+	assertRootNode(t,"someHost")
+}
+
+// Check that function always returns first one
+func TestDiscovery_GetMultiValidatingPeer(t *testing.T) {
+	viper.Set("peer.discovery.rootnode","a,b,c,d,e,f,g,h,i,j,k,l,m,n,o,p,q,r,s,t,u,v,w,x,y,z")
+	discovery.SetDiscoveryService(NewStaticDiscovery(true))
+
+	for i := 0; i < 100; i++ {
+		assertRootNode(t,"a")
+	}
+}
+
+// Check that function always returns first one
+func TestDiscovery_GetMultiNonValidatingPeer(t *testing.T) {
+	viper.Set("peer.discovery.rootnode","a,b,c,d,e")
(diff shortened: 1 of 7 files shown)


yacovm:

yacovm:
The build failed on:
`[31;1mThe command "go test -timeout=20m $(go list github.com/hyperledger/fabric/... | grep -v /vendor/ | grep -v /examples/)" failed and exited with 1 during .[0m
`
But it passes on my machine, attached terminal output below:
[goTest.txt](https://github.com/hyperledger/fabric/files/251959/goTest.txt)

Can anything be done with this?

yacovm:
OK it seems the tests passed now ( I guess, because it's Sunday and no one is running anything). 

Can someone please take a look?

integration/raft/cft_test.go
```diff
@@ -115,6 +120,12 @@ var _ = Describe("EndToEnd Crash Fault Tolerance", func() {
 			o1Proc.Signal(syscall.SIGKILL)
 			Eventually(o1Proc.Wait(), network.EventuallyTimeout).Should(Receive(MatchError("exit status 137")))
 
+			By("setting etcdraft log level to debug")
+			setOrdererLogLevel(`{"spec":"orderer.consensus.etcdraft=debug:info"}`, o2, network)
+
+			By("observing active nodes to shrink")
+			Eventually(o2Runner.Err(), time.Minute, time.Second).Should(gbytes.Say("Current active nodes in cluster are: \\[2 3\\]"))
```
sykesm: We try to use network.EventuallyTimeout where we can.

integration/raft/cft_test.go
```diff
@@ -115,6 +120,12 @@ var _ = Describe("EndToEnd Crash Fault Tolerance", func() {
 			o1Proc.Signal(syscall.SIGKILL)
 			Eventually(o1Proc.Wait(), network.EventuallyTimeout).Should(Receive(MatchError("exit status 137")))
 
+			By("setting etcdraft log level to debug")
+			setOrdererLogLevel(`{"spec":"orderer.consensus.etcdraft=debug:info"}`, o2, network)
```
sykesm: Why do we need to set the log level in this way? I'd much rather see the `OrdererRunner` function extended to look like `PeerRunner` and get passed a variadic list of environment variables. Then, when you start the runner, you can simply set the logspec you want using the config variable.

Gets rid of a bunch of boilerplate and removes moving parts.

integration/raft/cft_test.go
```diff
@@ -737,3 +748,36 @@ func configFromBootstrapBlock(bootstrapBlock []byte) *common.Config {
...
+
+	return &http.Client{
+		Transport: &http.Transport{
+			TLSClientConfig: &tls.Config{
+				Certificates: []tls.Certificate{clientCert},
+				RootCAs:      clientCertPool,
+			},
+		},
+	}
+}
```
sykesm: This should all go.

Issue #1364, opened 2016-05-05
PBFT crash tollerance needs additional test scenarios


jyellick:
Per comments on PR #1325 and the inclusion of @corecode's persistence changeset, additional tests should be added to exercise the crash fault tolerance of PBFT.

This has been partially addressed already via #1336, but this only addressed a very basic crash fault scenario.  Scenarios which include view change at startup, as well as state transfer should be included.

PR #1371, opened 2016-05-05
Add unit test for crash fault including view change

jyellick:
This is intended to help address #1364 

This test simulates a more complex crash fault than the existing units tests. It relies on the existence of recovered checkpoints to trigger a view change, so that the network can process a new request.

This test did not uncover any additional bugs, but should be useful to help prevent regression in the future, and because it is a unit test, adds negligible time to the test process.

Included outside of the test file are a few logging enhancements to make debugging easier.

Review from @corecode, @tuand27613, or @kchristidis would be helpful, though as this does not change any PBFT logic, PBFT expertise is less required.

### DIFF
--- a/consensus/obcpbft/pbft-core_test.go
+++ b/consensus/obcpbft/pbft-core_test.go
@@ -1144,3 +1144,89 @@ func TestReplicaCrash2(t *testing.T) {
 		}
 	}
 }
+
+// TestReplicaCrash3 simulates the restart requiring a view change
+// to a checkpoint which was restored from the persistance state
+// Replicas 0,1,2 participate up to a checkpoint, then all crash
+// Then replicas 0,1,3 start back up, and a view change must be
+// triggered to get vp3 up to speed
+func TestReplicaCrash3(t *testing.T) {
+	validatorCount := 4
+	net := makePBFTNetwork(validatorCount, func(pep *pbftEndpoint) {
+		pep.pbft.K = 2
+		pep.pbft.L = 2 * pep.pbft.K
+	})
+	defer net.stop()
+
+	twoOffline := false
+	threeOffline := true
+	net.filterFn = func(src int, dst int, msg []byte) []byte {
+		if twoOffline && dst == 2 { // 3 is 'offline'
+			return nil
+		}
+		if threeOffline && dst == 3 { // 3 is 'offline'
+			return nil
+		}
+		return msg
+	}
+
+	mkreq := func(n int64) *Request {
+		txTime := &gp.Timestamp{Seconds: n, Nanos: 0}
+		tx := &pb.Transaction{Type: pb.Transaction_CHAINCODE_DEPLOY, Timestamp: txTime}
+		txPacked, _ := proto.Marshal(tx)
+
+		return &Request{
+			Timestamp: &gp.Timestamp{Seconds: n, Nanos: 0},
+			Payload:   txPacked,
+			ReplicaId: uint64(generateBroadcaster(validatorCount)),
+		}
+	}
+
+	for i := int64(1); i <= 8; i++ {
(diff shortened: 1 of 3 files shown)


ghaskins:
This patch consists of either logging statements or a unit test and therefore is pretty low risk.  Unfortunately the CI test hit issue #1380 and aborted early.  I do observe that the unit test in question (consensus/obcpbft) still passes prior to running into #1380 so thats a good sign.

I'll roll this into a manual confirmation cycle, but otherwise LGTM.

Reviewed-by: Gregory Haskins gregory.haskins@gmail.com

ghaskins:
Could any of the SMEs referenced above review the test scenario and implementation and comment back in this thread?

consensus/obcpbft/pbft-core_test.go
```diff
@@ -1144,3 +1144,89 @@ func TestReplicaCrash2(t *testing.T) {
...
+
+	for i := int64(1); i <= 8; i++ {
+		net.pbftEndpoints[0].pbft.recvRequest(mkreq(i))
+	}
+	net.process() // vp0,1,2 should have a stable checkpoint for seqNo 8
+
+	// Create new pbft instances to restore from persistence
+	for id := 0; id < 2; id++ {
+		pe := net.pbftEndpoints[id]
+		os.Setenv("CORE_PBFT_GENERAL_K", "2") // TODO, this is a hacky way to inject config before initialization rather than after, address this in a future changeset
```
corecode, line 1193: I usually change the setting in a loaded config, instead of in the environment.

consensus/obcpbft/pbft-core_test.go
```diff
@@ -1144,3 +1144,89 @@ func TestReplicaCrash2(t *testing.T) {
...
+		pe.pbft = newPbftCore(uint64(id), loadConfig(), pe.sc)
+		pe.pbft.N = 4
+		pe.pbft.f = (4 - 1) / 3
+		pe.pbft.requestTimeout = 200 * time.Millisecond
+	}
+
+	threeOffline = false
+	twoOffline = true
+
+	// Because vp2 is 'offline', and vp3 is still at the genesis block, the network needs to make a view change
```
corecode, line 1203: The view change is required because we do not have eager recovery, right?  So in this case view change is not used to ditch a defective primary, but to supply a list of outstanding requests to the lagged peer.

corecode:
Good to go.

ghaskins:
Thank you @jyellick for the patch, and @corecode for the review.

I have just confirmed that a manual run of the unit-tests/behave (unsurprisingly) passes without any errors, so I will merge this.

consensus/obcpbft/pbft-core_test.go
jyellick, line 1193: I agree, the problem is that the persistence recovery happens before I can modify the environment (and would set the low watermark incorrectly), hence the hack, really think we should be passing in a 'config struct' of some sort, rather than Viper (though we could use Viper to construct it), that was these sorts of hacks wouldn't be necessary.

consensus/obcpbft/pbft-core_test.go
jyellick, line 1203: Yes, exactly, the primary is not defective, but because the peer has lagged, the network cannot make progress, so a view change is needed to catch that peer up.

consensus/obcpbft/pbft-core_test.go
corecode, line 1193: Oh I think I wasn't clear enough.  What I do is something like

```
config := loadConfig()
config.Set("general.K", "2")
pe.pbft = newPbftCore(id, config, pe.sc)
```

jyellick, line 1193: Ah, good note, I didn't realize you could do that, I'll use that from now on and update that `TODO` line in the future. Thanks!

Issue #1049, opened 2016-04-11
chaincode has a problem with stub.CreateTable


novusopt:
Hi,

currently i am looking at the development of chaincode and i tried to write a simple chaincode which creates a table in the Init callback function.

Some notes about my setup:
- 1 validating peer with security enabled & noops as consensus algorithm-> command to run the peer: `sudo CORE_SECURITY_ENABLED=true CORE_SECURITY_PRIVACY=true ./fabric peer --peer-chaincodedev`
- membership service is running
- command to deploy the chaincode: `CORE_CHAINCODE_ID_NAME=chaincodetest CORE_PEER_ADDRESS=127.0.0.1:30303 ./chaincodetest`
- command to call the init function: `./fabric chaincode deploy -n chaincodetest -u user1 -c  '{"Function":"init", "Args": ["arg1"]}'` -> user1 is logged in &  registered as a client

My Init function

```
func (t *ChainCode) Init(stub *shim.ChaincodeStub, function string, args []string) ([]byte, error) {
       err := stub.CreateTable("TempTable", []*shim.ColumnDefinition{
        &shim.ColumnDefinition{"Column1", shim.ColumnDefinition_BYTES, true},
        &shim.ColumnDefinition{"Column2", shim.ColumnDefinition_INT64, false}})
    if err != nil {
        return nil, errors.New("Failed creating TempTable")
    }
       return nil, nil
}
```

Output when i deploy the chaincode:

```
2016/04/11 14:24:20 Peer address: 127.0.0.1:30303
2016/04/11 14:24:20 os.Args returns: [./chaincodetest]
2016/04/11 14:24:20 Chaincode ID: chaincodetest
2016/04/11 14:24:20 Registering.. sending REGISTER
2016/04/11 14:24:20 []Received message REGISTERED from shim
2016/04/11 14:24:20 []Handling ChaincodeMessage of type: REGISTERED(state:created)
2016/04/11 14:24:20 Received REGISTERED, ready for invocations
```

Output after i called the Init function 

```
2016/04/11 14:31:37 Peer address: 127.0.0.1:30303
2016/04/11 14:31:37 os.Args returns: [./chaincodetest]
2016/04/11 14:31:37 Chaincode ID: chaincodetest
2016/04/11 14:31:37 Registering.. sending REGISTER
2016/04/11 14:31:37 []Received message REGISTERED from shim
2016/04/11 14:31:37 []Handling ChaincodeMessage of type: REGISTERED(state:created)
[...]

PR #1373, opened 2016-05-05
Protect against decrypting empty state when creating table

cp68:
When a table does not exist the state value returned is nil. The code then attempts to decrypt this nil value and throws an error. Noticed this behavior when attempting to create a table when privacy was enabled.


srderson:
Could you provide a test case to check for this issue?

srderson:
LGTM. Merging as this is needed to close issue #1049. Will add test case is separate PR.

cp68:
Signed-off by Chris Price (cp68@ntrs.com)

PR #1374, opened 2016-05-05
Use make to build the fabric

ghaskins:
Supersedes PR #1349.  Withholding DCO until discussion complete.

This version is based in PR #1349, except I rebased on the latest code and incorporated feedback 

---

This patch adds a top-level makefile for managing builds and dependencies

Why?

golang is just one part.

of our build system.

```
 git pull && make clean peer
```

```
 a) automatically managed

  OR

```

current process.


ghaskins:
Replaced by PR #1390

PR #1376, opened 2016-05-05
More 1290

christo4ferris:
corrected more license headers.


srderson:
LGTM. Comments and whitespace only, but also tests have run and passed.

orderer/common/server/main.go
```diff
@@ -113,58 +112,91 @@ func Main() {
...
+		if err := onboarding.ValidateBootstrapBlock(bootstrapBlock, cryptoProvider); err != nil {
+			logger.Panicf("Failed validating bootstrap block: %v", err)
+		}
+
+		// Are we bootstrapping with a genesis block? If yes, generate the system channel with a genesis block.
+		if len(lf.ChannelIDs()) == 0 && bootstrapBlock.Header.Number == 0 {
+			logger.Info("Bootstrapping the system channel")
+			initializeBootstrapChannel(bootstrapBlock, lf)
+		} else if len(lf.ChannelIDs()) > 0 {
+			logger.Info("Not bootstrapping because of existing channels")
```
C0rWin: I found it very difficult to comprehand this log message, what exactly you are trying to state here?

orderer/common/server/main.go
```diff
@@ -113,58 +112,91 @@ func Main() {
...
+			// Are we bootstrapping with a block with number >0 ?
+			// Only clusters that are equipped with a recent config block (number i.e. >0) can replicate.
+			// This will replicate all channels if the clusterBootBlock number > system-channel height (i.e. there is a gap in the ledger).
+			repInitiator = onboarding.NewReplicationInitiator(lf, clusterBootBlock, conf, clusterClientConfig.SecOpts, signer, cryptoProvider)
+			if conf.General.BootstrapMethod == "file" {
+				repInitiator.ReplicateIfNeeded(clusterBootBlock)
 			}
 		}
+	} else {
+		logger.Infof("Setting up cluster without a system channel")
```
C0rWin: Isn't that a duplication of line 138? This also makes me think that in general the flow is pretty tangled here with respect to way we handle different bootstrap procedures. Not sure whenever this PR is the right place, but ideally I think that we should have clearer logic segregation.

orderer/common/server/main.go
```diff
@@ -113,58 +112,91 @@ func Main() {
...
 
-				r = createReplicator(lf, bootstrapBlock, conf, clusterClientConfig.SecOpts, signer, cryptoProvider)
-				// Only clusters that are equipped with a recent config block can replicate.
-				if conf.General.BootstrapMethod == "file" {
-					r.replicateIfNeeded(bootstrapBlock)
-				}
+	// This happens only when the orderer is operating with a system channel.
+	if clusterBootBlock != nil {
+		consensusTypeName := consensusType(clusterBootBlock, cryptoProvider)
+		_, clusterType = clusterTypes[consensusTypeName]
```
C0rWin: a very small nit: but IMO `isClusterType` would be better here.

Issue #1368, opened 2016-05-05
'index out of range' panic using Sieve


bcbrock:
This is another failure mode of the setup described in #1366. Here is the tail of the log of the failing peer.

[logTail.txt](https://github.com/hyperledger/fabric/files/250691/logTail.txt)

corecode:
@jyellick should we just use `append()` instead of pre-allocating that array?  I'll have to look longer to figure out how this happens.

jyellick:
@corecode This is actually a bug because the Sieve thread is entering into `pbft-core.go` and modifying the state while the `pbft-core.go` thread is also doing so.  It causes the map size to change and produces this out of index bug.  This should be a straightforward fix, I'll submit a PR shortly.

Edit: Actually looks like you beat me to this one, should be fixed by your PR below.

vukolic:
what's the status? can this one be closed?

bcbrock:
@vukolic Yes, I haven't seen this in a while.

Issue #1366, opened 2016-05-05
'concurrent map read and map write' panic using Sieve


bcbrock:
In:

```
commit a56ad9b8e592119973d0290d70f537b628ddcf32
Merge: 1ca0c89 dad6b2a
Author: Sheehan Anderson <sranderson@gmail.com>
Date:   Tue May 3 13:41:19 2016 -0400

    Merge pull request #1325 from jyellick/consensus-threading-and-queues-rb

    Consensus threading and queues rebase
```

4 peers running sieve consensus. One peer crashes with a 'concurrent map read and map write'  panic. The logs are gigantic, here is the tail of the log of the crashed peer.

[logTail.txt](https://github.com/hyperledger/fabric/files/250680/logTail.txt)

To recreate:

Install https://github.ibm.com/bcbrock/busywork
cd busywork/scripts; make images
cd busywork/counters; make issue10xx

bcbrock:
Note that there is a chance, maybe 1 out of 4, that a run set up as above will pass.

corecode:
Thanks!  Fix coming soon.

vukolic:
can this one be closed?

bcbrock:
@vukolic Yes, I haven't seen this in awhile

PR #1384, opened 2016-05-06
sieve: do not bypass pbft state machine, #1366

corecode:


corecode:
Linking #1366

jyellick:
@corecode Looks good to me, just spotted this in reviewing #1368, this should fix that issue as well.

corecode:
This should also fix #1368.

srderson:
Is there a test case for this issue?

jyellick:
@srderson Both issues associated with this are races, and occur because two threads start concurrently modifying the internal pbft state.  Since it is a race, any test would most likely be non-deterministic.

srderson:
ok, LGTM

lindluni:
> Nit, but I think we should be updating the version in the vagrant box as well. While it's rarely used, it is a working, viable development environment.

Agreed, I just realized vagrant was on 1.14.2 so I missed it when I grepped for 1.14.1.

Opened #1389

PR #1288, opened 2016-04-28
Issue 1287

christo4ferris:
Fixes #1287 to correct the example license header text

[ci skip]

Signed-off-by: Christopher Ferris chrisfer@us.ibm.com

PR #1388, opened 2016-05-06
Adding non deterministic and semi nondeterministic tests

RicHernandez2:
## Description

Adding non-deterministic and semi-non-deterministic chaincode examples in order to test sieve version of pbft, negative test (all VPs return a different value) and positive test (majority return same value but some return different value.
## Motivation and Context

It adds test code for a pbft mode

Fixes #
## How Has This Been Tested?

Compiled the go code, no complaints
Ran a deploy with chaincode and got expected results (no consensus for non-deterministic, successfully sieve out non determinism in semi non-deterministic)

VP set of 4 running pbft - sieve mode

No effect, called in deploy.
## Checklist:

- [ x] If applicable, I have updated the documentation accordingly.
- [ x] If applicable, I have added tests to cover my changes.

Signed-off-by: Ric Hernandez  rhernan@us.ibm.com

### DIFF
--- a/examples/chaincode/go/README.md
+++ b/examples/chaincode/go/README.md
@@ -6,11 +6,24 @@ The purpose of this modified Chaincode_Example02 is to create non-deterministic 
 ###How does it work?###
 The Non-deterministic version of the chaincode takes the time in nanoseconds on the VP and it uses it to seed the Rand function in order to get a non-deterministic number, for although all VPs will be executing the same code, it is a given that they will do so at a different nanosecond, and thus have a different seed for the Rand function, producing a different number for each VP.
 
+    rand.Seed(time.Now().UnixNano())
+    B := rand.Int63()
+    fmt.Printf("Seed Number = ", B)
+
 ###How can I tell what number ends up getting used?###
 The seed number will appear in the VP Log, search for "Seed Number" if you want to retrieve the exact seed number used on a particular VP.
 
 ###What is Semi-nondeterministic chaincode?###
 The idea behind the sieve version of pbft was to sieve out nondeterminism, for that to happen the majority of VPs would need to have the same value and the minority value will be sieved out.
+
+    rand.Seed(time.Now().UnixNano())
+    B := rand.Int63()
+    if (B % 2) == 0 {
+      B = 2000
+    } else {
+      B = 2001
+    }
+    fmt.Println("Seed Number =", B)
 
 This is accomplished by performing a modulo operation on the nanosecond value previously used as a seed and generating one of two values depending on whether the number is odd or even on the given VP.  This doesn't guarantee the right mix, it may take a few tries to reach the desired effect, but here too you can search the log for "Seed Number" to find which of two values was used for a particular machine.
 
--- a/examples/chaincode/go/README.md
+++ b/examples/chaincode/go/README.md
@@ -1,0 +1,57 @@
+##Read Me for Non-deterministic Chaincode ##
+
+###Purpose###
+The purpose of this modified Chaincode_Example02 is to create non-deterministic results for the purpose of testing.  That is to say results that would mimic and even **exaggerate** some of the real world examples that would cause a non-deterministic result.  Things like, lag between VPs, garbled text, malicious fault injection, etc.
+
+###How does it work?###
+The Non-deterministic version of the chaincode takes the time in nanoseconds on the VP and it uses it to seed the Rand function in order to get a non-deterministic number, for although all VPs will be executing the same code, it is a given that they will do so at a different nanosecond, and thus have a different seed for the Rand function, producing a different number for each VP.
+
+###How can I tell what number ends up getting used?###
+The seed number will appear in the VP Log, search for "Seed Number" if you want to retrieve the exact seed number used on a particular VP.
+
+###What is Semi-nondeterministic chaincode?###
+The idea behind the sieve version of pbft was to sieve out nondeterminism, for that to happen the majority of VPs would need to have the same value and the minority value will be sieved out.
+
+This is accomplished by performing a modulo operation on the nanosecond value previously used as a seed and generating one of two values depending on whether the number is odd or even on the given VP.  This doesn't guarantee the right mix, it may take a few tries to reach the desired effect, but here too you can search the log for "Seed Number" to find which of two values was used for a particular machine.
+
+##Sample Deploy for ndt (non deterministic code):##
+
+```
+{
+  "jsonrpc": "2.0",
+  "method": "deploy",
+  "params": {
+    "type": 1,
+    "chaincodeID":{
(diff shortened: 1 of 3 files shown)


ghaskins:
This is a great idea, thank you.  Is there any way you could include a behave test that would drive these?  I think that would offer the most utility.

srderson:
Agree with @ghaskins, this PR seems to be missing the associated behave test. Also, please move the chaincodes to the `/bddtests/chaincode/go/`.

Did you run gofmt, golint, and go vet on these files? They seem to be missing some formatting, such as comments on public functions that start with the function name.

Finally, please add a comment in each chaincode describing its functionality or a readme associated with each chaincode.

Thanks

jeffgarratt:
@ghaskins @srderson @RicHernandez2 Could you also make explicit the distinction between this test mechanism and that employed through the pbft byzantine configuration test.   https://github.com/hyperledger/fabric/blob/master/bddtests/peer_basic.feature#L710

ghaskins:
@jeffgarratt I certainly could be missing some nuance here, but I think the difference is that the example you cite injects byzantine behavior through the framework level with an otherwise deterministic chaincode (example02 in this case).  This example would be geared towards leaving the framework in a more "production" type scenario and driving non-determinism at the chaincode level.  Assuming I am not misinterpreting anything, I think both methodologies have some merit.  If I had to choose between one or the other, I would favor this approach as I think it captures the risk/attack-surface most precisely.

Related to this, it might be interesting to drive this example through all consensus plugins to get a feel for what happens in the face of non-determinism.

muralisrini:
@ghaskins to add to the above comment, this is a step towards "chaincode practices" (good/bad).

One thing we could do is optionally take the seed as a parameter to Init (say if seedArg > 0). This would help make this non-deterministic process repeatable.

christo4ferris:
@RicHernandez2 this PR has the incorrect license header text. Please use the text [here](https://github.com/hyperledger/fabric/blob/master/docs/dev-setup/headers.txt) substituting IBM 2016 for the COPYRIGHT OWNER

ghaskins:
@RicHernandez2 Any thoughts on the request for behave coverage?

RicHernandez2:
Would love to oblige, I haven't touched behave tests yet so I'm eager to learn that new skill, but unfortunately Friday will be my last day and I have more work than I have time left in handing my areas of responsibility over to others.  I can maybe come in on the weekend if I'm allowed to get the behave tests done.  But mostly I just wanted to have the chaincode examples in the master so they get built into the continuous integration images that Ramesh's Travis CI generates, saves boatloads of time!

ghaskins:
@srderson In light of @RicHernandez2 imminent departure and lack of time to complete the additional work, I would vote to merge this as is and perhaps file a help-wanted issue to develop the tests.  It looks like I decent addition to the portfolio and I would hate to see it get lost completely.  Thoughts?

tuand27613:
+1 on the as-is merge.

Would like to ask @RicHernandez2  to add a README describing the chaincode logic and what to expect when running the example

binhn:
@RicHernandez2 @christo4ferris where did the incorrect header come from?  I'd like to fix that so that others don't run into the same problem.  I thought Chris had changed all file headers to the correct one.

@srderson @ghaskins agreed to merge, but keep in mind that SIEVE is a prototype and failures there should be just FYI in the test suite. I would recommend that we don't include SIEVE related tests in the mandatory set.

christo4ferris:
@binhn I fixed all existing headers, these are new files. I also fixed the sample header a couple weeks back #1288

muralisrini:
I think this good work will be more useful if we add two things to make tests repeatable 
- take a seed as argument. Generate one if not provided (default)
- print the seed used

Without the above it will be hard to recreate problems seen when running these random tests. Shouldn't be hard to do ?

srderson:
Could we just wait for someone to submit a PR on top of this one? Then we can merge once all issues have been addressed? It's not a prerequisite for this to be in the master branch for someone to fix it up.

corecode:
It may be better to introduce the non-determinism in the `invoke` function, rather than in `init`; `invoke` can be repeated more easily (especially for the binary non-determinism code).  Further, I think to test and demo non-determinism, the code can be significantly reduced.  We should also add a BIG WARNING MESSAGE THAT YOU SHOULD NOT WRITE SUCH CODE EVER.

RicHernandez2:
##Read Me for Non-deterministic Chaincode ##

###Purpose###
The purpose of this modified Chaincode_Example02 is to create non-deterministic results for the purpose of testing.  That is to say results that would mimic and even **exaggerate** some of the real world examples that would cause a non-deterministic result.  Things like, lag between VPs, garbled text, malicious fault injection, etc.

###How does it work?###
The Non-deterministic version of the chaincode takes the time in nanoseconds on the VP and it uses it to seed the Rand function in order to get a non-deterministic number, for although all VPs will be executing the same code, it is a given that they will do so at a different nanosecond, and thus have a different seed for the Rand function, producing a different number for each VP.

###How can I tell what number ends up getting used?###
The seed number will appear in the VP Log, search for "Seed Number" if you want to retrieve the exact seed number used on a particular VP.

###What is Semi-nondeterministic chaincode?###
The idea behind the sieve version of pbft was to sieve out nondeterminism, for that to happen the majority of VPs would need to have the same value and the minority value will be sieved out.

This is accomplished by performing a modulo operation on the nanosecond value previously used as a seed and generating one of two values depending on whether the number is odd or even on the given VP.  This doesn't guarantee the right mix, it may take a few tries to reach the desired effect, but here too you can search the log for "Seed Number" to find which of two values was used for a particular machine.

##Sample Deploy for ndt (non deterministic code):##

```
{
  "jsonrpc": "2.0",
  "method": "deploy",
  "params": {
    "type": 1,
    "chaincodeID":{
        "path":"github.com/hyperledger/fabric/examples/chaincode/go/ndt"
    },
    "ctorMsg": {
        "function":"init",
        "args":["a", "1000", "b", "2000"]
    },
    "secureContext": "lukas"
  },
  "id": 1
}
```

##Sample Deploy for sndt (Semi-nondeterministic):##

```
{
  "jsonrpc": "2.0",
  "method": "deploy",
  "params": {
    "type": 1,
    "chaincodeID":{
        "path":"github.com/hyperledger/fabric/examples/chaincode/go/sndt"
    },
    "ctorMsg": {
        "function":"init",
        "args":["a", "1000", "b", "2000"]
    },
    "secureContext": "lukas"
  },
  "id": 1
}
```

RicHernandez2:
Ok, readme.md completed, and I've added code to print seeds used to log, please complete merge.

Thanks!

---- Ric

ghaskins:
There seems to be disagreement on whether this should merge in its current state.  @srderson @binhn what do you want to do?

christo4ferris:
@hyperledger/fabric-maintainers I'm going to close this as there isn't much we can do with it in its current state.

PR #1390, opened 2016-05-06
Use make to build fabric

ghaskins:
## Description

```
This patch adds a top-level makefile for managing builds and dependencies
```
## Motivation and Context

```

golang is just one part.

of our build system.

     git pull && make clean peer

     a) automatically managed

      OR

current process.
```
## How Has This Been Tested?

Booted a clean vagrant environment, and using the new facilities:

make unit-test
make behave

All passed without any errors.
## Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,89 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+# -------------------------------------------------------------
+# This makefile defines the following targets
+#
+#   - peer (default) - builds the fabric ./peer/peer binary
+#   - membersrvc - builds the ./membersrvc/membersrvc binary
+#   - unit-test - runs the go-test based unit tests
+#   - behave - runs the behave test
+#   - images - ensures pre-requisites are availble for running behave manually
+#   - peer-image - ensures the peer-image is available (for behave, etc)
+#   - ca-image - ensures the ca-image is available (for behave, etc)
+#   - protos - generate all protobuf artifacts based on .proto files
+#   - clean - cleans the build area
+#   - dist-clean - superset of 'clean' that also removes persistent state
+
+
+PKGNAME = github.com/hyperledger/fabric
+CGO_LDFLAGS = -lrocksdb -lstdc++ -lm -lz -lbz2 -lsnappy
+
+EXECUTABLES = go docker
+K := $(foreach exec,$(EXECUTABLES),\
+	$(if $(shell which $(exec)),some string,$(error "No $(exec) in PATH: Check dependencies")))
+
+all: peer
+
+.PHONY: peer
+peer: base-image
+	cd peer; CGO_CFLAGS=" "	CGO_LDFLAGS="$(CGO_LDFLAGS)" go build
+
+.PHONY: membersrvc
+membersrvc:
+	cd membersrvc; CGO_CFLAGS=" " CGO_LDFLAGS="$(CGO_LDFLAGS)" go build
+
+unit-test: base-image
+	@echo "Running unit-tests"
+	@go test -timeout=20m $(shell go list $(PKGNAME)/... | grep -v /vendor/ | grep -v /examples/)
+	@touch .peerimage-dummy
+	@touch .caimage-dummy
+
+base-image: .baseimage-dummy
+peer-image: .peerimage-dummy
+ca-image: .caimage-dummy
+images: peer-image ca-image
+
+behave: images
+	@echo "Running behave tests"
+	@cd bddtests; behave
+
+.peerimage-dummy: .baseimage-dummy
+	go test $(PKGNAME)/core/container -run=BuildImage_Peer
+	@touch $@
+
+.caimage-dummy: .baseimage-dummy
+	go test $(PKGNAME)/core/container -run=BuildImage_Obcca
+	@touch $@
+
+.baseimage-dummy:
+	@echo "Building docker base-image"
+	@./scripts/provision/docker.sh 0.0.9
+	@touch $@
+
+protos:
+	./devenv/compile_protos.sh
+
+.PHONY: clean
+clean:
+	-@rm .*image-dummy ||:
+	-@rm -f ./peer/peer ||:
+	-@rm -f ./membersrvc/membersrvc ||:
+
+.PHONY: dist-clean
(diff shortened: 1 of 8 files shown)


ghaskins:
Supersedes PR #1374

ghaskins:
@srderson this is ready for a merge consideration.  Due to the proposed change to the fundamental interaction with the project, should probably have a consensus led by @binhn

ghaskins:
Hmm, this build failure is legit (deficient setup for travis).  I know whats wrong and will fix.

rameshthoomu:
@ghaskins : Could you please update `behave` command with `behave -D logs=Y` at `@cd bddtests; behave`.

ghaskins:
@rameshthoomu How about this idea?

```
commit 05a8090253f9344a396c27d02304bfc5947fc1d2
Author: Gregory Haskins <gregory.haskins@gmail.com>
Date:   Fri May 6 19:20:09 2016 -0400

    Allow behave to be parameterized

    Signed-off-by: Greg Haskins <gregory.haskins@gmail.com>

diff --git a/.travis.yml b/.travis.yml
index ce99b0b..33ab6c9 100755
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,14 +39,14 @@ before_script:
  - cd $HOME/gopath/src/github.com/$USER_NAME/$REPO_NAME/scripts && chmod +x foldercopy.sh && ./foldercopy.sh $TR_PULL_REQUEST $USER_NAME
  - sudo rm -rf /var/hyperledger/ && sudo mkdir /var/hyperledger/ && sudo chown $USER:$USER /var/hyperledger
  - cd /$HOME/gopath/src/github.com/hyperledger/fabric
- - make behave-deps unit-test
+ - make unit-test

 script:

  - echo "Executing Behave test scripts"
- - cd $HOME/gopath/src/github.com/hyperledger/fabric/bddtests
+ - cd $HOME/gopath/src/github.com/hyperledger/fabric
  - sed -i -e 's/172.17.0.1:4243\b/'"$ip:$port"'/g' $HOME/gopath/src/github.com/hyperledger/fabric/bddtests/compose-defaults.yml
- - behave -D logs=Y -o testsummary.log
+ - make behave BEHAVE_OPTS="-D logs=Y -o testsummary.log"

 after_failure:

diff --git a/Makefile b/Makefile
index dc07572..57b1452 100644
--- a/Makefile
+++ b/Makefile
@@ -64,7 +64,7 @@ images: peer-image ca-image
 behave-deps: images peer
 behave: behave-deps
        @echo "Running behave tests"
-       @cd bddtests; behave
+       @cd bddtests; behave $(BEHAVE_OPTS)

 .peerimage-dummy: .baseimage-dummy
        go test $(PKGNAME)/core/container -run=BuildImage_Peer
```

rameshthoomu:
@ghaskins I am good with this.

jyellick:
Looks good to me, this would certainly save me time and be a nice productivity boost.

srderson:
I wasn't sure about moving away from the Go commands, but this is actually quite slick given the increasing complexity of some of the commands. Nice job @ghaskins!

A couple of comments

In install.md, the doc still instructs developers to use `go test -timeout=20m $(go list github.com/hyperledger/fabric/... | grep -v /vendor/ | grep -v /examples/)` instead of `make unit-test`.

In install.md, `$GOPATH/src/github.com/hyperledger/fabric/devenv/compile_protos.sh` should like be changed to `make protos`. Also, I tried running `make protos` and it didn't seem to work. It just responded with `make: protos' is up to date.`

A bunch of -dummy images were created. What are these? They should likely be put in `.gitignore`

@binhn @jeffgarratt you may both wish to review

Makefile
```diff
@@ -0,0 +1,93 @@
...
+
+
+PKGNAME = github.com/hyperledger/fabric
+CGO_LDFLAGS = -lrocksdb -lstdc++ -lm -lz -lbz2 -lsnappy
+
+EXECUTABLES = go docker
+K := $(foreach exec,$(EXECUTABLES),\
+	$(if $(shell which $(exec)),some string,$(error "No $(exec) in PATH: Check dependencies")))
+
+all: peer
```
christo4ferris: I would think that all should build peer, unit-test and behave - no? Should we also run the go tools as part of the make, as well?

Makefile
```diff
@@ -0,0 +1,93 @@
...
+.caimage-dummy: .baseimage-dummy
+	go test $(PKGNAME)/core/container -run=BuildImage_Obcca
+	@touch $@
+
+.baseimage-dummy:
+	@echo "Building docker base-image"
+	@./scripts/provision/docker.sh 0.0.9
+	@touch $@
+
+protos:
```
christo4ferris, line 92: seems like we should have a dependency on the .proto files to trigger this whenever building the peer

christo4ferris:
me likey

Makefile
ghaskins: I hadn't thought of making the default build all + checks, but this makes a lot of sense.  The user can always use "make peer" if they want only the peer specifically.  I will add this in.

+1 on the go tools, I was going to propose we add the checks to this substrate: we can then just have travis invoke as appropriate as well as provide our users with a way to run the tooling the same way CI will.

Makefile
ghaskins, line 92: Im torn on whether the model should be

a) do not check in .pb files and run the protoc compiler as part of your build (this is how I run most projects).

b) check in .pb files and only run the protoc compiler on demand

I had never worked on a project until this one that used (b), but I did see the value in reducing the pre-requisite burden on the average user (no need to have protoc even installed in the vast majority of cases).  I can be convinced to go in either direction.  However, if we go route (a) I suggest we just abandon the script and write formal make rules that describe how to compile a .proto -> .pb.go.  I also suggest we tackle this as a discrete PR, for simplicity/clarity ;)

ghaskins:
@srderson Thanks for the review.

Regarding doc disparity, that somehow slipped through, will fix.

Regarding .dummy files: Make is designed to evaluate prerequisites based on comparing files (e.g. is "foo.o" older than "foo.cpp"?).  For targets that do not result in outputting a standard file (ala docker-build), you can integrate the subsystem by creating fake outputs by touching otherwise worthless files.  In this case, I was using .dummy to represent docker images.  A few comments about that
- If I can find a reliable way for the makefile to assess the artifact status directly (e.g. via docker-images) then we can probably get rid of the .dummy hacks, TBD
- If not, those files should certainly be in .gitignore.

I propose that we explore the former in a follow up effort.  Will fix the .gitignore problem as part of this PR

Makefile
christo4ferris, line 92: agree to an independent PR for the .proto -> .pb.go rule.

Makefile
christo4ferris: +1

ghaskins:
@srderson oh, and I will investigate whats going on with "make protos".  Sorry, missed that comment the first time.

srderson:
WRT to the dummy files, fixing up the .gitignore for this PR sounds like a good approach. Thanks for the explanation.

ghaskins:
@srderson @christo4ferris I pushed more patches that implement the requested changes.  Please give the latest a review at your earliest convenience.

ghaskins:
@srderson I just noticed I missed the compile_protos doc update..  Fixing now.

I should also mention, the latest iteration introduces two new make targets: "linter" and "checks".  "linter" is currently a no-op, waiting for a future PR that drives whatever we want as far as automated checking.  And "checks" is a composite target that runs unit-test+behave+linter.

srderson:
Yes, a "checks" is a great idea. I can investigate this as I believe we already install all tools in the vagrant image.

LGTM

PR #1391, opened 2016-05-06
peer: Fix possible nil dereference in PeerImpl Unicast

dubek:
## Description

In `PeerImpl`'s `Unicast` function, if the `receiveHandle` is not in the `handlerMap`, we'll try to call `SendMessage` on a nil, which will panic.  I'm not sure there's a way to actually trigger this bug but it seems cleaner this way.

The new extracted `getMessageHandler` function also takes care of unlocking with `defer`, which is safer than the previous code (against possible future changes).
## Motivation and Context

Possible panic if a bad `receiveHandle` is passed to the exported `Unicast` function.
## How Has This Been Tested?

Passes all unit tests on vagrant.
## Checklist:

- [ ] If applicable, I have updated the documentation accordingly. <- N/A

Signed-off-by: Dov Murik <dmurik@us.ibm.com>

### DIFF
--- a/core/peer/peer.go
+++ b/core/peer/peer.go
@@ -442,13 +442,23 @@ func (p *PeerImpl) Broadcast(msg *pb.Message, typ pb.PeerEndpoint_Type) []error 
 	return returnedErrors
 }
 
+func (p *PeerImpl) getMessageHandler(receiverHandle *pb.PeerID) (MessageHandler, error) {
+	p.handlerMap.Lock()
+	defer p.handlerMap.Unlock()
+	msgHandler, ok := p.handlerMap.m[*receiverHandle]
+	if !ok {
+		return nil, fmt.Errorf("Message handler not found for receiver %s", receiverHandle.Name)
+	}
+	return msgHandler, nil
+}
+
 // Unicast sends a message to a specific peer.
 func (p *PeerImpl) Unicast(msg *pb.Message, receiverHandle *pb.PeerID) error {
-	p.handlerMap.Lock()
-	msgHandler := p.handlerMap.m[*receiverHandle]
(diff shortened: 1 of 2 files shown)


dubek:
As a side note, while reviewing the `peer.go` file (and others), I noticed that locks are always acquired with `p.handlerMap.Lock()`, even when read-only access is need and `RLock()` should be enough. It might cause a performance issue if many readers are "serialized" for these read-only accesses even though they could access the handlerMap in parallel.

srderson:
Agreed that this looks like a cleaner approach. Thanks! Could you add a simple unit test to `peer_test.go` to check this? Something like

```
func TestBadReceiverHandleUnicast(t *testing.T) {
    testHandlerMap := handlerMap{m: make(map[pb.PeerID]MessageHandler)}
    peerImpl := PeerImpl{handlerMap: &testHandlerMap}
    err := peerImpl.Unicast(nil, &pb.PeerID{})
    if err == nil {
        t.Error("Expected error with bad receiver handle, but there was none")
    }
}
```

I agree that an `RLock()` seems like a better option. I'm not if there was a reason it was not used. @jeffgarratt may know.

dubek:
Thank you @srderson for the suggestions. I pushed a fix with the suggested unit test.

srderson:
LGTM. Thanks

dubek:
Thanks @srderson . Should I tackle the RLock/RUnlock in a separate PR?

srderson:
Yes, we're trying to keep PRs as small as possible so they're easily reviewable by the community.

jeffgarratt:
@dubek @srderson I see no reason not to use the RLock.  Seem to remember an issue with the unlock but may be mistaken.  Give it a go.

Issue #588, opened 2016-02-02
Transaction status tracking


manish-sethi:
A proper mechanism for application to track the status of a transaction is missing today. I was asked about this by one of my colleague at IRL (who is building a demo).

Initially, he planned to poll the rest api /transaction/{UUID}. However, it may have following issues
- The transactions results are not included in this API. So, the user will not be able to find out whether the transaction failed or succeeded
- Even if we extend the above API for including the tx result, user may run into another issue because this API pulls data from ledger and some tx may never appear on blockchain (e.g., a batch may be rolled back because the state hash did not match across peer). 

@srderson and I discussed about this and felt that using event mechanism may be right for this - We may need to emit an additional event at the end of batch execution. May be a single event can contain a map of guid -> status (for all the transactions included in the batch) where status can be one of the {committed, failed, rolledback}.

Finally, ideally these events should be stored for answering queries on historical data (at least for some time in past). For instance, a user may submit a transaction and may come back to check the status the next day. I am not sure whether this (storing the events and answering queries) should be within the scope of obc or left to apps.

corecode:
How would this address potentially malicious NVPs or VPs that drop the transaction?

Does the failed status mean "commited (without state change), but error during chaincode execution"?

What part of the system decides to send an event on rollback?  When we batch transactions, how does that affect transactions that have been batched together?  If we want to single out non-deterministic transactions, and retry execution of the block without the one non-deterministic transaction, will the re-executed transactions get duplicate events?

manish-sethi:
@corecode please see the comments below.

As per my understanding, in our current model - the clients connected to a malicious NVPs or VPs would be affected in many scenarios. Let me know if I am wrong here.

Yes, that's what I had in mind for the 'failed' status

No, in my opinion the multiple events for a single tx should be avoided - otherwise client is never sure whether he has received the final event (event at each stage could be useful for system monitoring point of view but not for the user). Keeping this in mind, I think that the consensus layer is the right place for generating this event.

kletkeman:
[...]

PR #1392, opened 2016-05-07
Transaction results event#588

tuand27613:
## Description

Set transaction results and add to nonHashData when committing a block
## Motivation and Context

Fixes first part of #588 
## How Has This Been Tested?

unit tests run.
behave tests run.
## Checklist:

- [ ] If applicable, I have updated the documentation accordingly.


srderson:
README should be an md file

block-listener.go needs gofmt, go vet, golint run. Public functions and structs require comments and they are missing in this file.

tuand27613:
golint/gofmt'ed block-listener.go

README now README.md

srderson:
Thanks @tuand27613! LGTM

srderson:
Also thanks @muralisrini for the readme and sample app! These are great for people learning to use the event mechanism.

common/viperutil/config_util.go
```diff
@@ -24,14 +28,230 @@ import (
...
 
-type viperGetter func(key string) interface{}
+// prototype for get environment variable
+type envGetter func(key string) interface{}
 
-func getKeysRecursively(base string, getKey viperGetter, nodeKeys map[string]interface{}, oType reflect.Type) map[string]interface{} {
+// default config file path
+const OfficialPath = "/etc/hyperledger/fabric"
+
+var SupportedExts []string = []string{"yaml", "yml"}
```
sykesm, line 42: The slice can be inlined in the range that's happening below. No need to have the exported field for it.

common/viperutil/config_util.go
```diff
@@ -24,14 +28,230 @@ import (
...
 
 var logger = flogging.MustGetLogger("viperutil")
 
-type viperGetter func(key string) interface{}
+// prototype for get environment variable
+type envGetter func(key string) interface{}
 
-func getKeysRecursively(base string, getKey viperGetter, nodeKeys map[string]interface{}, oType reflect.Type) map[string]interface{} {
+// default config file path
+const OfficialPath = "/etc/hyperledger/fabric"
```
sykesm, line 40: This was copied from fabric/core/config and I'm not really sure why.

common/viperutil/config_util.go
```diff
@@ -24,14 +28,230 @@ import (
...
+	// parsed config
+	config map[string]interface{}
+}
+
+// Creates a ConfigParser instance
+func New() *ConfigParser {
+	c := new(ConfigParser)
+	c.config = make(map[string]interface{})
+	return c
+}
```
sykesm, line 70: More idiomatically:
```
return &ConfigParser{
    config: map[string]interface{}{},
}
```

common/viperutil/config_util.go
```diff
@@ -24,14 +28,230 @@ import (
...
+	// configuration file to process
+	configPaths []string
+	configName  string
+	configFile  string
+	configType  string
+
+	// env variable prefix
+	envPrefix           string
+	automaticEnvApplied bool
+	envKeyReplacer      *strings.Replacer
```
sykesm, line 59: We don't need to preserve the generalizations of viper. We have one pattern and can simply inline a call to `strings.ReplaceAll` where we need to map `.` to `_`.

common/viperutil/config_util.go
```diff
@@ -24,14 +28,230 @@ import (
...
+
+	// configuration file to process
+	configPaths []string
+	configName  string
+	configFile  string
+	configType  string
+
+	// env variable prefix
+	envPrefix           string
+	automaticEnvApplied bool
```
sykesm, line 58: We don't need `automaticEnvApplied` or the associated `AutomaticEnv` method since we don't need the generalizations of viper.

common/viperutil/config_util.go
```diff
@@ -24,14 +28,230 @@ import (
...
+func New() *ConfigParser {
+	c := new(ConfigParser)
+	c.config = make(map[string]interface{})
+	return c
+}
+
+// Keeps a list of path to search the relevant
+// config file. Multiple paths can be provided.
+func (c *ConfigParser) AddConfigPath(in string) {
+	if in != "" {
```
sykesm, line 75: No need to wrap this with a zero value check. And, if we did need one, idiomatically we would return immediately on the condition and leave the real logic as the mainline path.

common/viperutil/config_util.go
```diff
@@ -24,14 +28,230 @@ import (
...
+				return
+			}
+		}
+		c.configPaths = append(c.configPaths, in)
+	}
+}
+
+// The config file name. The extension is not included
+func (c *ConfigParser) SetConfigName(in string) {
+	if in != "" {
```
sykesm, line 88: Sane here. No need to wrap with a zero value check.

common/viperutil/config_util.go
```diff
@@ -24,14 +28,230 @@ import (
...
+		c.configName = in
+	}
+}
+
+// The extension of the config file
+// Currently only "yaml" is supported.
+func (c *ConfigParser) SetConfigType(in string) {
```
sykesm, line 95: No need for this method at all. We only support yaml.

common/viperutil/config_util.go
```diff
@@ -24,14 +28,230 @@ import (
...
+	if in != "" {
+		c.configType = in
+	}
+}
+
+// Method to initialize default paths.
+// The defaults paths are CWD and OfficialPath
+// The env variable FABRIC_CFG_PATH can be used to override default
+func (c *ConfigParser) InitConfigPath(configName string) error {
```
sykesm, line 104: I believe this should be a function instead of a method that simply returns the paths to search. There's no need to do any validation here since searching does the same thing.

Probably something as simple as

```go
func ConfigPaths() []string {
    var paths []string
    if p := os.Getenv("FABRIC_CFG_PATH"); p != "" {
        paths = append(paths, p)
    }
    return append(paths, ".", "/etc/hyperledger/fabric")
}
```

common/viperutil/config_util.go
```diff
@@ -24,14 +28,230 @@ import (
...
+// Creates a ConfigParser instance
+func New() *ConfigParser {
+	c := new(ConfigParser)
+	c.config = make(map[string]interface{})
+	return c
+}
+
+// Keeps a list of path to search the relevant
+// config file. Multiple paths can be provided.
+func (c *ConfigParser) AddConfigPath(in string) {
```
sykesm, line 74: The method can be made plural and the arg list variadic. That would allow us to add multiple paths in one go.

If combined with the suggestions around removing `InitConfigPath`, we can set all paths in one go.

common/viperutil/config_util.go
```diff
@@ -24,14 +28,230 @@ import (
...
+	// Now set the configuration file.
+	c.SetConfigName(configName)
+
+	return nil
+}
+
+// Explicitly setting the full config file name.
+// In this case, configPaths search shall not be done
+func (c *ConfigParser) SetConfigFile(in string) {
+	if in != "" {
```
sykesm, line 141: No need for the zero value check

common/viperutil/config_util.go
```diff
@@ -24,14 +28,230 @@ import (
...
+}
+
+// Return the used configFile to fill config store
+func (c *ConfigParser) ConfigFileUsed() string { return c.configFile }
+
+// Set the env variable prefix
+// If "peer" is set here, all environment variable
+// searches shall be prefixed with "PEER_"
+func (c *ConfigParser) SetEnvPrefix(in string) {
+	if in != "" {
```
sykesm, line 153: no need for the zero value check

common/viperutil/config_util.go
```diff
@@ -24,14 +28,230 @@ import (
...
+		c.envPrefix = in
+	}
+}
+
+// Search for the existence of filename for all supported extensions
+func (c *ConfigParser) searchInPath(in string) (filename string) {
+	logger.Debugf("Searching for config in ", in)
+	for _, ext := range SupportedExts {
```
sykesm, line 161: SupportedExts can be an inline slice of `yaml` and `yml`.

common/viperutil/config_util.go
```diff
@@ -24,14 +28,230 @@ import (
...
+
+	cf := c.getConfigFile()
+	logger.Debugf("Attempting to read from config file : %s", cf)
+	file, err := ioutil.ReadFile(cf)
+	if err != nil {
+		logger.Errorf("Unable to read from config file : %s", cf)
+		return err
+	}
+
+	//c.config = make(map[string]interface{})
```
sykesm, line 208: remove commented out code, please.

common/viperutil/config_util.go
```diff
@@ -24,14 +28,230 @@ import (
...
+
+	c.configFile = c.findConfigFile()
+	return c.configFile
+}
+
+func (c *ConfigParser) ReadInConfig() error {
+
+	cf := c.getConfigFile()
+	logger.Debugf("Attempting to read from config file : %s", cf)
+	file, err := ioutil.ReadFile(cf)
```
sykesm, line 202: Please pass the file as a reader to `ReadConfig` instead of reading it fully into a slice and then creating a reader from the buffer. Not only does it make simpler, it makes it more efficient and removes the duplicated logic.

common/viperutil/config_util.go
```diff
@@ -24,14 +28,230 @@ import (
...
+	return yaml.Unmarshal(buf.Bytes(), c.config)
+}
+
+// Identify the configFile from the provided configPaths
+// and read and unmarshall and store in config.
+func (c *ConfigParser) ReadConfig(in io.Reader) error {
+	//c.config = make(map[string]interface{})
+	buf := new(bytes.Buffer)
+	buf.ReadFrom(in)
+	return yaml.Unmarshal(buf.Bytes(), c.config)
```
sykesm, line 222: Please use the decoder instead of reading the entire contents of the reader into a buffer and then passing it to `Unmarshal`. By creating a decoder instance, it can stream the data.

common/viperutil/config_util.go
```diff
@@ -24,14 +28,230 @@ import (
...
+// Used in searching env. variables.
+func (c *ConfigParser) SetEnvKeyReplacer(r *strings.Replacer) {
+	c.envKeyReplacer = r
+}
+
+// Set the environment variable search boolean
+func (c *ConfigParser) AutomaticEnv() {
+	c.automaticEnvApplied = true
+}
+
```
sykesm, line 235: No need for these. We don't need the generalizations of viper.

common/viperutil/config_util.go
```diff
@@ -24,14 +28,230 @@ import (
...
+// Get value for the key by searching
+// environment variables
+func (c *ConfigParser) GetFromEnv(key string) interface{} {
```
sykesm, line 238: There's no need to export this method as it's only called within this file. Please unexport it.

common/viperutil/config_util.go
```diff
@@ -24,14 +28,230 @@ import (
...
+	envKey := key
+	if c.envPrefix != "" {
+		envKey = strings.ToUpper(c.envPrefix + "_" + envKey)
+	}
+	if c.envKeyReplacer != nil {
+		envKey = c.envKeyReplacer.Replace(envKey)
+	}
```
sykesm, line 245: This is where we can inline the call to `strings.ReplacAll`.

common/viperutil/config_util.go
```diff
@@ -325,9 +553,8 @@ func EnhancedExactUnmarshal(v *viper.Viper, output interface{}) error {
 		return errors.Errorf("supplied output argument must be a pointer to a struct, but it is pointer to something else")
 	}
 
-	baseKeys := v.AllSettings()
-
-	getterWithClass := func(key string) interface{} { return v.Get(key) } // hide receiver
+	baseKeys := c.GetParsedConfig()
+	getterWithClass := func(key string) interface{} { return c.GetFromEnv(key) }
```
sykesm, line 557: No need for this function. Just pass `c.getFromEnv` (once unexported) directly to `getKeysRecursively`.

common/viperutil/config_util.go
```diff
@@ -24,14 +28,166 @@ import (
...
+		envKey = strings.ToUpper(c.envPrefix + "_" + envKey)
+	}
+	envKey = strings.ReplaceAll(envKey, ".", "_")
+	return os.Getenv(envKey)
+}
+
+// getParsedConfig returns the parsed config store
+func (c *ConfigParser) getParsedConfig() map[string]interface{} {
+	return c.config
+}
```
sykesm: This is unexported and only returns a field; we don't need a method to access one internal bit of state with no decorations.

common/viperutil/config_util.go
sykesm: This doesn't quite address the comment. The code is still fully reading the file instead of opening it and passing the file as the reader to ReadConfig.

```go
...
    file, err := os.Open(cf)
    if err != nil {
        return err
    }
    defer file.Close()

    return c.ReadConfig(file)
```

common/viperutil/config_util.go
```diff
@@ -24,14 +28,166 @@ import (
...
+
+	var paths []string
+	if len(c.configPaths) > 0 {
+		paths = c.configPaths
+	} else {
+		// pick the default paths
+		paths = ConfigPaths()
+	}
+	for _, cp := range paths {
+		file := c.searchInPath(cp)
```
sykesm, line 179: We like flat conditions when possible. There's no need for if/else.

```go
paths := c.configPaths
if len(paths) == 0 {
    paths = ConfigPaths()
}

for _, p := range Paths {
...
}
```

common/viperutil/config_util.go
```diff
@@ -24,14 +28,166 @@ import (
...
+			logger.Debugf("Found: ", fullPath)
+			return fullPath
+		}
+	}
+	return ""
+}
+
+// Search for the configName in all configPaths
+func (c *ConfigParser) findConfigFile() string {
+
```
sykesm, line 175: nit: we can drop the blank lines after method declarations. There are a few in here.

common/viperutil/config_util.go
```diff
@@ -24,14 +28,166 @@ import (
...
 	"github.com/pkg/errors"
-	"github.com/spf13/viper"
+	"gopkg.in/yaml.v2"
 )
 
 var logger = flogging.MustGetLogger("viperutil")
 
-type viperGetter func(key string) interface{}
+// prototype for get environment variable
+type envGetter func(key string) interface{}
```
sykesm, line 37: This is a detail of getKeys and very far away from it. Please move it close to where it's used.

common/viperutil/config_util.go
```diff
@@ -24,14 +28,166 @@ import (
...
 	"github.com/mitchellh/mapstructure"
 	"github.com/pkg/errors"
-	"github.com/spf13/viper"
+	"gopkg.in/yaml.v2"
 )
 
 var logger = flogging.MustGetLogger("viperutil")
 
-type viperGetter func(key string) interface{}
+// prototype for get environment variable
```
sykesm, line 36: Common godoc nit. The documentation should be complete sentences when possible. They won't hold up merging but if they're being updated, it's good to ensure they are complete and end with periods.

PR #1393, opened 2016-05-07
Add logging APIs for chaincodes

### LINKED ISSUE #618: We need to be able to control logging levels in chaincodes

### LINKED ISSUE #1303: There are 3 loggers logging as the "chaincode" module

bcbrock:
## Description

 APIs are added that allow Go Language chaincodes to control the logging level of the _shim_. APIs are also provided that allow a Go language chaincode to create and manage logging objects that are consistent with the shim logger, but do not require the chaincode to import any new packages. The APIs are fully documented in the code and accompanying documentation. A new test file with 2 test cases has been added.

I have also removed a redundant logging object from the chaincode package (#1303)

I also modified the logger used by the _shim_ to log as "shim" rather than as "chaincode" to remove the ambiguity with logs from the _chaincode_ package.

I also set up formatted logging for the _shim_ and chaincodes.
## Motivation and Context

Fixes #1303 
Fixes #618
## How Has This Been Tested?

Passes Go language tests and Behave tests.

This will not cause any current chaincodes to fail or require recompile.
## Checklist:

- [ x] If applicable, I have updated the documentation accordingly.
- [x ] If applicable, I have added tests to cover my changes.


bcbrock:
@muralisrini Here is my rework of the chaincode logging PR.

muralisrini:
@bcbrock : very cool, love it,.... :-)  LGTM

christo4ferris:
@bcbrock there are a number of gometalinter warnings in the changed files that should be cleaned up. Otherwise LGTM, though will need to be revisited when we finally catch up to with more recent version of go-logging (this is one ugly piece technical debt we are carrying, IMO)

bcbrock:
@christo4ferris I've offered my opinion of the `gometalinter` requirement in #1414. I could run `gofmt`, but even then, I would prefer not to see random formatting changes from other peoples code mixed in with my functional changes. It seems to me we need a plan to establish a well-formatted, consistent code base, rather than imposing it one PR at a time.

To your other point, as you can see I used the "f" forms of formatted I/O functions for use by the chaincode loggers and other techniques to protect chaincode developers from changes to the core code's logging infrastructure.

core/chaincode/shim/chaincode.go
```diff
@@ -881,3 +888,99 @@ func (stub *ChaincodeStub) insertRowInternal(tableName string, row Row, update b
...
+	level := LoggingLevel(l)
+	if err != nil {
+		level = LOG_ERROR
+	}
+	return level, err
+}	
+
+// ------------- Chaincode Loggers ---------------
+
+type ChaincodeLogger struct {
```
srderson, line 931: Please add a comment to public types as indicated by golint

core/chaincode/shim/chaincode.go
```diff
@@ -881,3 +888,99 @@ func (stub *ChaincodeStub) insertRowInternal(tableName string, row Row, update b
...
+}
+
+// Note: We're only creating the 'f' forms of the logging functions here for
+// consistency with Go language conventions around formatted I/O routines, and
+// to avoid confusion with the conventions used in the core code. It is
+// possible that some day the core code will also change from using
+// logger.Debug() to logger.Debugf() etc., and we want to protect chaincode
+// writers from that hiccup if it occurs.
+
+func (c *ChaincodeLogger) Debugf(format string, args ...interface{}) {
```
srderson, line 964: Please add a comment to public functions as indicated by golint

bcbrock:
@srderson I am a novice at git, and I have screwed up somehow trying to integrate the changes suggested by golint. I am going to delete this branch/PR and start over. When I come back, though, all of the changes will be there.

orderer/common/multichannel/registrar.go
```diff
@@ -446,3 +453,64 @@ func (r *Registrar) RemoveChannel(channelID string, removeStorage bool) error {
...
+
+// validateJoinBlock returns whether this block can be used as a join block for the channel participation API
+// as well as whether the orderer is a member of the consenter set.
+// A join block is a block of either a system channel containing a ConsoritumsConfig, or an application channel
+// containing an ApplicationConfig.
+func (r *Registrar) validateJoinBlock(channelID string, configBlock *cb.Block) (isAppChannel bool, isMember bool, err error) {
+	if configBlock == nil {
+		return false, false, errors.New("nil block")
+	}
+
```
tock-ibm: Extract the channel header from the block and verify that the ChannelId it is equal to the method parameter channelID.  May also check the type (should be HeaderType_CONFIG) before trying to create a bundle below.

PR #1398, opened 2016-05-07
Add a godoc icon to README

srderson:
## Description

This adds a godoc icon to the README
## Motivation and Context

This change allows users to easily access the GoDoc for the Hyperledger Fabric project
## How Has This Been Tested?

There are no new test cases as this is only a change to the README
## Checklist:

- [x] I have run [gometalinter](https://github.com/alecthomas/gometalinter), [gofmt](https://golang.org/cmd/gofmt/), and [goimports](https://godoc.org/golang.org/x/tools/cmd/goimports) and have resolved all errors and warnings.
- [ ] If applicable, I have added tests to cover my changes.

@binhn I know you were discussing how to share GoDocs with users. I think this may be a good approach. We may wish to add links elsewhere to more specific areas of the GoDoc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-[![Build Status](https://travis-ci.org/hyperledger/fabric.svg?branch=master)](https://travis-ci.org/hyperledger/fabric)
+[![Build Status](https://travis-ci.org/hyperledger/fabric.svg?branch=master)](https://travis-ci.org/hyperledger/fabric) [![GoDoc](https://godoc.org/github.com/hyperledger/fabric?status.svg)](https://godoc.org/github.com/hyperledger/fabric)
 
 # Incubation Notice
 This project is a Hyperledger project in _Incubation_. It was proposed to the community and documented [here](https://goo.gl/RYQZ5N). Information on what _Incubation_ entails can be found in the [Hyperledger Project Lifecycle document](https://goo.gl/4edNRc).


binhn:
@srderson This is a step forward with API doc, but have you been able to find a way to enhance the contents so that `godoc` doesn't output the "public" functions that are only intended for internal packages but not external APIs?  For example, the package [shim](https://godoc.org/github.com/hyperledger/fabric/core/chaincode/shim) contains chaincode APIs but also other "junks" that we don't want chaincode developers to use as they are not APIs. Maybe we have to re-org the chaincode APIs into another package by itself.

muralisrini:
@binhn The only _public_ method that chaincode.go has that is not a "chaincode API" is StartInProc. This can be moved to inprocstream.go (which is a consistent change) so all public methods in godoc are user APIs. Then we can pare it down so chaincode.go is the only file in the shim package.

christo4ferris:
@binhn IMO, this is a function of reorganizing/refactoring the code to correctly reflect what is public and what is not. This PR is just about adding the badge to the README. 

LGTM, suggest it be merged. If there is a need to address your concern, that should be dealt with in a separate issue(s).

binhn:
@muralisrini I see other protobuf stuff toward the end of the file...  but I agreed with @christo4ferris that we should discuss the contents of the godoc on another issue #1410.

PR #1402, opened 2016-05-08
Rebrand our docker image under the hyperledger name

ghaskins:
Rebrand all remaining baseimage mechanics to the hyperledger name
## Description

This patch introduces a lateral move of the assets on dockerhub from openblockchain/baseimage to hyperledger/fabric-baseimage.

Notable changes aside from the obvious renaming:
- We now use "fabric-baseimage" rather than the more general "baseimage" to reflect there might be other projects under the hyperledger umbrella beside the fabric project.
- We include the $arch in the tag ("x86_64-0.0.9" vs "0.0.9") to lay the groundwork for supporting other architectures.  Today, only x86_64 is available, with others planned for the near future.

Note that the image was, in fact, recompiled rather than copied.  This is because the branding impacts the image with things such as /etc/hyperledger-baseimage-release.  However, it is not believed that the resulting image differed from the original artifact in any substantial way since there have been no relevant changes since v0.0.9.  Therefore, the version number remains unchanged despite a regeneration cycle.
## Motivation and Context

The references to "openblockchain" were a legacy remnant from the project history prior to HL/LF acceptance as an incubated project.  This reference remained longer than most other rebranding efforts because of initial issues with permissions/access to the hyperledger namespace on dockerhub, which has since been resolved.
## How Has This Been Tested?

Standard boot fresh vagrant and run unit-tests/behave.  All passed
## Checklist:
- [ ] I have run [gometalinter](https://github.com/alecthomas/gometalinter), [gofmt](https://golang.org/cmd/gofmt/), and [goimports](https://godoc.org/golang.org/x/tools/cmd/goimports) and have resolved all errors and warnings.

### DIFF
--- a/consensus/docker-compose-files/Dockerfile
+++ b/consensus/docker-compose-files/Dockerfile
@@ -1,4 +1,4 @@
-from openblockchain/baseimage:latest
+from hyperledger/fabric-baseimage:latest
 
 # Copy GOPATH src and install Peer
 RUN mkdir -p /var/hyperledger/db
--- a/core/ledger/genesis/genesis_test.yaml
+++ b/core/ledger/genesis/genesis_test.yaml
@@ -46,7 +46,7 @@ peer:
     networkId: dev
 
     Dockerfile:  |
-        from openblockchain/baseimage
+        from hyperledger/fabric-baseimage
         # Copy GOPATH src and install Peer
         COPY src $GOPATH/src
         RUN mkdir -p /var/hyperledger/db
@@ -186,7 +186,7 @@ chaincode:
 
         # This is the basis for the Golang Dockerfile.  Additional commands will be appended depedendent upon the chaincode specification.
         Dockerfile:  |
-            from openblockchain/baseimage
+            from hyperledger/fabric-baseimage
             COPY src $GOPATH/src
             WORKDIR $GOPATH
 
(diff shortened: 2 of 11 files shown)


ghaskins:
@srderson @christo4ferris I checked the "added tests" box because I know the unit-tests/behave effectively test this change by virtue of they depend on the dockerhub mechanism.  It's not clear to me if this is the most accurate way to capture the actions taken.

Consider:
- Not checking the box implies there is no test coverage, which is not the case
- Checking the box implies I added something new, which is also not the case.

I am just wondering if there are any checklist level changes that should try to capture the above scenario.  Or perhaps this is not a scenario that will be common and we don't need to worry about it.

ghaskins:
@binhn @bcbrock

christo4ferris:
I don't think that these are related, but there are still a few vestigial references to 'openblockchain' remaining.

grep -r "openblockchain" .
./bddtests/utxo.feature:      When I deploy chaincode "github.com/openblockchain/obc-peer/examples/chaincode/go/utxo" with ctor "init" to "vp0"
./core/chaincode/exectransaction.go:        txUUID, err := ledger.GetState(chaincodeID, "github.com_openblockchain_obc-peer_chaincode_id", true)
./docs/API/Samples/Sample_1.js:            "path":"github.com/openblockchain/obc-peer/openchain/example/chaincode/chaincode_example02"
./examples/chaincode/go/asset_management/asset_management.go:// https://github.com/openblockchain/obc-docs/blob/master/tech/application-ACL.md
./examples/chaincode/go/asset_management_with_roles/asset_management_with_roles.go:// https://github.com/openblockchain/obc-docs/blob/master/tech/application-ACL.md

./bddtests/utxo.feature I am not even sure I understand how this is relevant until @srderson and @tamasblummer merge their changes from the hackathon

./core/chaincode/exectransaction.go seems like a bug... what I am unclear on is what that value should be, and regardless, it shouldn't be hard-coded;-)

./docs/API/Samples/Sample_1.js is dated and needs to be tested and updated.

the remaining need to be updated to reflect the new location of these files. 

What we may want to do for this merge, though is give a heads up on mailing list, slack what we have done. 

I will fix the examples with a PR and raise a separate issue. 

I guess what I am saying is that other than these vestigial references to 'openblockchain' which I don't believe are related, that this LGTM thanks.

ghaskins:
Thanks @christo4ferris.  Regarding the check-list verbage: perhaps "If applicable, I have ensured there is unit-test/behave coverage for these changes" ?

ghaskins:
Superseded by #1411

PR #1407, opened 2016-05-09
Multiple RootNodes support

yacovm:
## Description

All peers can now start with an array of root nodes instead of a single one,
which removes the SPOF of having a peer depend on a single root node for joining the cluster.
The GetRootNode() method is not present anymore and instead is replaced by GetRandomNode() which returns a random root node. 
## How Has This Been Tested?

Ran the unit tests (go test -timeout=20m $(go list github.com/hyperledger/fabric/... | grep -v /vendor/ | grep -v /examples/))
and if the travis job finished successfully then I can say that I also ran the behave tests (indirectly).

Unit tests for code are included inside and they check the possible scenarios of (whether the peer is validator or not) X (input of core.peer.rootnode config)
## Checklist:


core/discovery.go
```diff
@@ -19,9 +19,43 @@ under the License.
...
 
-// GetRootNode place holder function for providing a boostrap address for a Validating Peer.
-func GetRootNode() (string, error) {
-	return viper.GetString("peer.discovery.rootnode"), nil
+	"github.com/spf13/viper"
+
+	
+)
+
+type StaticDiscovery struct {
```
srderson, line 26: Please add comments to public types and functions in the form `// StaticDiscovery ...`

discovery/discovery.go
```diff
@@ -0,0 +1,43 @@
...
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+*/
+
+package discovery
+
+type Discovery interface {
```
srderson, line 24: Please add comments to public interfaces and functions in the form `// Discovery ...`

srderson:
@jeffgarratt could you give this a quick review also?

core/discovery.go
```diff
@@ -19,9 +19,48 @@ under the License.
...
+	"fmt"
+	"math/rand"
+	"strings"
+	"time"
 
-// GetRootNode place holder function for providing a boostrap address for a Validating Peer.
-func GetRootNode() (string, error) {
-	return viper.GetString("peer.discovery.rootnode"), nil
+	"github.com/spf13/viper"
+
```
corecode: how come `go fmt` kept these empty lines?

core/discovery.go
```diff
@@ -19,9 +19,48 @@ under the License.
...
+// StaticDiscovery is an implementation of Discovery
+type StaticDiscovery struct {
+	rootNodes   []string
+	random      *rand.Rand
+	isValidator bool
+}
+
+// NewStaticDiscovery is a constructor of a Discovery implementation
+func NewStaticDiscovery(isValidator bool) *StaticDiscovery {
+	staticDisc := StaticDiscovery{}
```
corecode: for uniformity, you may want to call this `sd` like in the functions below

core/discovery.go
```diff
@@ -19,9 +19,48 @@ under the License.
...
+	return &staticDisc
+}
+
+// GetRootNode returns the first root node defined in peer.discovery.rootnode config
+// for validating peers, and a random one for non validating peers.
+func (sd *StaticDiscovery) GetRootNode() (string, error) {
+	if sd.isValidator {
+		return sd.rootNodes[0], nil
+	}
+	return sd.rootNodes[sd.random.Int()%(len(sd.rootNodes))], nil
```
corecode: How about using `sd.random.Intn(len(sd.rootNodes))`?

core/discovery.go
```diff
@@ -19,9 +19,48 @@ under the License.
...
+}
+
+// GetRootNodes returns a comma separated array of all the nodes defined in peer.discovery.rootnode config
```
corecode: what do you mean by "comma separated"?

core/discovery.go
yacovm: Didn't check but I guess it's from the amend that I added the comments, didn't run gofmt after that.

core/discovery.go
yacovm: to call what exactly? that variable staticDisc?

core/discovery.go
```diff
@@ -19,9 +19,48 @@ under the License.
...
+func (sd *StaticDiscovery) GetRootNodes() ([]string, error) {
+	var nodes []string = make([]string, len(sd.rootNodes))
+	if copied := copy(nodes, sd.rootNodes); copied != len(sd.rootNodes) {
```
corecode: how about using:
`return append([]string(nil), sd.rootNodes...), nil`?
In that case, the error return can be removed as well.

core/discovery.go
yacovm: Sounds better.

core/discovery_test.go
```diff
@@ -19,14 +19,121 @@ under the License.
...
+
+func TestDiscovery_GetEmptyRootNodes(t *testing.T) {
+	discovery.SetDiscoveryService(NewStaticDiscovery(true))
+	if rootNodes, err := discovery.GetRootNodes(); err != nil {
+		fail(t, "Error getting rootnodes:  %s", err)
+	} else {
+		if size := len(rootNodes); size != 1 || rootNodes[0] != "" {
+			fail(t, "Needed input is not ['']")
+		}
+
```
corecode: extra empty line

core/discovery_test.go
```diff
@@ -19,14 +19,121 @@ under the License.
...
+func inArray(element string, array []string) bool {
+	for _, val := range array {
+		if val == element {
+			return true
+		}
+	}
+	return false
+}
 
+func fail(t *testing.T, s string, a ...interface{}) {
```
corecode: You can use `t.Fatalf` instead of creating a separate function.

core/discovery_test.go
```diff
@@ -19,14 +19,121 @@ under the License.
...
+func TestDiscovery_GetEmptyRootNode(t *testing.T) {
+	discovery.SetDiscoveryService(NewStaticDiscovery(true))
+	assertRootNode(t, "")
+}
+
+func TestDiscovery_GetEmptyRootNodes(t *testing.T) {
+	discovery.SetDiscoveryService(NewStaticDiscovery(true))
+	if rootNodes, err := discovery.GetRootNodes(); err != nil {
+		fail(t, "Error getting rootnodes:  %s", err)
+	} else {
```
corecode: If you make `fail` abort (or use `t.Fatalf`), you can hoist the else branch.

core/discovery_test.go
```diff
@@ -19,14 +19,121 @@ under the License.
...
+	"github.com/hyperledger/fabric/discovery"
+	"github.com/spf13/viper"
+)
+
+func TestDiscovery_GetEmptyRootNode(t *testing.T) {
+	discovery.SetDiscoveryService(NewStaticDiscovery(true))
+	assertRootNode(t, "")
+}
+
+func TestDiscovery_GetEmptyRootNodes(t *testing.T) {
```
corecode, line 30: Won't these empty rootnode tests fail if somebody enters a rootnode in their local config file?  I guess we should remove the rootnode config explicitly for this test.

core/discovery_test.go
```diff
@@ -19,14 +19,121 @@ under the License.
...
+func TestDiscovery_GetSingleValidatingPeer(t *testing.T) {
+	viper.Set("peer.discovery.rootnode", "someHost")
+	discovery.SetDiscoveryService(NewStaticDiscovery(true))
+	assertRootNode(t, "someHost")
+}
+
+func TestDiscovery_GetAllPeers(t *testing.T) {
+	s := "a,b,c,d,e,f,g,h,i,j,k,l,m,n,o,p,q,r,s,t,u,v,w,x,y,z"
+	viper.Set("peer.discovery.rootnode", s)
+	for _, validator := range []bool{true, false} {
```
corecode: that's cute :)

core/discovery_test.go
```diff
@@ -19,14 +19,121 @@ under the License.
...
+	} else {
+		if size := len(rootNodes); size != 1 || rootNodes[0] != "" {
+			fail(t, "Needed input is not ['']")
+		}
+
+	}
+}
+
+func TestDiscovery_GetSingleValidatingPeer(t *testing.T) {
+	viper.Set("peer.discovery.rootnode", "someHost")
```
corecode: Won't this test race with the null test if both tests are executed in parallel?

core/peer/peer.go
```diff
@@ -233,8 +234,16 @@ func NewPeerWithHandler(secHelperFunc func() crypto.Peer, handlerFact HandlerFac
 		return nil, fmt.Errorf("Error constructing NewPeerWithHandler: %s", err)
 	}
 	peer.ledgerWrapper = &ledgerWrapper{ledger: ledgerPtr}
-	go peer.chatWithPeer(viper.GetString("peer.discovery.rootnode"))
-	return peer, nil
+
+	if rootNodes, err := discovery.GetRootNodes(); err == nil {
+		for _, rootNode := range rootNodes {
+			go peer.chatWithPeer(rootNode)
```
corecode: Why connect to all rootnodes?

peer/main.go
```diff
@@ -494,14 +501,14 @@ func serve(args []string) error {
 		go rest.StartOpenchainRESTServer(serverOpenchain, serverDevops)
 	}
 
-	rootNode, err := core.GetRootNode()
+	rootNodes, err := discovery.GetRootNodes()
 	if err != nil {
-		grpclog.Fatalf("Failed to get peer.discovery.rootnode valey: %s", err)
+		grpclog.Fatalf("Failed to get discovery rootnodes valey: %s", err)
```
corecode: `value`

core/discovery.go
```diff
@@ -19,9 +19,48 @@ under the License.
...
+	staticDisc := StaticDiscovery{}
+	staticDisc.rootNodes = strings.Split(viper.GetString("peer.discovery.rootnode"), ",")
+	staticDisc.random = rand.New(rand.NewSource(time.Now().Unix()))
+	staticDisc.isValidator = isValidator
+	return &staticDisc
+}
+
+// GetRootNode returns the first root node defined in peer.discovery.rootnode config
+// for validating peers, and a random one for non validating peers.
+func (sd *StaticDiscovery) GetRootNode() (string, error) {
```
corecode: Why is the behavior different for validating peers?  Why not always return a random node?

peer/main.go
```diff
@@ -392,11 +395,15 @@ func serve(args []string) error {
 		viper.Set("ledger.blockchain.deploy-system-chaincode", "false")
 		viper.Set("validator.validity-period.verification", "false")
 	}
+
+	// Install discovery
+	discovery.SetDiscoveryService(core.NewStaticDiscovery(viper.GetBool("peer.validator.enabled")))
```
corecode: Why do we use a singleton, which even gets initialized the same way?  I think (explicitly) passing the new discovery instance to the peer is better (than implicitly sharing it via a global singleton).

core/discovery.go
yacovm: You're right, this needs to be rephrased. I meant an array that was split on a comma separated string.

core/discovery.go
yacovm: In that case shouldn't I remove the error on the single form (GetRootNode) ?

core/discovery_test.go
yacovm, line 30: Why? If the test runs- how can the config file be loaded without the main.go not being invoked?

core/discovery_test.go
yacovm: I didn't know that we execute tests in parallel... do we?

core/discovery.go
yacovm: I fail to see a use case in which a validating peer would want to connect to a random peer. So I want to enforce deterministic behavior.

core/discovery.go
corecode: Yes

core/discovery.go
corecode: I'd love to remove error returns where there don't have to be error returns.

core/discovery_test.go
corecode, line 30: I assumed that the config file is always loaded.  At least for most tests this seems to be the case.

core/discovery_test.go
corecode: `go test` by default executes tests in parallel.

core/discovery_test.go
yacovm, line 30: hmm, but when I ran the tests via 

> go test -timeout=20m $(go list github.com/hyperledger/fabric/... | grep -v /vendor/ | grep -v /examples/)           

It went OK and no test failed.

corecode, line 30: right, not all races actually show up immediately.

yacovm:
OK @corecode , thanks for the CR, I'll resubmit the PR in the upcoming week.

Issue #645, opened 2016-02-10
State transfer should be migrated out of the consensus package


jyellick:
As discussed with @binhn @jeffgarratt @kchristidis @corecode and others, the state transfer code is very obc/ledger specific, and has applications outside of consensus.

In particular, the state transfer code can/should be utilized for synchronizing block state to NVPs, as well as for the `noops` and each of the `obcpbft` plugins.

Therefore, the state transfer code should be migrated into a package outside of consensus, possibly under `peer`.

Then, the `Stack` interface (formerly `CPI`) should be modified to expose a `StateTransfer` interface, which the consensus plugins can utilize for their needs.

When NVP catchup is implemented, it should be able to be done trivially by hooking into the new state transfer package (with an additional dependency on #640 )

PR #1416, opened 2016-05-09
Remove helper dependencies from statetransfer

jyellick:
This is the first of several PRs decoupling statetransfer from consensus.
## Description

State transfer needs to be moved from `consensus/statetransfer` to `core/peer/statetransfer`.  This first changeset decouples statetransfer from `consensus/helper/helper.go` and instead uses the interfaces provided in `peer/peer.go`.
## Motivation and Context

State transfer was originally developed as part of PBFT, as it became more complex, it was realized the code could be more generally useful, and was pulled out of pbft and into consensus.  As more applications of state transfer became apparent, such as performing state transfer to a non-validating peer which does not participate in consensus, it became apparent that state transfer should be moved from `consensus` and into `peer`.

Because statetransfer is no longer a part of consensus, the related APIs have been removed from the `consensus.Stack` interface, simplifying the consensus plugin API surface noticeably.

Although it does not fix, it contributes to #645 
## How Has This Been Tested?

Since this PR adds no new feature and exercises the same underlying APIs (though accesses them slightly differently) the existing unit and behave tests are sufficient to verify its correctness.
## Checklist:
- [X] I have run [gofmt](https://golang.org/cmd/gofmt/), and [goimports](https://godoc.org/golang.org/x/tools/cmd/goimports)
- [X] If applicable, I have updated the documentation accordingly.

@corecode @tuand27613 @kchristidis Review is welcome, though consensus expertise is not needed.

@jeffgarratt Since the bulk of the code additions are into `peer.go` I'd love your vote approval.

### DIFF
--- a/consensus/consensus.go
+++ b/consensus/consensus.go
@@ -71,6 +71,7 @@ type Executor interface {
 	SkipTo(tag uint64, id []byte, peers []*pb.PeerID)
 }
 
+// StatePersistor is used to store consensus state which should survive a process crash
 type StatePersistor interface {
 	StoreState(key string, value []byte) error
 	ReadState(key string) ([]byte, error)
--- a/consensus/consensus.go
+++ b/consensus/consensus.go
@@ -17,7 +17,6 @@ limitations under the License.
 package consensus
 
 import (
-	"github.com/hyperledger/fabric/core/ledger/statemgmt"
 	pb "github.com/hyperledger/fabric/protos"
 )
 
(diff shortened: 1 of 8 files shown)


jeffgarratt:
@jyellick Can you make sure RLock is Lock when a 'write' lock is desired?  Besides that, looks good and make sense.

jyellick:
@jeffgarratt Thanks! Good catch, I didn't realize the locking was different for reads vs. writes.  I've added a commit to address this.

consensus/consensus.go
```diff
@@ -94,13 +71,6 @@ type Executor interface {
...
 }
 
-// RemoteLedgers is used to interrogate the blockchain of other replicas
-type RemoteLedgers interface {
-	GetRemoteBlocks(replicaID *pb.PeerID, start, finish uint64) (<-chan *pb.SyncBlocks, error)
-	GetRemoteStateSnapshot(replicaID *pb.PeerID) (<-chan *pb.SyncStateSnapshot, error)
-	GetRemoteStateDeltas(replicaID *pb.PeerID, start, finish uint64) (<-chan *pb.SyncStateDeltas, error)
-}
-
 type StatePersistor interface {
```
srderson, line 75: Please add a comment to `StatePersistor`

consensus/statetransfer/statetransfer.go
```diff
@@ -46,9 +46,12 @@ func init() {
 // =============================================================================
 
 type PartialStack interface {
```
srderson, line 49: Please add a comment to the public interface of the form `// PartialStack...`

consensus/statetransfer/statetransfer.go
```diff
@@ -46,9 +46,12 @@ func init() {
...
-	consensus.Inquirer
+	peer.BlockChainAccessor
+	peer.BlockChainModifier
+	peer.BlockChainUtil
+	GetPeers() (*protos.PeersMessage, error)
+	GetPeerEndpoint() (*protos.PeerEndpoint, error)
+	GetRemoteLedger(receiver *protos.PeerID) (peer.RemoteLedger, error)
 }
 
 type Listener interface {
```
srderson, line 59: Please add a comment to the public interface of the form `// Listener...`

srderson:
LGTM

PR #1423, opened 2016-05-10
Split statetransfer tests from obcpbft

jyellick:
## Description

This changeset splits the common infrastructure between state transfer and PBFT in two, and moves the statetransfer tests into the statetransfer package.  This continues the work begun in PR #1416 and is the second in a series of PRs designed to move statetransfer to `core/peer/statetransfer`.  This should not be considered for inclusion until #1416 has been pulled (the corresponding diff should shrink noticeably).
## Motivation and Context

Because state transfer was initially developed as a part of PBFT, its mock testing infrastructure was shared with the `obcpbft` package.  Because go does not allow tests to be shared across packages, this led to the awkward infrastructure of having to run the statetransfer tests as part of the obcpbft package.  As statetransfer is moved out of consensus, this becomes even more awkward and needed to be fixed.

Does not fix, but partially address #645 
## How Has This Been Tested?

This changeset only changes test files, and does not affect the actual codebase (other than removing a single unused method signature from an interface).  Because it is simply restructuring the tests themselves, no new test case is included.
## Checklist:
- [X] I have run [gofmt](https://golang.org/cmd/gofmt/), and [goimports](https://godoc.org/golang.org/x/tools/cmd/goimports) as well as golint and go vet and cleaned up any errors in the files modified by this changeset.
- [X] If applicable, I have updated the documentation accordingly.


srderson:
LGTM. Thanks!

common/ledger/blkstorage/blockstore_provider.go
```diff
@@ -125,6 +125,22 @@ func (p *BlockStoreProvider) Exists(ledgerid string) (bool, error) {
...
+	if err != nil {
+		return err
+	}
+	if !exists {
+		return nil
+	}
+	if err := p.leveldbProvider.Remove(ledgerid); err != nil {
+		return err
+	}
+	return os.RemoveAll(p.conf.getLedgerBlockDir(ledgerid))
```
cendhu: We need to check whether this is an atomic operation (i.e, whether the removed [inodes](https://geek-university.com/linux/inode/) info is flushed and persisted to disk). For the safety, I would better get the parent directory of the ledgerID and call https://github.com/hyperledger/fabric/blob/da0c6efff04b244ce0963ba4dba82150a1ef7768/common/ledger/blkstorage/file_util.go#L48 after the removal.

common/ledger/blkstorage/blockstore_provider_test.go
```diff
@@ -191,6 +191,17 @@ func TestBlockStoreProvider(t *testing.T) {
 	assert.NoError(t, err)
 	assert.Equal(t, false, exists)
 
+	// remove a channel and verify blockindex and blocks dir are deleted
```
cendhu: - Shouldn't we create a block store with a few block + index entries and then remove it? 
- Shouldn't we also check whether all types of index entries associated with the ledgerID are removed when we call `Remove()` (simply to check whether the DB has no entries for the lederID)?  The test for `DeleteAll()` added in the `leveldbhelper` pkg is more of a unit-test only for the `DeleteAll` introduced at that package but it does not ensure that the `Remove()` calls the index store to remove all entries associated with the ledgerID.

common/ledger/blkstorage/blockstore_provider.go
```diff
@@ -125,6 +125,27 @@ func (p *BlockStoreProvider) Exists(ledgerid string) (bool, error) {
 	return exists, err
 }
 
+// Remove block index and blocks for the given ledgerid (channelID). It is not an error if the channel does not exist.
+// The opened BlockStore for the removed channel should not be used any more because it will return invalid data.
+func (p *BlockStoreProvider) Remove(ledgerid string) error {
```
manish-sethi, line 132: I would add something like following to the comments.

This function is not error safe. If this function returns an error or a crash takes place, this is highly likely that the data for this ledger is left in an inconsistent state and opening the ledger again can show unknown behavior.

Having said this, I am just wondering that maintaining a basic list of unusable channels, in this code may not be a bad idea. But we can see how it evolves in usage and come back to this later. For now, this can be merged as is.

PR #1433, opened 2016-05-10
Update chaincode APIs

### LINKED ISSUE #875: we need the shim to provide the transaction timestamp to the contract

binhn:
Passing transaction timestamp to the chaincode context and update the APIs
## Description

Chaincode needs the transaction creation timestamp. The fabric includes the timestamp in the context passed to the chaincode on init, invoke, and query calls. This change also includes update to the chaincode API document and godoc comments. Unit-test for the timestamp is added to the chaincode_example01, which is run as part of the test-suite. 
## Motivation and Context

Fixes #875
## How Has This Been Tested?

Run the unit-tests or the chaincode_example01
## Checklist:

- [x ] I have added a [Signed-off-by](https://github.com/hyperledger/fabric/blob/master/CONTRIBUTING.md#legal-stuff)
- [x ] Either no new documentation is required by this change, OR I added new documentation
- [ x] Either no new tests are required by this change, OR I added new tests
- [x ] I have run [goimports](https://godoc.org/golang.org/x/tools/cmd/goimports), [go vet](https://golang.org/cmd/vet/), and [golint](https://github.com/golang/lint). I have cleaned up all valid errors and warnings in code I have added or modified. These tools may generate false positives. Don't be worried about ignoring some errors or warnings. The goal is clean, consistent, and readable code.

Signed-off-by: binhn@us.ibm.com


core/chaincode/shim/chaincode.go
```diff
@@ -252,39 +264,39 @@ func (stub *ChaincodeStub) init(uuid string, secContext *pb.ChaincodeSecurityCon
 
 // ------------- Call Chaincode functions ---------------
 
-// InvokeChaincode function can be invoked by a chaincode to execute another chaincode.
+// InvokeChaincode locally calls the specified chaincode `Invoke`.
```
srderson: People always ask me if this creates a second transaction. Should we state that it does not in the comment? Maybe that is what you were trying to imply by locally?

core/chaincode/shim/chaincode.go
```diff
@@ -252,39 +264,39 @@ func (stub *ChaincodeStub) init(uuid string, secContext *pb.ChaincodeSecurityCon
...
 func (stub *ChaincodeStub) InvokeChaincode(chaincodeName string, function string, args []string) ([]byte, error) {
 	return handler.handleInvokeChaincode(chaincodeName, function, args, stub.UUID)
 }
 
-// QueryChaincode function can be invoked by a chaincode to query another chaincode.
+// QueryChaincode locally calls the specified chaincode `Query`.
```
srderson: People always ask me if this creates a second transaction. Should we state that it does not in the comment?

binhn:
@srderson yes, that's what I meant by "locally", but i agreed what it's worth more detail explanation here. I'll provide an update

binhn:
@srderson i rebased and force push to update the pr.  Please merge

srderson:
LGTM

PR #1436, opened 2016-05-10
Add logging APIs for Go chaincodes

### LINKED ISSUE #618: We need to be able to control logging levels in chaincodes

### LINKED ISSUE #1303: There are 3 loggers logging as the "chaincode" module

bcbrock:
## Description

This changeset adds chaincode shim APIs and other facilities that allows
Go chaincodes to create multiple, independent and levelled logging objects
that log consistently with the shim logger. APIs are also provided that allow
the chaincode to control the shim's logging level.

This changeset also removes a redundant logger from the "chaincode" package, and changes the logger used by the Go chaincode _shim_ to log as "shim" (rather than as "chaincode").
## Motivation and Context

Fixes #618
Fixes #1303
## How Has This Been Tested?

All Go unit tests and behave tests pass. I've also tested it with my own chaincode that uses the new facilities to make sure that both the shim and chaincode logs are generated and controlled correctly.

All of the 'golint' warnings from my new code have been addressed. I also went ahead and gofmt-ed core/chaincode/shim/chaincode.go.
## Checklist:

- [ x] Either no new documentation is required by this change, OR I added new documentation
- [x ] Either no new tests are required by this change, OR I added new tests
- [ x] I have run [goimports](https://godoc.org/golang.org/x/tools/cmd/goimports), [go vet](https://golang.org/cmd/vet/), and [golint](https://github.com/golang/lint). I have cleaned up all valid errors and warnings in code I have added or modified. These tools may generate false positives. Don't be worried about ignoring some errors or warnings. The goal is clean, consistent, and readable code.


ghaskins:
LGTM

PR #1440, opened 2016-05-11
Multi-arch substrate

ghaskins:
Provide a framework for supporting different processor architectures
## Description

This PR lays the groundwork for supporting more than just x86_64 processors as platforms for building/deploying hyperledger/fabric.  It consists of the following major points:
- We move ./devenv/baseimage to ./baseimage to reflect that it encompases more than the development environment
- We convert the baseimage Docker build to using raw Make+Dockerfile from packer.  Vagrant remains with packer.
- We add new automated build logic that will dynamically compile the baseimage locally when a suitable architecture specific image cannot be retrieved from dockerhub
### Whats _NOT_ Included
- Explicit support for any other architectures.  These will be added in follow-up PRs.
## Motivation and Context

The current code base makes many assumptions about x86_64 and its operation is predicated on baseimages being published to dockerhub.  There is emerging interest in supporting this project on other processor architectures such as s390x and POWER.  We need several changes to address the assumptions/limitations, and this patch is one step of several required to get there.

One of the major changes is to drop packer as the engine to build docker baseimages.  Packer was originally selected in an attempt to unify the vagrant and docker builds.  But as it turns out, packer was somewhat limiting (lack of support for ENV directives, only provides x86 binaries, etc) and the vast majority of our commonality was derived from our ./scripts anyway.  So we simplify the tooling and gain the ability to more easy build docker on non-x86.  We retain packer based builds for vagrant because it works well there and also promises the ability to generate other targets like AMI, GCE, or Azure for us in the future. 

We also add a new feature to the newly introduced build system: The ability to dynamically build the baseimage from source if a suitable binary cannot be found on dockerhub.  This is powerful because it decouples platform teams from the dockerhub publishing process, currently a gate.  It means the system will treat dockerhub simply as an optimization, but can continue forward if images are not available.  Not having the dockerhub cache does substantially increase the time for a build (30+ minutes) but is better than not being able to build at all.
## How Has This Been Tested?
- 'make checks' on x86_64 have all passed
- 'make unit-test' on s390x (support not included in this PR) have passed, behave test support emerging
- forced a local baseimage build by referencing a non-existent dockerhub image version
- ensured dockerhub image is used when referencing a valid target
- ensured "make vagrant" still functions properly
## Checklist:


ghaskins:
@bcbrock

christo4ferris:
@ghaskins thanks for this. I haven't had a chance to look in-depth but you might also want to make an announce in slack and on hyperledger-technical-discuss that you are doing this to invite any who share an interest in multi-architecture support to review and provide feedback.

baseimage/Makefile
```diff
@@ -1,18 +1,37 @@
 NAME="hyperledger/fabric-baseimage"
+VERSION=0.0.9
+ARCH=$(shell uname -m)
+DOCKER_NAME ?= hyperledger/fabric-baseimage
```
ghaskins: The redundancy with line 1 should be cleaned up

baseimage/Makefile
```diff
@@ -1,18 +1,37 @@
...
+	BASEIMAGE_RELEASE=$(VERSION) \
 	packer build -only virtualbox-iso packer-local.json
 
-docker: packer-local.json
-	packer build -only docker packer-local.json
+Dockerfile:
+
+docker: Dockerfile.in Makefile
+	@echo "Generating docker"
+	@cat Dockerfile.in | \
```
ghaskins, line 28: I should emit a "Do not edit!" warning into the resulting Dockerfile, or switch to piping the file in rather than writing it to the filesystem.

scripts/provision/docker.sh
```diff
@@ -48,12 +48,32 @@
...
 
-docker pull $FQN
+echo "BUILD-CACHE: Pulling \"$DOCKERHUB_NAME\" from dockerhub.."
+docker pull $DOCKERHUB_NAME
+docker inspect $DOCKERHUB_NAME 2>&1 > /dev/null
+if [ "$?" == "0" ]; then
+   echo "BUILD-CACHE: Success!"
+   BASENAME=$DOCKERHUB_NAME
+else
+    echo "BUILD-CACHE: WARNING - Build-cache unavailable, attempting local build but this should be fixed ASAP"
```
ghaskins, line 63: Whitespace discrepancy between conditional clauses.  Will fix.

bcbrock:
If you think it's working for X86/Vagrant I would say go ahead and merge, and we can try it on POWER after that. The fact that you have s390x partially working is a good sign.

bcbrock:
Or - if you have the time and energy to open a whole new can of worms - how could this infrastructure be extended to different Linux distributions, and should we go ahead and put in some plumbing now? Can all the uses of "apt-get" be isolated, or replaced with some kind of portable package manager? My personal near-term interest is running the fabric on an OpenStack cloud with POWER8 nodes running Fedora 23.

christo4ferris:
@bcbrock @ghaskins yeah, I am also hearing interest in RHEL However, this isn't something we can simply parameterize, I don't think.

ghaskins:
@christo4ferris @bcbrock My feelings on the multi-distro support are that I think we can strive to get things like vagrant/travis/linuxone/etc as more or less just a set of packages that need to be installed and the build system will take care of everything else (installing go-tools like golint/protoc-gen-go/etc, building docker images, etc).  We can then use something like puppet-apply (https://docs.puppet.com/puppet/latest/reference/man/apply.html) rather than bash to abstract the distro specifics for the former portion.  Those tools are already designed to deal with distro abstractions, so they might be able to offer some simplicity/sanity in this regard.  To be clear, I am not proposing that this would all be part of a distributed puppet network per se, but rather we use puppet-apply with manifests stored locally in the tree.  TBD.

ghaskins:
Fixes #1455, though a more targeted fix is possible if this PR is not believed to be on a fast-track for merging.

srderson:
@vpaprots This may be of interest to you

baseimage/Makefile
```diff
@@ -1,18 +1,37 @@
...
-	packer build packer.json
+all: vagrant docker
 
 $(VAGRANTIMAGE): packer-local.json
+	BASEIMAGE_RELEASE=$(VERSION) \
 	packer build -only virtualbox-iso packer-local.json
 
-docker: packer-local.json
-	packer build -only docker packer-local.json
+Dockerfile:
```
ghaskins: This looks like a vestige of an earlier design and should be cleaned up.

srderson:
@ghaskins does the "in progress" tag imply that a PR should not be merged because there's still ongoing work? That was my assumption, but now I'm not sure based on your last comment.

ghaskins:
@srderson yeah, though I probably should have talked to you guys about that, heh.  I was trying to figure a good way to mark the PRs that had the ball back in the court of the submitter so I didn't keep looking at them.  In this case, I have patches to fix the issues I flagged above that I will submit shortly so I will remove the tag then.

Bigger question is: do you and @binhn etc, like that approach?  I am open to other ideas.

ghaskins:
@srderson And, I just pushed those fixes and remove the tag....should be considered a candidate again.

christo4ferris:
@ghaskins such an approach will only work for someone with write access, I think. I am not sure that anyone submitting a PR can do this. Maybe you could confirm whether anyone without write access can assign a label?

ghaskins:
@christo4ferris understood, sorry.  I was more thinking along the lines of a maintainer operation.  E.g. "I looked at this and it needs more work".  Right now, we can look at the PR dashboard and see details like whether travis worked, whether someone has been assigned (presumably a maintainer), how much activity there is in the comments, but no real indication of "ready for merge consideration".  I was thinking perhaps an "in-progress" label could help to mark those that got pushed back out for more work.

baseimage/Makefile
```diff
@@ -1,18 +1,36 @@
...
-docker: packer-local.json
-	packer build -only docker packer-local.json
+Dockerfile: Dockerfile.in
+	@echo "# Generated from Dockerfile.in.  DO NOT EDIT!" > $@
+	@cat Dockerfile.in | \
+	sed -e  "s|_DOCKER_BASE_|$(DOCKER_BASE)|" >> $@
+
+docker: Dockerfile Makefile
+	@echo "Generating docker"
+	@docker build -t $(NAME):$(DOCKER_TAG) .
```
vpaprots, line 33: @ghaskins One thing that did bug me, is the use of @ in the makefile, though I think this docker build command is the only one that I would really remove it from.. Unless docker doesnt report failures via $? properly and should be wrapped in some sort of grep?

One could argue the same about @ cat since if it failed, dockerbuild would foul up too without a generated dockerfile to follow...

ghaskins, line 33: I have no problem removing them, but I want to make sure we are talking about the same thing:

@ just suppresses the makefile echo of the command

It sounds like you are taking about "-" which has Make ignore any errors in $?.

Or did I misunderstand you?

vpaprots:
One more comment.. .gitignore needs updating to not consider generated Dockerfile:

```
[root@blockchain fabric]# git status
# On branch s390x-build-support
# Untracked files:
#   (use "git add <file>..." to include in what will be committed)
# 
#   baseimage/Dockerfile
#   devenv/setupLinuxONE.sh
```

baseimage/Makefile
vpaprots, line 33: Ah.. yes, you are right.. ignore that comment then, we really dont need to see the full dump of docker printouts.. I thought I double-checked the syntax before posting, apparently not

ghaskins:
@vpaprots good catch, thank you.  I think I also need to make Dockerfile dep on Makefile technically.  Will fix both.

ghaskins:
@christo4ferris I was just thinking about this some more:  To your point, its all well and good if a maintainer marks it "in-progress" but if it isn't bi-directional its of limited value.  E.g. how does the submitter than get noticed again that its ready if they can't clear the flag.

Oh well.  Bad idea in general and drop it?  Or good idea with more refinement needed?

ghaskins:
CI failure is the 680 issue (i.e. unrelated).  I rebased this and addressed remaining comments, ready for consideration again.

scripts/provision/docker.sh
```diff
@@ -49,11 +49,31 @@
...
-docker pull $FQN
+echo "BUILD-CACHE: Pulling \"$DOCKERHUB_NAME\" from dockerhub.."
+docker pull $DOCKERHUB_NAME
+docker inspect $DOCKERHUB_NAME 2>&1 > /dev/null
+if [ "$?" == "0" ]; then
+    echo "BUILD-CACHE: Success!"
+    BASENAME=$DOCKERHUB_NAME
+else
+    echo "BUILD-CACHE: WARNING - Build-cache unavailable, attempting local build but this should be fixed ASAP"
+    (cd $CURDIR/../../baseimage && make docker DOCKER_TAG=localbuild)
```
ghaskins, line 64: This probably needs to have its retval checked for error.  We wouldn't want to proceed if this fails.

srderson:
Had a converstation with @ghaskins on slack in the #fabric-dev-env channel. My main concern was with /baseimage being moved out of the /devenv folder. The baseimage has lots of useful development stuff in it, but when running in production I think many users may wish to use a lightweight image such as alpine for the peer or chaincode.

We discussed using /images/base instead so that additional images can be added in the future.

Other than that, LGTM

ghaskins:
Superseded by #1502

PR #1441, opened 2016-05-11
marshal chaincode return message as payload to calling chaincode

### LINKED ISSUE #1425: In CC to CC calls, the calling CC will get killed by FSM if called CC returns Error.
## Description

If you set up a chaincode (cc) that calls another cc, then if the called cc returns a fmt.Error (which can be common), then the cc FSM intercepts that Error and kills the calling cc.  This does not happen if you call the cc from a client (ie outside the chain).  The behavior should be consistent across calling methods (ie should not depend on who is calling the cc) and the common use case of returning a fmt.Errror should be supported.
## Describe How to Reproduce

Attached are two chaincodes: dns.go and dns-test.go that demonstrate this problem.  To reproduce follow these steps:
- Run cc dns.go and dns-test.go on the peer in dev mode in terms 1 and 2.
  - term1: `CORE_CHAINCODE_ID_NAME=dns CORE_PEER_ADDRESS=0.0.0.0:30303 ./dns`
  - term2: `CORE_CHAINCODE_ID_NAME=dns-test CORE_PEER_ADDRESS=0.0.0.0:30303 ./dns-test`
- Deploy both cc's from term3.
  - `./peer chaincode deploy -n dns -c '{"Function":"FooBar", "Args":[]}'`
  - `./peer chaincode deploy -n dns-test -c '{"Function":"FooBar", "Args":[]}'`
- Create a namespace in term3
  - `./peer chaincode invoke -l golang -n dns-test -c '{"Function": "registerNamespace", "Args":["NS1", "owner1"]}'`
- Create a name mapping in term3
  - `./peer chaincode invoke -l golang -n dns-test -c '{"Function": "register", "Args":["owner1", "NS1", "test", "1234"]}'`
- Call query for existing name (should not fail)
  - `./peer chaincode query -l golang -n dns-test -c '{"Function": "lookup", "Args":["NS1", "test"]}'`
- Call query for non-existing name (will fail)
  - `./peer chaincode query -l golang -n dns-test -c '{"Function": "lookup", "Args":["NS1", "HelloWorld"]}'`

The error seen in term2 (which is running dns-test) is:

`Query: (lookup)
Calling (dns), function (lookup) with args ([NS1 HelloWorld])
2016/05/05 22:24:13 [683d6d42]Sending INVOKE_QUERY
2016/05/05 22:24:13 [683d6d42]Received message ERROR from shim
2016/05/05 22:24:13 [683d6d42]Handling ChaincodeMessage of type: ERROR(state:ready)
Error starting dns-test: Error handling message: [683d6d42-5912-4a28-9d9e-4b8475586b46]Chaincode handler FSM cannot handle message (ERROR) with payload size (31) while in state: ready`

And the dns-test cc process exits as a result.

muralisrini:
Chaincode calling chaincode treated all errors - fabric as well as error returned from called chaincode - as errors. This caused calling chaincode to error out. 
## Description

chaincode_support.go - Chaincode's execute method returns response and error. It used to introspect response and if a chaincode error,  set the returned error as well. This is unnecessary as the fabric treats chaincode errors correctly anyway. By not setting the error in the first place the way is paved for chaincode-calling-chaincode path. 

With Execute not returning error for everything, rest of the fix can rely upon nil error to marshalls the response from the called chaincode in the peer and unmarshalls it in the calling shim. The response is then checked for error in calling chaincode.
## Motivation and Context

Conflating fabric errors with chaincode errors causes these to be indistiguishable in the called chaincode. Called chaincode treats all errors as Fabric errors and exit with error.  The fix disambiguates the two cases and the shim correctly  looks for fabric errors, chaincode errors and chaincode success responses.

Fixes #1425
## How Has This Been Tested?

Tested with testcase supplied in #1425 
Added unit tests
## Checklist:

### DIFF
--- a/core/chaincode/chaincode_support.go
+++ b/core/chaincode/chaincode_support.go
@@ -327,7 +327,7 @@ func (chaincodeSupport *ChaincodeSupport) launchAndWaitForRegister(ctxt context.
 	return alreadyRunning, err
 }
 
-//StopChaincode stops a chaincode if running
+//Stop stops a chaincode if running
 func (chaincodeSupport *ChaincodeSupport) Stop(context context.Context, cds *pb.ChaincodeDeploymentSpec) error {
 	chaincode := cds.ChaincodeSpec.ChaincodeID.Name
 	if chaincode == "" {
@@ -359,7 +359,7 @@ func (chaincodeSupport *ChaincodeSupport) Stop(context context.Context, cds *pb.
 	return err
 }
 
-// LaunchChaincode will launch the chaincode if not running (if running return nil) and will wait for handler of the chaincode to get into FSM ready state.
+// Launch will launch the chaincode if not running (if running return nil) and will wait for handler of the chaincode to get into FSM ready state.
 func (chaincodeSupport *ChaincodeSupport) Launch(context context.Context, t *pb.Transaction) (*pb.ChaincodeID, *pb.ChaincodeInput, error) {
 	//build the chaincode
 	var cID *pb.ChaincodeID
@@ -497,7 +497,7 @@ func (chaincodeSupport *ChaincodeSupport) getVMType(cds *pb.ChaincodeDeploymentS
 	return container.DOCKER, nil
 }
 
-// DeployChaincode deploys the chaincode if not in development mode where user is running the chaincode.
+// Deploy deploys the chaincode if not in development mode where user is running the chaincode.
 func (chaincodeSupport *ChaincodeSupport) Deploy(context context.Context, t *pb.Transaction) (*pb.ChaincodeDeploymentSpec, error) {
 	//build the chaincode
 	cds := &pb.ChaincodeDeploymentSpec{}
@@ -595,9 +595,8 @@ func (chaincodeSupport *ChaincodeSupport) Execute(ctxt context.Context, chaincod
 	var ccresp *pb.ChaincodeMessage
 	select {
 	case ccresp = <-notfy:
-		if ccresp.Type == pb.ChaincodeMessage_ERROR || ccresp.Type == pb.ChaincodeMessage_QUERY_ERROR {
-			err = fmt.Errorf(string(ccresp.Payload))
-		}
+		//response is sent to user or calling chaincode. ChaincodeMessage_ERROR and ChaincodeMessage_QUERY_ERROR
+		//are typically treated as error
 	case <-time.After(timeout):
 		err = fmt.Errorf("Timeout expired while executing transaction")
 	}
--- a/core/chaincode/exectransaction.go
+++ b/core/chaincode/exectransaction.go
@@ -128,8 +128,8 @@ func Execute(ctxt context.Context, chain *ChaincodeSupport, t *pb.Transaction) (
 
 //ExecuteTransactions - will execute transactions on the array one by one
 //will return an array of errors one for each transaction. If the execution
-//succeeded, array element will be nil. returns []byte of state hash or 
-//error 
+//succeeded, array element will be nil. returns []byte of state hash or
+//error
 func ExecuteTransactions(ctxt context.Context, cname ChainName, xacts []*pb.Transaction) (stateHash []byte, txerrs []error, err error) {
 	var chain = GetChain(cname)
 	if chain == nil {
--- a/core/chaincode/exectransaction_test.go
+++ b/core/chaincode/exectransaction_test.go
@@ -21,6 +21,7 @@ import (
 	"net"
 	"os"
 	"strconv"
+	"strings"
 	"sync"
 	"testing"
 	"time"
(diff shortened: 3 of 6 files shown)


ghaskins:
Thank you @muralisrini 

LGTM

orderer/common/filerepo/filerepo.go
```diff
@@ -0,0 +1,102 @@
...
+
+	if err := w.Flush(); err != nil {
+		return err
+	}
+
+	if err := tmpFile.Close(); err != nil {
+		return err
+	}
+
+	return os.Rename(tmpFile.Name(), dest)
```
cendhu: **Close and rename do not imply sync** -- a good deck that exactly talk about this scenario https://www.slideshare.net/nan1nan1/eat-my-data

As you might know, `Flush` flushes the data present in the write buffer only to the disk (no fsync).

orderer/common/filerepo/filerepo.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+// Scan parses the directory and produce a list of file names, filtered by suffix.
+func (r *Repo) Scan() ([]string, error) {
```
cendhu: Don't we need synchronization between `Scan()`, `Read()`, and `Save()` using a lock? If not, can you clarify how these operations can interleave and still provide consistency?

orderer/common/filerepo/filerepo.go
```diff
@@ -0,0 +1,102 @@
...
+	"bufio"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/hyperledger/fabric/common/configtx"
+)
+
+type Repo struct {
```
cendhu, line 29: Don't we need a `Remove()` method on the Repo? I assume that it is missing because the consumer of this pkg does not need it.

orderer/common/filerepo/filerepo.go
```diff
@@ -0,0 +1,102 @@
...
+	Options Options
+}
+
+type Options struct {
+	FileSuffix          string
```
cendhu: I assume the file suffix as a type of entry stored?

orderer/common/filerepo/filerepo.go
```diff
@@ -0,0 +1,102 @@
...
+	TransientFilePrefix string
```
cendhu: Isn't this field internal to this pkg? Should we need to export it?

orderer/common/filerepo/filerepo.go
```diff
@@ -0,0 +1,102 @@
...
+
+	tmpFile, err := ioutil.TempFile(fileRepoDir, r.Options.TransientFilePrefix+channelName)
+	if err != nil {
+		return err
+	}
+
+	defer func() {
+		if err != nil {
+			tmpFile.Close()
+			os.Remove(tmpFile.Name())
```
cendhu: It doesn't guarantee a removal right? Given that there is an error already, I am not sure whether the removal/sync would even succeed.

orderer/common/filerepo/filerepo.go
tock-ibm: That is correct, we need `Remove` when on boarding is done. I think it should be added.

orderer/common/filerepo/filerepo.go
tock-ibm: I think - leave it exposed but apply a default `~` for when it is not set.

orderer/common/filerepo/filerepo.go
tock-ibm: Thread safety is not just thread races - concurrent access - but also visibility. In the channel participation API, the `Scan` is used by the main go-routine, when initializing the server, `Save` by the http server request go-routine (which probably may spin multiple go-routines if there are multiple requests?), and the `Remove` by the internal go-routine servicing the follower.Chain, one for each chain. Since this is used by multiple threads, I think a thread safe implementation is needed.

orderer/common/filerepo/filerepo.go
```diff
@@ -0,0 +1,102 @@
...
+	FileSuffix          string
+	TransientFilePrefix string
+	FileRepoDir         string
+}
+
+// Save atomically persists the content to the channelName+suffix file by first writing it
+// to a tmp file prefixed by the TransientFilePrefix and then moves the file to the final
+// destination indicated by the FileSuffix.
+func (r *Repo) Save(channelName string, content []byte) error {
+	err := configtx.ValidateChannelID(channelName)
```
tock-ibm: I don't think we need to validate the channel name... channel names are very strict, and this would limit the generality of this package. W.r.t the channel participation API, we check the validity of the channel name when we first process the request, so no need to do it here.

orderer/common/filerepo/filerepo.go
```diff
@@ -0,0 +1,102 @@
...
+	if err != nil {
+		return err
+	}
+
+	fileRepoDir := r.Options.FileRepoDir
+	dest := filepath.Join(fileRepoDir, channelName+r.Options.FileSuffix)
+
+	tmpFile, err := ioutil.TempFile(fileRepoDir, r.Options.TransientFilePrefix+channelName)
```
tock-ibm: I have second thoughts on this... maybe it would be much cleaner to add the transient marker `~` as appended to the  suffix? like:
`mychannel.joinblock` vs. `mychannel.joinblock~` 
This seems to be closer to what many other programs do, and, this way there will never be a confusion with object names (i.e. channel names) starting with a `~`?

orderer/common/filerepo/filerepo.go
tock-ibm: You have to use `Sync()` on the file, I think, to write both data and metadata to disk.

orderer/common/filerepo/filerepo.go
tock-ibm: Plus, why a temp file? as we have the transient marker?

orderer/common/filerepo/filerepo.go
```diff
@@ -0,0 +1,102 @@
...
+// to a tmp file prefixed by the TransientFilePrefix and then moves the file to the final
+// destination indicated by the FileSuffix.
+func (r *Repo) Save(channelName string, content []byte) error {
+	err := configtx.ValidateChannelID(channelName)
+	if err != nil {
+		return err
+	}
+
+	fileRepoDir := r.Options.FileRepoDir
+	dest := filepath.Join(fileRepoDir, channelName+r.Options.FileSuffix)
```
tock-ibm: I think we want to fail early if `dest` above already exists, otherwise the os.Rename() will fail us later.

orderer/common/filerepo/filerepo.go
tock-ibm: This may leave transient files on the dir after crash. When we start using the repo again, after a crash, we should clean all the transient files. In the participation API we use `Scan` first, so that might be a good place to do it (once!). However, I think a better way is to (optionally) make `options` private, and build a constructor that cleans transient files. This way we force the user towards a safe and correct usage pattern.

```go
type Repo struct {
  options Options
}

func NewRepo( options Options) (*Repo, error) {
  //Create the directory if not there
  //Scan and remove transient files
  //Build a Repo and return it
}
```

orderer/common/filerepo/filerepo.go
tock-ibm, line 29: I would consider being more specific about the file permissions:

```go
	repoFilePermPrivateRW os.FileMode = 0600
	repoDirPermPrivateRWX os.FileMode = 0700
```

In addition, consider creating a separate directory for every suffix, under the parent dir. This way the `*.joinblock` and `*.remove` files will be in different sub-directories. This would speed up the scan and avoid filesystem sync issues if we use two file-repo's in the same process with different suffixes. 
Another way to do this is to explicitly define a different file repo dir for different uses in the orderer.yaml, for example.

orderer/common/filerepo/filerepo_test.go
```diff
@@ -0,0 +1,116 @@
...
+	"testing"
+
+	"github.com/golang/protobuf/proto"
+	"github.com/hyperledger/fabric-protos-go/common"
+	"github.com/hyperledger/fabric/orderer/common/filerepo"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestFileRepo_Save(t *testing.T) {
```
tock-ibm: Add a test for saving and loading empty files with a `nil` or empty []byte content, for the `channel.remove` use case.

orderer/common/filerepo/filerepo_test.go
```diff
@@ -0,0 +1,116 @@
...
+		assert.True(t, os.IsNotExist(err))
+	})
+
+	t.Run("Failed save, invalid channel name", func(t *testing.T) {
+		err := r.Save("InvalidChannelName", []byte{})
+		assert.EqualError(t, err, "'InvalidChannelName' contains illegal characters")
+	})
+}
+
+func TestFileRepo_Scan(t *testing.T) {
```
tock-ibm: In the context of the above discussion about cleaning temp files after a crash, add a test case that puts some dummy temp files in the directory and test that the first `Scan` or `NewRepo` cleans them up (I prefer the latter).

orderer/common/filerepo/filerepo.go
```diff
@@ -0,0 +1,199 @@
...
+
+	// Remove existing transient files in the repo
+	for _, f := range files {
+		transientFilePattern := "*" + repo.fileSuffix + repo.transientFileMarker
+		isTransientFile, err := filepath.Match(transientFilePattern, f.Name())
+		if err != nil {
+			return nil, err
+		}
+		if isTransientFile {
+			os.Remove(filepath.Join(repo.fileRepoDir, f.Name()))
```
cendhu: We need to call `sync()` on the parent directory.

orderer/common/filerepo/filerepo.go
```diff
@@ -0,0 +1,194 @@
...
+func New(repoParentDir, fileSuffix string, opts ...RepoOption) (*Repo, error) {
+	// default transientFileMarker to ~, can modify via opts
+	repo := &Repo{
+		transientFileMarker: "~",
+		fileSuffix:          fileSuffix,
+		fileRepoDir:         filepath.Join(repoParentDir, fileSuffix),
+	}
+
+	for _, opt := range opts {
+		if err := opt(repo); err != nil {
```
cendhu: additional response to @sykesm question on why I find this hard to read. 

1. I did go to definition by keeping the cursor on `opt`. Nothing happened.
2. I needed to track to opts and reach `ReadOption`
3. Then, I find that ReadOption is a function pointer and it provides a list of functions which the consumer could have called to get a function implementation that sets a value for a variable. 
4. The returned function implementation is called here. 

I understand that this allow the caller to configure only the required variables (i.e., optional variables) but adds more time to the code readability (& multi-level functions). As I do not see any strong reason for the caller to set `transientFileMarker` (as it is internal to the `filerepo` pkg), I am not convinced that this added code complexity is required. The default `~` looks fine to me. If others have a strong opinion that the added code complexity is justified and `transientFileMarker` needs to be configured by the caller, I will step aside.

orderer/common/filerepo/filerepo.go
```diff
@@ -0,0 +1,169 @@
...
+// All file system operations on the returned file repo are thread safe.
+func New(repoParentDir, fileSuffix string) (*Repo, error) {
+	// Check that fileSuffix doesn't contain the os path separator
+	if err := validateFileSuffix(fileSuffix); err != nil {
+		return nil, err
+	}
+
+	fileRepoDir := filepath.Join(repoParentDir, fileSuffix)
+
+	// Make the file repo if it doesn't already exist
```
cendhu: Nit. Would be good to have comments to explain a decision or reasoning or complex side-effects. Other types of comments are distraction while reading the code.

orderer/common/filerepo/filerepo.go
```diff
@@ -0,0 +1,169 @@
...
+
+	"github.com/hyperledger/fabric/internal/fileutil"
+	"github.com/pkg/errors"
+)
+
+const (
+	repoFilePermPrivateRW os.FileMode = 0600
+)
+
+type Repo struct {
```
cendhu, line 29: Nit. Doc comment for this exported type is missing.

orderer/common/filerepo/filerepo.go
```diff
@@ -0,0 +1,169 @@
...
+
+	// Remove existing transient files in the repo
+	for _, f := range files {
+		transientFilePattern := "*" + fileSuffix + "~"
+		isTransientFile, err := filepath.Match(transientFilePattern, f.Name())
+		if err != nil {
+			return nil, err
+		}
+		if isTransientFile {
+			os.Remove(filepath.Join(fileRepoDir, f.Name()))
```
cendhu: Better to check the err returned by `os.Remove()`

orderer/common/filerepo/filerepo.go
```diff
@@ -0,0 +1,169 @@
...
+
+// FileToBaseName strips the suffix from the file name to get the associated channel name.
+func (r *Repo) FileToBaseName(fileName string) string {
+	baseFile := filepath.Base(fileName)
+
+	return strings.TrimSuffix(baseFile, "."+r.fileSuffix)
+}
+
+func validateFileSuffix(fileSuffix string) error {
+	if len(fileSuffix) <= 0 {
```
cendhu: Nit. as it can never be < 0
```suggestion
	if len(fileSuffix) == 0 {
```

orderer/common/filerepo/filerepo.go
```diff
@@ -0,0 +1,169 @@
...
+	tmpFileName := fileName + r.transientFileMarker
+	if err := fileutil.CreateAndSyncFileAtomically(r.fileRepoDir, tmpFileName, fileName, content, repoFilePermPrivateRW); err != nil {
+		return err
+	}
+
+	return fileutil.SyncDir(r.fileRepoDir)
+}
+
+// Remove removes the file associated with baseName from the file system.
+func (r *Repo) Remove(baseName string) error {
```
cendhu, line 106: Unit test is missing.

orderer/common/filerepo/filerepo.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+
+	files, err := ioutil.ReadDir(fileRepoDir)
+	if err != nil {
+		return nil, err
+	}
+
+	// Remove existing transient files in the repo
+	for _, f := range files {
+		transientFilePattern := "*" + fileSuffix + "~"
```
cendhu: Nit. Can be kept outside the loop instead of creating it newly for each file.

cendhu: Nit. `~` -- till we reach the line 74, the reader cannot understand what this mean. One way to fix this is by using a const variable to keep the default transientFileMarker.

orderer/common/filerepo/filerepo.go
```diff
@@ -0,0 +1,169 @@
...
+
+// Save atomically persists the content to suffix/baseName+suffix file by first writing it
+// to a tmp file marked by the transientFileMarker and then moves the file to the final
+// destination indicated by the FileSuffix.
+func (r *Repo) Save(baseName string, content []byte) error {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	fileName := fmt.Sprintf("%s.%s", baseName, r.fileSuffix)
+	dest := filepath.Join(r.fileRepoDir, fileName)
```
cendhu: Nit. Given that these two lines occurs in three methods on `Repo`, it would be good to move them to a separate method called `baseToFileName()` (which would do an opposite of the existing `FileToBaseName()`)

orderer/common/filerepo/filerepo.go
```diff
@@ -0,0 +1,169 @@
...
+
+// New initializes a new file repo at repoParentDir/fileSuffix.
+// All file system operations on the returned file repo are thread safe.
+func New(repoParentDir, fileSuffix string) (*Repo, error) {
+	// Check that fileSuffix doesn't contain the os path separator
+	if err := validateFileSuffix(fileSuffix); err != nil {
+		return nil, err
+	}
+
+	fileRepoDir := filepath.Join(repoParentDir, fileSuffix)
```
cendhu, line 43: Nit. Using `fileSuffix` both as a sub-directory name and also as a suffix to the file name within the sub directory kind of confuses me while reading the code. I think the fileSuffix denotes the type of content being stored. One such type is the join block. We then create a sub-folder for the type and also add the type as a suffix to each file stored in the sub-folder.

Anyway, I do not have an alternative to the current folder structure that is shown below. Maybe the joinblock suffix looks redundant. 
```
/repo/joinblock
/repo/joinblock/channel1.joinblock
/repo/joinblock/channel2.joinblock
/repo/joinblock/channel3.joinblock~
```

PR #1445, opened 2016-05-11
Fix statetransfer failure on back to back execution

jyellick:
## Description

This changeset adds a test case for multiply invoked statetransfer, as well as fixes the logic of state transfer in the pbft test framework.  It also modifies the behavior of state transfer to more closely match the expectations of its new usage in consensus.  Principally, this means eliminating the `Initiate()` call from state transfer and having state transfer implicitly started whenever a target is provided.
## Motivation and Context

In the significant churn from last week, the state transfer logic was damaged.  The logic was changed to initiate state transfer once, then add targets indefinitely.  Once state transfer completed, this would cause state transfer to never complete again.  This was not captured by the existing tests, because they only tested one instance of state transfer.  The included changeset adds a test which includes back to back state transfer.

Fixes half of #1091 (where state transfer was observed not to recover nearly enough blocks)
## How Has This Been Tested?

Ran the unit and behave tests, added new test for multiple invocations of statetransfer.
## Checklist:

### DIFF
--- a/consensus/consensus.go
+++ b/consensus/consensus.go
@@ -23,8 +23,9 @@ import (
 // Consenter is used to receive messages from the network
 // Every consensus plugin needs to implement this interface
 type Consenter interface {
-	RecvMsg(msg *pb.Message, senderHandle *pb.PeerID) error
-	StateUpdate(tag uint64, id []byte)
+	RecvMsg(msg *pb.Message, senderHandle *pb.PeerID) error // Called serially with incoming messages from gRPC
+	StateUpdated(tag uint64, id []byte)                     // Called when state transfer completes, serial with StateUpdating
+	StateUpdating(tag uint64, id []byte)                    // Called when SkipTo causes state transfer to start serial with StateUpdated
 }
 
 // Inquirer is used to retrieve info about the validating network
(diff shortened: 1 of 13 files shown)


srderson:
@jyellick Should @corecode or @tuand27613 review this?

srderson:
@jyellick There's a merge conflict. Could you rebase on master please? I believe this happened when I merged PR #1277

jyellick:
I believe the test failure is due to the new custody test in #1277 and unrelated to this changeset.  Currently investigating a fix.

corecode:
This is a race condition where sometimes first and then second request are committed, and sometimes second and then first request are committed.  I've been puzzling over a good solution.  If this persists, please revert #1277.

jyellick:
@corecode I'm not sure what you mean by your comment? The test looks for the existence of a single block, I don't see why the request ordering should matter?

This actually looks to me like another bug of modifying internal PBFT state not through the PBFT thread.  Fixed this locally, but still trying to track down why we are seeing multiple view changes.

srderson:
Agree that the Travis failure looks like it was introduced by PR #1277, not the PR. I will manually run behave

srderson:
LGTM. Confirmed behave and unit tests are passing locally.

Issue #1183, opened 2016-04-20
Fix ChaincodeDeploymentSpec and ChaincodeInvocationSpec API


srderson:
The ChaincodeDeploymentSpec and ChaincodeInvocationSpec API as defined in chaincode.proto needs to be reworked. The main issue is that `ChaincodeDeploymentSpec` and `ChaincodeInvocationSpec` both share `ChaincodeSpec` which results in unnecessary fields when when either deploying or invoking a chaincode. Also, some fields such as `ctorMsg` need to be renamed so that they are more clear.

I will post a proposal soon.

ghaskins:
Just a note to capture that chaincodeType is superfluous in Invoke/Query operations, so ideally cleaned up as part of this effort.

adecaro:
Actually, the chaincode type needs to be at the transaction level in order for the crypto layer to work properly.

ghaskins:
@adecaro this isn't obvious to me why this would be.  Could you elaborate?

srderson:
not clear to me either. Seems we should be able to lookup this information from the DB on an invoke/query if it is needed.

adecaro:
@ghaskins @srderson Actually, deploy/execute/query transactions are encrypted in different ways because they contain different information. Then, once the crypto layer has to pre-validate and then pre-exececute a transaction at the validator side, having the transaction type field allows the crypto layer to know quickly how to decrypt the transaction. Without that field, the crypto layer should try all the possible ways to decrypt a transaction to find out which type the transaction is. This is quite inefficient!

Notice that, the type of a transaction cannot be hidden because the size of a transaction leaks this kind of information. So, for efficiency reason I would prefer to have that field there. If you decide anyway to remove that field then the crypto layer needs to be updates as I said above.

ghaskins:
@adecaro ok, we might be talking about different fields.  chaincodeType is what specifies the type of language the chaincode is implemented with (today only "GOLANG" is supported).  It really has no meaning for anything except deploy, as we invoke/query all chaincode types the same way once its running.  It sounds like you are concerned about the transaction type of deploy/invoke/query, not the chaincode type of golang/other/other.  I would imagine the transaction type field should remain present indefinitely .  In light of this new info, is this ok?

adecaro:
@ghaskins Yes, it is fine :)

srderson:
The first pass is changing `ChaincodeSpec.ctorMsg` to `ChaincodeSpec.input`.

[...]

PR #1461, opened 2016-05-12
Rename ChaincodeSpec.ctor to ChaincodeSpec.input

srderson:
## Description

This change renames `ChaincodeSpec.ctor` to `ChaincodeSpec.Input` in `chaincode.proto`. This causes API changes when deploying, invoking, or querying a transaction. The changes are to the REST API, the gRPC API, and the CLI where `ctor` `c` has been renamed to `input` `i`. All documentation has been updated to reflect these changes.
## Motivation and Context

This is part of issue #1183. The goal is to have a cleaner and more understandable API. `ctorMsg` is a leftover from the days when `ChaincodeSpec` was intended to define a chaincode. At some point this was overloaded and also used for invocation. At this point, we have decided to use `ChaincodeSpec` for all elements that are common between `ChaincodeDeploymentSpec` and `ChaincodeInvocationSpec`. Specific elements will be moved to either `ChaincodeDeploymentSpec` or `ChaincodeInvocationSpec`.

While breaking API changes are unfortunate, this change should result in a better user experience over the long run.
## How Has This Been Tested?

The tests have all been run and passed. There are no new tests because there are no feature changes. Some tests were modified to support the new API.
## Checklist:

### DIFF
--- a/bddtests/peer_basic.feature
+++ b/bddtests/peer_basic.feature
@@ -31,7 +31,7 @@ Feature: lanching 3 peers
     Given we compose "docker-compose-1.yml"
       When requesting "/chain" from "vp0"
       Then I should get a JSON response with "height" = "1"
-      When I deploy chaincode "github.com/hyperledger/fabric/examples/chaincode/go/map" with ctor "init" to "vp0"
+      When I deploy chaincode "github.com/hyperledger/fabric/examples/chaincode/go/map" with input "init" to "vp0"
       ||
       ||
 
@@ -90,7 +90,7 @@ Feature: lanching 3 peers
     Given we compose "docker-compose-1.yml"
       When requesting "/chain" from "vp0"
       Then I should get a JSON response with "height" = "1"
-      When I deploy chaincode "github.com/hyperledger/fabric/bddtests/chaincode/go/table" with ctor "init" to "vp0"
+      When I deploy chaincode "github.com/hyperledger/fabric/bddtests/chaincode/go/table" with input "init" to "vp0"
       ||
       ||
       Then I should have received a chaincode name
@@ -319,7 +319,7 @@ Feature: lanching 3 peers
 	    Given we compose "docker-compose-1.yml"
 	    When requesting "/chain" from "vp0"
 	    Then I should get a JSON response with "height" = "1"
-	    When I deploy chaincode "github.com/hyperledger/fabric/examples/chaincode/go/chaincode_example02" with ctor "init" to "vp0"
+	    When I deploy chaincode "github.com/hyperledger/fabric/examples/chaincode/go/chaincode_example02" with input "init" to "vp0"
 		     | arg1 |  arg2 | arg3 | arg4 |
 		     |  a   |  100  |  b   |  200 |
 	    Then I should have received a chaincode name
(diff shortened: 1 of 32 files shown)


ghaskins:
@srderson Thanks for taking this on.  While I recognize that ABI affecting changes should not be taken lightly, I also recognize we currently have some suboptimal representations of things in the spec.  To this point, better to try to get it right as early as possible while we still have the option without requiring compatibility baggage.  I would therefore like to voice my enthusiastic support for this effort.

While I fully agree with everything you are proposing here, I still am a little uncertain about specific things that remain in the spec even after your changes that we should discuss.  For example, I am bothered by the current practice of specifying the chaincodeType in invoke.  It seems to me this only belongs in the deployment.

Assuming we agree that additional changes may be ideal, my question is: should we try to identify and tackle them all at once so we only break the ABI once?  Or should we tackle them as isolated efforts for simplicity at the cost of more red-flag events?

srderson:
I also plan to move `chaincodeType` as you suggest, but was trying to avoid dropping too much at once to make review a bit easer. I agree that we should try to break all at once rather than slowly overtime.

Option 1) You could review this change, but not merge, and then I can add a second commit with the other mentioned change. Then you could review again and merge.

Option 2) I submit a second PR with the other change. You could wait to merge this change until the second one is submitted, but I may need to rebase quickly after this one is merged.

@christo4ferris thoughts on how to handle this case?

srderson:
spoke with @jyellick and he suggested stacking PRs which has worked for him. This GitHub UI doesn't handle this perfectly because the files tab shows all changes in the latest PR, but it seems like the best approach based on his experience.

jyellick:
To expound on what @srderson said, my flow has been:

```
git checkout upstream/master
git checkout -b bigfeature-pr1
# do changes and commits push, submit PR
git checkout -b bigfeature-pr2
# do changes and commits push, submit PR
...
git checkout -b bigfeature-prn
...
```

Then if there ends up being a merge conflict at some point, it's as easy as walking back through the tree

```
git checkout bigfeature-pr2
git rebase upstream/master
git checkout bigfeature-pr3
git rebase bigfeature-pr2
git push -f
git checkout bigfeature-pr4
git rebase bigfeature-pr3
git push -f
...
```

Github PRs seem to handle the `push -f` pretty gracefully.  Though the diffs in the UI won't initially be great, they can always be viewed through the git cli.

ghaskins:
This is my first time using github as the patch review mechanism, so I suspect there will be a little bit of a learning curve and trial+error until we come up with a good workflow.  But FWIW, I've done lots of work on other projects like the Linux kernel that (at least at the time) used more crude methods such as patches mailed to the ML.  In that environment, the email was the "PR", and the one or more patches in the thread were the individual commits for an atomic, individually reviewable change.  This is of course analogous to the patches in the branch that are submitted in the PR.

It seems the same basic principle _should_ work here as long as people adhere a small amount of discipline in how they aggregate the patches.  If people push PRs with willy-nilly merges and incremental changes, this wont work.  However, if people curate a clean, rebased, patch series with the goals of making each patch an individual, sensible, and self-contained changeset, I see no reason this would be difficult for reviewers to see the breakdown:  Yes, the top "Files Changed" tab shows an aggregate of the series, but its straight forward to review individual patches as well I have found.

Personally, I use stgit to help curate series in this manner, but there are other workflows that work just as well.

This is the round-about way of saying "Option 1", with the intent of spurring a dialog regarding general methods going forward.  Thoughts?

binhn:
I think option 1 would likely be less work for submitters while option 2 would be easier for reviewers but high potential of merge conflicts. So I think we should go with option 1.

christo4ferris:
@binhn why optimize for the submitter?

binhn:
@christo4ferris Hope to encourage more contribution. Resolving merge conflicts some time takes a lot of time and some time has to be done manually, so option 1 is to hopefully avoid that possibility.

ghaskins:
@srderson We should ideally get this cleanup done ASAP.  How would you like to proceed?

srderson:
Closing. Will be optimized as part of V2

PR #1469, opened 2016-05-13
Fix spurious TestBatchCustody test failure

corecode:
## Description

Do not test for the state of the byzantine node.
## Motivation and Context

`TestBatchCustody` configures replica 0 to only receive messages, but to be unable to send messages.  In a sense, this is byzantine behavior.  Depending on message order, replica 0 may apply all blocks, or it may stay in a view change and not apply the transactions.  Given that replica 0 can be considered byzantine, we should not expect any specific state.  Therefore we should ignore its state when considering whether the test failed or was successful.

This test issue had been introduced in #1277.
## How Has This Been Tested?

Repeated unit tests, bdd tests.
## Checklist:

### DIFF
--- a/consensus/obcpbft/obc-batch_test.go
+++ b/consensus/obcpbft/obc-batch_test.go
@@ -102,7 +102,11 @@ func TestBatchCustody(t *testing.T) {
 	time.Sleep(6 * time.Second)
 	net.stop()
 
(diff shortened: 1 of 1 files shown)


srderson:
LGTM. Thanks

Issue #936, opened 2016-04-01
Chaincode shim `GetState` routine should return an indication of whether the key was found in the database


bcbrock:
If a chaincode calls `stub.GetState` with a key that does not exist, no error or presence indication is returned. This can lead to crashes if the chaincode was expecting the key to be present and non-empty. Since the empty byte array is valid data this can not be used as a check for the presence of of key. It seems that the "Go way" to handle this would be to include a 3rd, 'ok' return value separate from the error return that would be set true if and only if the key were actually found in the database.

ibmmark:
@srderson Can you add help wanted to this?

zmanian:
I've encountered this as a challenge in writing chaincode. 

For most applications protocols, I've been experimenting treating an array of len(0) as the equivalent to "key not found" has been worked fine.

The API is ambiguous on first encounter and I believe it will lead to errors.

I like the @bcbrock suggestion the third return value.

bcbrock:
@srderson - I'd like to volunteer to address this issue. I need to learn how the state database works for my research.

mastersingh24:
@bcbrock - are you on this one?

bcbrock:
Yes

murali-katipalli-dtcc:
I can help. Can you assign the issue to me

bcbrock:
I can't assign it, but I understand you are going to do it.

murali-katipalli-dtcc:
Would returning the third value be the appropriate way? 
If we do return the third value, wouldn't it break existing GetState API. And returning multiple error return types (error and ok) doesnt seem to be the 'Go' way of doing it.
Instead can we do the following -  When key is not found, shim returns as an Error. Client can check on that error and take appropriate action. This way we can maintain our existing API and provide the requisite functionality.
If we choose this way, should we define GetStateError type? or should we keep it simple where it just returns the generic Error with appropriate message.

Pl. advice and I can implement it accordingly

Murali (DTCC)

bcbrock:
@murali-katipalli-dtcc I'm personally not worried about changing the API. This is a very new project that has not published a single release. The compiler will flag the error, people will update their chaincodes and move on. In fact the 3-return-value way will ensure that people really understand and code for the new semantics because the compiler won't allow them to continue until they have acknowledged the change in their code.

[...]

PR #1471, opened 2016-05-13
Empty array support

manish-sethi:
This PR allows storing an empty array as a value of a key in the ledger. However, an attempt to store a 'nil' would result in an error - see issue# 936

### DIFF
--- a/core/ledger/ledger.go
+++ b/core/ledger/ledger.go
@@ -207,6 +207,9 @@ func (ledger *Ledger) GetStateRangeScanIterator(chaincodeID string, startKey str
 
 // SetState sets state to given value for chaincodeID and key. Does not immideatly writes to DB
 func (ledger *Ledger) SetState(chaincodeID string, key string, value []byte) error {
+	if key == "" || value == nil {
+		return fmt.Errorf("A nil key or a nil value is not supported. key=%s, value=%#v", key, value)
+	}
 	return ledger.state.Set(chaincodeID, key, value)
 }
 
--- a/core/ledger/ledger.go
+++ b/core/ledger/ledger.go
@@ -40,8 +40,8 @@ var ledgerLogger = logging.MustGetLogger("ledger")
 type ErrorType string
 
 const (
-	//ErrorTypeWrongInput used to indicate the wrong input to ledger method
-	ErrorTypeWrongInput = ErrorType("WrongInput")
+	//ErrorTypeInvalidArgument used to indicate the invalid input to ledger method
+	ErrorTypeInvalidArgument = ErrorType("InvalidArgument")
 	//ErrorTypeOutOfBounds used to indicate that a request is out of bounds
 	ErrorTypeOutOfBounds = ErrorType("OutOfBounds")
 	//ErrorTypeResourceNotFound used to indicate if a resource is not found
@@ -238,7 +238,7 @@ func (ledger *Ledger) GetStateRangeScanIterator(chaincodeID string, startKey str
 // SetState sets state to given value for chaincodeID and key. Does not immideatly writes to DB
 func (ledger *Ledger) SetState(chaincodeID string, key string, value []byte) error {
 	if key == "" || value == nil {
-		return newLedgerError(ErrorTypeWrongInput,
+		return newLedgerError(ErrorTypeInvalidArgument,
 			fmt.Sprintf("An empty string key or a nil value is not supported. Method invoked with key='%s', value='%#v'", key, value))
 	}
 	return ledger.state.Set(chaincodeID, key, value)
--- a/core/ledger/ledger.go
+++ b/core/ledger/ledger.go
@@ -18,7 +18,6 @@ package ledger
 
 import (
 	"bytes"
-	"errors"
 	"fmt"
 	"reflect"
 	"sync"
@@ -37,12 +36,43 @@ import (
 
 var ledgerLogger = logging.MustGetLogger("ledger")
 
+//ErrorType represents the type of a ledger error
+type ErrorType string
+
+const (
+	//ErrorTypeWrongInput used to indicate the wrong input to ledger method
+	ErrorTypeWrongInput = ErrorType("WrongInput")
+	//ErrorTypeOutOfBounds used to indicate that a request is out of bounds
+	ErrorTypeOutOfBounds = ErrorType("OutOfBounds")
+	//ErrorTypeResourceNotFound used to indicate if a resource is not found
+	ErrorTypeResourceNotFound = ErrorType("ResourceNotFound")
+)
+
+//Error can be used for throwing an error from ledger code.
+type Error struct {
+	errType ErrorType
+	msg     string
+}
+
+func (ledgerError *Error) Error() string {
+	return fmt.Sprintf("LedgerError - %s: %s", ledgerError.errType, ledgerError.msg)
+}
+
+//Type returns the type of the error
+func (ledgerError *Error) Type() ErrorType {
+	return ledgerError.errType
+}
+
+func newLedgerError(errType ErrorType, msg string) *Error {
+	return &Error{errType, msg}
+}
+
 var (
 	// ErrOutOfBounds is returned if a request is out of bounds
-	ErrOutOfBounds = errors.New("ledger: out of bounds")
+	ErrOutOfBounds = newLedgerError(ErrorTypeOutOfBounds, "ledger: out of bounds")
 
 	// ErrResourceNotFound is returned if a resource is not found
-	ErrResourceNotFound = errors.New("ledger: resource not found")
+	ErrResourceNotFound = newLedgerError(ErrorTypeResourceNotFound, "ledger: resource not found")
 )
 
 // Ledger - the struct for openchain ledger
@@ -208,7 +238,8 @@ func (ledger *Ledger) GetStateRangeScanIterator(chaincodeID string, startKey str
 // SetState sets state to given value for chaincodeID and key. Does not immideatly writes to DB
 func (ledger *Ledger) SetState(chaincodeID string, key string, value []byte) error {
 	if key == "" || value == nil {
-		return fmt.Errorf("A nil key or a nil value is not supported. key=%s, value=%#v", key, value)
+		return newLedgerError(ErrorTypeWrongInput,
+			fmt.Sprintf("An empty string key or a nil value is not supported. Method invoked with key='%s', value='%#v'", key, value))
 	}
 	return ledger.state.Set(chaincodeID, key, value)
 }
--- a/core/ledger/ledger_test.go
+++ b/core/ledger/ledger_test.go
@@ -915,11 +915,41 @@ func TestLedgerEmptyArrayValue(t *testing.T) {
 
 	value, _ := l.GetState("chaincodeID1", "key1", true)
 	if value == nil || len(value) != 0 {
-		t.Fatal("An empty array expected in value. Found = %#v", value)
+		t.Fatalf("An empty array expected in value. Found = %#v", value)
 	}
 
 	value, _ = l.GetState("chaincodeID1", "non-existing-key", true)
 	if value != nil {
-		t.Fatal("A nil value expected. Found = %#v", value)
+		t.Fatalf("A nil value expected. Found = %#v", value)
 	}
+}
+
+func TestLedgerWrongInput(t *testing.T) {
+	ledgerTestWrapper := createFreshDBAndTestLedgerWrapper(t)
+	l := ledgerTestWrapper.ledger
+	l.BeginTxBatch(1)
+	l.TxBegin("txUUID")
+
+	// nil value input
+	err := l.SetState("chaincodeID1", "key1", nil)
+	ledgerErr, ok := err.(*Error)
+	if !(ok && ledgerErr.Type() == ErrorTypeWrongInput) {
+		t.Fatal("A 'LedgerError' of type 'ErrorTypeWrongInput' should have been thrown")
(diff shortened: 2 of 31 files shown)


corecode:
Commit 715b2ca095e7b651da7aabd39044a0657e627992 is a mechanical pass over several files and should be labeled as such.  This commit is also independent of the contents of the PR.

core/ledger/ledger.go
```diff
@@ -207,6 +207,9 @@ func (ledger *Ledger) GetStateRangeScanIterator(chaincodeID string, startKey str
 
 // SetState sets state to given value for chaincodeID and key. Does not immideatly writes to DB
 func (ledger *Ledger) SetState(chaincodeID string, key string, value []byte) error {
+	if key == "" || value == nil {
```
corecode, line 240: why is a nil key or value not supported?

manish-sethi:
@corecode, when I ran goimports etc. tools, it picked up existing files also. May be, I should have run these tools only on selective files that were originally changed in this PR. Is that how you suggest?

core/ledger/ledger.go
manish-sethi, line 240: This was never supported and tested. I have added the check explicitly here to avoid issues down stream. We can add support for these later but that requires changes down stream.

corecode:
I think somebody should run goimports for the whole repo, do one monster commit (making sure that chaincode hashes are fixed appropriately), and then have CI to fail if commits have not been passed through goimports.

binhn:
@corecode we probably should take goimports related discussion on another issue. A lot of problems reported are on public function comments, which I think would be more appropriate for the authors to do.

What are the use-cases for empty or nil keys?

mastersingh24:
I agree with @binhn  - why do we need empty or nil keys and/or valuies?

bcbrock:
I've made the case for empty values in #936. As for empty keys: The key is specified to be a string, "" is a unique string, so it should be accepted as a key. From a specification point of view there is no reason to disallow it; there is no ambiguity. As far as I can tell RocksDB allows the empty string as a key. Also compare specifications of other modern key/value stores like Redis

```
Redis keys are binary safe, this means that you can use any binary sequence as a key, 
from a string like "foo" to the content of a JPEG file. The empty string is also a valid key.
```

I've seen examples where keys are stored as fields of tables. If a field is "" it indicates a default value, which is represented by a key/value pair ""/<default>.

I think that `PutState` from Go should allow empty keys, empty values and `nil` values (see #936), so I think @corecode comment above is correct, and the test at `ledger.go:210` should be eliminated.

I don't understand the hesitation about downstream changes. Why wait to fix issues related to handling empty strings as keys? Why not fix them now?

manish-sethi:
@bcbrock - For making the changes, we all need to agree on what we want to support specially about the 'nil' values.

binhn:
@bcbrock That's not a use-case. I know and understand that generic key-value DBs support empty keys, but what use-case would a blockchain app use empty (or nil) keys? Fabric is an app of RocksDB, but that doesn't mean it has to support all the capabilities and features of RocksDB.

bcbrock:
@binhn I gave a use case based on indirection of keys stored in a table:

_I've seen examples where keys are stored as fields of tables. If a field is "" it indicates a default value, which is represented by a key/value pair ""/<default>._

Is there some fundamental difficulty with allowing the empty string as a key?

srderson:
Can we agree this PR is a step in the right direction?

@bcbrock I think @manish-sethi answered your question at https://github.com/hyperledger/fabric/pull/1471/files#r63202854. Empty keys are broken today, so we respond with an error. This does not prevent us from adding support in a future PR.

binhn:
@bcbrock thanks for clarifying though I still see that as marginal use-case

@srderson I agreed to merge this pr and address empty key in a separate pr/issue

core/ledger/ledger_test.go
```diff
@@ -902,3 +902,24 @@ func TestCopyState(t *testing.T) {
 	values, _ := l.GetStateMultipleKeys("chaincodeID2", []string{"key1", "key2", "key3"}, true)
 	testutil.AssertEquals(t, values, [][]byte{[]byte("value1"), []byte("value2"), []byte("value3")})
 }
+
+func TestLedgerEmptyArrayValue(t *testing.T) {
```
srderson, line 906: Could you add tests to confirm the error is returned on a empty key or a nil value?

core/ledger/ledger.go
```diff
@@ -207,6 +207,9 @@ func (ledger *Ledger) GetStateRangeScanIterator(chaincodeID string, startKey str
...
+		return fmt.Errorf("A nil key or a nil value is not supported. key=%s, value=%#v", key, value)
```
srderson: This error should specify an empty key instead of nil.

core/ledger/ledger_test.go
manish-sethi, line 906: Added a test for wrong input (last test in ledger_test.go) in commit aa93cf4

core/ledger/ledger.go
```diff
@@ -37,12 +36,43 @@ import (
 
 var ledgerLogger = logging.MustGetLogger("ledger")
 
+//ErrorType represents the type of a ledger error
+type ErrorType string
+
+const (
+	//ErrorTypeWrongInput used to indicate the wrong input to ledger method
+	ErrorTypeWrongInput = ErrorType("WrongInput")
```
srderson: Minor issue, but can we use `InvalidArgument` instead of `WrongInput`?

manish-sethi: fixed in commit 2d272c8

srderson:
LGTM. We will potentially address other issues (support for empty key and nil value) in a separate PRs.

integration/nwo/network.go
```diff
@@ -413,6 +413,26 @@ func (n *Network) userCryptoDir(org *Organization, nodeOrganizationType, user, c
...
+func (n *Network) PeerOrgCADir(o *Organization) string {
+	return filepath.Join(
+		n.CryptoPath(),
+		"peerOrganizations",
+		o.Domain,
+		"ca",
+	)
+}
+
+func (n *Network) OrdererOrgCADir() string {
```
sykesm: Nit: godoc

integration/nwo/network.go
```diff
@@ -413,6 +413,26 @@ func (n *Network) userCryptoDir(org *Organization, nodeOrganizationType, user, c
 	)
 }
 
+// PeerOrgCADir returns the path to the folder containing the CA certificate(s) and/or
+// keys for the specified organization
```
sykesm: nit: complete sentences in godoc

integration/pkcs11/p11key.go
```diff
@@ -0,0 +1,64 @@
...
+	"crypto/ecdsa"
+	"encoding/asn1"
+	"fmt"
+	"io"
+	"math/big"
+
+	"github.com/miekg/pkcs11"
+)
+
+// P11ECDSAKey test implementation of crypto.Signer
```
sykesm: nit: throughout, any godoc should be complete sentences.

integration/pkcs11/p11key.go
```diff
@@ -0,0 +1,64 @@
...
+	privateKeyHandle pkcs11.ObjectHandle
+}
+
+// Public returns the corresponding public key for the
+// private key
+func (k *P11ECDSAKey) Public() crypto.PublicKey {
+	return k.publicKey
+}
+
+// Sign returns a crypto.Signer
```
sykesm: This implements crypto.Signer; it doesn't return one.

integration/pkcs11/p11key.go
```diff
@@ -0,0 +1,64 @@
...
+func (k *P11ECDSAKey) Sign(rand io.Reader, digest []byte, opts crypto.SignerOpts) (signature []byte, err error) {
+	hash := opts.HashFunc()
+	hashLen := hash.Size()
+	if len(digest) != hashLen {
```
sykesm: Why the temporary vars that are not referenced beyond this block instead of?
```go
if len(digest) != opts.HashFunc().Size() {
...
}
```

integration/pkcs11/pkcs11_suite_test.go
```diff
@@ -0,0 +1,53 @@
...
+	. "github.com/onsi/gomega"
+)
+
+func TestPKCS11(t *testing.T) {
+	RegisterFailHandler(Fail)
+	lib, pin, label := bpkcs11.FindPKCS11Lib()
+	if lib != "" && pin != "" && label != "" {
+		RunSpecs(t, "PKCS11 Suite")
+	}
+}
```
sykesm: So, I think it's preferred to explicitly state that the suite is being skipped instead of just omitting the call to `RunSpecs`. Since this gets the `*testing.T` we can simply call `Skip(...)` on that.

So flip the test and call skip when any of the args is nil?

integration/pkcs11/pkcs11_test.go
```diff
@@ -0,0 +1,411 @@
...
+				Library:  lib,
+			},
+		}
+
+		By("update bccsp peer config")
+		peerConfig := network.ReadPeerConfig(peer)
+		peerConfig.Peer.BCCSP = myBCCSP
+		network.WritePeerConfig(peer, peerConfig)
+
+		By("Update bccsp orderer config")
```
sykesm: subject-verb agreement

integration/pkcs11/pkcs11_test.go
```diff
@@ -0,0 +1,411 @@
...
+		ordererConfig := network.ReadOrdererConfig(orderer)
+		ordererConfig.General.BCCSP = myBCCSP
+		network.WriteOrdererConfig(orderer, ordererConfig)
+
+		network.Bootstrap()
+
+		ctx, sess := setupPKCS11Ctx(lib, label, pin)
+		defer ctx.CloseSession(sess)
+		Expect(err).NotTo(HaveOccurred())
```
sykesm: The order of these are switched.

integration/pkcs11/pkcs11_test.go
```diff
@@ -0,0 +1,411 @@
...
+		Eventually(process.Wait(), network.EventuallyTimeout).Should(Receive())
+
+		if network != nil {
+			network.Cleanup()
+		}
+		os.RemoveAll(tempDir)
+	})
+
+	It("Creates and join channel successfully", func() {
+
```
sykesm: nit: blank lines

integration/pkcs11/pkcs11_test.go
```diff
@@ -0,0 +1,411 @@
...
+		orderer := network.Orderer("orderer")
+		network.CreateAndJoinChannels(orderer)
+
+	})
+
```
sykesm: nit: blank line

integration/pkcs11/pkcs11_test.go
```diff
@@ -0,0 +1,411 @@
...
+})
+
+func configurePeerPKCS11(ctx *pkcs11.Ctx, sess pkcs11.SessionHandle, network *nwo.Network) {
+
+	peer := network.Peers[0]
```
sykesm: Why just one peer?

integration/pkcs11/pkcs11_test.go
```diff
@@ -0,0 +1,411 @@
...
+
+		orderer := network.Orderer("orderer")
+		network.CreateAndJoinChannels(orderer)
+
+	})
+
+})
+
+func configurePeerPKCS11(ctx *pkcs11.Ctx, sess pkcs11.SessionHandle, network *nwo.Network) {
+
```
sykesm, line 116: nit: blank line

integration/pkcs11/pkcs11_test.go
```diff
@@ -0,0 +1,411 @@
...
+	peer := network.Peers[0]
+	orgName := peer.Organization
+
+	peerPubKey, peerCSR, peerSerial := createCSR(ctx, sess, orgName, "peer")
+	adminPubKey, adminCSR, adminSerial := createCSR(ctx, sess, orgName, "admin")
+
+	// Retrieves org CA cert
+	orgCAPath := network.PeerOrgCADir(network.Organization(orgName))
+	caBytes, err := ioutil.ReadFile(filepath.Join(orgCAPath, "ca.org1.example.com-cert.pem"))
```
sykesm: this is assuming that org1 and peer0 are together, isn't it?
Also, the organization domain should be resolved by calling `network.Organization(orgName)` and getting the `Domain` field.

integration/pkcs11/pkcs11_test.go
```diff
@@ -0,0 +1,411 @@
...
+	Expect(err).NotTo(HaveOccurred())
+
+	By("Updating the peer signcerts")
+
```
sykesm, line 126: nit: more blank lines.

(I'm cool with vertical whitespace between functional blocks of code but when there's a lot of it, it actually makes things harder to read.)

integration/pkcs11/pkcs11_test.go
```diff
@@ -0,0 +1,411 @@
...
+	orgAdminMSPPath := network.PeerUserMSPDir(peer, "Admin")
+	updateMSPFolder(orgAdminMSPPath, "Admin@org1.example.com-cert.pem", newAdminPemCert)
+
+}
+
+func configureOrdererPKCS11(ctx *pkcs11.Ctx, sess pkcs11.SessionHandle, network *nwo.Network) {
+
+	orderer := network.Orderer("orderer")
+	orgName := "orderer"
```
sykesm: the orderer org should be obtained from the org.

Issue #1466, opened 2016-05-13
behave test scenarios issue_#680 and issue_#1000 fail with latest commit tag


gongsu832:
```
commit 8b6eb2ff1f253d3a3546d99d1ff9f6a79442192c
Merge: 509b2d6 4e19fa4
Author: Sheehan Anderson <sranderson@gmail.com>
Date:   Thu May 12 15:13:12 2016 -0400

    Merge pull request #1463 from novusopt/issue-1462

    Use new peer command in the docker-compose files
```

Issue_#1000 passes when I increase wait time from 60 seconds to 240 seconds.

Issue_#680 is a bit more interesting. When I increase wait time from 60 seconds to 240 seconds, @1.1  passes but @1.2 and @1.3 fail. When I increase wait time to 480 seconds, @1.1 and @1.2 pass but @1.3 still fails. Maybe if I increase the wait time further and @1.3 will eventually pass. But I tend to think that the slowness isn't due to the machine but rather some inherent problem with the protocol that warrants a closer look. I'm attaching the test run log, as well as logs of the containers, for both 240 second and 480 second runs.

[behave_issue_680_240sec.zip](https://github.com/hyperledger/fabric/files/262571/behave_issue_680_240sec.zip)
[behave_issue_680_480sec.zip](https://github.com/hyperledger/fabric/files/262572/behave_issue_680_480sec.zip)

gongsu832:
After moving to a machine with 32GB memory, even with the default 60 seconds wait time, I no longer get timeouts for issue_680. Instead I'm getting indexError. All logs attached (note in this particular run @1.1 still had a timeout).

[behave_issue_680_60sec.zip](https://github.com/hyperledger/fabric/files/263840/behave_issue_680_60sec.zip)

gongsu832:
Per Tuan's request, log files for issue_1000 behave test with @doNotDecompose uncommented.

[behave_issue_1000_60sec.zip](https://github.com/hyperledger/fabric/files/263868/behave_issue_1000_60sec.zip)

tuand27613:
PR #1476 fixes the indexErrors problem

gongsu832:
I can confirm that with the patch I no longer get an IndexError for issue_1000 @1.1 behave test. Instead, I'm getting a timeout, which is still being investigated.

tuand27613:
running on @gongsu832 's machine,  I'm consistently seeing a timeout in the @issue_1000 test case during the batch run and not the classic run.

[...]

PR #1476, opened 2016-05-13
correctly parse docker output from z debian systems in behave functions

tuand27613:
## Description
- Correctly parse docker-compose output to pull container names when stopping/restarting peers. 
- add timestamp when printing out REST calls
## Motivation and Context

on certain z systems running debian ( @gongsu832 ) , we get an indexError message when running testcase @issue1000.
This is because when behave uses docker-compose to stop and restart the peers as part of the test, docker will return an output string with a different format from what is returned in a Vagrant/Ubuntu system. Behave will fail as it cannot parse the container names correctly.
This fixes an interim problem while investigating #1466
## How Has This Been Tested?

`behave` tests run on vagrant and z systems
## Checklist:

### DIFF
--- a/bddtests/steps/peer_basic_impl.py
+++ b/bddtests/steps/peer_basic_impl.py
@@ -49,9 +49,16 @@ def parseComposeOutput(context):
     containerNamePrefix = os.path.basename(os.getcwd()) + "_"
     containerNames = []
     for l in context.compose_error.splitlines():
-    	print(l.split())
-    	containerNames.append(l.split()[1])
+        tokens = l.split()
+        print(tokens)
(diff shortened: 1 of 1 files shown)


srderson:
@gongsu832 could you add a comment confirming that this fixes the issue you saw? It LGTM, but I do not have a Z Debian system for testing the PR.

gongsu832:
I can confirm that with the patch I no longer get an IndexError for
issue_1000 @1.1 behave test. Instead, I'm getting a timeout, which is still
being investigated.

srderson:
Thanks @gongsu832, LGTM

Issue #1280, opened 2016-04-27
Deploying invalid chaincode gives unclear error message in logs and response


srderson:
I deployed a chaincode with an invalid interface (missing the Init function) and the log message printed

```
ERR - 21:04:53.383 [rest] ProcessChaincode -> INFO 04c[0m REST processing chaincode request...
ERR - 21:04:53.384 [rest] processChaincodeDeploy -> INFO 04d[0m REST deploying chaincode...
ERR - 21:04:53.384 [rest] processChaincodeDeploy -> INFO 04e[0m Local user 'user_type0_1c2ce7506d' is already logged in. Retrieving login token.
ERR - [31m21:04:54.995 [devops] getChaincodeBytes -> ERRO 04f[0m Error getting chaincode package bytes: Error getting code process done with error = exit status 2
ERR - [31m21:04:54.995 [devops] Deploy -> ERRO 050[0m Error deploying chaincode spec: type:GOLANG chaincodeID: ctorMsg: secureContext:"user_type0_1c2ce7506d" 
ERR - 
ERR -  error: Error getting chaincode package bytes: Error getting code process done with error = exit status 2
ERR - [31m21:04:54.996 [rest] processChaincodeDeploy -> ERRO 051[0m Error when deploying chaincode: Error getting chaincode package bytes: Error getting code process done with error = exit status 2
ERR - 21:04:54.996 [rest] ProcessChaincode -> INFO 052[0m REST sucessfully deploy chaincode: {"jsonrpc":"2.0","error":{"code":-32001,"message":"Deployment failure","data":"Error when deploying chaincode: Error getting chaincode package bytes: Error getting code process done with error = exit status 2"},"id":1}
```

This made me think it couldn't reach the chaincode in GitHub. Ideally it would have logged a clear message indicating
1) That it was an invalid chaincode
2) Why it was invalid
3) The response to the client contained the unclear error message, but was a 200 instead of a 500

muralisrini:
High level notes - 

The error "process done with error" stems from "getCodeFromHTTP" in https://github.com/hyperledger/fabric/blob/master/core/chaincode/platforms/golang/hash.go. Investigation would likely be around scraping output (stdout or stderr) from the preceding "go get" exec command and returning it as Error

spomytkin:
I can help

christo4ferris:
@spomytkin go to https://github.com/hyperledger and accept the invite so that @srderson can assign you

mastersingh24:
[...]

PR #1477, opened 2016-05-14
Fix for Issue #1280 "Deploying invalid chaincode gives unclear error"

spomytkin:
Fix for Issue #1280 
## Description

```
per conversation with Sheehan Anderson  changed message to the:
"go get command failed in getCodeFromHTTP.  This may be an issue with
the network or compilation"
```
## How Has This Been Tested?

deployments of  chaincode via co-located peer process in latest vagrant environment
- from local filesystem: valid& compilable, non-compilable, missing 
- from http:valid& compilable, non-compilable, missing
## Checklist:

Signed-off-by: Sergey Pomytkin  spomytkin@gmail.com


christo4ferris:
@srderson might be worthwhile cleaning up all of the error handling/logging here. none of the error messages are likely clear because they lack context. also, why aren't we using go-logging for this? Seems like we should be preparing for use of syslog or similar interface for collecting logs when we run this in production at scale.

Otherwise, LGTM

muralisrini:
@spomytkin : the following is what I meant by scraping output. Key is to extract the string from the bytes.  If it works it should provide readable context to help users turnaround chaincode errors quickly. 

@christo4ferris : the error from this function does get go-logged up the stack. But agree we need a coherent approach to error handling.

```
        // Use a 'go get' command to pull the chaincode from the given repo
        cmd := exec.Command("go", "get", path)
        ...
        var errBuf bytes.Buffer
        cmd.Stderr = &errBuf //capture Stderr and print it on error
        err = cmd.Start()
         ...
         ...
        select {
         ....
        case err = <-done:
                // If we're here, the 'go get' command must have finished
                if err != nil {
                        err = fmt.Errorf("'go get' failed with error\n\"%s\"\n", err, string(errBuf.Bytes()))
                }
        }
```

Sample output from the above

image

srderson:
@muralisrini ah, @spomytkin showed the me the bytes, did not realize there was a useful string hidden in there. @spomytkin - would it be possible to convert the bytes to string a @muralisrini suggests?

spomytkin:
Yes. I'll update the code to include converted string

spomytkin:
did the change, tested and created new pull reuest: https://github.com/hyperledger/fabric/pull/1506
Should I just close this one?

srderson:
@spomytkin Thanks! Yeah, I'll close this one.

PR #1482, opened 2016-05-16
Multi rootnode support

yacovm:
### 2nd attempt after CR comments by @corecode

## Description

All peers can now start with an array of root nodes instead of a single one, for example (from core.yaml): 

```
    discovery:

        # The root nodes are used for bootstrapping purposes, and generally
        # supplied through ENV variables
        # It can be either a single host or a comma separated list of hosts.
        rootnode: 172.17.0.2,172.17.0.3
```

This removes the SPOF of having a peer depend on a single root node for joining the cluster.
The original GetRootNode() method is not present anymore and instead is replaced by GetRandomNode() which returns a random root node. 
- In addition, NVP nodes now query a random validating peer out of the root node list instead of the same one each time
- In terms of discovery process, validating peers now connect to all root nodes at startup, and non-validating peers seek a servicing peer to connect to.
## How Has This Been Tested?

- Ran the unit tests (go test -timeout=20m $(go list github.com/hyperledger/fabric/... | grep -v /vendor/ | grep -v /examples/))
- Behave tests ran by Travis
## Checklist:


jeffgarratt:
@yacovm @srderson Can you give me a bit of time to review this PR?

core/discovery_test.go
```diff
@@ -16,14 +16,81 @@ limitations under the License.
...
+	}
+}
+
+func TestDiscovery_GetSinglePeer(t *testing.T) {
+	assertRandomRootNode(t, "someHost", NewStaticDiscovery("someHost"))
+}
+
+func TestDiscovery_GetAllPeers(t *testing.T) {
+	s := "a,b,c,d,e,f,g,h,i,j,k,l,m,n,o,p,q,r,s,t,u,v,w,x,y,z"
+	discovery := (NewStaticDiscovery(s))
```
corecode: extra `()`

corecode:
nice!

ghaskins:
@jeffgarratt Is your review still outstanding on this or are you ok with it going in?

jeffgarratt:
@ghaskins I have concerns about the concept of load balancing in the non-validating peer.  I think the concept of multi-rootnode is sound, but not sure that we would want to load balance in this scenario.  Particularly since TXs require multiple calls to verify.

yacovm:
Can you elaborate on the multiple calls for verifying TXs? Who's verifying it, and how? I didn't see multiple gRPC calls with the NVP (when I forked the code, I don't know if it changed).

BTW- maybe one can think of a more complex approach- to load balance with a "time" based policy- if the last discovery lookup was made more than T seconds ago, apply random logic and otherwise resolve the same VP you resolved for the last call. What do you think?

jeffgarratt:
@yacovm @binhn Trying to understand the need for chatTokens?

yacovm:
@jeffgarratt , the chat tokens is an implementation detail, nothing more.
It's practically a semaphore that ensures that chat will be established:
- Only to one VP, if you're an NVP, but it iterates over all VPs until a servicing one is found.
- To all VPs, if you're a VP, since you need to chat with everyone anyway. 

I did it this way so the change will be minimal and the logic flow will be the same.

bmandler:
The gist of this PR is to support an array of rootnodes rather than a single one, to avoid such a single point of failure. Having a single rootnode means that a peer coming up cannot join the cluster if the specified rootnode happens to be down at the moment. Furthermore, currently an NVP wishing to communicate with a VP will only do so via the single rootnode defined for it, thus if that rootnode is down, the NVP cannot communicate with the VPs. The change is about having an array of rootnodes rather than a single one. 
As for the load balancing aspect, at the moment it amounts to having an NVP choose randomly from among the array of rootnodes the VP it shall communicate with. If this proves to be a problem and there's a need for an NVP to always talk to the same VP, we can do that by communicating with a designated VP and switch to another VP only in the case that the designated one has failed

ghaskins:
My thoughts are: overall, I think that the concept of having multiple seed nodes is a good idea.  I will point out that I think eventually we need to migrate to a design where the yaml config is only secondary to the whitelist that we store on/alongside the ledger.  Therefore, the importance of the rootnode declaration would only come into play when the whitelist is unhelpful (such as when it doesn't exist i.e. genesis/pow or has become so stale that the current set of whitelisted members have since gone dark).  In that scenario, having multiple nodes that can be used as seeds is a positive advancement, so thank you for this patch.

yacovm:
Hi @ghaskins , I confirm the comment about the white-listing, and I think that the discovery interface in the patch can be extended to support this quite easily.

jeffgarratt:
@yacovm @ghaskins discussed with @yacovm .  I think this PR is good to go after a rebase.

yacovm:
Reopened as: https://github.com/hyperledger/fabric/pull/1686

PR #1485, opened 2016-05-16
Fix PBFT request custody bugs

jyellick:
## Description

This changeset moves new view processing for batch onto the batch thread, clears the pbft core outstanding request map on view change (as it is maintained by custody now), retains complained requests until explicit removal, and uses the PBFT thread via function injection to initiate view change from complaints.
## Motivation and Context

After the failures in `TestBatchCustody` observed after the merging of PR #1277, it was clear that the `TestBatchCustody` needs to run faster, and more deterministically.  Additionally, to enhance this test, the terminal view for the test was added, which exposed other bugs in the code.
1. The batch thread was being used to invoke `op.pbft.sendViewChange()` which modifes the internal PBFT state, possibly causing data corruption, or even concurrent access panics.
2.  The `outstandingReqs` map in PBFT would complain that a request had not been executed, triggering unnecessary view changes.  This was wrong, because a batch request is a collection of requests, and although a particular batch is not executed does not mean that its component requests have not been.
3. The new primary was only processing the requests for which it was the original receiver, but was also discarding the complaints it had received in a previous view.  This required yet another round of complaints which slows recovery.
4. Once a request in custody's timer had expired, it was removed from custody, and a view change was initiated.  Once view change completed, the request was no longer available to be resubmitted.

All of these problems could be observed in the view being inappropriately advanced or replica 0 becoming out of sync.

Potentially fixes many bugs associated with PBFT batch, but because of the wide variety of potential failures, they will each need to be reviewed and will be linked to this PR later if associated.
## How Has This Been Tested?

These fixes are all targeted at the `TestBatchCustody` test.
## Checklist:

### DIFF
--- a/consensus/obcpbft/complainer.go
+++ b/consensus/obcpbft/complainer.go
@@ -121,12 +121,14 @@ func (c *complainer) SuccessHash(hash string) {
 // The custody queue timeouts are reset.  Restart returns all requests
 // that are maintained in custody.
 func (c *complainer) Restart() map[string]*Request {
-	c.complaints.RemoveAll()
+	complaints := c.complaints.RemoveAll()
 	custody := c.custody.RemoveAll()
 	reqs := make(map[string]*Request)
-	for _, pair := range custody {
-		c.custody.Register(pair.ID, pair.Data)
-		reqs[pair.ID] = pair.Data.(*Request)
+	for _, custodian := range [][]custodian.CustodyPair{complaints, custody} {
+		for _, pair := range custodian {
+			c.custody.Register(pair.ID, pair.Data)
+			reqs[pair.ID] = pair.Data.(*Request)
+		}
 	}
 	return reqs
 }
--- a/consensus/obcpbft/complainer.go
+++ b/consensus/obcpbft/complainer.go
@@ -119,16 +119,20 @@ func (c *complainer) SuccessHash(hash string) {
 // Restart resets custody and complaint queues without calling into
 // the complaintHandler.  The complaint queue is drained completely.
 // The custody queue timeouts are reset.  Restart returns all requests
-// that are maintained in custody.
+// that are maintained in custody, or have been received from other replicas
 func (c *complainer) Restart() map[string]*Request {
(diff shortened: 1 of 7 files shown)


jyellick:
Please let @corecode validate these changes before pulling.

consensus/obcpbft/complainer.go
```diff
@@ -121,12 +121,14 @@ func (c *complainer) SuccessHash(hash string) {
...
 // that are maintained in custody.
 func (c *complainer) Restart() map[string]*Request {
-	c.complaints.RemoveAll()
+	complaints := c.complaints.RemoveAll()
 	custody := c.custody.RemoveAll()
 	reqs := make(map[string]*Request)
-	for _, pair := range custody {
-		c.custody.Register(pair.ID, pair.Data)
-		reqs[pair.ID] = pair.Data.(*Request)
+	for _, custodian := range [][]custodian.CustodyPair{complaints, custody} {
```
corecode: I don't think we should add all our complaints (i.e. received from other replicas) to our custody.

corecode:
Looks good except for that one line I pointed out.

consensus/obcpbft/complainer.go
jyellick: Ah, yes, that's a side effect I didn't intend, I'll fix that, thanks!

jyellick:
@corecode The above commit should address your comment.

@srderson I think this should be ready to merge.

jyellick:
Rebased because of conflict

consensus/obcpbft/pbft-core.go
```diff
@@ -43,7 +43,7 @@ func init() {
 }
 
 const (
-	// An ugly thing, we need to create timers, then stop them before they expire, so use a large timeout
+	// UnreasonableTimeout is an ugly thing, we need to create timers, then stop them before they expire, so use a large timeout
 	UnreasonableTimeout = 100 * time.Hour
```
srderson, line 47: This variable name made me lol

srderson:
LGTM

PR #1486, opened 2016-05-16
Update identity_management_FAQ (language edit)

ghost:
I've edited the answers for clarity and grammar while maintaining the same structure and whenever possible the same verbiage. 
## Description

Mainly this involves smoothing out some rough edges. Changing the order of certain things, removing duplicated words and phrases so things flow better. 
## Motivation and Context

A FAQ is often the first place someone will go when they're not intimately familiar with the product or the issues at hand. It's therefore particularly important for it to be written as clearly and concisely as possible. 
## How Has This Been Tested?

N/A
## Checklist:

Signed-off-by: Joe Alewine (jalewine@gmail.com)

The changes I believe help the readability and therefore the usability of the FAQ.

### DIFF
--- a/docs/FAQ/identity_management_FAQ.md
+++ b/docs/FAQ/identity_management_FAQ.md
@@ -1,18 +1,18 @@
 ## Identity Management (Membership Service)
 &nbsp;
 ##### What is unique about the fabric's Membership Service module?
-The design and implementation of the Membership Service module encompass many of the latest advances in cryptography, which we believe make it stand apart from other alternatives.
+One of the things that makes the Membership Service module stand out from the pack is our implementation of the latest advances in cryptography.
 
-In addition to supporting generally-expected requirements such as preserving the privacy of transactions and making them auditable, the fabric's membership service module also introduces the concept of enrollment and transaction certificates. Infinite number of transaction certificates can be generated from their parent enrollment certificates (issued to validated users). This design ensures that asset tokens (which can be represented by transaction certificates) can only be created by verified owners, and private keys of asset tokens can be regenerated if lost.
+In addition to ensuring private, auditable transactions, our Membership Service module introduces the concept of enrollment and transaction certificates. This innovation ensures that only verified owners can create asset tokens, allowing an infinite number of transaction certificates to be issued through parent enrollment certificates while guaranteeing the private keys of asset tokens can be regenerated if lost. 
 
-Furthermore, the design of the system allows transaction certificates to both expire and be revoked, which allows issuers to have greater control over the asset tokens that they issued on a distributed chain.
+Issuers also have the ability revoke transaction certificates or designate them to expire within a certain timeframe, allowing greater control over the asset tokens they have issued. 
 
-Finally, like most other modules on the fabric, you can always replace the default membership service implementation with another option.
+Like most other modules on the fabric, you can always replace the default module with another membership service option should the need arise.
 
 
 &nbsp;
 ##### Does its Membership Service make the fabric a centralized solution?
 
-No, because the fabric's Membership Service does not provide, deploy, validate, or execute transactions or business logic. The only role of the Membership Service is to issue digital certificates to validated entities that want to participate in the network. The service is not aware of how or when these certificates are used in any particular network.
+No. The only role of the Membership Service module is to issue digital certificates to validated entities that want to participate in the network. It does not execute transactions nor is it aware of how or when these certificates are used in any particular network.
 
-The fabric's Membership Service does, however, serve as the central regulator of the networks that it services, because the certificates that the service issues are used by the networks to regulate and manage their users.
+However, because certificates are the way networks regulate and manage their users, the module serves a central regulatory and organizational role.


docs/FAQ/identity_management_FAQ.md
```diff
@@ -1,18 +1,18 @@
...
+In addition to ensuring private, auditable transactions, our Membership Service module introduces the concept of enrollment and transaction certificates. This innovation ensures that only verified owners can create asset tokens, allowing an infinite number of transaction certificates to be issued through parent enrollment certificates while guaranteeing the private keys of asset tokens can be regenerated if lost. 
 
-Furthermore, the design of the system allows transaction certificates to both expire and be revoked, which allows issuers to have greater control over the asset tokens that they issued on a distributed chain.
+Issuers also have the ability revoke transaction certificates or designate them to expire within a certain timeframe, allowing greater control over the asset tokens they have issued. 
 
-Finally, like most other modules on the fabric, you can always replace the default membership service implementation with another option.
+Like most other modules on the fabric, you can always replace the default module with another membership service option should the need arise.
```
christo4ferris, line 10: suggest:
As is the case with other capabilities of the fabric, the membership services framework allows for alternate implementations to replace the default behavior.

docs/FAQ/identity_management_FAQ.md
```diff
@@ -1,18 +1,18 @@
 ## Identity Management (Membership Service)
 &nbsp;
 ##### What is unique about the fabric's Membership Service module?
-The design and implementation of the Membership Service module encompass many of the latest advances in cryptography, which we believe make it stand apart from other alternatives.
+One of the things that makes the Membership Service module stand out from the pack is our implementation of the latest advances in cryptography.
```
christo4ferris, line 4: this is a bit bold... sounds like marketing material. Shouldn't we back this claim up with some evidence?

docs/FAQ/identity_management_FAQ.md
ghost, line 10: I struggled with this line because it wasn't entirely clear to me whether it was trying to say that fabric users could replace the service module **entirely** or whether it was possible to change only certain functionalities. 

If it can be replaced entirely -- what other module frameworks are available? And if it's only possible to modify the framework, which capabilities are fungible and which are not? Also, who has the power to change the framework (whether entirely or in part) and how is that done?

docs/FAQ/identity_management_FAQ.md
ghost, line 4: Replacing "that" with "we believe" in the first sentence would soften it I think. Changing it to:

"One of the things we believe makes the Membership Service module stand out from the pack is our implementation of the latest advances in cryptography."

christo4ferris, line 4: who is "we"? Please keep in mind that this is a Linux Foundation document. I still think it should be struck unless we can back up the claim.

ghost, line 4: Well it's not for me to say but I doubt a long digression into competing versions of membership service is what anyone wants to see in that FAQ. 

Is it possible to change the wording of the question? Because I'm just not sure how you answer

**"What is unique about the fabric's Membership Service module?"** 

Without drawing a comparison, even if it's an implicit one. 

If the question was worded:

**"What are some of the security features of the fabric's Membership Service module?"** 

You could provide exactly the same information in the answer without the need to make a comparison.

ghaskins:
I assume this patch still has outstanding work before merge consideration?

ghost:
ghaskins --

I've asked Josh Horton to contribute to this matter so I can make the appropriate edits. As of right now some issues need to be resolved before this can be merged.

ghaskins:
@joe-alewine thank you for the update

docs/FAQ/identity_management_FAQ.md
christo4ferris, line 4: Maybe we just omit the "What is unique..." Q and A. I like the rephrasing, but would not want to put the onus on you to do the research. Maybe remove and create an issue that suggests we should add the rephrased question with a suitable answer.

docs/FAQ/identity_management_FAQ.md
christo4ferris, line 10: I'm not clear that we have actually described how that can be achieved beyond hand waving. Maybe soften to say "As is the case with other capabilities of the fabric, the membership services framework is intended to permit alternate implementations. As the fabric matures, the precise means by which the default membership services can be replaced will be improved and hardened."

docs/FAQ/identity_management_FAQ.md
ghost, line 4: Omit the entire question and the entire answer or just that phrase? 

I think simply by rephrasing the question without a reference to a direct or implicit comparison you have kept what's useful about the answer without materially changing the question all that much. 

I don't know how much the questions have been written in stone, however.

docs/FAQ/identity_management_FAQ.md
ghost, line 10: That works. I like it better than what it was originally or what I changed it to. 

I do wonder how important of an issue this is to address at all. Are potential users particularly worried that membership service won't be customizable (as compared to the fabric's other features)?

joshhus:
Who wrote the original FAQ - my instincts say we should at least try to engage that person(s) ... and also to compare what this "FAQ" says to the protocol spec content in this area. ... (As an aside, Joe, we will want to split up protocol spec into separate documents.) ... (i.e. would this "FAQ" content do better for the user in a clear and integrated and separate "membership services" document. Then we can get rid of the question of identifying a "unique" feature, and simply document what is.)

ghost:
So the plan is to do away with a centralized FAQ and establish separate FAQs for each topic? Or would the kinds of specific process questions that make up the FAQ right now be folded into documents related to the various topics while a new, broader centralized FAQ would be created from scratch and take its place?

christo4ferris:
LGTM thanks @joe-alewine

srderson:
Thanks, LGTM

PR #1490, opened 2016-05-17
Move statetransfer to peer package

### LINKED ISSUE #645: State transfer should be migrated out of the consensus package

jyellick:
## Description

This patch moves statetransfer from consensus to peer.  The only modifications are imports, and one relative path change for finding the config during testing.
## Motivation and Context

This is the last in a series of changesets (including #1416  and #1423) designed to move state transfer out of consensus.  It closes #645.
## How Has This Been Tested?

This is a code organizational change only, no logic is changed in this patch, so only the CI tests are important.
## Checklist:


srderson:
LGTM

PR #1495, opened 2016-05-17
Add correlation ID to peer statetransfer for blocks and state deltas

### LINKED ISSUE #974: State synchronization API should be able to report an error or other ending
Today, two pieces of the state synchronization API, `SYNC_GET_BLOCKS` and `SYNC_STATE_GET_DELTAS`, have no way of informing the other end of the connection that they are done communicating.  

This was a conscious design decision to simplify the implementation because blocks and state deltas have an associated block number, and their value should never change.  If a read receives an incorrect block number, it can safely ignore it as being from an unrelated request.

However, this produces multiple problems.  First and foremost, in an issue like #906, where a request has been made for a state delta which has been discarded by the ledger, the peer has no recourse but to stop returning data and abort the transfer. This works today, because eventually the other side times out, but relying on a timeout to communicate an end condition is definitely suboptimal.

The immediately thought would be to use an unused value, such a `nil` state delta to indicate 'end', but this has a problem, that although a reader can safely ignore an incorrect block number, can it safely ignore an 'end' value? Without some sort of correlation ID, this is not clear.

My inclination would be to introduce a correlation ID for all state transfer type requests.  The reading peer would only send requests with the correct correlation ID onto the channel, and could close the channel at 'end' in much the same say the `SYNC_STATE_GET_SNAPSHOT` call works.  This would solve the above problem, and would also decrease the accounting necessary for timeouts in the consuming code. Today there are several `TODO`s in the state transfer code around counting incorrect values off the channel against the timeout, as otherwise a malicious peer could send bogus information indefinitely without state transfer timing out, a correlation ID would fix this as well.

I'd like to hear your opinion on this @jeffgarratt.

jyellick:
## Description

This changeset adds a correlation ID to `SyncBlockRange`, it also modifies the peer handler to populate this correlation ID field, and filter extraneous messages with the wrong correlation ID.
## Motivation and Context

Because of the peer's stream based approach, it is not possible to determine which `SYNC_GET_BLOCKS` request a `SYNC_BLOCKS` message corresponds to.  Therefore, if multiple request are issued to the same peer, the requester cannot differentiate between that peer sending garbage, and that peer merely having other traffic on the wire.  This requires that consumers of the state transfer rely on timeouts, rather than being able to detect errors.  In the case of a 'missed' block (do to, say an buffer overflow), the requesting peer must read all pending blocks or wait until a timeout in order to retry.  By including a correlation ID, this allows the caller to recognize that any out of order sequence number indicates an error, and the call should be aborted immediately.

This change applies to both block synchronization and state delta synchronization.

Fixes #974 
## How Has This Been Tested?

This has been tested using the existing state transfer behave test (@issue_680).  I'm unaware of any better way to test this function.  Maybe @jeffgarratt could weigh in here? (I'd also like Jeff's sign-off on the validity of these changes)
## Checklist:

Note, for the lint-ing, I have intentionally ignored the errors in the `config.go` as these are formulaic Get-ers to allow caching of Viper results.

### DIFF
--- a/core/peer/config.go
+++ b/core/peer/config.go
@@ -52,6 +52,7 @@ var peerEndpointError error
 // Cached values of commonly used configuration constants.
 var syncStateSnapshotChannelSize int
 var syncStateDeltasChannelSize int
+var syncBlocksChannelSize int
 var validatorEnabled bool
 var tlsEnabled bool
 
@@ -116,6 +117,7 @@ func CacheConfiguration() (err error) {
 
 	syncStateSnapshotChannelSize = viper.GetInt("peer.sync.state.snapshot.channelSize")
 	syncStateDeltasChannelSize = viper.GetInt("peer.sync.state.deltas.channelSize")
+	syncBlocksChannelSize = viper.GetInt("peer.sync.blocks.channelSize")
 	validatorEnabled = viper.GetBool("peer.validator.enabled")
 	tlsEnabled = viper.GetBool("peer.tls.enabled")
 
@@ -173,6 +175,13 @@ func SyncStateDeltasChannelSize() int {
 		cacheConfiguration()
 	}
 	return syncStateDeltasChannelSize
+}
+
+func SyncBlocksChannelSize() int {
+	if !configurationCached {
+		cacheConfiguration()
+	}
+	return syncBlocksChannelSize
 }
 
 func ValidatorEnabled() bool {
--- a/core/peer/handler.go
+++ b/core/peer/handler.go
@@ -43,6 +43,7 @@ type Handler struct {
 	syncBlocks                    chan *pb.SyncBlocks
 	snapshotRequestHandler        *syncStateSnapshotRequestHandler
 	syncStateDeltasRequestHandler *syncStateDeltasHandler
+	syncBlocksRequestHandler      *syncBlocksRequestHandler
 }
 
 // NewPeerHandler returns a new Peer handler
@@ -58,6 +59,7 @@ func NewPeerHandler(coord MessageHandlerCoordinator, stream ChatStream, initiate
 
 	d.snapshotRequestHandler = newSyncStateSnapshotRequestHandler()
 	d.syncStateDeltasRequestHandler = newSyncStateDeltasHandler()
+	d.syncBlocksRequestHandler = newSyncBlocksRequestHandler()
 	d.FSM = fsm.NewFSM(
 		"created",
 		fsm.Events{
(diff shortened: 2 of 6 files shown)


corecode:
Maybe we should switch to a sync RPC approach (re-using the same peer grpc connection), so that we don't have to manually implement de-multiplexing over and over.

jyellick:
@corecode I'm not strictly opposed to this in principle, though I know @jeffgarratt and others supported the use of asynchronous chat for a number of reasons.  For instance, when requesting 500 blocks, would you suggest that this be 500 synchronous calls? Or would this be one call that returned the 500 blocks? If so, I'm unsure how to enable processing of the results as they arrive in a synchronous type call (rather than requiring all 500 blocks be retrieved, and then processed).  

If we decide that we should migrate this to a synchronous RPC interface, I would still advocate using the correlation ID for the moment, as migration to a different RPC interface will take time, and this changeset addresses problems which exist today.

binhn:
@jyellick @corecode I would stay with the current chat stream interface as this scenario covers cases where huge amount of data (entire blockchain) might be transferred

corecode:
@binhn how does the interface impact the amount of data?

binhn:
@corecode my understanding is that rpc (in grpc) is request/response, not streaming, so it can't handle large volume of data well

corecode:
you can run multiple requests in parallel - internally it will stream these requests over the same grpc connection, but it will also provide demultiplexing (and a blocking interface)

jeffgarratt:
@jyellick  Reviewed, looks good to me.  Nice job!!

jyellick:
@corecode But then to retrieve 10k blocks, you would need to issue 10k requests, no?

corecode:
Or ranged requests.  But yes, that's what we're doing anyways (just in a stream format)?

jyellick:
@corecode Today, to retrieve 10k blocks, you can conceivably send a single 'give me those 10k blocks' message. 

Because of the risk of overflowing the buffer, you should generally not ask for more blocks than you have room to store (see issue #1496), but still, with a buffer of 100, you can start processing block 1 as soon as it arrives, rather than having to wait for for a full 100 block chunk.  This seems technically superior, even if the code is a little more complex (especially as the code is already written).

binhn:
@jyellick @corecode I am ready to merge this unless you want to discuss further; however, I maintain that using stream for blocks transfer is appropriate in this PR. We could take the discussion on another issue to capture the dialog there.

jyellick:
@binhn Pull is fine by me

Issue #1496, opened 2016-05-17
State transfer should never request more blocks than the buffer can handle


jyellick:
## Description

The peer has a buffered channel for accepting blocks for state transfer.  Today, state transfer requests whatever range of blocks it requires, regardless of whether that range exceeds the size of the buffer.  In periods of high stress, when the replica is behind, this can cause the replica to overflow the buffer, and discard necessary data, requiring retry.

This would also be a good opportunity to clean up some of the cruft which has accumulated over the evolution of the state transfer code, including adapting to the more predictable channel error behavior introduced by #974

PR #1498, opened 2016-05-17
Fix state transfer to fail on out of order

jyellick:
## Description

Change the state transfer code to abort on incorrectly ordered messages, rather than tolerate them.
## Motivation and Context

When the underpinning mechanisms of state transfer (`SYNC_GET_BLOCKS` and `SYNC_GET_STATE_DELTAS`) were introduced, they had no notion of correlation ID.  Consequently, when the invoking code received messages, there was no guarantee that message was associated with any particular call from the caller.  Therefore, the code would check for inappropriately ordered messages, but could not treat them as errors.  Now thanks to issue #974 and PR #1495 (which should be merged before this one), state transfer can treat these bad messages as errors.

This fixes some outstanding TODOs in the code.  It can be seen as contributing to implementation of #1496, or as more general cleanup.

The bulk of the changes are to the tests, and only general golang knowledge is required for review.
## How Has This Been Tested?

A new failure type, `OutOfOrder` was introduced to the state transfer mock testing framework.  All existing state transfer tests that check for survival of failure were enhanced to also test for tolerance of an `OutOfOrder` failure.
## Checklist:

### DIFF
--- a/core/peer/statetransfer/statetransfer.go
+++ b/core/peer/statetransfer/statetransfer.go
@@ -502,18 +502,14 @@ func (sts *StateTransferState) syncBlocks(highBlock, lowBlock uint64, highHash [
 
 				if syncBlockMessage.Range.Start < syncBlockMessage.Range.End {
 					// If the message is not replying with blocks backwards, we did not ask for it
-					// TODO, this should count against the timeout
-					logger.Debug("Received block potentially from a different request, start=%d, end=%d", syncBlockMessage.Range.Start, syncBlockMessage.Range.End)
-					continue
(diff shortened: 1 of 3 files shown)


srderson:
@tuand27613 @corecode @kchristidis please review. Thanks.

corecode:
I don't feel I understand state transfer enough to give a confident review.

tuand27613:
@jyellick explained the code path to me in slack.  LGTM.

binhn:
@tuand27613 Thanks
LGTM  too

srderson:
LGTM

Issue #1286, opened 2016-04-28
many golint and go vet warnings in fabric/membersrvc/ca


christo4ferris:
fabric/membersrvc/ca contains two pages of golint and vet warnings. I'm willing to do a PR to address them but concerned that there may be active work on this package and that whomever is working it should address this in their PR.

JonathanLevi:
While we're at it, can somebody please assign this to me? Will do it one file at a time.

srderson:
Is it possible to run golint and vet in the CI pipeline? Sometimes they give false positives, but a list of warnings would help with review.

christo4ferris:
@srderson yes, it is possible... agree that there can be false positives. I wonder if it can be done as informative (eg not fail the build). I'll investigate.

christo4ferris:
@JonathanLevi it's yours!

JonathanLevi:
As discussed, we waited with these until the end of Sprint 5 of last week. Currently there are 349 lint errors, for the interested/avid reader, we enclose the output here [2016-05-16_membersrvc_lint_errors.txt](https://github.com/hyperledger/fabric/files/268588/2016-05-16_membersrvc_lint_errors.txt)

To minimize intervention, will work to tackle them a file at a time adopting @christo4ferris's approach.

JonathanLevi:
Following

``````
vagrant@hyperledger-devenv:v0.0.9-048fc48:/opt/gopath/src/github.com/hyperledger/fabric/membersrvc/ca$ git log -1
commit cb011329cd3ef5715d36eb0dff96ffad59de642c
Merge: 35e7f3e 55c9175
Author: Sheehan Anderson <sranderson@gmail.com>
Date:   Tue May 31 20:37:09 2016 -0400

    Merge pull request #1669 from JonathanLevi/issue-1558-go-vet

    Issue 1558 Fix go vet issues in fabric/membersrvc/ca```
``````

Executing:

```
vagrant@hyperledger-devenv:v0.0.9-048fc48:/opt/gopath/src/github.com/hyperledger$ golint ./fabric/membersrvc/ca/... | wc -l
0
vagrant@hyperledger-devenv:v0.0.9-048fc48:/opt/gopath/src/github.com/hyperledger$ go vet ./fabric/membersrvc/ca/... | wc -l
0
```

JonathanLevi:
[...]

PR #1499, opened 2016-05-17
Address golint warnings and errors in ca.go (Fixes #1499)

### LINKED ISSUE #1409: The enrollID should not be included in the subject of transaction certificates
## Description

Transaction certificates currently include the enrollID as part of the subject in the X509 public certificate.  Given that tcerts are used for unlink-ability, this needs to be fixed.  
## Describe How to Reproduce

Enable security, use transaction certificates and decode the X509 certs.  Easiest way to do this is to simply extract the cert bytes from the block/transaction API response and decode them

JonathanLevi:
Fixing **golint** errors and warnings in ca.go (and a small change in tca.go to reflect the GroupAffiliation change)
## Description

See an impressive list of golint errors here: [This is part of #[1286](https://github.com/hyperledger/fabric/issues/1286)]
## How Has This Been Tested?

`golint + make all 
`
## Checklist:

Signed-off-by: JonathanLevi

### DIFF
--- a/membersrvc/ca/ca.go
+++ b/membersrvc/ca/ca.go
@@ -32,8 +32,8 @@ import (
 	"time"
 
 	"github.com/hyperledger/fabric/core/crypto/primitives"
+	_ "github.com/mattn/go-sqlite3" // TODO: not use a blank import!
 	pb "github.com/hyperledger/fabric/membersrvc/protos"
-	_ "github.com/mattn/go-sqlite3"
 )
 
 // CA is the base certificate authority.
@@ -47,7 +47,7 @@ type CA struct {
 	raw  []byte
 }
 
-// The certificate spec defines the parameter used to create a new certificate.
+// CertificateSpec defines the parameter used to create a new certificate.
 type CertificateSpec struct {
 	id           string
 	commonName   string
@@ -62,12 +62,12 @@ type CertificateSpec struct {
 // AffiliationGroup struct
 type AffiliationGroup struct {
 	name     string
-	parentId int64
+	parentID int64
 	parent   *AffiliationGroup
 	preKey   []byte
 }
 
-// Create a new certificate spec
+// NewCertificateSpec creates a new certificate spec
 //
 func NewCertificateSpec(id string, commonName string, serialNumber *big.Int, pub interface{}, usage x509.KeyUsage, notBefore *time.Time, notAfter *time.Time, opt ...pkix.Extension) *CertificateSpec {
 	spec := new(CertificateSpec)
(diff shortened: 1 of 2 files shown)


membersrvc/ca/ca.go
```diff
@@ -33,7 +33,7 @@ import (
 
 	"github.com/hyperledger/fabric/core/crypto/primitives"
 	pb "github.com/hyperledger/fabric/membersrvc/protos"
-	_ "github.com/mattn/go-sqlite3"
+	_ "github.com/mattn/go-sqlite3" // TODO: not use a blank import!
```
ghaskins: Do we know why this was here and why it is now marked TODO?

Blank imports are generally used for one purpose to my knowledge: to force the side-effects of running the modules func init().  

So, my first question is: Did a human or the tooling decide to add that comment.  If a human, what was the rationale?

Second question: is the side-effect inducing behavior desired, or not?  If desired, this needs to stay the way it was.  If it was not desired, we should just remove the import outright.

JonathanLevi: Hi @ghaskins,
1. Yes I manually added the comment and not the tool. So yes, without loss of generality ;-) and under some assumption, it is relatively safe to claim that it was a **human**.
2. It's true that blank imports are generally used to force the side-effects of running the modules func init()... 

As for the second question, I do understand why it is currently there: Specifically in this case I believe the blank import is used for registering the **sqlite3** driver as a database driver without importing any other functions, kind of a:

`sql.Register("sqlite3", &SQLiteDriver{})`

I believe that the coder's original desired side-effects are/were that further (standard) `sql` calls will refer to the registered `sqlite3` driver, e.g., in:

`db, err := sql.Open("sqlite3", filepath.Join(ksPath, ks.node.conf.getKeyStoreFilename()))
`
in [/core/crypto/node_ks.go#L358](https://github.com/JonathanLevi/fabric/blob/ca8135b1a0f95127c96478518628a65a377ec123/core/crypto/node_ks.go#L358)

Which is why I haven't removed the import altogether.

Keeping in mind that the background to this suggested fix is that I wanted to quickly address the go lint message:

`ca/ca.go:35:2: a blank import should be only in a main or test package, or have a comment justifying it`.

JonathanLevi: I hope I addressed the questions/concern, otherwise do let me know?
Even though I personally don't like blank imports (or at least avoid them whenever possible), I have also changed the import comment as follows:

```
-       _ "github.com/mattn/go-sqlite3" // TODO: not use a blank import!
+       _ "github.com/mattn/go-sqlite3" // TODO: justify this blank import or remove
```

Please kindly let me know if these address the questions and concerns, or otherwise? Thanks.

srderson: Looks good. Agree it should be justified to help people understand why it was used.

membersrvc/ca/ca.go
```diff
@@ -33,7 +33,7 @@ import (
 
 	"github.com/hyperledger/fabric/core/crypto/primitives"
 	pb "github.com/hyperledger/fabric/membersrvc/protos"
-	_ "github.com/mattn/go-sqlite3" // TODO: not use a blank import!
+	_ "github.com/mattn/go-sqlite3" // TODO: justify this blank import or remove
```
JonathanLevi, line 36: @srderson Yes, didn't want to simply 'hide' this lint warning...

I like action-able TODOs ;-)

srderson:
LGTM. Thanks!

PR #1516, opened 2016-05-19
Merge pull request #1 from hyperledger/master

marvinchin:
## Motivation and Context

Fixes #
## Checklist:

- [ ] I have added a [Signed-off-by](https://github.com/hyperledger/fabric/blob/master/CONTRIBUTING.md#legal-stuff)
- [ ] Either no new documentation is required by this change, OR I added new documentation
- [ ] Either no new tests are required by this change, OR I added new tests

Signed-off-by:

asd


common/ledger/util/leveldbhelper/leveldb_provider.go
```diff
@@ -204,6 +204,16 @@ func (h *DBHandle) DeleteAll() error {
...
 }
 
+// IsEmpty returns true if no data exists for the DBHandle
+func (h *DBHandle) IsEmpty() (bool, error) {
+	itr, err := h.GetIterator(nil, nil)
+	if err != nil {
+		return false, err
+	}
+	defer itr.Release()
+	return !itr.Next(), errors.WithMessagef(itr.Error(), "internal leveldb error while obtaining next entry from iterator")
```
cendhu: Nit. 

When an error occurs, I thought we need to return the default `false` value. That's why in the earlier version of the code, I had this check explicitly. Anyway, it does not matter as we always needs to check the error first.

common/ledger/util/leveldbhelper/leveldb_provider_test.go
```diff
@@ -372,6 +372,79 @@ func TestClose(t *testing.T) {
 	require.Equal(t, map[string]*DBHandle{}, p.dbHandles)
 }
 
+func TestIsEmpty(t *testing.T) {
+	var env *testDBProviderEnv
+	var db1, db2 *DBHandle
```
cendhu, line 377: We have two types of independent sub-tests:

1. Make it independent that even if we change the order, the test result would not change.
2. Make it independent that even if we run them in parallel using `t.Parallel()`, the test result should not change.

Here, (1) is satisfied which is good enough.

common/ledger/util/leveldbhelper/leveldb_provider_test.go
```diff
@@ -372,6 +372,79 @@ func TestClose(t *testing.T) {
...
+
+	setup := func() {
+		env = newTestProviderEnv(t, testDBPath)
+		p := env.provider
+		db1 = p.GetDBHandle("db1")
+		db2 = p.GetDBHandle("db2")
```
cendhu, line 383: Do we need a test with `db1` and `db2`? Can't we use `db1` alone? It will help in reducing the LOC without making the test weaker.

Issue #1304, opened 2016-04-29
Default peer address is hard-coded in chaincode shim


bcbrock:
From fabric/core/chaincode/shim at line 129:

```
func getPeerAddress() string {
    if peerAddress != "" {
        return peerAddress
    }

    if peerAddress = viper.GetString("peer.address"); peerAddress == "" {
        // Assume docker container, return well known docker host address
        peerAddress = "172.17.42.1:30303"
    }

    return peerAddress
}
```

The hard-coded default peer address is a very bad assumption. The Docker ".1" address is the Docker bridge itself, not any peer, and there is no guarantee that Docker is bridged at 172.17.42.1. The Start() call should fail explicitly if the peer address is unknown, as using this address will lead to a certain, more obscure failure.

christo4ferris:
@bcbrock good catch, and yes, very bad assumption. There are a bunch of similar functions in peer/config.go and specifically, Cache.Configuration.getLocalAddress. Does this apply here?

bcbrock:
@christo4ferris I'm not sure what you mean  re: peer/config.go, Cache.Configuration.getLocalAddress. I don't see any hard-coded addresses in there.

mastersingh24:
@gennadylaventman  - can you add this to the list of issue around connection management?

gennadylaventman:
@mastersingh24 - I will.

bcbrock:
There is a similar hard-coded default in fabric/core/chaincode/chaincode_support.go

```
const (
    // DefaultChain is the name of the default chain.
    DefaultChain ChainName = "default"
    // DevModeUserRunsChaincode property allows user to run chaincode in development environment
    DevModeUserRunsChaincode       string = "dev"
    chaincodeStartupTimeoutDefault int    = 5000
    chaincodeInstallPathDefault    string = "/opt/gopath/bin/"
    peerAddressDefault             string = "0.0.0.0:30303"
)

```

lehors:
@bcbrock This issue was addressed by PR #1520 and can now be closed.

bcbrock:
Thanks Arnaud @lehors

PR #1520, opened 2016-05-19
Connection code centralization.

### LINKED ISSUE #1362: Moving connection creation code to one place
Code contains 4 methods that create connections:
core/peer/peer.go: NewPeerClientConnectionWithAddress
events/consumer/consumer.go: newEventsClientConnectionWithAddress 
core/crypto/node_grpc.go: getClientConn
core/caincode/shim/chaincode.go: newPeerClientConnection  
and code in all methods do almost same work. 
It worth to move all same code to one place, to avoid code duplication and make code more readable. 

This change is first step in communication code consolidation to easy the introduction new capabilities such as flow control.

Inspired by Gari Singh comments.

gennadylaventman:
## Description

Moving connection creation code to central place
## Motivation and Context
1. Connection creation code moved from
   core/chaincode/shim/chaincode.go, core/peer/peer.go, events/consumer/consumer.go and core/crypto/node_grpc.go to new package core/comm (connection.go file) - issue #1362
2. Cached configuration ("peer.tls.enabled") moved to comm/config.go instead peer/config.go, to avoid circular dependency.
3. Added method to read peer tls credentials InitTLSForPeer to comm/connection.go, instead reading it in each file (peer.go, chaincode.go, consumer.go)
4. Fixed hardcoded peer address in core/chaincode/shim/chaincode.go - issue #1304

Fixes #1362 
## How Has This Been Tested?

Run unit tests and behave tests
## Checklist:

Signed-off-by: Gennady Laventman gennady@il.ibm.com


core/chaincode/shim/chaincode.go
```diff
@@ -148,40 +147,20 @@ func getPeerAddress() string {
 	}
 
 	if peerAddress = viper.GetString("peer.address"); peerAddress == "" {
+		os.Exit(-1)
```
ghaskins: This strikes me as something that probably warrants a log/stderr message before exiting.  IIUC, we are basically trying to say that omitting peer.address is fatal, right?

gennadylaventman: Sure, will add error log.
How about:
`chaincodeLogger.Fatalf("Peer address not configured")`
instead
`os.Exit(-1)`

muralisrini:
In the interest of loose coupling, I would suggest not having to import fabric into shim ? I know we do some fabric imports already IMHO but we should really be finding ways to decouple more. At some point when we stop including the entire fabric with chaincode image, these will become thorny. Also there is the issue of multi-language chaincode support. Suggest we consider making shim an exception ?

core/chaincode/shim/chaincode.go
```diff
@@ -148,40 +147,18 @@ func getPeerAddress() string {
 	}
 
 	if peerAddress = viper.GetString("peer.address"); peerAddress == "" {
-		// Assume docker container, return well known docker host address
-		peerAddress = "172.17.42.1:30303"
+		chaincodeLogger.Fatalf("Peer address not configured, can't connect to peer")
```
ghaskins: This is certainly an improvement, thank you.  Sorry to nit pick, but let me ask one question: Do you think this will sufficiently inform the user precisely what is wrong?  (I'm not saying it doesn't, I'm sincerely asking).  What I am wondering is whether referring to the configuration element more precisely may help a (perhaps novice) user know what to do.  For instance "peer.address is not configured ... " might be clearer. I can be convinced otherwise.

mastersingh24:
@muralisrini  - not sure exactly what you mean on the "decouple" comment?   I think part of the problem is that everything related to the fabric project goes under the "fabric" root.  Looking at things that way, fabric no longer really means just the peer, but everything else as well.

BUT - to your point, there is way too much cross calling of "packages" with needless cross-dependencies.  So there needs to be a much more modular approach to packages (which by the way should all be self-testable).

But I don't think that any of that directly relates to this PR as the goal here was to create a universal comms package which I think this PR does nicely.

core/chaincode/shim/chaincode.go
gennadylaventman: Will change it, np.

srderson:
@muralisrini @mastersingh24 IMO this is one reason why a deploy transaction should only require the chaincode and not the shim also. The chaincode would just indicate its shim dependency and the peer would supply it if it's one of the included shims.

muralisrini:
@mastersingh24   more the import dependencies we add to shim, less isolated it becomes... more the difficulty in converting to other languages.

And fabric package import dependencies add a twist - we have to be extra careful in changing the common code so we don't introduce changes that other languages will also have to keep up with.  Bottomline, IMO, keeping the shim self-contained is a good thing and so worth making a principle of.

@srderson dependency between shim and chaincode is unavoidable - there is  a shim side chaincode is tied to... and chaincode is developed on top of. So regardless of who supplies (comes with the chaincode or peer provides it)  the dependency has to be carried from dev to deployment.  This is more about reducing those dependencies.

mastersingh24:
@muralisrini - while I kinda see what you are saying here, I don't quite understand how you want this to be handled.  The fabric/core/comm package created here is actually a standalone package - meaning it does not have any other dependencies on any other fabric packages / components.  The idea here is to provide a common module for all Go code used by / with the fabric to handle client connections rather than coding up client connections everywhere they are used.   If you import this package, you are not bringing in any other parts of the fabric.   Now if you are saying that not everything belongs under core, perhaps that's OK but up to this point (at least not that I have seen) no one has really made any rules on what goes under core.   But in any case, even if at a later date this package was moved, it would still function

mastersingh24:
@gennadylaventman  - I noticed that there is not a standalone test for the comm package.  I'd suggest creating a test file within this package so that it is testable on its own.

gennadylaventman:
@mastersingh24 
Will add it.

muralisrini:
@mastersingh24 It would still function if it was go. But if the chaincode is Java and the common go package was changed, we'll have to change the java.  And you might change the core in a way that is not needed for the chaincode interaction. All I'm saying is the more we add, the more these subtle differences and book keeping we have to do.

gennadylaventman:
@muralisrini 
Agreed in general, but this is not a case. It doesn't really matter if shim connection code located it shim/chaincode.go or in totally standalone package - if connection code changed in go - you have to change it in java.

muralisrini:
@gennadylaventman the hidden assumption with that statement is that all changes will be needed for all the parts.  Peer connections are qualitatively different from chaincode connections and we may find some changes to one do not apply to the other.

IMO its good to adhere to the principle of keeping number of imports as small as possible on the chaincode side and use protos.ChaincodeMessage to encapsulate protocol/communication as much as possible.

But I don't want to belabor the point. We should go with majority decision which I think is to accept the PR as is :-)

srderson:
@gennadylaventman LGTM and it sounds like we're ready to merge the change. Could you rebase on master as there seems to be a conflict? Thanks.

gennadylaventman:
@srderson Done, I think.

srderson:
Yes, LGTM. Thanks!

PR #1525, opened 2016-05-19
Linux on Z build support

vpaprots:
S390X build support
## Description

Two parts:
- Top level makefile can be now run on z (docker tags sensitive to s390x)
- development on z is done on native OS, not in vagrant. Added a script to set it up RHEL on Z, make it ready for development, similar to devenv/setup.sh which does the same for vagrant
## Motivation and Context

With the eventual move to Jenkins, we want to dispatch to a z machine to test as well as x86. For that to happen, builds on s390x must happen out of the box, as similar to x86 as possible
## How Has This Been Tested?

Ran unit tests and behave tests, all pass
## Checklist:

Signed-off-by:Volodymyr Paprotski vpaprots@ca.ibm.com

### DIFF
--- a/devenv/setupRHELonZ.sh
+++ b/devenv/setupRHELonZ.sh
@@ -1,5 +1,10 @@
 #!/bin/bash
 
+# Development on Z is done on the native OS, not in Vagrant. This script can be 
+# used to set things up in RHEL on Z, similar to devenv/setup.sh which does the 
+# same for Vagrant. 
+# See https://github.com/hyperledger/fabric/blob/master/docs/dev-setup/install.md
+#
 # To get started:
 #       sudo su
 #       yum install git
--- a/devenv/setupRHELonZ.sh
+++ b/devenv/setupRHELonZ.sh
@@ -81,8 +81,9 @@ rm -rf /tmp/rocksdb
 
 ################
 # PIP
-wget http://dl.fedoraproject.org/pub/epel/7/x86_64/p/python-pip-7.1.0-1.el7.noarch.rpm
-rpm -ivh python-pip-7.1.0-1.el7.noarch.rpm
+yum install python-setuptools
+curl "https://bootstrap.pypa.io/get-pip.py" -o "get-pip.py"
+python get-pip.py
 pip install --upgrade pip
 pip install behave nose docker-compose
 
--- a/devenv/setupRHELonZ.sh
+++ b/devenv/setupRHELonZ.sh
@@ -1,0 +1,96 @@
+#!/bin/bash
+
+# To get started:
+#       sudo su
+#       yum install git
+#       mkdir -p $HOME/git/src/github.com/hyperledger
+#       cd $HOME/git/src/github.com/hyperledger
+#       git clone https://github.com/vpaprots/fabric.git
+#       source fabric/devenv/setupRHELonZ.sh
+#       ~/build.sh
+
+if [ xroot != x$(whoami) ]
+then
+   echo "You must run as root (Hint: sudo su)"
+   exit
+fi
+
+if [ -n -d $HOME/git/src/github.com/hyperledger/fabric ]
+then
+    echo "Script fabric code is under $HOME/git/src/github.com/hyperledger/fabric "
+    exit
+fi
+
+#TODO: should really just open a few ports..
+iptables -I INPUT 1 -j ACCEPT
+sysctl vm.overcommit_memory=1
+
+##################
+# Install Docker
+cd /tmp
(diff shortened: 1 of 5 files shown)


images/base/scripts/common/setup.sh
```diff
@@ -90,9 +109,14 @@ apt-get install -y build-essential libtool
 #./configure
 ./configure --prefix=/usr
 
-make
-make check
-make install
+if [ x$MACHINE = xs390x ]
+then
+    echo FIXME: protobufs wont compile on 390, missing atomic call
```
ghaskins, line 114: IIUC, the only implication of this is that we cannot invoke the protoc compiler on Z (today).

vpaprots, line 114: I should really go fix it.. The protobuf header files did something weird looking for atomic functions... The GCC names are the same as elsewhere, just refused to use them on z.. Either patch inline and/or submit a fix.. Either way I wanted more time to understand the  ifdef spaghetti and it wasn't making tests fail, so I saw that as future item.. Probably should open an issue so that I don't forget

vpaprots:
Signed-off-by: Volodymyr Paprotski vpaprots@ca.ibm.com

images/base/scripts/common/setup.sh
ghaskins, line 114: I am fine with leaving it as a FIXME if the only implication is the lack of protoc compilation.  I might have guessed there would be run-time implications as well w.r.t. the package installing .so files that the golang protobuf integration depended on if I didn't know better.  However, the evidence is to the contrary since we know it builds and can run the tests, etc.

ghaskins:
This LGTM.

Given that I worked closely with @vpaprots on its development, I should probably abstain from approving the merge unless directed otherwise by one of the other maintainers.

Note to @srderson some of the conditionals in the setup script are opting for simple apt-get because X is using Xenial as opposed to Trusty (s390x/ubuntu only has xenial images).  I mention this because, those conditionals could be collapsed when we move x86 to Xenial too, TBD.

srderson:
Could you add a "Building in Z' section to https://github.com/hyperledger/fabric/blob/master/docs/dev-setup/install.md and reference devenv/setupRHELonZ.sh?

devenv/setupRHELonZ.sh
```diff
@@ -0,0 +1,96 @@
+#!/bin/bash
+
+# To get started:
```
srderson, line 8: Could you add a comment explaining why this file exists? Something like "Development on Z is done on the native OS, not in Vagrant. This script can be used to set things up in RHEL on Z, similar to devenv/setup.sh which does the same for Vagrant. See https://github.com/hyperledger/fabric/blob/master/docs/dev-setup/install.md"

srderson:
We should not move "Building outside of Vagrant" and other build instructions to dev-env in this PR. If we want to make that change, please make a second issue/PR.

docs/dev-setup/install.md
```diff
@@ -107,13 +107,30 @@ go test github.com/hyperledger/fabric/core/container -run=BuildImage_Peer
 go test github.com/hyperledger/fabric/core/container -run=BuildImage_Obcca
 ```
 
-## Building outside of Vagrant <a name="vagrant"></a>
```
srderson, line 110: Could you add back the anchor tag? It should not be removed.

docs/dev-setup/install.md
```diff
@@ -107,13 +107,30 @@ go test github.com/hyperledger/fabric/core/container -run=BuildImage_Peer
...
-While not recommended, it is possible to build the project outside of Vagrant (e.g., for using an editor with built-in Go toolking). In such cases:
+## Building outside of Vagrant
+It is possible to build the project outside of Vagrant (e.g. when one already runs Linux and wants to develop natively). Generally speaking, one has to match the vagrant [setup file](https://github.com/hyperledger/fabric/blob/master/devenv/setup.sh).
```
srderson, line 111: This text seems to imply that it will only run natively on Linux. 
"when one already runs Linux and wants to develop natively"
That is not true. Maybe just "when one wants to develop or run natively"

docs/dev-setup/install.md
```diff
@@ -107,13 +107,30 @@ go test github.com/hyperledger/fabric/core/container -run=BuildImage_Peer
 go test github.com/hyperledger/fabric/core/container -run=BuildImage_Obcca
 ```
 
-## Building outside of Vagrant <a name="vagrant"></a>
-While not recommended, it is possible to build the project outside of Vagrant (e.g., for using an editor with built-in Go toolking). In such cases:
+## Building outside of Vagrant
```
srderson, line 110: Anchor tag should be here. `<a name="vagrant"></a>`

srderson:
LGTM. @vpaprots please help with cleanup on the few doc issues I noted.

PR #1526, opened 2016-05-19
Remove vestiges of consensus imports from statetransfer tests

jyellick:
## Description

This removes all consensus imports from statetransfer, and replaces them with the relevant peer import.
## Motivation and Context

The statetransfer code was modified to no longer depend on the consensus APIs, but the internal state transfer mock testing framework still relied on some of them.  Because the goal was to completely separate statetransfer from consensus, this dependency needed to be removed.

One unrelated change included is passing the mock ledger an instance of `*t.Testing` so that it can create a test failure, rather than simply panic-ing and dying.

No particular expertise is needed to review this changeset, it is generic go cleanup.

This is considered general cleanup, though was noticed during work on #1496 
## How Has This Been Tested?

Since this does not actually test production code, but only restructures the test themselves, rerunning the unit tests is sufficient.
## Checklist:

### DIFF
--- a/core/peer/statetransfer/statetransfer_mock_test.go
+++ b/core/peer/statetransfer/statetransfer_mock_test.go
@@ -24,9 +24,6 @@ import (
 	"sync"
 	"testing"
 
-	"github.com/golang/protobuf/proto"
-
-	"github.com/hyperledger/fabric/consensus"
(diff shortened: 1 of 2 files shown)


kchristidis:
(This is good to go. Slowly catching up with a huge backlog.)

PR #1527, opened 2016-05-20
Match statetransfer max transfer size to buffer

### LINKED ISSUE #1496: State transfer should never request more blocks than the buffer can handle

jyellick:
## Description

This changeset sets a maximum number of block blocks and state deltas to attempt to retrieve.  This value is equal to the buffer size for each respective type.

This PR builds on #1526 so to view the 'true diff', please look only at the second commit.

Review is welcome from all, though I'd ask @tuand27613 specifically to take a look.
## Motivation and Context

It has been observed that in times of high stress, a peer which initiates state transfer will not be able to read blocks as fast as they are retrieved, causing blocks to fall off the end of the buffer, and be lost.  This problem is partially addressed via #1498 and #1495, but more generally, after discussion with @jeffgarratt, state transfer should not be requesting more blocks/deltas than could potentially be stored in the buffers.  This is both an optimization for the reading and sending side, as a sender is necessarily unaware of a buffer overflow on the receiving side, and will therefore waste time sending blocks or deltas which will be discarded.

This fixes #1496 
## How Has This Been Tested?

Two unit tests have been added, one which tests the behavior for blocks only, while the other tests the behavior for both blocks and deltas.  This would be extremely difficult to test in a behave test.

Further, the issue_680 behave test on CI should test for any generic state transfer regression.
## Checklist:

### DIFF
--- a/core/peer/statetransfer/statetransfer.go
+++ b/core/peer/statetransfer/statetransfer.go
@@ -24,6 +24,8 @@ import (
 	"sync"
 	"time"
 
+	_ "github.com/hyperledger/fabric/core" // Logging format init
+
 	"github.com/hyperledger/fabric/core/ledger/statemgmt"
 	"github.com/hyperledger/fabric/core/peer"
 	"github.com/hyperledger/fabric/protos"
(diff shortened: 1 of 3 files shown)


ghaskins:
Excellent work as always @jyellick.  If we could get an SME or two to sign off on a reviewed-by, I can merge this.

jeffgarratt:
@jyellick @ghaskins looks good!!  This is a nice addition and a good pattern to follow in general wrt buffers and our async requests model.

srderson:
LGTM

PR #1529, opened 2016-05-20
Forward the host REST port to 5000 instead of 3000. Fixes #1528

### LINKED ISSUE #1528: Forward host REST port to 5000 instead of 3000
## Description

Forward the REST host port to 5000 instead of 3000 so that it matches to REST port defined by the peer. The different ports has led to confusion.
## Describe How to Reproduce

Start the peer in Vagrant and from your host try to access http://localhost:5000. There is no response because you must use http://localhost:3000 instead.

srderson:
## Description

This changes the host REST port from 3000 to 5000. This matches the peer REST port.
## Motivation and Context

The fact that the peer REST port used 5000, but if using Vagrant the host port was mapped to 3000 caused confusion. It was mentioned by @ghaskins and @bmos299 as a point of confusion. There is no reason why port 5000 cannot be used on both the Vagrant host and in the peer.
Fixes #1528 
## How Has This Been Tested?

Ran `make checks`, No new tests are required because this PR does not add new functionality.
## Checklist:

To maintainer: after review and before clicking merge
a) Send an email to the Fabric mailing list and post in the #Fabric channel in Slack
b) Contact me and ask me to do so.
In the message, we'll need to indicate that running `vagrant halt` `vagrant up` is necessary due to the port change.


docs/API/CoreAPI.md
```diff
@@ -136,7 +136,7 @@ For additional information on the available CLI commands, please see the [protoc
 
 You can work with the REST API through any tool of your choice. For example, the curl command line utility or a browser based client such as the Firefox Rest Client or Chrome Postman. You can likewise trigger REST requests directly through [Swagger](http://swagger.io/). You can utilize the Swagger service directly or, if you prefer, you can set up Swagger locally by following the instructions [here](#to-set-up-swagger-ui).
 
-**Note:** The REST interface port is defined as port 5000 in the [core.yaml](https://github.com/hyperledger/fabric/blob/master/peer/core.yaml). If you are sending REST requests to the peer node from inside Vagrant, use port 5000. If you are sending REST requests through Swagger, the port specified in the Swagger file is port 3000. The different port emphasizes that Swagger will likely run outside of Vagrant. To send requests from the Swagger interface, set up port forwarding from host port 3000 to Vagrant port 5000 on your machine, or edit the Swagger configuration file to specify another  port number of your choice.
```
ghaskins, line 139: Ah, at least we understand why it might have been that way to begin with.

core/rest/rest_api.json
```diff
@@ -5,7 +5,7 @@
         "description": "Interact with the enterprise blockchain through Hyperledger Fabric API",
         "version": "1.0.0"
     },
-    "host": "127.0.0.1:3000",
+    "host": "127.0.0.1:5000",
```
ghaskins, line 8: I'm a little confused by this change, but its probably because I don't know Swagger.  Did we have something listening on port 3000?  I was under the impression we were listening on port 5000 but Vagrant was simply remapping things.

docs/API/CoreAPI.md
srderson, line 139: Yes, but I think that having them on the same port leads to less confusion overall.

core/rest/rest_api.json
srderson, line 8: Swagger gives you a UI in the browser you can use to test out an API. Because it was assumed that you would run Swagger outside of Vagrant and that you would be testing against a peer running in Vagrant, the default port was set to 3000. Now that it's 5000 everywhere, it should work no matter what your setup is, unless you've customized the REST port. @angrbrd is the Swagger expert so she can double check that I'm not making stuff up :)

ghaskins:
This LGTM but I will wait for @angrbrd review, and then coordinate with @srderson for the red-flag changeover before merging.

srderson:
@angrbrd suggested swapping this fix so that 3000 is the default port on both sides. This would result in no impact for people running a peer in Vagrant and connecting to it from their host, but would impact people running the peer outside of Vagrant. Any thoughts @ghaskins or @mastersingh24?

ghaskins:
My tools already assume 3000, so this is the best case scenario for me.

+1

bmos299:
@srderson @angrbrd Why have an unnecessary port mapping?  When someone starts a peer and observes a log message for port 5000 and the core.yaml has port 5000, and they have to connect to Port 3000, that is opening the door for a permanent usability question.   If we change the Port to 5000, there will be a short-term issue for some people (agreed).  However, it is a long-term win.

ghaskins:
@bmos299 I think the proposal is to get rid of the remapping.  Now it's just a question of what is the port we want to use consistently across the environments.  I don't have a strong opinion, but picking 3000 is slightly less work for me since I already use that as the default in two tools.

bmos299:
@ghaskins thanks.  I think as a whole the community is more accustomed to 5000.  I think we are better off changing a mapping parm than the system (fabric) default.

angrbrd:
@bmos299 

@ghaskins is correct. The point of this PR is to remove the port re-mapping in order to eliminate any confusion on the client side. However, as @ghaskins also states, we already have applications hitting port 3000 from outside of Vagrant, which is why I am suggesting to keep that as the exposed port. That would remove any client side impact.

bmos299:
@angrbrd would it be best to open a new issue for this topic?  We have a lot of good detail here.  My point is I can't see changing a product default to accommodate tooling.   Many people (including the team in the lab) use port 5000 directly.

angrbrd:
@bmos299 I don't think the issue is accommodating tooling. The issue is disrupting existing applications that already use port 3000 as the default REST interface port. IMHO that would take priority over any tooling that we have. That being said, I would defer the answer to this question to @mastersingh24 @mrshah-at-ibm @masterDev1985 as that team has built legitimate applications on top of this port and unless they have changed it in the Bluemix deployment, it effects that service also. I would like to hear their opinion.

mrshah-at-work:
We use port 5000 on the peer as our API port. We'd like to keep it 5000.

If we change the port, an update to the this version will need an update in all the provisioning scripts for  the port mappings.

angrbrd:
@mrshah-at-ibm thanks. @srderson @ghaskins then I think it makes sense to keep the change as is, but send out a note to the Slack channel to alert people prior to this change going in. The change LGTM.

core/rest/rest_api.json
angrbrd, line 8: @srderson you're right. If we're remapping the host port to 5000, the Swagger UI would now be accessed on port 5000 on the host. So this change makes sense.

docs/API/SandboxSetup.md
```diff
@@ -70,7 +70,7 @@ The chaincode console will display the message "Received REGISTERED, ready for i
 
 #### **Note on REST API port**
 
-The REST interface port is defined as port 5000 in the [core.yaml](https://github.com/hyperledger/fabric/blob/master/peer/core.yaml). If you are sending REST requests to the peer node from inside Vagrant, use port 5000. If you are sending REST requests through Swagger, the port specified in the Swagger file is port 3000. The different port emphasizes that Swagger will likely run outside of Vagrant. To send requests from the Swagger interface, set up port forwarding from host port 3000 to Vagrant port 5000 on your machine, or edit the Swagger configuration file to specify another port number of your choice.
+The REST interface port is defined as port 5000 in the [core.yaml](https://github.com/hyperledger/fabric/blob/master/peer/core.yaml). If you are sending REST requests to a peer node running in Vagrant from outside of Vagrant, also use port 5000. This mapping is defined in [Vagrantfile](https://github.com/hyperledger/fabric/blob/master/devenv/Vagrantfile)
```
ghaskins: Minor nit: "peer node running in Vagrant from outside of Vagrant" seems to be missing punctuation.  I assume it was meant to say something like "peer node running in Vagrant.  From outside of Vagrant ..".  That said, I question whether we should even mention this or just leave the first sentence and drop the part of the vagrant at all.  I realize that someone that changes core.yaml would need to understand to also update the Vagrantfile, but not sure that needs/should-be mentioned here.

docs/API/CoreAPI.md
```diff
@@ -136,7 +136,7 @@ For additional information on the available CLI commands, please see the [protoc
...
+**Note:** The REST interface port is defined as port 5000 in the [core.yaml](https://github.com/hyperledger/fabric/blob/master/peer/core.yaml). If you are sending REST requests to a peer node running in Vagrant from outside of Vagrant, also use port 5000. This mapping is defined in [Vagrantfile](https://github.com/hyperledger/fabric/blob/master/devenv/Vagrantfile)
```
ghaskins: Same punctuation issue as noted below.

docs/API/SandboxSetup.md
srderson: How about

The default REST interface port is 5000. It can be configured in [core.yaml](https://github.com/hyperledger/fabric/blob/master/peer/core.yaml) using the `rest.address` property. If using Vagrant, the REST port mapping is defined in [Vagrantfile](https://github.com/hyperledger/fabric/blob/master/devenv/Vagrantfile).

ghaskins: +1

ghaskins:
LGTM, Thanks!

PR #1533, opened 2016-05-20
Build-system/devenv improvements

ghaskins:
This patch series attempts to further improve the build system / devenv.
## Description

This patch series introduces 4 modifications to our build and devenv systems.
- We move the download/build/install of go tools from the vagrant provisioner to the makefile
- We introduce ./scripts/provision/host.sh, intended to help provision host environments, that manages the installation of all packages common to hosts such as vagrant and travis.
-  We introduce a TLD file ./baseimage_release that contains the baseimage version.  This allows us to have one common definition of the version that is consumed by vagrant and docker alike, and allows us to depend on changes to the file to govern our makefile prerequisite triggers.
- We update the "make base-image" dependencies so that it can properly detect when it needs to be rebuilt.
## Motivation and Context

We have the goal of moving as much build-logic as possible away from vagrant and into common fabric infrastructure such that all environments (vagrant, travis, etc) can take advantage of updates.  To this point, the goal is to get the vagrant provisioner only minimally involved with external dependency installation and move everything else to either ./scripts/provision or the Makefile build.

As a side note to this goal, one interesting aspect of this change is that we can now have rules like "make linter" properly install its own dependencies (e.g. golint) without relying on vagrant.

We also had the goal to try to make the baseimage version definition be common between Docker and Vagrant to eliminate DRY violations.

And lastly, we wanted to make the build smarter about prerequisites to try to avoid needing the big hammer "make clean" for every git-pull.
## How Has This Been Tested?
- Vagrant destroy/up a fresh environment, 'make checks' runs without error.
- Ran various by hand checks such as touching baseimage prerequisites and ensuring that a rebuild was triggered.
- Pushed changes to public ghaskins/hyperledger-fabic tree and ensured TravisCI environment still worked.
## Checklist:


ghaskins:
@christo4ferris Of relevance to you is "make linter" will now ensure any go-tool that you add to $(GOTOOLS) is present before running.

@corecode Of relevance to you is that the baseimage build now has much smarter prerequisites like you asked for.

ghaskins:
@bcbrock Of relevance to you: tcl is now installed in the travis environment.

@binhn This has the common definition of the baseimage release ("0.0.9") that you asked for a while back.

ghaskins:
@rameshthoomu

bcbrock:
@ghaskins Would it make sense to consolidate all testing, linting etc. done by Travis into a single `make ci-test` target? That might make it easier to maintain, and easier for developers to run the suite before making a PR.

ghaskins:
@bcbrock I _think_ I may have captured that in "make checks", or at least that was my intention unless I am misunderstanding your point.  However, travis.yml still calls unit/behave/linter as discrete steps because I was trying to retain @rameshthoomu's log granularity.  I will defer to @rameshthoomu on whether he would like to retain the granularity of prefer to close the loop on that particular maintenance burden.

ghaskins:
Fixes #1539

bcbrock:
@ghaskins I didn't realize that `make checks` was there; Would have assumed Travis would be using it if it were.

ghaskins:
My rationale for leaving it discrete steps was that @rameshthoomu went through the trouble of making the travis log collapsable per section...e.g. each step in the process end up with its own log that can be optionally hidden/shown by the user in the UI.  If I converted travis to use "make checks", the entire process would run under one log session and partially defeat his setup.  I didn't want to be the one to propose to break that UI element, even though doing so violates DRY (make checks and travis have to be manually synchronized).  I don't have a strong opinion, as I understand both sides of the argument.

ghaskins:
@bcbrock related to your point, I would like to see us eventually advocate that "make checks" should be part of the PR checklist.  I see it as ultimately replacing the one we have now that explicitly lists goimports, golint, etc.  However, until we implement "make linter", it's too early and should remain as it is now.

Makefile
```diff
@@ -85,20 +108,37 @@ linter:
...
+	cp -r $(GOPATH)/src/github.com/hyperledger/fabric/vendor/github.com/golang/protobuf/ $(GOPATH)/src/github.com/golang/
+	go install github.com/golang/protobuf/protoc-gen-go
+	rm -rf $(GOPATH)/src/github.com/golang/protobuf
+
+# Default rule for gotools uses the name->path map for a generic 'go get' style build
+gotool.%:
+	$(eval TOOL = ${subst gotool.,,${@}})
+	go get ${go.fqp.${TOOL}}
+
+$(GOPATH)/bin/%:
```
ghaskins: I think I can do this part better.  Hold off on merging for now.

ghaskins: Nevermind, my experiment failed to produce a better approach.  Good to go as is.

srderson:
Would it make sense to move /baseimage_release to /images/base/release? Guess I don't fully understand why it's a top level file.

ghaskins:
@srderson this is a great suggestion, I admit it just simply didn't occur to me.  Will make the change and push an update.

srderson:
LGTM. Tested and confirmed this is working. Thanks.

PR #1547, opened 2016-05-21
Maintain white list of validating peers

### LINKED ISSUE #756: Consensus should run with whitelisted VPs when security is enabled

tuand27613:
## Description

Remove requirement of having to name pbft validating peers vp0,1,2,3,...
Keep a white list of the pbft N peers and use the index values of that list as IDs in pbft code
## Motivation and Context

Fixes #756
## How Has This Been Tested?

unit tests run. (fabric/consensus/obcpbft unit tests updated to use the white list methods)
behave test run except for @issue_1000 noted in #1466 . (consensus_4 docker-compose files in fabic/bddtests updated to use peers names other than vp0/1/2/3)
## Checklist:


tuand27613:
Please excuse the large number of commits.  I hung on to the code too long and kept having to do rebases along the way.

tuand27613:
@corecode  @jyellick  @kchristidis   please take a look. Thanks !

ghaskins:
@tuand27613 first off, let me say that I am very glad you are tackling this problem and I look forward to having this capability in the tree.

That said, I am a little concerned about the patch series presentation itself with the rebase records, etc (though I do appreciate the note with the explanation).  I reviewed the CONTRIBUTING.md/Github Flow documentation, and while there is nothing strictly in violation of the workflow advocated, I do worry about accepting a series like this unless it were intended to be a squash merge.  My first and biggest concern is that its not clear if someone were to land a bi-sect in the middle of this that they would end up with a sane tree.  My secondary concern is its really hard to figure out whats going on patch to patch as a reviewer.

I guess it really comes down to a debate between "gitflow branches with actual/natural history (merge warts and all)" and "curated patch series".  I am not sure what way the community wants to run this repository and I am not saying I would NACK this work for this reason if the SMEs and other community members felt it was fine.  I just wanted to voice my opinion that I am concerned about the general precedent of accepting PRs in this state and to start a dialog about what the policy should be going forward.

corecode:
I don't understand the development workflow that would create such resolve messages...  In any case, I think they should be squashed into their corresponding commits, or everything should be squashed together.

srderson:
+1 for squashing commits in this situation. We could consider turing on the squash feature in GitHub. https://github.com/blog/2141-squash-your-commits

tuand27613:
I am closing this PR and will rebuild  from Kostas' original branch. Stay tuned.

core/ledger/confighistory/mgr_test.go
```diff
@@ -167,6 +167,34 @@ func TestMgr(t *testing.T) {
...
 	})
+
+	t.Run("test-api-Remove()", func(t *testing.T) {
+		// remove ledgerid1 and verify ledgerid1 entries are deleted and ledger2 returns collection config as is
+		mgr.Remove("ledgerid1")
+		retriever1 := mgr.GetRetriever("ledgerid1", dummyLedgerInfoRetriever)
+		itr, err := retriever1.dbHandle.GetIterator(nil, nil)
+		require.NoError(t, err)
+		defer itr.Release()
+		require.False(t, itr.Next())
```
manish-sethi: - It's more maintainable if we write subtests such that they are independent of each other. For instance, if you change the order of the subtests, they should not break. The other subtests in this Test are for verifying the query results and hence a common set up that includes the pre-populated data works fine. However, this subtest removes the data. So, I would suggest to have a separate test for this one. In order to convey the scope, you may like to change the name of the existing test from `TestMgr` to `TestQueries`.
- In addition to test for iterator not returning any data, add assertions for high level queries as well - in a way, that's the real contract. The deletion is internal detail that could be implemented differently as well (e.g., setting a marker and actual deletion of data in background)

core/ledger/kvledger/history/db_test.go
```diff
@@ -409,6 +409,57 @@ func TestName(t *testing.T) {
...
+		require.NoError(t, err)
+		defer store.Shutdown()
+		bg, gb := testutil.NewBlockGenerator(t, ledgerid, false)
+		txid := util2.GenerateUUID()
+		simulator, _ := env.txmgr.NewTxSimulator(txid)
+		simulator.SetState("ns1", "key1", []byte("value1"))
+		simulator.SetState("ns2", "key2", []byte("value2"))
+		simulator.Done()
+		simRes, _ := simulator.GetTxSimulationResults()
+		pubSimResBytes, _ := simRes.GetPubSimulationBytes()
```
manish-sethi: Better to check for `require.NoError`

core/ledger/kvledger/history/db_test.go
```diff
@@ -409,6 +409,57 @@ func TestName(t *testing.T) {
...
+		block1 := bg.NextBlock([][]byte{pubSimResBytes})
+
+		historydbs[i] = env.testHistoryDBProvider.GetDBHandle(ledgerid)
+		require.NoError(t, store.AddBlock(gb))
+		require.NoError(t, historydbs[i].Commit(gb))
+		require.NoError(t, store.AddBlock(block1))
+		require.NoError(t, historydbs[i].Commit(block1))
```
manish-sethi: Nit: You can make this more readable by using a local variable instead of `historydbs[i]` here. After having populated the data, ouside this loop, you can use `env.testHistoryDBProvider.GetDBHandle(ledgerid)` to access specific ledger instance.

core/ledger/kvledger/history/db_test.go
```diff
@@ -409,6 +409,57 @@ func TestName(t *testing.T) {
...
+		testutilVerifyResults(t, historydbQEs[i], "ns2", "key2", []string{"value2"})
+	}
+
+	require.NoError(t, env.testHistoryDBProvider.Remove("ledger1"))
+
+	// verify ledger1 historydb has no entries and ledger2 historydb remains same
+	itr, err := historydbs[0].levelDB.GetIterator(nil, nil)
+	require.NoError(t, err)
+	defer itr.Release()
+	require.False(t, itr.Next())
```
manish-sethi: same comment as for the collection config mgr

core/ledger/kvledger/txmgmt/privacyenabledstate/db.go
```diff
@@ -113,6 +113,18 @@ func (p *DBProvider) Close() {
...
+// Remove removes channel-specific data from the statedb
+func (p *DBProvider) Remove(ledgerid string) error {
+	statedb, err := p.GetDBHandle(ledgerid, nil)
+	if err != nil {
+		return err
+	}
+	if err = statedb.Remove(); err != nil {
+		return err
+	}
+	return nil
```
manish-sethi: `return statedb.Remove()`

core/ledger/kvledger/txmgmt/privacyenabledstate/db_test.go
```diff
@@ -588,6 +590,79 @@ func generateLedgerID(t *testing.T) string {
 	return fmt.Sprintf("x%s", hex.EncodeToString(bytes))
 }
 
+func TestRemove_StateLevelDBDB(t *testing.T) {
+	env := &LevelDBTestEnv{}
+	env.Init(t)
+	defer env.Cleanup()
```
manish-sethi: If you notice, the existing tests in this package run the same test on both the environments. The main reason of maintaining that structure is to ensure the same behavior from both leveldb and couchdb. If you broke this pattern because of specific things that you wanted to check then I would like to explore moving them as a function on the env itself (see function `StopExternalResource()`).

core/ledger/kvledger/txmgmt/statedb/stateleveldb/stateleveldb_test.go
```diff
@@ -251,6 +251,52 @@ func TestImportStateErrorPropagation(t *testing.T) {
 	})
 }
 
+func TestRemove(t *testing.T) {
+	env := NewTestVDBEnv(t)
+	defer env.Cleanup()
+	channelName1 := "testremove"
+	channelName2 := "testremove-anotherchannel"
+
+	// create test data for channelName1 and channelName2
```
manish-sethi: If you look at other tests in this file, they follow a pattern such that the the actual test is defined at a common place to ensure same test on leveldb and couchdb. Only specific behavior related tests are present in this package

core/ledger/kvledger/ledger_data_remover.go
```diff
@@ -0,0 +1,67 @@
...
+		bookkeepingProvider: provider.bookkeepingProvider,
+		historydbProvider:   provider.historydbProvider,
+	}
+}
+
+// Remove removes all the ledger data for a given ledgerID (channelID), which includes
+// stateleveldb/statecouchdb, confighistorydb, bookkeeperdb, historydb, block index and blocks directory.
+// This function can be called multiple times for the same ledgerID. It is not an error if the ledger
+// does not exist. The data consistency and concurrency control will be handled outside of this function.
+func (r *ledgerDataRemover) Remove(ledgerID string) error {
```
cendhu: We could consistently use the term `Drop()` rather than using all the following `Remove()`, `Drop()`, and `DeleteAll()` inconsistently.

core/ledger/kvledger/ledger_data_remover.go
```diff
@@ -0,0 +1,67 @@
...
+		return err
+	}
+
+	if err = r.configHistoryMgr.Remove(ledgerID); err != nil {
+		logger.Errorf("failed to remove config historydb data for channel %s, error: %s", ledgerID, err)
+		return err
+	}
+
+	if err := r.bookkeepingProvider.Remove(ledgerID); err != nil {
+		logger.Errorf("failed to remove config bookkeepingdb data for channel %s, error: %s", ledgerID, err)
```
cendhu: nit. Just bookkeepingDB, no config.

core/ledger/kvledger/ledger_data_remover.go
```diff
@@ -0,0 +1,67 @@
...
+
+func newLederDataRemover(provider *Provider) *ledgerDataRemover {
+	return &ledgerDataRemover{
+		blkStoreProvider:    provider.blkStoreProvider,
+		statedbProvider:     provider.dbProvider,
+		configHistoryMgr:    provider.configHistoryMgr,
+		bookkeepingProvider: provider.bookkeepingProvider,
+		historydbProvider:   provider.historydbProvider,
+	}
+}
```
cendhu: We have the following methods on the `Provider`. 
Screenshot 2020-07-08 at 4 37 12 PM
To me, it is natural to add `Drop(ledgerID string) error` as a method on the `Provider` struct itself rather than passing the `Provider` here to create another struct called `ledgerDataRemover` and a `Remove(ledgerID string) error` method. 

If the goal of `ledgerDataRemover` is to work irrespective of peer status, i.e., online/offline, I am not sure about passing the `Provider` struct as an argument. Instead, we should pass a separate initializer struct holding the required list of providers.

core/ledger/kvledger/bookkeeping/provider.go
```diff
@@ -52,3 +54,14 @@ func (provider *provider) GetDBHandle(ledgerID string, cat Category) *leveldbhel
...
+
+// Remove removes channel-specific data from the config history db
+func (provider *provider) Remove(ledgerid string) error {
+	if err := provider.GetDBHandle(ledgerid, PvtdataExpiry).DeleteAll(); err != nil {
+		return err
+	}
+	if err := provider.GetDBHandle(ledgerid, MetadataPresenceIndicator).DeleteAll(); err != nil {
+		return err
+	}
+	return nil
```
cendhu: `return provider.GetDBHandle(ledgerid, MetadataPresenceIndicator).DeleteAll()`

core/ledger/kvledger/bookkeeping/provider_test.go
```diff
@@ -9,16 +9,36 @@ package bookkeeping
...
 func TestProvider(t *testing.T) {
 	testEnv := NewTestEnv(t)
 	defer testEnv.Cleanup()
 	p := testEnv.TestProvider
-	db := p.GetDBHandle("TestLedger", PvtdataExpiry)
-	assert.NoError(t, db.Put([]byte("key"), []byte("value"), true))
-	val, err := db.Get([]byte("key"))
-	assert.NoError(t, err)
-	assert.Equal(t, []byte("value"), val)
+	db1 := p.GetDBHandle("TestLedger", PvtdataExpiry)
```
cendhu: nit. can name `db1` and `db2` as `pvtdataExpiryDB` and `metadataIndicatorDB` to improve readability.

core/ledger/kvledger/bookkeeping/provider_test.go
```diff
@@ -9,16 +9,36 @@ package bookkeeping
...
+	require.NoError(t, db1.Put([]byte("key1"), []byte("value1"), true))
+	val, err := db1.Get([]byte("key1"))
+	require.NoError(t, err)
+	require.Equal(t, []byte("value1"), val)
+	db2 := p.GetDBHandle("TestLedger", MetadataPresenceIndicator)
```
cendhu: nit. an empty line before this statement for better grouping of statements.

core/ledger/kvledger/ledger_data_remover.go
```diff
@@ -0,0 +1,67 @@
...
+
+	if err = r.historydbProvider.Remove(ledgerID); err != nil {
+		logger.Errorf("failed to remove historydb data for channel %s, error: %s", ledgerID, err)
+		return err
+	}
+
+	if err = r.blkStoreProvider.Remove(ledgerID); err != nil {
+		logger.Errorf("failed to remove blockstore data for channel %s, error: %s", ledgerID, err)
+		return err
+	}
```
cendhu: Nit. Not related to this PR. `Channel`, `ChannelID`, `ledgerID`, `Chain` -- we need to decide on a single name and use it consistently.

core/ledger/kvledger/ledger_data_remover.go
```diff
@@ -0,0 +1,67 @@
...
+
+// Remove removes all the ledger data for a given ledgerID (channelID), which includes
+// stateleveldb/statecouchdb, confighistorydb, bookkeeperdb, historydb, block index and blocks directory.
+// This function can be called multiple times for the same ledgerID. It is not an error if the ledger
+// does not exist. The data consistency and concurrency control will be handled outside of this function.
+func (r *ledgerDataRemover) Remove(ledgerID string) error {
+	logger.Infof("Removing ledger data for channel %s", ledgerID)
+	var err error
+	if err = r.statedbProvider.Remove(ledgerID); err != nil {
+		logger.Errorf("failed to remove statedb data for channel %s, error: %s", ledgerID, err)
```
cendhu: We have decided to start using `logger.Errorw` in the ledger component. https://github.com/hyperledger/fabric/blob/4fc4674a79ef941cbe9c1034e2b9f790578819ce/core/ledger/pvtdatastorage/store.go#L816

core/ledger/kvledger/ledger_data_remover_test.go
```diff
@@ -0,0 +1,227 @@
...
+	"testing"
+	"time"
+
+	"github.com/hyperledger/fabric-protos-go/common"
+	"github.com/hyperledger/fabric/common/ledger/blkstorage"
+	"github.com/hyperledger/fabric/common/ledger/dataformat"
+	"github.com/hyperledger/fabric/common/ledger/testutil"
+	"github.com/hyperledger/fabric/common/ledger/util/leveldbhelper"
+	"github.com/hyperledger/fabric/core/ledger"
+	lgr "github.com/hyperledger/fabric/core/ledger"
```
cendhu: duplicate imports

core/ledger/kvledger/txmgmt/statedb/statecouchdb/statecouchdb.go
```diff
@@ -721,6 +722,28 @@ func (vdb *VersionedDB) Close() {
...
+		// do not drop metadataDB until all other dbs are dropped
+		if dbInfo.DBName == metadataDBName {
+			continue
+		}
+		if _, err := dropDB(vdb.couchInstance, dbInfo.DBName); err != nil && !strings.Contains(err.Error(), "not_found") {
+			logger.Errorf("Error dropping database for channel %s, namespace %s, error: %s", vdb.chainName, dbInfo.Namespace, err)
+			return err
+		}
+	}
+	if _, err := dropDB(vdb.couchInstance, metadataDBName); err != nil && !strings.Contains(err.Error(), "not_found") {
```
cendhu: This function has `Remove()`, `Drop()`, and `DeleteAll()`.

core/ledger/kvledger/bookkeeping/provider.go
```diff
@@ -28,6 +28,8 @@ type Provider interface {
 	GetDBHandle(ledgerID string, cat Category) *leveldbhelper.DBHandle
 	// Close closes the BookkeeperProvider
 	Close()
+	// Drop drops channel-specific data from the config history db
+	Drop(ledgerid string) error
```
cendhu: Super nit. `ledgerID` instead of `ledgerid` everywhere -- refer to line 28.

core/ledger/kvledger/history/db_test.go
```diff
@@ -409,6 +409,73 @@ func TestName(t *testing.T) {
...
+	for _, ledgerid := range []string{"ledger1", "ledger2"} {
+		store, err := provider.Open(ledgerid)
+		require.NoError(t, err)
+		defer store.Shutdown()
+		bg, gb := testutil.NewBlockGenerator(t, ledgerid, false)
+		txid := util2.GenerateUUID()
+		simulator, err := env.txmgr.NewTxSimulator(txid)
+		require.NoError(t, err)
+		simulator.SetState("ns1", "key1", []byte("value1"))
+		simulator.SetState("ns2", "key2", []byte("value2"))
```
cendhu: `require.NoError(t, simulator.SetState("ns1", "key1", []byte("value1")))`.

common/ledger/util/leveldbhelper/leveldb_provider_test.go
```diff
@@ -199,6 +199,24 @@ func TestBatchedUpdates(t *testing.T) {
 	}
 }
 
+func TestDrop(t *testing.T) {
```
cendhu, line 201: We need to move the code present in `TestDeleteAll()` here. This is because if we remove `return dbHandle.deleteAll()` in the future from `Drop()`, all tests in the pkg would pass but some tests might fail outside this pkg. As pkg tests have to be self-sufficient, it would be better to move `deleteAll()` test here.

core/ledger/confighistory/mgr_test.go
```diff
@@ -169,6 +169,66 @@ func TestMgr(t *testing.T) {
...
+	defer os.RemoveAll(dbPath)
+	mockCCInfoProvider := &mock.DeployedChaincodeInfoProvider{}
+	mgr, err := NewMgr(dbPath, mockCCInfoProvider)
+	require.NoError(t, err)
+	chaincodeName := "chaincode1"
+	maxBlockNumberInLedger := uint64(2000)
+	dummyLedgerInfoRetriever := &dummyLedgerInfoRetriever{
+		info: &common.BlockchainInfo{Height: maxBlockNumberInLedger + 1},
+		qe:   &mock.QueryExecutor{},
+	}
```
cendhu: This code has become repetitive. Not because of this PR alone. Already, we had a few repetitions in this file as we didn't have a testEnv. 

We had introduced `newTestEnvForSnapshot()` for snapshot testing to avoid this kind of repetition. We could introduce something similar for other tests too. Having said that, I do not want to block this PR for this change and can be done in a subsequent PR too (leave the decision to you). 

If we do that, we can make this test look more like https://github.com/hyperledger/fabric/blob/9a6c729563347609429f5250ddf41225b620010c/core/ledger/confighistory/mgr_test.go#L360 where each sub-test is independent and also helps in improving the readability.

core/ledger/confighistory/mgr_test.go
```diff
@@ -169,6 +169,66 @@ func TestMgr(t *testing.T) {
...
+			require.Equal(t, expectedConfig, retrievedConfig.CollectionConfig)
+			require.Equal(t, expectedHeight, retrievedConfig.CommittingBlockNum)
+		}
+
+		// drop again is not an error
+		require.NoError(t, mgr.Drop("ledger1"))
+	})
+
+	t.Run("test-Drop-db-closed", func(t *testing.T) {
+		mgr.Close()
```
cendhu: - If we change these sub-tests order, things would break. The current setting makes these sub-tests not really independent.
- The major reason for making it independent is to improve the readability and extendability. With the current setting, if we want to add a new sub-test in the future, we need to look into existing sub-tests to decide where to place them (definitely we will have to place them before this sub-test as it calls `mrg.Close()`). Consider a test function with 10s of non-independent sub-tests -- will introduce problem during extendability/reduces productivity.
- When we make each sub-test independent, we can directly go to the end of the function and add a new sub-test. 

with this, you can revisit all other tests in this PR.

core/ledger/confighistory/mgr_test.go
```diff
@@ -169,6 +169,66 @@ func TestMgr(t *testing.T) {
...
+
+	// Populate collection config versions
+	for _, ledgerid := range ledgerIds {
+		for _, committingBlockNum := range configCommittingBlockNums {
+			// for each ledgerid and commitHeight combination, construct a unique collConfigPackage and induce a stateUpdate
+			collConfigPackage := sampleCollectionConfigPackage(ledgerid, committingBlockNum)
+			testutilEquipMockCCInfoProviderToReturnDesiredCollConfig(mockCCInfoProvider, chaincodeName, collConfigPackage)
+			mgr.HandleStateUpdates(&ledger.StateUpdateTrigger{LedgerID: ledgerid, CommittingBlockNum: committingBlockNum})
+		}
+	}
```
cendhu: Only applicable for the first sub-test and not for the second one. However, the code is present in a common layer. This code needs to be moved to the first sub-test.

core/ledger/pvtdatastorage/store_test.go
```diff
@@ -712,6 +712,79 @@ func TestCollElgEnabled(t *testing.T) {
...
+	require.NoError(err)
+	require.Nil(retrievedData)
+
+	// pvt data retrieval for block 1 should return full pvtdata
+	retrievedData, err = store.GetPvtDataByBlockNum(1, nilFilter)
+	require.NoError(err)
+	for i, data := range retrievedData {
+		require.Equal(data.SeqInBlock, testData[i].SeqInBlock)
+		require.True(proto.Equal(data.WriteSet, testData[i].WriteSet))
+	}
```
manish-sethi, line 764: add a check  `len(retrievedData) > 0` before this. Else, this may pass even when there is no data.

core/ledger/kvledger/txmgmt/statedb/statedb.go
```diff
@@ -23,6 +23,8 @@ type VersionedDBProvider interface {
 	GetDBHandle(id string, namespaceProvider NamespaceProvider) (VersionedDB, error)
 	// Close closes all the VersionedDB instances and releases any resources held by VersionedDBProvider
 	Close()
+	// Drop drops the channel-specific ledger data
+	Drop(channeName string) error
```
manish-sethi: All over in the ledger code, `ledgerID/ id` is being used. Though, we understand that the channel name and `ledgerID` are the same but it may not be clear to someone who will look at this code for the first time. For instance, `GetDBHandle` function above uses `id`

core/ledger/confighistory/mgr_test.go
```diff
@@ -169,6 +169,66 @@ func TestMgr(t *testing.T) {
...
+		mgr.Drop("ledger1")
+		retriever1 := mgr.GetRetriever("ledger1", dummyLedgerInfoRetriever)
+		retrievedConfig, err := retriever1.MostRecentCollectionConfigBelow(math.MaxUint64, chaincodeName)
+		require.NoError(t, err)
+		require.Nil(t, retrievedConfig)
+		itr, err := retriever1.dbHandle.GetIterator(nil, nil)
+		require.NoError(t, err)
+		defer itr.Release()
+		require.False(t, itr.Next())
+
```
manish-sethi: Here and at other places, you can replace this by `retriever1.dbHandle.IsEmpty()`

PR #1555, opened 2016-05-23
Remove dead code paths from handler.go

jyellick:
## Description

This changeset removes the handling of transactions/queries from `handler.go` as it is now handled via `engine.go` (and does not come through the chat interface).  It also embeds the PeerHandler, to remove a number unnecessary pass through methods.
## Motivation and Context

Dead code is a burden to maintain and misleading, and should be removed.
## How Has This Been Tested?

Ran some behave tests locally to verify that this is indeed a dead code path.  CI should be sufficient to guarantee this is as well.

This should be easy to review, but I've discussed this some with @jeffgarratt as he introduced the new `engine.go` path and he confirmed that this path should be dead, so his signoff might be good.
## Checklist:

### DIFF
--- a/consensus/helper/handler.go
+++ b/consensus/helper/handler.go
@@ -19,13 +19,10 @@ package helper
 import (
 	"fmt"
 
-	"github.com/golang/protobuf/proto"
 	"github.com/op/go-logging"
 	"github.com/spf13/viper"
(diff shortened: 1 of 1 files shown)


corecode:
looks good

ghaskins:
LGTM

Issue #737, opened 2016-02-25
Test chaincode deploy for proper error messaging when chanicodes exists locally as well as in remote locations


2016Nishi:
Tried to send Deploy command using RestAPI from SoapUI/Postman using chaincodeexample_02 on a local peer without security setup as: [http://127.0.0.1:3000/devops/deploy]

{
  "type": "GOLANG",
  "chaincodeID":{
      "path": "https://github.com/openblockchain/obc-peer/openchain/example/chaincode/chaincode_example02"
  },
  "ctorMsg": {
      "function":"init",
      "args":["a", "100", "b", "200"]
  }
 }

Expected Result:

It to return a unique hash code for the sample request in response from request.

Actual Result:

Response from Request:
20:00:51.278 [rest] Deploy -> ERRO 013 {"Error": "Deploying Chaincode -- Error getting chaincode package bytes: Error generating hashcode: code does not exist Download failer open /opt/gopath/_usercode_/925645731/src/github.com/openblockchain/obc-peer/openchain/example/chaincode/chaincode_example02: no such file or directory"}

This was on today's build:  
394d5bed1e3a4e24efc6dec76413ee23f010c073

2016Nishi:
## Here are the combinations where Deploy responds successfully vs a failure in response:
1. With Security Enabled/Disabled : 
      having http:' , 'https' in the beginning of path value causes a failure.
2. With Security Enabled/Disabled: 
   "//github.com/openblockchain/obc-peer/openchain/example/chaincode/chaincode_example02"
   gets chaincode deployed successfully on peer.

Error Message:

## When security was enabled seeing similar kind of error message as in one without-security

20:57:10.105 [devops] getChaincodeBytes -> ERRO 014 Error getting chaincode package bytes: Error generating hashcode: code does not exist Download failer open /opt/gopath/_usercode_/189210238/src/github.com/openblockchain/obc-peer/openchain/example/chaincode/chaincode_example02: no such file or directory
20:57:10.106 [devops] Deploy -> ERRO 015 Error deploying chaincode spec: type:GOLANG chaincodeID:<path:"http://github.com/openblockchain/obc-peer/openchain/example/chaincode/chaincode_example02" > ctorMsg:<function:"init" args:"a" args:"100" args:"b" args:"200" > secureContext:"jim" confidentialityLevel:CONFIDENTIAL

2016Nishi:
[...]

PR #1556, opened 2016-05-23
Fix Issue 866 and more

### LINKED ISSUE #866: Tests assume `GOPATH` is one element and fail if it's not
```
2016/03/23 11:54:07 Chaincode support using peerAddress: 0.0.0.0:40303
getting deployment spec for chaincode spec: type:GOLANG chaincodeID:<path:"github.com/openblockchain/obc-peer/openchain/example/chaincode/chaincode_example01" > ctorMsg:<function:"init" args:"a" args:"100" args:"b" args:"200" > 
--- FAIL: TestExecuteDeployTransaction (2.00s)
        exectransaction_test.go:165: Error deploying <>: Error generating hashcode: code does not exist Download failer open /home/matthew/src/obc/Go:/home/matthew/src/Go_external_1.6/src/github.com/openblockchain/obc-peer/openchain/example/chaincode/chaincode_example01: no such file or directory
```

Whilst many people do run `GOPATH` as a single element, it is not a requirement, and best practise is not to rely on it being a single element. There are several successful ways of configuring `GOPATH`.

lehors:
## Description

This PR cleans up the way GOPATH is handled when deploying a chaincode so that when GOPATH contains multiple elements, double slashes, trailing spaces, etc, deployment can still succeed.
In case where GOPATH contains several elements only the first one is searched for the chaincode. So the chaincode must be found in the first element of GOPATH for deployment to succeed. The documentation has been updated to warn about this limitation.
## Motivation and Context

Fixes #866 as well as several edge cases that can be met when GOPATH contains unusual values.
## How Has This Been Tested?

Two new unit tests were added:
- one that sets GOPATH to contain multiple elements and a trailing slash
- one to test deployment of a chaincode retrieved over HTTP

**05/25/16 Note: The second one depends on an external source so it will be submitted as part of a separate PR.**
## Checklist:

### DIFF
--- a/core/chaincode/exectransaction_test.go
+++ b/core/chaincode/exectransaction_test.go
@@ -205,12 +205,14 @@ func TestGopathExecuteDeployTransaction(t *testing.T) {
 }
 
 // Test deploy of a transaction with a chaincode over HTTP.
+/*
 func TestHTTPExecuteDeployTransaction(t *testing.T) {
 	// The chaincode used here cannot be from the hyperledger repo
 	// itself or it won't be downloaded because it will be found
 	// in GOPATH, which would defeat the test
 	executeDeployTransaction(t, "http://github.com/lehors/fabric/examples/chaincode/go/chaincode_example01")
 }
+*/
 
 // Check the correctness of the final state after transaction execution.
 func checkFinalState(uuid string, chaincodeID string) error {
--- a/core/chaincode/exectransaction_test.go
+++ b/core/chaincode/exectransaction_test.go
@@ -190,9 +190,17 @@ func executeDeployTransaction(t *testing.T, url string) {
 	closeListenerAndSleep(lis)
 }
 
-
 // Test deploy of a transaction
 func TestExecuteDeployTransaction(t *testing.T) {
+	executeDeployTransaction(t, "github.com/hyperledger/fabric/examples/chaincode/go/chaincode_example01")
+}
+
+// Test deploy of a transaction with a GOPATH with multiple elements
+func TestGopathExecuteDeployTransaction(t *testing.T) {
+	// add a trailing slash to GOPATH
+	// and a couple of elements - it doesn't matter what they are
+	os.Setenv("GOPATH", os.Getenv("GOPATH") + string(os.PathSeparator) + string(os.PathListSeparator) + "/tmp/foo" + string(os.PathListSeparator) + "/tmp/bar")
+	fmt.Printf("set GOPATH to: \"%s\"\n", os.Getenv("GOPATH"))
 	executeDeployTransaction(t, "github.com/hyperledger/fabric/examples/chaincode/go/chaincode_example01")
 }
 
--- a/core/chaincode/exectransaction_test.go
+++ b/core/chaincode/exectransaction_test.go
@@ -206,6 +206,9 @@ func TestGopathExecuteDeployTransaction(t *testing.T) {
 
 // Test deploy of a transaction with a chaincode over HTTP.
 func TestHTTPExecuteDeployTransaction(t *testing.T) {
+	// The chaincode used here cannot be from the hyperledger repo
+	// itself or it won't be downloaded because it will be found
+	// in GOPATH, which would defeat the test
 	executeDeployTransaction(t, "http://github.com/lehors/fabric/examples/chaincode/go/chaincode_example01")
 }
 
--- a/core/chaincode/exectransaction_test.go
+++ b/core/chaincode/exectransaction_test.go
@@ -138,8 +138,7 @@ func closeListenerAndSleep(l net.Listener) {
 	time.Sleep(2 * time.Second)
 }
 
-// Test deploy of a transaction.
-func TestExecuteDeployTransaction(t *testing.T) {
+func executeDeployTransaction(t *testing.T, url string) {
 	var opts []grpc.ServerOption
 	if viper.GetBool("peer.tls.enabled") {
 		creds, err := credentials.NewServerTLSFromFile(viper.GetString("peer.tls.cert.file"), viper.GetString("peer.tls.key.file"))
@@ -174,7 +173,6 @@ func TestExecuteDeployTransaction(t *testing.T) {
 
 	var ctxt = context.Background()
 
-	url := "github.com/hyperledger/fabric/examples/chaincode/go/chaincode_example01"
 	f := "init"
 	args := []string{"a", "100", "b", "200"}
 	spec := &pb.ChaincodeSpec{Type: 1, ChaincodeID: &pb.ChaincodeID{Path: url}, CtorMsg: &pb.ChaincodeInput{Function: f, Args: args}}
@@ -190,6 +188,17 @@ func TestExecuteDeployTransaction(t *testing.T) {
 
 	GetChain(DefaultChain).Stop(ctxt, &pb.ChaincodeDeploymentSpec{ChaincodeSpec:spec})
 	closeListenerAndSleep(lis)
+}
+
+
+// Test deploy of a transaction
+func TestExecuteDeployTransaction(t *testing.T) {
+	executeDeployTransaction(t, "github.com/hyperledger/fabric/examples/chaincode/go/chaincode_example01")
+}
+
+// Test deploy of a transaction with a chaincode over HTTP.
+func TestHTTPExecuteDeployTransaction(t *testing.T) {
+	executeDeployTransaction(t, "http://github.com/lehors/fabric/examples/chaincode/go/chaincode_example01")
 }
 
 // Check the correctness of the final state after transaction execution.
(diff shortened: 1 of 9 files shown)


lehors:
The [ci skip] was only for the last commit which changes a comment and that was done after the tests passed.

ghaskins:
@lehors Thank you for the fix, this is sorely needed.

FYI, as a policy please don't ci-skip any patches that have .go files (or any source, really), even if the change appears to be innocuous such as comments.  We had issues earlier where the license header in the comments was skipped and then subsequently broke unit-tests b/c it changed the chaincode hash.  There could be classes of other mistakes too, such as unintentionally commenting out active logic, etc.   We therefore decided to never skip any source updates.  I think it's more conservative to let the CI system do extra work than the alternative.

That said, I agree with you that this looks harmless, so I would have made the same call if not aware of the policy.  I'll manually restart Travis just to be on the safe side.

lehors:
Sorry, didn't know.

ghaskins:
This LGTM, but it should probably be reviewed by @muralisrini or @jeffgarratt before merging.  Thanks @lehors

core/chaincode/platforms/golang/hash.go
```diff
@@ -138,7 +135,11 @@ func getCodeFromHTTP(path string) (codegopath string, err error) {
 	//     . more secure
 	//     . as we are not downloading OBC, private, password-protected OBC repo's become non-issue
 
-	env[gopathenvIndex] = "GOPATH=" + codegopath + ":" + origgopath
+	os.Setenv("GOPATH", codegopath+":"+origgopath)
+	// Get a copy of that new env for the go get command
+	env := os.Environ()
+	// and reset GOPATH to its original value
+	os.Setenv("GOPATH", origgopath)
```
muralisrini: @lehors the original code in old line 141 is correct. We don't want to override the systems GOPATH with the new Env (new lines 138-142). We want to just change the "GOPATH" element in the original "env" var locally to be passed to exec.Command.

core/chaincode/platforms/golang/hash.go
```diff
@@ -169,29 +169,21 @@ func getCodeFromHTTP(path string) (codegopath string, err error) {
...
-	env := os.Environ()
-	var gopath string
-	for _, v := range env {
-		if strings.Index(v, "GOPATH=") == 0 {
-			p := strings.SplitAfter(v, "GOPATH=")
-			gopath = p[1]
-			break
-		}
-	}
-
```
muralisrini, line 189: @lehors good catch removing copy paste from the HTTP method and simplifying, thanks!

core/chaincode/platforms/golang/hash.go
lehors: I understand. The result is the same though. I'm merely changing the env temporarily to get a copy ready for the exec.Command. Line142 resets the gopath so nothing has changed.

But this way I can take advantage of os.Getenv and os.Setenv rather than doing it the hard way with string manipulations. :-)

muralisrini: Right. However, concurrent requests will get different GOPATH results, won't they ?

ghaskins: @muralisrini makes a good point.  It's not clear from reading the golang library spec if this operation would be considered thread/go-routine safe.  If it only impacts the current go-routine, this is probably fine, but I am skeptical this is the case.

muralisrini: Right @ghaskins. And even if the Setenv is threadsafe, we would need a serial lock covering the two Setenvs.

lehors: Bummer. I was quite proud of my trick but I think you're right so,, I'll reverse that change.
Thanks!

core/chaincode/platforms/golang/hash.go
lehors: Ok, I think I found a happy medium. Reusing the old way of setting the env for the subcommand in a threadsafe manner while still ending up with a simpler piece of code.
And with the new commit, travis kicked in so that will also address any possible failure related to the comment I previously added with ci skip.

core/chaincode/platforms/golang/hash.go
lehors: Ok, I think I found a happy medium. Reusing the old way of setting the environment for the subcommand in a threadsafe manner while still ending up with a simpler piece of code.
And with the new commit the ci tests kicked in so that will address any possible problems with the addition of a comment I previously made with ci skip.

core/chaincode/platforms/golang/hash.go
```diff
@@ -135,11 +135,13 @@ func getCodeFromHTTP(path string) (codegopath string, err error) {
...
-	// Get a copy of that new env for the go get command
 	env := os.Environ()
-	// and reset GOPATH to its original value
-	os.Setenv("GOPATH", origgopath)
+	for i, v := range env {
+		if strings.Index(v, "GOPATH=") == 0 {
+			env[i] = "GOPATH=" + codegopath + string(os.PathListSeparator) + origgopath
+			break
+		}
+	}
```
muralisrini, line 150: If we are going to resort to using os.Environ, I'd say we just go with the original implementation (for this func) of one os.Environ call instead of a os.Getenv and a os.Environ ?

lehors, line 150: Ok, I submitted another change to address your point.

muralisrini:
LGTM...perhaps @ghaskins to do a last once-over ? Thanks, @lehors !

core/chaincode/exectransaction_test.go
```diff
@@ -192,6 +190,28 @@ func TestExecuteDeployTransaction(t *testing.T) {
...
+	fmt.Printf("set GOPATH to: \"%s\"\n", os.Getenv("GOPATH"))
+	executeDeployTransaction(t, "github.com/hyperledger/fabric/examples/chaincode/go/chaincode_example01")
+}
+
+// Test deploy of a transaction with a chaincode over HTTP.
+func TestHTTPExecuteDeployTransaction(t *testing.T) {
+	// The chaincode used here cannot be from the hyperledger repo
+	// itself or it won't be downloaded because it will be found
+	// in GOPATH, which would defeat the test
+	executeDeployTransaction(t, "http://github.com/lehors/fabric/examples/chaincode/go/chaincode_example01")
```
ghaskins, line 213: This seems like a liability to have a dependency on an external, unversioned URL.  I realize that there is a requirement to have it not be in hyperledger/fabric, but I wonder if we can either 
1. Use something more stable (e.g. hyperledger/somethingelse/v1.0), OR
2. Drop this test outright

I suppose if github.com/lehors/fabric ever stops working, we can remove the test then.

Thoughts? @srderson @muralisrini

muralisrini:
@ghaskins agreed, will be nice to point to something generic and stable. But would be ok with leaving it in here too (assume @lehors won't remove the chaincode). It does fill a void... ?

core/chaincode/exectransaction_test.go
lehors, line 213: @ghaskins I totally agree with the undesirable aspect of this dependency and it's only reluctantly that I put it in. Unless we are ready to change the current logic of the software (and there was a discussion a while ago that seems to end with agreement not to change that - see issue #737) there is no way around having _some_ external dependency.

Given that, I feel that depending on my fork is pretty safe. I have no reason to remove that any time soon and as long as I'm involved in the project, which I expect to remain true for the foreseeable future, it will keep being sync'ed with the main repo.
I do think it is better to have to live with that dependency than to have such a critical piece of code not part of CI.

ghaskins, line 213: Ok, that seems reasonable.  Is it possible to at least use a versioned URL rather than leaving it wide open?

E.g. https://github.com/lehors/fabric/tree/ca8135b1a0f95127c96478518628a65a377ec123/examples/chaincode/go/chaincode_example01

Or is there a reason this is a bad idea?

lehors, line 213: I thought about that too but decided to use the latest one because it's the closest to the ideal situation where we could simply use the one from the main repo.
Presumably, if we changed something in the fabric that required the examples to change we would want to use the latest version of the example rather than an old one.

ghaskins, line 213: The potential problem is one of repeatability.  If you couple a version of your repo, say X, to version A of fabric, then as long as your repo doesnt go away, A->X should be repeatable and deterministic.  Over time, your repo may progress to X->Y->Z while fabric progresses through A->B->C.  If we had a coupling like [A->X, B->Y, C->Z], then going back to A should always result in A->X coupling and work.  Otherwise, going back to A in the future may be A->Z and not work.  I'm not sure how important this all is in this case, but it seems like the more prudent thing to do.

lehors, line 213: I understand. It seems that both solutions have their pros and cons but I agree repeatability is important, and arguably more so than protecting ourselves against possible future changes which could be dealt with as part of those changes.

I very much care about having such a test be part of CI - I don't care so much how we get it done. I just wouldn't have felt comfortable touching that code without a way to test and get reasonably confident that I didn't break anything.

Given the implications maybe the right thing to do is for me to take it out of this PR and raise a separate issue and PR.

And since we're into fine-tuning this now, I think we should create a repo dedicated to holding that code, something like hyperleger/fabric-test, that we could more reliably depend on rather that someone else's repo - be mine or any other.

Every now and then or when necessary one will need to update it and update the test accordingly.

ghaskins:
This LGTM.  Thank you @lehors 

Issue #1331, opened 2016-05-03
ledger data become inconsistent across nodes after restarting a peer (using PBFT batch)


sachikoy:
I am testing on an environment with 4 VPs (vp0, .. vp3) with PBFT batch and security enabled.
When I stop one peer and restart it, the peer's state becomes inconsistent with other peers, yet all peers keep making consensus.

Here is the precise steps I did:
1. invoke several transactions on all of peers (vp0, .., vp3)
2. stop vp3, and make sure other nodes can still process transactions
3. start vp3, and make sure that vp3 can process transactions (and all other peers)
4. then vp3's the block height (from /chain REST API endpoint) is different from other peers.
5. If I query the state, vp3's result is different from other peers.

tuand27613:
it looks like vp3 has not done state transfer yet.  State transfer is triggered by a checkpoint event from the other peers and checkpoint is in turn  controlled by CORE_PBFT_GENERAL_K  in the ./consensus/obcpbft/config.yaml file. 

An example of the flow is in the behave test in file `./bddtests/peer_basic.feature` ( search for string "#680") .  We run this test through all 3 PBFT variations batch/classic/sieve  and do a similar flow to what you are doing:
- deploy
- invoke x times
- stop vp3
- invoke y times to get vp3 out of synch
- start vp3
- invoke z times
- check that vp0-3 have same blockchain height

corecode:
This should be addressed by #1000 and related PRs.

corecode:
This should be fixed in latest master.  Please confirm and close if you can no longer reproduce this bug.

sachikoy:
@corecode I confirmed that the issue is reproduced on the latest code, with the commit level - 80337e286d9c3a910346d3858e84e0729dc71e51  (this should be the version after the merge of PR #1325)

corecode:
Oh I see.  The crashed peer will eventually catch up.  You are not guaranteed to always read the most recent state on every peer.  To verify that, keep submitting transactions (say, 100).

sachikoy:
@corecode I tried 200 transactions but still the peers do not sync. Checkpoint period K=10 in hyperledger/fabric/consensus/obcpbft/config.yaml. BTW, security & privacy are enabled on our environment, if that would make a difference.

tuand27613:
@sachikoy could you run this behave test ?
in directory ./fabric/bddtests, run  `behave -n "#680" -D logs-y`

sachikoy:
[...]

PR #1557, opened 2016-05-23
Reject queries when out of date

jyellick:
## Description

This changeset adds a flag to `helper.go` which tracks whether consensus believes the state to be valid or not.  When consensus believes state is no longer valid, it calls `InvalidateState()`, once it believe the replica has caught up, it calls `ValidateState()`.  When the peer receives a query, it first checks whether the helper believes the state is valid or not, and it not, rejects the query because the state may be inconsistent.

Note, presently, the state is assumed to be good at startup, and is only invalidated by consensus at a later time.  We may wish to flip this assumption at some point in the future, but for now it is still an improvement.

Note also that this change intentionally does nothing to change the processing of transaction submission.  Even though a peer is out of date, it does not mean that it cannot submit new transactions to the network.
## Motivation and Context

Before this change, clients would receive old state with no indication that there was a problem, even if the peer knew it was thousands of blocks out of date.  After discussions with @binhn and @jeffgarratt it was concluded that queries should be rejected in this scenario.

This is intended to partially address issue #1091.
## How Has This Been Tested?

This has been tested locally with behave test issue_680 and with a new behave test tagged as issue_1091.

For the new test, it gets a peer (vp3) into an inconsistent state, then queries, and verifies that an error is returned.  The test for issue_680 covers that once state transfer has completed that the 'invalid' flag is unset, and queries are replied to properly once more.

Relying on CI to execute the other tests.
## Checklist:

### DIFF
--- a/bddtests/steps/peer_basic_impl.py
+++ b/bddtests/steps/peer_basic_impl.py
@@ -231,6 +231,10 @@ def step_impl(context):
     assert context.transactionID != ""
     pass
 
+@when(u'I unconditionally query chaincode "{chaincodeName}" function name "{functionName}" on "{containerName}"')
+def step_impl(context, chaincodeName, functionName, containerName):
+    invokeChaincode(context, "query", functionName, containerName)
+
 @when(u'I query chaincode "{chaincodeName}" function name "{functionName}" on "{containerName}"')
 def step_impl(context, chaincodeName, functionName, containerName):
     invokeChaincode(context, "query", functionName, containerName)
@@ -411,9 +415,15 @@ def step_impl(context, chaincodeName, functionName):
         responses.append(resp)
     context.responses = responses
 
+@when(u'I unconditionally query chaincode "{chaincodeName}" function name "{functionName}" with value "{value}" on peers')
+def step_impl(context, chaincodeName, functionName, value):
+    query_common(context, chaincodeName, functionName, value, False)
 
 @when(u'I query chaincode "{chaincodeName}" function name "{functionName}" with value "{value}" on peers')
 def step_impl(context, chaincodeName, functionName, value):
+    query_common(context, chaincodeName, functionName, value, True)
+
+def query_common(context, chaincodeName, functionName, value, failOnError):
     assert 'chaincodeSpec' in context, "chaincodeSpec not found in context"
     assert 'compose_containers' in context, "compose_containers not found in context"
     assert 'table' in context, "table (of peers) not found in context"
@@ -438,14 +448,12 @@ def step_impl(context, chaincodeName, functionName, value):
         request_url = buildUrl(context, container.ipAddress, "/devops/{0}".format(functionName))
         print("{0} POSTing path = {1}".format(currentTime(), request_url))
         resp = requests.post(request_url, headers={'Content-type': 'application/json'}, data=json.dumps(chaincodeInvocationSpec), timeout=30, verify=False)
-        assert resp.status_code == 200, "Failed to POST to %s:  %s" %(request_url, resp.text)
+        if failOnError:
+            assert resp.status_code == 200, "Failed to POST to %s:  %s" %(request_url, resp.text)
         print("RESULT from {0} of chaincode from peer {1}".format(functionName, container.containerName))
         print(json.dumps(resp.json(), indent = 4))
         responses.append(resp)
     context.responses = responses
-
-
-
 
 @then(u'I should get a JSON response from all peers with "{attribute}" = "{expectedValue}"')
 def step_impl(context, attribute, expectedValue):
--- a/bddtests/peer_basic.feature
+++ b/bddtests/peer_basic.feature
@@ -495,7 +495,7 @@ Feature: lanching 3 peers
             | vp3  |
         And I wait "15" seconds
 
-        # Invoke 6 more txs, this will trigger a state transfer, set a target, and execute new outstanding transactions
+        # Invoke 10 more txs, this will trigger a state transfer, set a target, and execute new outstanding transactions
         When I invoke chaincode "example2" function name "invoke" on "vp0" "10" times
 			|arg1|arg2|arg3|
 			| a  | b  | 10 |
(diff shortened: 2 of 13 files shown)


bcbrock:
@jyellick I am going to poke at the checkbox that says no new documentation is required. This seems like a major change to the semantics of the network that would require a significant documentation effort. What will this rejected query look like from a user's perspective? (I'm thinking a low-level user like me who is calling the REST API directly). How do we distinguish this failure from any other?  Is there a parameter to control "how far out of date" a query has to be to be rejected? I also assume it now opens the can of worms that queries on a node may _never_ work, which also needs to be documented.

kchristidis:
Having just reviewed the code for this PR, I'll venture a couple of answers before @jyellick chimes in.

> What will this rejected query look like from a user's perspective? (I'm thinking a low-level user like me who is calling the REST API directly).

The user will get the following message: `Error: state may be inconsistent, cannot query`

> How do we distinguish this failure from any other?

This error message is unique to this case.

> Is there a parameter to control "how far out of date" a query has to be to be rejected?

It's not an issue of "how far out of date"; it's an issue of whether the node realizes it has fallen behind, which happens when it hears about f+1 checkpoints above its high watermark. I assume that a smaller checkpoint period means that, in theory, the chances of you finding out that you're falling behind are higher. (Of course, a checkpoint period of 1 if you're partitioned won't help you.)

bddtests/steps/peer_basic_impl.py
```diff
@@ -231,6 +231,10 @@ def step_impl(context):
     assert context.transactionID != ""
     pass
 
+@when(u'I unconditionally query chaincode "{chaincodeName}" function name "{functionName}" on "{containerName}"')
```
kchristidis, line 234: Does this look identical to the "I query chaincode" step definition or am I missing something? (And I guess you break it down to a separate step because you want to keep the semantics right?)

bcbrock:
@kchristidis Thanks - but what I was suggesting is that those answers need to be added at least to the core API document before this PR should be accepted.

jyellick:
@bcbrock I'm by no means opposed to adding, as a future enhancement to the message format, something to indicate that a replica is behind.  This changeset uses the generic "Error" response, and I disagree that this PR is incomplete without a more sophisticated reporting mechanism.

Although @kchristidis is correct when he says you can distinguish this error from the others by that string, pattern matching against an English language arbitrary string does not seem like the sort of thing we want defined in our API. It may be fine for code correctness testing, but to actually make this consumable from an API perspective, I think we would need to introduce some broader concept or error codes, or at the very least classification (say, transient, or fatal).

As I see it, the only existing contract we have for the query interface is the binary "OK and here is your value" or "ERROR and here is an English string as to why we could not get you your value".  Today, a replica which knows that it is out of date replies to queries with "OK" and that is a bug.  It needs to reply with "ERROR" and that is what this changeset does.  Since this does not modify the existing query contract, I'm not sure what documentation would be modified.  To my knowledge, we do not enumerate all of the possible error strings that could be returned by a query, if I'm mistaken, please correct me, and I'll happily document this there.

bddtests/steps/peer_basic_impl.py
jyellick, line 234: This is nearly identical, essentially the "unconditionally" keyword acts as a way to more readable set the boolean `failOnError` in the now refactored `query_common` function.  Since there are only two possible values for that boolean, it did make sense to me to add it as a more formalized bracketed parameter (and this would have caused a larger changeset in the `peer_basic.feature` file).

kchristidis:
@bcbrock: I get you, point taken.

To the committers: @bcbrock's documentation comment aside, the code on this one looks good to go.

corecode:
Shouldn't state transfer set this flag, instead of the consensus plugin?  I.e. when state transfer gets triggered, state transfer signals this fact with the ledger; when state transfer is done (with all outstanding transfer requests), it removes the flag from the ledger.  This has the advantage of keeping the staleness metadata where the data lives, and other routines that may query the ledger (other code paths, local code querying ledger) will receive the same information about staleness.

Also, although I like the simplicity of this solution, I wonder whether it is the right one.  We're trying to build a highly available system - we should not have to reject queries just because we are bulk transferring blocks from elsewhere.

Multiple solutions come to my mind:
- block query until state transfer finishes (not much better)
- return local state, together with metadata on how stale it is (local block height, known block height)
- forward query to other, more up to date peer (either locally, or return a "redirect")

bcbrock:
This PR is not ready to be committed. As @corecode points out, this should be part of a comprehensive architecture for the semantics of queries. What is the urgency behind adding this one piece of that architecture now? 

@jyellick has drawn attention to the fact that the chaincode APIs and documentation need to be improved. I agree with him that it not his responsibility to rearchitect and redocument the chaincode APIs. However an improvement in API error reporting needs to be implemented before this change should be accepted. What is being proposed here is that a syntactically valid query, targeting a deployed and fully functional chaincode will be rejected, under the assumption  that it might produce the incorrect answer. This is a _bug_, not a feature, unless the APIs allow the caller to definitively interpret this rejection as a recoverable error, with options on how to recover (e.g., @corecode suggestions).

Finally, based on @kchristidis explanation, this behavior is triggered as a function of a "watermark". The word "watermark" does not appear in any file in the fabric/docs directory. How is a user supposed to understand and interpret this behavior? So I still believe there is a documentation component missing here.

jyellick:
@corecode With regards to state transfer setting this flag, I did consider this as an option, but ultimately, decided this needed to be done from consensus for a number of reasons.  First off, state transfer being complete, does not indicate that the replica is up to date, it simply means that state transfer has reached a target consensus has supplied to it.  Only consensus knows whether this target is correct or not.  That is why it is the consensus's responsibility to validate the state as current, and you can see in the PBFT code, that this is done after state transfer completes.  I did not include it in this changeset, but my intention is to add code to PBFT to track minimum acceptable block heights, so that it does not incorrectly mark the state as valid, simply because state transfer has completed.

With regards to the solutions you outlined:
- Blocking until state transfer completes : This is simply not practical, there is no upper bound on the amount of time a replica takes to become current.
- return local state with a flag indicating how stale it is : This was my first inclination, but after more thought, I realized this too is simply no good.  It is possible that the state is not only stale, it could be completely incorrect (anytime we must drop state and retrieve a full state snapshot).  Perhaps these conditions could be differentiated, and there's no reason this could not be modified in the future, but it is a significantly more complex solution.
- Forward the query to another more up to date peer : This is yet another solution that was considered, and I wouldn't rule out doing it in a later PR.  It would be simple to plug in here, simply instead of replying with an error, forward the request.  The key to this PR is simply the framework to recognize when we should not be replying.

jyellick:
@bcbrock I suppose the 'urgency' is simply attempts to address the confusion caused by the current behavior, as seen in issues like #1091 and #1331 .

As I pointed out to @corecode, this is not merely an issue of stale data, the data might be complete gibberish, and in fact, your query might fail today because we simply don't realize your chaincode even exists.  I would argue, we are simply making the behavior more predictable.

With the APIs defined as they are today, we have two options.
1. When we know we are out of date, pretend that nothing's wrong, and reply with sometimes correct, sometimes arbitrarily old, sometimes incorrect, and sometimes error.
2. When we know we are out of date, acknowledge that we can make no promises about our reply, and consistently return error.

You say it's a bug to return an error when our answer _might_ be correct.  I can certainly see the validity of this argument, however, I would say it's a much more grievous bug to return success when our answer _might_ be wrong.

If it would be a compromise that would allow this PR to be processed, I'd be willing to implement the "forward to another peer" behavior, rather than replying with error.  I'd have preferred to have done that in a separate PR, to minimize the diff and make review easier, but I will do what is needed.

corecode:
I don't think we should fix something we haven't even decided whether it is broken.

We're facing two separate issues:
- stale data.  we need to define what is means to be stale, and how to deal with it, both for queries, and for internal use.
- incorrect data or incorrect replies (#1091).  incorrect data must never happen.

jyellick:
This work was originally undertaken with urging from @binhn, maybe he can be more articulate than I as to why this is broken.

binhn:
@corecode This is to address part of #1331, where inconsistent state was observed from querying different peers.

This solution is a stopgap, and I think a better solution and more time could be had. For example, we could add another status code to the `Response` object (similar to HTTP codes) to give more detail than just 500 or 200. Checking the error code would be much easier for the application than the error msg.

```
message Response {
    enum StatusCode {
        UNDEFINED = 0;
        SUCCESS = 200;
        FAILURE = 500;
    }
    StatusCode status = 1;
    bytes msg = 2;
}
```

Someone could help defining and documenting all potential error codes and messages. This is certainly the work of a protocol, which I am working on setting that up.

corecode:
Even as a stopgap, I don't like the solution.  I think this should be fixed in the ledger.

jyellick:
@corecode I'm not sure what you mean by 'fixed in the ledger', whether the peer is up to date or not is something only consensus can be aware of.

corecode:
Yes, but #1331 is because the ledger returns invalid blocks.  This is something the ledger needs to handle correctly.

jyellick:
@corecode I think you've got it backwards, #1091 is because the ledger returns invalid blocks, and this should be fixed separately.  #1331 is because clients expect all replicas to return the same data, and one is behind, and returns old data.  This is more appropriately framed in the context of #1331, absolutely, the #1091 ledger bug should be fixed additionally.

corecode:
If we say "old data", we need to define what "old data" means.

I don't think that returning an error is preferable over returning old data.

jyellick:
@corecode Today, there is no promise that the data returned will be 'old', there are state transfer scenarios where the data could be 'wrong'.  Returning 'wrong' data, and reporting it as successful is a serious bug, and I think it's worth overzealously returning error for 'stale' data as well.

corecode:
Incorrect data must not returned.  In this case, we need to return a 5xx HTTP error code, and not an inline error.

Before we start returning errors for old data, we should decide what "old" means.  Peer performing state transfer?  Peer not caught up to weak checkpoint?  Peer not done executing most recent block?  Peer not done executing most recently proposed request?  Where is the line, and why?

I think staleness is a fundamental issue and needs to be addressed conceptually first.

bcbrock:
As noted in #1566 I would also want to see at least a dozen or so 5XX codes defined so that I could differentiate what happened to better help the user understand what to do next in the event of an error.

jyellick:
I've opened #1608 which describes the scenario in which the queries can return inconsistent or wrong data.  With some ledger enhancements, we should be able to eliminate that scenario, but I believe this PR is necessary as a stopgap measure.

I've opened #1607 above to continue this discussion for a more complete handling of the 'stale data' issue.

My vote is to merge this as it stands.  Without it, we risk returning wrong data, which is in my opinion serious enough to warrant a stopgap measure.

srderson:
@bcbrock @corecode do you agree with @jyellick that this is ready to merge or are you still debating? Wasn't sure based on your last comments. Thanks.

bcbrock:
@srderson @corecode @jyellick I understand better from a conversation with Jason today why this change is required. I agree that without this or other fixes there is a chance of a query returning an answer that never was, is or could be correct in any state. In an ideal world this and related issues would be addressed at once along with a better error reporting mechanism. But here we are...

Bottom line: I am OK with merging this.

tuand27613:
code LGTM.

can we add an entry to the FAQ to describe the error message and the issue ?

jyellick:
@tuand27613 Did you have a particular part of the FAQ this should go in? Don't currently see any reference to queries there.

tuand27613:
I copy-edited my original comment ... must learn to type and read.

[Usage FAQ](https://github.com/hyperledger/fabric/blob/master/docs/FAQ/usage_FAQ.md) seems like a good spot.

jyellick:
@tuand27613 Could you verify you're happy with the doc change in e3f1cfb?

srderson:
Thanks, LGTM

PR #1571, opened 2016-05-24
Update consensus_FAQ (language)

ghost:
Tightening up the grammar and readability of the answers. 
## Description

Same as above. There's nothing wrong with the answers as they were per se (beyond grammatical issues), but improvements in language can only help. 
## Motivation and Context

Helping out where I can. 
## How Has This Been Tested?

n/a
## Checklist:

Signed-off-by:
Joe Alewine (jalewine@gmail.com)

Cleaning/tightening up a few things here and there.

### DIFF
--- a/docs/FAQ/consensus_FAQ.md
+++ b/docs/FAQ/consensus_FAQ.md
@@ -2,15 +2,13 @@
 
 &nbsp;
 ##### Which Consensus Algorithm is used in the fabric? 
-The fabric is built on a pluggable architecture such that developers can configure their deployment with the consensus module that best suits their needs. The initial release package will offer three consensus implementations for users to select from: 1) No-op (consensus ignored); 2) Classic PBFT; and 3) SIEVE (an enhanced version of classic PBFT). 
+The fabric is built on a pluggable architecture to enable developers to configure deployment with the consensus module that best suits their needs. The initial release package will offer three consensus implementations for users to select from: 1) no-op (consensus ignored); 2) classic PBFT, and 3) SIEVE (an enhanced version of classic PBFT). 
 
 &nbsp;
 ##### What is the SIEVE Consensus Algorithm?
-There are many limitations in existing consensus algorithms, especially for solving challenges around security, performance, efficiency, and scalability. To solve these problems and make blockchain ready for business, the team has been researching several approaches to improve the existing algorithms. 
+To manage the challenges presented by the security, performance, efficiency and scalability required of a for-business consensus algorithm, the team has been researching several approaches. One of them, called “SIEVE”, is a consensus algorithm inspired by classic PBFT [Castro and Liskov, OSDI’99] and the Eve consensus protocol [Kapritsos et al., OSDI’2012]. 
 
-One such improvement is “SIEVE”, a consensus algorithm inspired by classic PBFT [Castro and Liskov, OSDI’99] and the Eve consensus protocol [Kapritsos et al., OSDI’2012]. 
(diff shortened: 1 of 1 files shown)


srderson:
LGTM. @joshhus any comments?

docs/FAQ/consensus_FAQ.md
```diff
@@ -2,15 +2,13 @@
...
-
-A research paper will be published by IBM Research to describe this new algorithm in further detail, and we will continue to make further research investments in the subject and share our results with the community.
+SIEVE is derived from PBFT in a modular way (inspired by ideas described in [Aublin et al., TOCS'15]) by reusing the PBFT view-change protocol. This lowers SIEVE's complexity and avoids having to develop a new consensus protocol from scratch. A paper will be published by IBM Research to describe this new algorithm in further detail. We're committed to making further research investments in the subject and sharing our results with the community.
```
christo4ferris, line 13: if this paper has yet to be published, why is this still referencing the intent? The pov of this statement is as if it were written by IBM, not the hyperledger project. It needs to be presented from the POV of Hyperledger.

christo4ferris, line 13: for expedience, suggest that the last two sentences be omitted.

tuand27613, line 13: The Sieve paper is published. It's reference [7] in the [protocol-spec](https://github.com/hyperledger/fabric/blob/master/docs/protocol-spec.md) doc.

christo4ferris, line 13: @joe-alewine @tuand27613 then the FAQ should be changed to link to the paper, and really should be written from the POV of this project.

ghost, line 13: In that case I think it would be reasonable to delete the offending two sentences and then create an issue about making broader changes to the answer to fit the paper on Sieve.

kchristidis:
This is definitely a step in the right direction (thanks @joe-alewine!) but (a) I agree with @christo4ferris's additional comments, and (b) I find the entire wording in this FAQ problematic to begin with.

My vote is for the following process:
1. Remove this page from the `docs` entirely
2. Spend some time to turn the [Consensus wiki](https://github.com/hyperledger/fabric/wiki/Consensus) in the central point of reference for all things consensus; I started doing this a few hours ago, and have much more work ahead of me)
3. Once the content in the Wiki is in good-enough shape, we can move some of that content over to the `docs` folder*

As things stand right now, the consensus-related documentation on the docs is either stale, or suffers from vague language that is geared towards business folks. I believe the audience here is developers and end users.

(*) FWIW I think that this is madness as it'll mean keeping two copies of information current. I vote for just maintaining the Wiki (which is fast and easy to work with... this stuff matters) and removing the `docs` folder entirely, but I understand that this is not the way most big projects work, and this suggestion will be shot down :grin:

kchristidis:
Thoughts on this? ^^ https://github.com/hyperledger/fabric/pull/1571#issuecomment-223749205

joshhus:
Deleting the "We're committed to ... " sentence should solve the pov issue. 

(HL docs generally need to be scrubbed by the authors and/or the community, for prospective goals vs. how it will work at the next HL "release" ...  and for IBM vs. Linux Foundation pov.)

christo4ferris:
@joe-alewine where does this stand? Have we transferred this to Gerrit? Should we just close?

ghost:
@christo4ferris After the comments made by @kchristidis I was left unsure where this stood. Making decisions on those points is beyond my role here. I'm happy to make the edits you requested but it was left ambiguous whether it was necessary or whether there was going to be some kind of systemic change not just to this particular documents but to the FAQ section in general.

christo4ferris:
Ok, I am closing then. I think @kchristidis had the right idea. Thanks!

ghost:
No problem. Happy to help.

PR #1576, opened 2016-05-25
Adding deprecation warning messages for /devops endpoint

### LINKED ISSUE #1101: Deprecated devops REST API ->  chaincode REST API.  What's the transition plan?
I am in the process of converting the obcsdk to use the new POST /chaincode API instead of the 
/devops/<op> API.  

The old API works fine, which started me wondering how long it will be supported?   Forever?  For a couple weeks? 
This has bearing on things important to testers.  For example, should we write and maintain a regression bucket for the deprecated API?  Should I dual-flavor obcsdk (let you choose your API)?  Or should we write devops out of obcsdk, and stop testing devops entirely?  Or will we test them forever?

Here is something that happened in PureApplication: a deprecated process was supported for two releases,  but every time the deprecated interface was used, a log message would appear warning you to modernize the call.  After two releases, the old calls simply failed. 

Barry Mosakowski, Nishi Nidamarty and I discussed this.  We decided a defect is the best way to initiate dialogue about the transition strategy from the old to the new.

angrbrd:
## Description

Adding three warning level messages that will appear in the peer logs whenever the deprecated /devops endpoint receives a REST request. Specifically, whenever the /devops/deploy, /devops/invoke, or /devops/query endpoints receive a request the below messages will be seen in the logs.

> [rest] Deploy -> WARN 017 /devops/deploy endpoint has been deprecated. Use /chaincode endpoint instead.
> [rest] Invoke -> WARN 01b /devops/invoke endpoint has been deprecated. Use /chaincode endpoint instead.
> [rest] Query -> WARN 01e /devops/query endpoint has been deprecated. Use /chaincode endpoint instead.
## Motivation and Context

This change is necessary to warn users of the REST API that the /devops endpoint is being deprecated and is now superceeded by the /chaincode endpoint. Users should transition to using the /chaincode endpoint as the /devops endpoint will eventually be removed.

Fixes #1101
## How Has This Been Tested?

I tested this change by sending requests to the /devops/deploy, /devops/invoke, and /devops/query endpoints and confirming that the correct warning messages are seen in the peer logs. This issue does not require new test cases as the code change simply adds new logging messages, not new functionality.
## Checklist:

Adding a warning message whenever the deprecated /devops endpoint
receives a REST request.

Fixes #1101.

### DIFF
--- a/core/rest/rest_api.go
+++ b/core/rest/rest_api.go
@@ -669,6 +669,7 @@ func (s *ServerOpenchainREST) GetTransactionByUUID(rw web.ResponseWriter, req *w
 // blockchain.
 func (s *ServerOpenchainREST) Deploy(rw web.ResponseWriter, req *web.Request) {
 	restLogger.Info("REST deploying chaincode...")
+	restLogger.Warning("/devops/deploy endpoint has been deprecated. Use /chaincode endpoint instead.")
 
 	// Decode the incoming JSON payload
 	var spec pb.ChaincodeSpec
@@ -812,6 +813,7 @@ func (s *ServerOpenchainREST) Deploy(rw web.ResponseWriter, req *web.Request) {
 // Invoke executes a specified function within a target Chaincode.
 func (s *ServerOpenchainREST) Invoke(rw web.ResponseWriter, req *web.Request) {
 	restLogger.Info("REST invoking chaincode...")
+	restLogger.Warning("/devops/invoke endpoint has been deprecated. Use /chaincode endpoint instead.")
 
 	// Decode the incoming JSON payload
 	var spec pb.ChaincodeInvocationSpec
@@ -948,6 +950,7 @@ func (s *ServerOpenchainREST) Invoke(rw web.ResponseWriter, req *web.Request) {
 // Query performs the requested query on the target Chaincode.
 func (s *ServerOpenchainREST) Query(rw web.ResponseWriter, req *web.Request) {
 	restLogger.Info("REST querying chaincode...")
+	restLogger.Warning("/devops/query endpoint has been deprecated. Use /chaincode endpoint instead.")
 
 	// Decode the incoming JSON payload
 	var spec pb.ChaincodeInvocationSpec


srderson:
Not sure that logging this on the server is the correct approach. The client will not see the message and the server log could be filled with these messages due to a misbehaving client.

As I believe the format of the POST has changed and a simple 301 redirect will not work, I'd suggest a [warning header](https://tools.ietf.org/html/rfc7234#section-5.5) be added to the response. This can be used to inform the client of the deprecation without filling up the server's log.

christo4ferris:
not that it needs to be addressed here and now, but we really should be considering some mechanism that a client can express the version of an API that it is expecting and that the API endpoint can respond back with some sort of compatibility expression. What I am suggesting is that we need more than some obscure log warning to manage versioning of the APIs and clients/SDKs that consume them. @srderson @binhn @ghaskins @tamasblummer

tamasblummer:
fabric-api has, but awaits implementation.

```
/**
 * retrieves client version
 *
 * @return client version e.g. 4.0.0
 */
String getClientVersion() throws HLAPIException;

/**
 * retrieves server version
 *
 * @return server version e.g. 4.0.0
 */
String getServerVersion() throws HLAPIException;
```

Tamas Blummer
CHIEF LEDGER ARCHITECT

T: +36 1 883 0300 
E: tamas@digitalasset.com mailto:address@digitalasset.com | W: digitalasset.com http://www.digitalasset.com/

> On 25 May 2016, at 19:34, Christopher Ferris notifications@github.com wrote:
> 
> not that it needs to be addressed here and now, but we really should be considering some mechanism that a client can express the version of an API that it is expecting and that the API endpoint can respond back with some sort of compatibility expression. What I am suggesting is that we need more than some obscure log warning to manage versioning of the APIs and clients/SDKs that consume them. @srderson https://github.com/srderson @binhn https://github.com/binhn @ghaskins https://github.com/ghaskins @tamasblummer https://github.com/tamasblummer
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub https://github.com/hyperledger/fabric/pull/1576#issuecomment-221646878

## 

This message, and any attachments, is for the intended recipient(s) only, 
may contain information that is privileged, confidential and/or proprietary 
and subject to important terms and conditions available at 
http://www.digitalasset.com/emaildisclaimer.html. If you are not the 
intended recipient, please delete this message.

angrbrd:
@srderson, thank you for the suggestion. That's a very good point. I will rework this PR in this way and let you have a look again. Will update later today.

@christo4ferris, I understand your point and agree that it's important. But I feel like this is not something that will be addressed immediately by this PR. Feel free to open another issue for that discussion separately and assign to me, as we have had no discussion on that point prior to this PR. The objective of this PR is to help the user understand that the /devops endpoint they are using has already been deprecated. And honestly, it has been done quite some time back now... that I am not even sure how much of a difference it would make right now to put this warning in. The superceeding /chaincode endpoint was introduced and all the docs have been updated a long time ago. So this is just to catch any remaining cases where the deprecated endpoint is used before it's removed entirely several sprints from now.

angrbrd:
I am opening an updated PR for this issue. Closing this one.

PR #1595, opened 2016-05-25
Move batch to pbft event manager

jyellick:
## Description

This changeset removes the batch plugin thread, and instead has it rely on the event manager thread introduced in #1586.
## Motivation and Context

This is the second in a series of changesets (see PR #1586) designed to eliminate the plugin threads from PBFT.  As this is in a series, it does not explicitly fix any particular issue, but it is work towards eliminating the race conditions (by eliminating go routines).

As with #1586, this changeset contains some hacky things and TODOs, pieces marked for eventual removal, etc.  This is intentional.

If @corecode could please review.
## How Has This Been Tested?

This has passed the pbft unit tests locally, and the CI should confirm the changeset doesn't break anything.  Because there is no new function, just re-routing of execution flow, the existing tests should be adequate.
## Checklist:

### DIFF
--- a/consensus/obcpbft/obc-batch.go
+++ b/consensus/obcpbft/obc-batch.go
@@ -33,16 +33,13 @@ type obcBatch struct {
 
 	batchSize        int
 	batchStore       []*Request
-	batchTimer       *time.Timer
+	batchTimer       eventTimer
 	batchTimerActive bool
 	batchTimeout     time.Duration
 	inViewChange     bool
 
-	incomingChan     chan *batchMessage // Queues messages for processing by main thread
-	custodyTimerChan chan custodyInfo   // Queues complaints
-	execChan         chan *execInfo     // Signals an execution event
-	viewChanged      chan struct{}      // Signals that a view change has occurred
-	idleChan         chan struct{}      // Used in unit testing to check for idleness
+	incomingChan chan *batchMessage // Queues messages for processing by main thread
+	idleChan     chan struct{}      // Idle channel, to be removed
 
 	complainer   *complainer
 	deduplicator *deduplicator
@@ -78,6 +75,11 @@ func newObcBatch(id uint64, config *viper.Viper, stack consensus.Stack) *obcBatc
 	logger.Debug("Replica %d obtaining startup information", id)
 
 	op.pbft = newPbftCore(id, config, op)
+	op.pbft.manager = newEventManagerImpl(op) // TODO, this is hacky, eventually rip it out
+	etf := newEventTimerFactoryImpl(op.pbft.manager)
+	op.pbft.newViewTimer.halt()
+	op.pbft.newViewTimer = etf.createTimer()
+	op.pbft.manager.start()
 
 	op.batchSize = config.GetInt("general.batchSize")
 	op.batchStore = nil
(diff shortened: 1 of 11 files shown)


jyellick:
Rebased, because of conflict.

consensus/obcpbft/pbft-core.go
```diff
@@ -439,25 +440,52 @@ func (instance *pbftCore) committed(digest string, v uint64, n uint64) bool {
 // =============================================================================
 
 // handle new consensus requests
+func (instance *pbftCore) requestSync(msgPayload []byte, senderID uint64) error {
+	msg := &Message{&Message_Request{&Request{Payload: msgPayload,
+		ReplicaId: senderID}}}
+	instance.manager.inject(pbftMessageEvent{
```
corecode, line 446: Uh hacky.  How would this work in a non-hacky version?

consensus/obcpbft/obc-batch.go
```diff
@@ -407,83 +394,78 @@ func (op *obcBatch) resubmitStaleRequest(c custodyInfo) {
...
-					logger.Debug("Batch replica %d complaint timeout expired for %s while in view change", op.pbft.id, c.hash)
-				}
+				logger.Debug("Batch replica %d complaint timeout expired for %s while in view change", op.pbft.id, c.hash)
 			}
-		case execInfo := <-op.execChan:
-			op.executeImpl(execInfo.seqNo, execInfo.raw)
-		case op.idleChan <- struct{}{}:
-			// Only used to detect thread idleness during unit tests
 		}
+	default:
```
corecode, line 452: I think we should have two synchronous state machines (each with their own event queue), which are serviced by a single state machine manager, instead of doing this event pass-through.

consensus/obcpbft/pbft-core.go
jyellick, line 446: The way I envision these state machine objects working, is an event comes in, is processed, and if it needs to, it emits a new event.

So, for instance, the batch timer expires, it builds the batch, and then would return/emit a `pbftMessageEvent`, this gets received, decoded, and it turns out it is a request, so it returns a `pbftRequestEvent`, this gets processed, and, if primary, returns a `pbftPrePrepareEvent`, which returns a `pbftPrepareEvent`, and finally, the chain is done.

This way it's much easier to inject specific events into the state machine, and the entry path would be the same for all types.  As the code stands today, it's very odd that we inject requests through so many different method calls, ultimately, I see all internal injection through `processEvent` and external injection carefully curated via calls from the consensus API.

consensus/obcpbft/obc-batch.go
jyellick, line 452: I'm not strictly opposed to this, but wrapping seems to match the current plugin architecture more closely (as `RecvMsg` filters and wraps to `recvMsg` for instance).  I think splitting this later should not be too much work, but splitting it now might make the diffs a little incomprehensible.

consensus/obcpbft/pbft-core.go
corecode, line 446: How would the batch state machine add a request in the pbft state machine?

consensus/obcpbft/obc-batch.go
corecode, line 452: I agree.

consensus/obcpbft/pbft-core.go
jyellick, line 446: Whatever is returned by `processEvent` is the _next_ event to be processed (regardless of any other events which might be 'in queue').  This is equivalent to directly calling the function, but without the tight coupling it creates.  So, today, for batch to inject an event into PBFT, it would simply return one from `processEvent`.

As you discuss later, we may want to split these into distinct state machines, rather than stacked ones.  If that were the case, then some modification would likely be required, maybe the return would need to include the target state machine, or maybe some other measure, but it doesn't seem insurmountable.

tuand27613:
LGTM

srderson:
@jyellick This one has a merge conflict

srderson:
@tuand27613 @corecode thanks for the review. LGTM

PR #1596, opened 2016-05-25
Protect against nil pointer dereference in execDoneSync

### LINKED ISSUE #1538: Consensus sieve has error on both LoZ and X86 platforms -- one peer cannot get block
## Description

Get the latest hyper ledger code (commit 66ca505deb06f4df024f0540d18bfdecb04584fa )
Run with busywork and get error from peers. After tested with @bcbrock , he confirmed that this is bug with  sieve on LoZ platform.
## Describe How to Reproduce

1) Get the latest hyper ledger code;
2) Build peer ;
3) run "make stress2s" from folder "$GPATHsrc/github.com/hyperledger/fabric/tools/busywork/counters"

Error msg:

```
 ...
 driver:     Client 0 is subprocess 49966
 driver: Waiting (indefinitely) for subprocesses to complete
 fabricLogger:ERROR: Block 2206 JSON response does not parse
 fabricLogger:ERROR: HTTP 'ncode' = '200'
 fabricLogger:ERROR: Full HTTP token dump below; Aborting
 fabricLogger:ERROR:     sock socka8d15a80
 fabricLogger:ERROR:     http HTTP/1.1 200 OK
 fabricLogger:ERROR:     querylength 0
 fabricLogger:ERROR:     -queryblocksize 8192
 fabricLogger:ERROR:     state body
 fabricLogger:ERROR:     charset iso8859-1
 fabricLogger:ERROR:     type application/json
 fabricLogger:ERROR:     -binary false
 fabricLogger:ERROR:     url http://0.0.0.0:48106/chain/blocks/2206
 fabricLogger:ERROR:     -keepalive 0
 fabricLogger:ERROR:     -timeout 0
 fabricLogger:ERROR:     socketinfo 0.0.0.0:48106
 fabricLogger:ERROR:     queryoffset 0
 fabricLogger:ERROR:     binary 1
 fabricLogger:ERROR:     -strict 1
 fabricLogger:ERROR:     totalsize 5
 fabricLogger:ERROR:     -validate 0
 fabricLogger:ERROR:     -protocol 1.1
 fabricLogger:ERROR:     connection close
 fabricLogger:ERROR:     -queryprogress
 fabricLogger:ERROR:     -headers
 fabricLogger:ERROR:     -blocksize 8192
 fabricLogger:ERROR:     coding
 fabricLogger:ERROR:     status ok
 fabricLogger:ERROR:     body null

 fabricLogger:ERROR:     currentsize 5
```

Then use `url http://localhost:<peer REST port>/chain/blocks/3` to find it returns "null" from vp3 only,
All 4 vps are running.

Here is error
[vps_log.zip](https://github.com/hyperledger/fabric/files/274937/vps_log.zip)
 log from vps

jyellick:
## Description

This changeset directly protects against a nil pointer dereference in `execDoneSync`
## Motivation and Context

The code in `execDoneSync` did not protect against dereferencing the `currentExec` pointer even if it was nil, sometimes resulting in crashes.

This should fix #1538, though confirmation from @bmos299 would be appreciated.
## How Has This Been Tested?

This is a simple change, and should be obviously correct.  Although the value of a new test case for this is dubious, I've included a very short unit test case which exhibits the behavior before the change, but not after.
## Checklist:

### DIFF
--- a/consensus/obcpbft/pbft-core.go
+++ b/consensus/obcpbft/pbft-core.go
@@ -894,7 +894,9 @@ func (instance *pbftCore) execDoneSync() {
 		}
 
 	} else {
-		logger.Debug("Replica %d had execDoneSync called, but there is no current execution", instance.id)
+		// XXX This masks a bug, this should not be called when currentExec is nil
+		logger.Warning("Replica %d had execDoneSync called, flagging ourselves as out of date", instance.id)
+		instance.skipInProgress = true
 	}
 	instance.currentExec = nil
 
--- a/consensus/obcpbft/pbft-core_test.go
+++ b/consensus/obcpbft/pbft-core_test.go
@@ -1291,3 +1291,8 @@ func TestReplicaPersistDelete(t *testing.T) {
 		t.Error("expected no persisted entry")
 	}
 }
+
+func TestNilCurrentExec(t *testing.T) {
+	p := newPbftCore(1, loadConfig(), &omniProto{})
+	p.execDoneSync() // Per issue 1538, this would cause a Nil pointer dereference
+}


corecode:
This is the wrong fix, it masks a race condition, rather than fixing the race condition.

jyellick:
@corecode I agree this is masking a race condition, as I noted in #1538, this directly prevents the nil pointer dereference, so that @bmos299 and others can potentially make progress while the single-threading work in #1586 #1595 and others works to eliminate the possibility of races.

corecode:
What I am saying is that it masks the problem.  What does it mean if `doneExecSync` is called without `currentExec` set?  Probably something between consensus and execution got out of order, and will result in an incorrect blockchain.  This will re-emerge later during the run as all kinds of defects, which will likely send us on a wild goose chase.

jyellick:
Possibly.  I see the merits in excluding this, as it might just generate new problems.  It also seems likely the the replica will simply recover via state transfer.  This struck me as a non-invasive change which fixes a crash, and not-crashing seems preferable.

corecode:
In that case we should flag ourselves "somehow incorrect", and perform a recovery of our blockchain via state transfer.

jyellick:
@corecode That's a great idea, fixed and pushed.

tuand27613:
LGTM

srderson:
Thanks, LGTM

PR #1599, opened 2016-05-26
Modifying chaincode invoke response message

### LINKED ISSUE #835: Always return ok when invoke chaincode even I give a wrong function name.
I am testing the chaincode_example02, find the REST API /devops/invoke returns ok message and create new transaction even if I provide a wrong function name in the ctorMsg.

But from the source code logic, I think we should get error message like: "Received unknown function invocation".

And the transaction has been created which really does not make sense.

angrbrd:
## Description

Modifying existing chaincode invoke response message and log message to clarify that the invoke transaction has been _submitted_, but not completed. 
## Motivation and Context

This change is made to clarify that the submission of a transaction does not guarantee it's successful execution. The previous response message has always returned "success", which misled the end user and resulted in issue #835.

Fixes #835
## How Has This Been Tested?

I verified that the new response and log message were visible whenever a chaincode invoke request took place. All the current UTs pass and no new tests are needed for this change.
## Checklist:

Modifying chaincode invoke response message to clarify that the
transaction has been successfully submitted, but not completed.

Fixes #835

### DIFF
--- a/core/rest/rest_api.go
+++ b/core/rest/rest_api.go
@@ -437,7 +437,7 @@ func (s *ServerOpenchainREST) GetEnrollmentCert(rw web.ResponseWriter, req *web.
 
 		rw.WriteHeader(http.StatusOK)
 		fmt.Fprintf(rw, "{\"OK\": \"%s\"}", urlEncodedCert)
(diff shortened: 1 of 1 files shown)


srderson:
LGTM. Thanks

PR #1605, opened 2016-05-26
pbft: add null request "keep alives"

### LINKED ISSUE #1454: Allow non-byzantine replica to always catch up
## Description

Per the classical PBFT implementation, the network is always able to make progress ordering and processing transactions, and the current PBFT implementation supports this.  However, the PBFT protocol does not make any guarantees that just because a node is behaving in a non-byzantine fashion, that it will be able to participate in the network, or that its state will be in sync with the network.  This was discussed somewhat in issue #1120.
## Describe How to Reproduce

During any sort of 4 node non-byzantine stress test, one replica will invariably be slower than the other four.  Because of this, three of the replicas will move ahead of the slow replica, and although the fourth replica is no longer actively participating in the ordering, the network will continue to make progress (as it must, otherwise a byzantine node could halt the network).  The lagging replica will initiate state transfer, but by the time state transfer completes, the network will have moved beyond the state which was transferred to, and the replica will remain out of date.  If the network stops receiving requests, the lagging replica is left in a state where it is not up to date, and because it cannot eavesdrop on the network (as there is no activity on the network), it cannot catch up.  This is the scenario observed using the stress tests from @bcbrock.
## Simple Solutions
1. One simple solution to the specific scenario mentioned, would be for the primary to be obligated to issue a null request periodically if a real request is not available.  This would allow the existing eavesdropping mechanism to attempt to recover and rejoin the network.
2. Unfortunately, issuing null requests by themselves in insufficient to recover from all failing scenarios.  In the event that the network has moved on to a new view, the replica needs to change views to this new view and become active in it.  The replica could eavesdrop on the view for the requests to attempt to make this determination, say, picking the highest view to which f+1 replicas have sent messages in in the same way that we eavesdrop for checkpoints today.
3.  Today, if a state transfer is already in progress, state transfer will accept additional targets, but will not transfer to them unless something goes wrong moving to the current target.  We could introduce some stronger notion of hard and soft targets, where hard targets are minimum targets which must be transferred to, and soft targets can optionally be transferred to.  (The usefulness of a soft target is, imagine with k=10, L=40, targets for seqNo=30, 40, 50 could all be added, and, if successfully transferred to, transactions from that sequence number could be played, as the log overlaps all of them.  If state transfer were guaranteed to be successful to a target, this logic could be simplified considerably).  
## More Complex Solutions
1. In order to support dynamically adding additional replicas to the network, and changing the network size, new protocol support will need to be added.  It might be possible to leverage whatever protocol support is added, such as forcing a view change (in a rate limited way) to perform the catchup.
2. Some other sort of solicitation for the networks current view and last stable checkpoint?  The original paper included a variant of this, that when data for a checkpoint was requested, replicas would reply with higher checkpoints if that checkpoint had already been garbage collected, but this was not implemented because blocks, corresponding to checkpoints are never garbage collected and still provide valid blockchain targets, even if the associated blockchain state is not transferrable.

The first set of solutions would fall under eavesdropping and are attractive because they require very little in ways of protocol modification.  The second approach may be superior technically however.  As this is a protocol problem, input from @vukolic especially is requested, but thoughts from @corecode @kchristidis and @tuand27613 would be welcomed as well.

corecode:
## Description

Optionally send periodic null request "keep alives" when the pbft network is idle.
## Motivation and Context

Fixes #1454
## How Has This Been Tested?

unit tests, behave
## Checklist:

### DIFF
--- a/consensus/obcpbft/config.yaml
+++ b/consensus/obcpbft/config.yaml
@@ -49,6 +49,10 @@ general:
 
         # How long may a view change take
         viewchange: 2s
+
+        # Interval to send "keep-alive" null requests.  Set to 0 to disable.
+        nullrequest: 0s
+
 ################################################################################
 #
 #   SECTION: EXECUTOR
--- a/consensus/obcpbft/pbft-core_test.go
+++ b/consensus/obcpbft/pbft-core_test.go
@@ -140,7 +140,6 @@ func TestIncompletePayload(t *testing.T) {
 	checkMsg(&Message{}, "Expected to reject empty message")
 	checkMsg(&Message{&Message_Request{&Request{ReplicaId: broadcaster}}}, "Expected to reject empty request")
 	checkMsg(&Message{&Message_PrePrepare{&PrePrepare{ReplicaId: broadcaster}}}, "Expected to reject empty pre-prepare")
-	checkMsg(&Message{&Message_PrePrepare{&PrePrepare{SequenceNumber: 1, ReplicaId: broadcaster}}}, "Expected to reject incomplete pre-prepare")
 }
 
 func TestNetwork(t *testing.T) {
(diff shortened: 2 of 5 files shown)


srderson:
@corecode unit tests failed with the following output

```
2016/05/26 14:49:42 Replica 1 prepare count for view=0/seqNo=1: 1
panic: Unimplemented

goroutine 6915 [running]:
panic(0xa8c500, 0xc8202e98e0)
    /home/travis/.gimme/versions/go1.6.linux.amd64/src/runtime/panic.go:464 +0x3e6
github.com/hyperledger/fabric/consensus/obcpbft.(*omniProto).sign(0xc82026a900, 0xc820355600, 0x71, 0x80, 0x0, 0x0, 0x0, 0x0, 0x0)
    /home/travis/gopath/src/github.com/hyperledger/fabric/consensus/obcpbft/mock_utilities_test.go:394 +0x111
github.com/hyperledger/fabric/consensus/obcpbft.(*pbftCore).sign(0xc82026b500, 0x2ac451f55580, 0xc820355580, 0x0, 0x0)
    /home/travis/gopath/src/github.com/hyperledger/fabric/consensus/obcpbft/sign.go:35 +0x10c
github.com/hyperledger/fabric/consensus/obcpbft.(*pbftCore).sendViewChange(0xc82026b500, 0x0, 0x0)
    /home/travis/gopath/src/github.com/hyperledger/fabric/consensus/obcpbft/viewchange.go:163 +0x6c0
github.com/hyperledger/fabric/consensus/obcpbft.(*pbftCore).main(0xc82026b500)
    /home/travis/gopath/src/github.com/hyperledger/fabric/consensus/obcpbft/pbft-core.go:295 +0x3f0
created by github.com/hyperledger/fabric/consensus/obcpbft.newPbftCore
    /home/travis/gopath/src/github.com/hyperledger/fabric/consensus/obcpbft/pbft-core.go:268 +0x185a
FAIL    github.com/hyperledger/fabric/consensus/obcpbft 61.174s
```

corecode:
If I knew which test fails... I can't reproduce it here.  Most likely again some test race condition.

corecode:
This is a race in `TestReplicaPersistQSet`:
- somehow it takes more than 2 seconds for the test to run
- the request we inject times out and triggers a view change
- view change requires a signature, which is not mocked in the test
- panic!

jyellick:
@corecode I switched this test to run only on the test thread (not to even start up the timer threads), so this should be a non-issue now.

srderson:
@corecode There's a merge conflict now. Could you rebase? That will kick off Travis again.

corecode:
rebased

consensus/obcpbft/config.yaml
```diff
@@ -49,6 +49,10 @@ general:
 
         # How long may a view change take
         viewchange: 2s
+
+        # Interval to send "keep-alive" null requests.  Set to 0 to disable.
+        nullrequest: 0s
```
jyellick, line 54: Does this make sense as a default? Certainly it is more classically PBFT, but most users likely expect eventual consistency.

corecode, line 54: We can change it later if we want to change the default.  We probably will also want to change batch size and consensus default (noop).

consensus/obcpbft/pbft-core_test.go
```diff
@@ -1296,3 +1295,59 @@ func TestNilCurrentExec(t *testing.T) {
...
+	net := makePBFTNetwork(validatorCount, func(pe *pbftEndpoint) {
+		pe.pbft.nullRequestTimeout = 200 * time.Millisecond
+		pe.pbft.requestTimeout = 500 * time.Millisecond
+	})
+	defer net.stop()
+
+	msg := createPbftRequestWithChainTx(1, 0)
+	net.pbftEndpoints[0].pbft.manager.queue() <- msg
+
+	go net.processContinually()
```
jyellick, line 1310: `processContinually` seems to be a source of many of our race conditions in the tests.  Why not:

```
net.process()
// Test the current state
for {
    time.Sleep(1 * time.Second)
    net.process()
    // Test the current state and break on success, or fail on enough retries
}
```

Ultimately this should be cleaned up by the deterministic timer wheel stuff, but I still like this approach better because of the first pass testing.

consensus/obcpbft/pbft-core_test.go
```diff
@@ -1296,3 +1295,59 @@ func TestNilCurrentExec(t *testing.T) {
...
+		pe.pbft.requestTimeout = 500 * time.Millisecond
+	})
+	defer net.stop()
+
+	net.pbftEndpoints[0].pbft.nullRequestTimeout = 0
+
+	msg := createPbftRequestWithChainTx(1, 0)
+	net.pbftEndpoints[0].pbft.manager.queue() <- msg
+
+	go net.processContinually()
```
jyellick, line 1339: ~~Same comment as with the previous test~~ Not sure how this got associated with this line, it was intended for the next `go processContinually`

consensus/obcpbft/pbft-core_test.go
corecode, line 1310: I don't understand what you mean by that.

jyellick:
Since we'll be restructuring the mock tests shortly anyhow, I'm fine with this PR as is.

srderson:
LGTM. Thanks

PR #1611, opened 2016-05-26
Add assertion to test case confirming non-existent block is nil

srderson:
## Description

As part of debugging issue #1091, a one line addition was added to the PutRawBlock test case to assert that a non-existent block is nil.
## Motivation and Context

If `ledger.PutRawBlock` is called for a value higher than the current chain height and then `ledger.GetBlockByNumber` is called for a value between the previous chain height and the newly added raw block, the returned block should be `nil`. There was no test confirming this.
## How Has This Been Tested?

This is an addition to an existing test case. The test case has been run and passes.
## Checklist:


christo4ferris:
LGTM

binhn:
thanks.

PR #1616, opened 2016-05-26
Update baseimage to v0.0.10

ghaskins:
This update references a new baseimage release, v0.0.10 for both Docker and Vagrant
## Description

This update encompasses the following changes since release v0.0.9
1. We upgrade protobufs to 3.0.0-beta-3, released May 16 2016
2. We introduce support for the s390x platform
## Motivation and Context

Multi-arch requirements are emerging for fabric, and this release adds support for s390x in addition to x86_64.  We also updated protobufs to the latest stable release, 3.0.0-beta-3.
## How Has This Been Tested?
1. Full vagrant destroy/up cycle, make checks passed.
2. Ran "make protos" and ensured there were no problems/changes in the protobuf code.
## Checklist:


images/base/scripts/common/setup.sh
```diff
@@ -93,7 +85,7 @@ fi
 cd /tmp
 git clone https://github.com/google/protobuf.git
```
srderson, line 86: Maybe for a future PR, but I think we could just download the release from https://github.com/google/protobuf/releases/download/v3.0.0-beta-3/protoc-3.0.0-beta-3-linux-x86_64.zip

ghaskins, line 86: I had thought of that, but it would only support x86 so I was thinking src-release was better in this case.  I was thinking since this image gets cached it had a negligible performance impact.  I could be convinced otherwise.

srderson:
LGTM. Tested locally and confirmed this is working.

PR #1617, opened 2016-05-26
Add a test for chaincode over HTTP

### LINKED ISSUE #1597: Missing unit-test for HTTP based chaincode
## Description

Chaincode can be specified locally or over HTTP however we currently have no unit-test for the latter.
## Describe How to Reproduce

There currently are no unit test covering getCodeFromHTTP in core/chaincode/platforms/golang/hash.go

lehors:
## Description

This PR adds a unit test to cover the code in getCodeFromHTTP in core/chaincode/platforms/golang/hash.go
The test currently depends on my fork of the project because the code must reside outside the Fabric repo. It would however be better to have it located in a separate repo within the HLP.
See related discussion on PR #1556 and Issue #737.
## Motivation and Context

Fixes #1597
## How Has This Been Tested?

make unit-test
## Checklist:

### DIFF
--- a/core/chaincode/chaincodetest.yaml
+++ b/core/chaincode/chaincodetest.yaml
@@ -380,7 +380,7 @@ chaincode:
     startuptimeout: 1000
 
     #timeout in millisecs for deploying chaincode from a remote repository.
-    deploytimeout: 30000
+    deploytimeout: 60000
 
     #mode - options are "dev", "net"
     #dev - in dev mode, user runs the chaincode after starting validator from
--- a/core/chaincode/exectransaction_test.go
+++ b/core/chaincode/exectransaction_test.go
@@ -365,7 +365,7 @@ func TestHTTPExecuteDeployTransaction(t *testing.T) {
 	// The chaincode used here cannot be from the hyperledger repo
 	// itself or it won't be downloaded because it will be found
 	// in GOPATH, which would defeat the test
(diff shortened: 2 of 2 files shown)


ghaskins:
@christo4ferris This is the work that originally inspired the question from @lehors about adding a test repository.  Do you have any thoughts here?  If I may summarize the previous/relevant discussion:
1. The unit-test coverage provided here is important, so we don't want to just drop the test.
2. An external URL (!= https://github.com/hyperledger/fabric) is required for the test
3. It would be ideal if this were not tied to a personal github URL since that _could_ theoretically disappear inadvertently without consensus and/or control of HL/LF.
4. In this case, @lehors URL is probably more stable than an arbitrary URL because he understands it needs to be kept around indefinitely and is motivated to keep the system working.

Assuming we all agree with (1) and (2), we just need to decide between (3) and (4).

christo4ferris:
@ghaskins @lehors ah, so the example chaincode needs to be external to hyperledger/fabric, but the test is fine where it is? I have no problem with a repo for that. Just ask @ryjones and he can hook you up. Have the new repo added to the hyperledger "fabric maintainers" team.

ghaskins:
@christo4ferris That is correct.  Unit-test itself will be as this PR patch indicates.  External repo will just hold an example chaincode.  Propose hyperledger/fabric-test.

ghaskins:
Thank you for going the extra mile to get this coverage in.

LGTM

PR #1614, opened 2016-05-26
Convert PBFT event manager to rely on type inference

jyellick:
## Description

This changeset removes the temporary integer enumeration mechanism for events introduced in #1586 and switches to processing arbitrary `interface{}` events, based on type inference.
## Motivation and Context

This is part of a series PRs designed to remove the multithreading from PBFT and to turn it into more of a simple state machine.  This change was always intended, and is also intended to address some feedback from @corecode in the original PR.

This also includes a quick fix of an incorrect copyright statement seen in `deduplicator.go`
## How Has This Been Tested?

This has been tested against the obcpbft unit tests locally, and should be covered by CI.

This PR does not require any new tests, as it effectively leaves the entire flow the same, just changes some type definitions.
## Checklist:
- [X] I have added a [Signed-off-by](https://github.com/hyperledger/fabric/blob/master/CONTRIBUTING.md#legal-stuff)
- [X] Either no new documentation is required by this change, OR I added new documentation
- [X] Either no new tests are required by this change, OR I added new tests
- [X] I have run [goimports](https://godoc.org/golang.org/x/tools/cmd/goimports), [go vet](https://golang.org/cmd/vet/), and [golint](https://github.com/golang/lint). I have cleaned up all valid errors and warnings in code I have added or modified. These tools may generate false positives. Don't be worried about ignoring some errors or warnings. The goal is clean, consistent, and readable code.

Signed-off-by: Jason Yellick jyellick@us.ibm.com

PR #1622, opened 2016-05-27
Begin removing legacy PBFT methods

jyellick:
## Description

This changeset is yet another in the long series of #1586 #1595 and #1614.

It begins to remove the non-event based mechanisms which were used to inject work into PBFT.  It moves some of them into a legacy structure for continued support of classic/sieve, and removes dependency on those methods from the PBFT core tests.
## Motivation and Context

This is part of a series PRs designed to remove the multithreading from PBFT and to turn it into more of a simple state machine.  There is very little new code in this changeset, just restructuring and API modifcations.  As with the other PRs in this series, review from @corecode especially is requested, though review from @kchristidis @tuand27613 and others is always welcome.

(And, as with the previous in the patch series, this is not the final patch, there remains considerable ugliness, which will be removed in the following series)
## How Has This Been Tested?

This has been tested via the obcpbft unit tests locally.  Relying on CI for the remainder.
## Checklist:

### DIFF
--- a/consensus/obcpbft/pbft-core.go
+++ b/consensus/obcpbft/pbft-core.go
@@ -267,26 +267,26 @@ func (instance *pbftCore) close() {
 }
 
 // allow the view-change protocol to kick-off when the timer expires
-func (instance *pbftCore) processEvent(e event) event {
+func (instance *pbftCore) processEvent(e interface{}) interface{} {
 
 	logger.Debug("Replica %d processing event", instance.id)
 
-	switch e.eventType() {
-	case viewChangeTimerEventID:
+	switch et := e.(type) {
+	case viewChangeTimerEvent:
 		logger.Info("Replica %d view change timer expired, sending view change", instance.id)
 		instance.timerActive = false
 		instance.sendViewChange()
-	case pbftMessageEventID:
-		msg := e.(pbftMessageEvent)
+	case pbftMessageEvent:
+		msg := et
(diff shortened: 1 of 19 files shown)


consensus/obcpbft/pbft-core.go
```diff
@@ -279,7 +281,29 @@ func (instance *pbftCore) processEvent(e interface{}) interface{} {
 	case pbftMessageEvent:
 		msg := et
 		logger.Debug("Replica %d received incoming message from %v", instance.id, msg.sender)
-		instance.recvMsg(msg.msg, msg.sender)
+		next, err := instance.recvMsg(msg.msg, msg.sender)
+		if err != nil {
```
corecode, line 287: Is there ever a case where we don't want to return a nil error?

corecode:
i think it is silly that we have to split up these PRs if they get merged together at the same time anyways.

This is good refactoring work.  Let's get it in.

tuand27613:
LGTM

joshhus:
Note when we are removing things, the documentation checklist item becomes "checked for, and removed references to item/function from documentation." Thanks.

jyellick:
@joshhus Good point.  In this case though, this only concerns internal implementation details, (none of the deprecated methods are even exposed outside of the obcpbft package), so I don't believe there is any documentation which refers to them (other than the internal code documentation, which has been updated).

srderson:
Thanks! LGTM

PR #1623, opened 2016-05-27
Use a whitelist for peers participating on consensus

### LINKED ISSUE #756: Consensus should run with whitelisted VPs when security is enabled

tuand27613:
## Description

Create a list of peers that can participate in consensus. The list is build from CORE_PBFT_GENERAL_N. Consensus does not start until N peers have connected. After that, only peers in the whitelist can reconnect.
The restriction on peers having to be named vp0/1/2/3,etc ... is also lifted.
## Motivation and Context

Fixes #756
## How Has This Been Tested?

Unit tests run ( unit tests were modified to use the whitelist methods )
Behave tests run ( the compose-consensus-4 docker-compose files in bddtests were modified so that peer names are not vp0/1/2/3 )
## Checklist:


tuand27613:
@jyellick @corecode  @kchristidis please take a look.

core/peer/statetransfer/statetransfer.go
```diff
@@ -25,11 +25,11 @@ import (
 	"time"
 
 	_ "github.com/hyperledger/fabric/core" // Logging format init
+	"github.com/op/go-logging"
```
corecode, line 28: why?

core/peer/whitelist.go
```diff
@@ -0,0 +1,244 @@
...
+	"strings"
+
+	"github.com/hyperledger/fabric/core/db"
+	pb "github.com/hyperledger/fabric/protos"
+	"github.com/spf13/viper"
+
+	"github.com/golang/protobuf/proto"
+)
+
+const whitelistFile = "/tmp/whitelist.dat"
```
corecode, line 34: not needed

core/peer/whitelist.go
```diff
@@ -0,0 +1,244 @@
...
+
+// CheckWhitelistExists blocks until we have N validators in the whitelist
+func (p *PeerImpl) CheckWhitelistExists() {
+	<-p.whitelistCreated
+}
+
+// LoadWhitelist loads the validator's whitelist of validating peers from disk into memory
+// Sets p.whitelist and p.whitelistedMap
+func (p *PeerImpl) LoadWhitelist() (err error) {
+
```
corecode, line 56: empty line

core/peer/whitelist.go
```diff
@@ -0,0 +1,244 @@
...
+	if !p.isValidator {
+		return nil
+	} // if I'm an NVP, I don't care
+
+	// pbft plugin waits on this channel for whitelist to complete
+	p.whitelistCreated = make(chan struct{}, 3) // make non-blocking. We only need to wake up the receiver once
```
corecode, line 62: why 3?

core/peer/whitelist.go
```diff
@@ -0,0 +1,244 @@
...
+	}
+	peerLogger.Debug("Whitelist now reads: %+v", p.whitelist)
+
+	return err
+}
+
+// SaveWhitelist stores to disk the list of validating peers this validator should connect to
+// Is called when the handlerMap is locked by RegisterHandler
+func (p *PeerImpl) SaveWhitelist() (err error) {
+
```
corecode, line 95: empty line

bddtests/docker-compose-4-consensus-sieve.yml
```diff
@@ -8,7 +8,7 @@ vp0:
     file: docker-compose-4-consensus-base.yml
     service: vpSieve
   environment:
-    - CORE_PEER_ID=vp0
```
corecode, line 11: why do these names change?

consensus/consensus.go
```diff
@@ -30,8 +30,18 @@ type Consenter interface {
 
 // Inquirer is used to retrieve info about the validating network
 type Inquirer interface {
-	GetNetworkInfo() (self *pb.PeerEndpoint, network []*pb.PeerEndpoint, err error)
-	GetNetworkHandles() (self *pb.PeerID, network []*pb.PeerID, err error)
+	GetOwnID() (id uint64)
```
corecode, line 33: What do all of these interfaces do?  I have a suspicion that we don't need that many.

consensus/obcpbft/obc-batch.go
```diff
@@ -90,20 +92,37 @@ func newObcBatch(id uint64, config *viper.Viper, stack consensus.Stack) *obcBatc
...
+	go op.waitForID(config)
+
 	go op.main()
 	return op
 }
 
+func (op *obcBatch) waitForID(config *viper.Viper) {
+	var id uint64
+
+	op.stack.CheckWhitelistExists()
```
corecode, line 111: I presume this is blocking?  Not obvious from the name.

consensus/obcpbft/obc-batch.go
```diff
@@ -90,20 +92,37 @@ func newObcBatch(id uint64, config *viper.Viper, stack consensus.Stack) *obcBatc
...
-	op.deduplicator = newDeduplicator()
-
 	// create non-running timer
 	op.batchTimer = time.NewTimer(100 * time.Hour) // XXX ugly
 	op.batchTimer.Stop()
 
 	op.idleChan = make(chan struct{})
 	op.viewChanged = make(chan struct{})
 
+	go op.waitForID(config)
```
corecode, line 102: why does this go here, and not at the very top of the constructor, and have the rest of the setup be performed later?

consensus/obcpbft/obc-batch.go
```diff
@@ -90,20 +92,37 @@ func newObcBatch(id uint64, config *viper.Viper, stack consensus.Stack) *obcBatc
...
+	id = op.stack.GetOwnID()
+	logger.Debug("Replica ID = %v", id)
+
+	// instantiate pbft-core
+	op.pbft = newPbftCore(id, config, op)
+
+	op.complainer = newComplainer(op, op.pbft.requestTimeout, op.pbft.requestTimeout)
+	op.deduplicator = newDeduplicator()
+
+	op.isSufficientlyConnected <- true
```
corecode, line 121: i suggest `close(op.isSufficientlyConnected)`

consensus/obcpbft/obc-batch.go
```diff
@@ -408,6 +418,11 @@ func (op *obcBatch) resubmitStaleRequest(c custodyInfo) {
 
 // allow the primary to send a batch when the timer expires
 func (op *obcBatch) main() {
+
```
corecode, line 421: if `op.isSufficientlyConnected` is closed, this reduces to an unconditional `<- op.isSufficientlyConnected`

core/peer/whitelist.go
```diff
@@ -0,0 +1,244 @@
...
+	if p.whitelist.Cap == noWhitelistNeeded {
+		peerLogger.Info("whitelist will not be saved. whitelist.Cap = %v", p.whitelist.Cap)
+		return nil
+	}
+
+	vpMap := filterHandlers(p.handlerMap.m) // who are we connected to currently ?
+
+	if p.whitelist.Persisted {
+		peerLogger.Debug("Whitelist has already been saved. This is a restart.")
+		minNeeded := ((p.whitelist.Cap - 1) / 3) * 2 //2f
```
corecode, line 110: This has to be calculated by consensus and not by the peer.

core/util/utils.go
```diff
@@ -88,3 +97,53 @@ func GenerateHashFromSignature(path string, ctor string, args []string) []byte {
 	hash := ComputeCryptoHash(b)
 	return hash
 }
+
+// LoadFromDisk loads a file from disk
+func LoadFromDisk(filename string) (data []byte, err error) {
```
corecode, line 102: not used

corecode:
I couldn't see any tests introduced for the whitelist functionality.

corecode:
Overall this feels like a big hack (which it is), and it is quite intrusive as well.  I don't like the way it ties up consensus and the rest of the stack even more.

I think we need to take a different approach that does not need such deep changes, especially as essentially a convenience stop-gap measure.

If the stack exposes `PeerConnect()` and `PeerDisconnect()` events to consensus, most of the membership management can be taken care of in consensus itself.

kchristidis:
I agree with @corecode's review. @tuand27613, I'll be happy to tackle this in a proper fashion, so if you want it off your hands let me know - don't want to get it in your way though.

tuand27613:
I'll refactor on top of Jason's event manager work.  Thanks @kchristidis !

ghost:
Can one of the admins verify this patch?

tuand27613:
we're re-designing how we should handle whitelisting of peers via #756.

closing this PR.

PR #1624, opened 2016-05-27
Support "peer network ls" besides list

yeasy:
Let `peer network` cmd support `ls` as well, besides `list`
## Description

Support `ls` as alias to `list`, which is popular in Linux world.

To keep compatible with old version, still keep `list` cmd, too.
## Motivation and Context

No open issue.
## How Has This Been Tested?

Passed `make check`, while the change needs no additional test cases.
## Checklist:

### DIFF
--- a/peer/main.go
+++ b/peer/main.go
@@ -155,8 +155,8 @@ var networkLoginCmd = &cobra.Command{
 // }
 
 var networkListCmd = &cobra.Command{
-	Use:   "ls",
-	Aliases: []string{"list"},
+	Use:   "list",
+	Aliases: []string{"ls"},
 	Short: "Lists all network peers.",
 	Long:  `Returns a list of all existing network connections for the target peer node, includes both validating and non-validating peers.`,
 	RunE: func(cmd *cobra.Command, args []string) error {
--- a/peer/main.go
+++ b/peer/main.go
@@ -156,7 +156,7 @@ var networkLoginCmd = &cobra.Command{
 
 var networkListCmd = &cobra.Command{
 	Use:   "ls",
-	Aliases:   ["list"],
+	Aliases: []string{"list"},
 	Short: "Lists all network peers.",
 	Long:  `Returns a list of all existing network connections for the target peer node, includes both validating and non-validating peers.`,
 	RunE: func(cmd *cobra.Command, args []string) error {
--- a/peer/main.go
+++ b/peer/main.go
@@ -155,7 +155,8 @@ var networkLoginCmd = &cobra.Command{
 // }
 
 var networkListCmd = &cobra.Command{
-	Use:   "list",
+	Use:   "ls",
+	Aliases:   ["list"],
 	Short: "Lists all network peers.",
 	Long:  `Returns a list of all existing network connections for the target peer node, includes both validating and non-validating peers.`,
 	RunE: func(cmd *cobra.Command, args []string) error {


yeasy:
Still no found error, and no restart ci button found ~~
Anyone could help?
Thanks!

ghaskins:
From the raw log, it looks like this is the failure:

```
--- FAIL: TestReplicaPersistQSet (0.00s)
    pbft-core_test.go:1281: did not restore qset properly
```

Which I am guessing has nothing to do with your change, but we need to get a more definitive answer.

ghaskins:
@corecode I recall you submitting a PR recently regarding Qset.  I am not sure if it was merged or not, but perhaps you might know what is going on in this test failure.

corecode:
That is most puzzling.  I cannot reproduce this here.  Will have to investigate.

corecode:
Timing issue with the test - should be fixed in master now.

ghaskins:
@christo4ferris @binhn @jeffgarratt @srderson any opinion on this?

srderson:
I personally prefer `list` over `ls`. The rest of the commands are full words so `list` also seems more consistent. If we do merge this, docs need to be updated. There's at least one location I found at https://github.com/hyperledger/fabric/blob/master/docs/API/CoreAPI.md#cli

srderson:
If people want `ls`, maybe we could keep `list` as the default and make `ls` an alias?

yeasy:
@srderson Yes, your suggestion is a more mitigatory change, I can update first to see how other people like it.
Thanks!

christo4ferris:
I had the same thought as @srderson but as both forms are still valid, thought there was little harm. However, as @yeasy agreed to go along with @srderson's suggestion, it's all good.

LGTM

srderson:
This LGTM now. Just going to leave it open over the weekend to see if there are any other comments. Thanks @yeasy

ghaskins:
LGTM

srderson:
@yeasy It looks like you did not run `goimports` (which runs `gofmt`) when submitting this change. Could you please run `goimports` on `main.go` and submit a new PR? Thanks.

yeasy:
@srderson aha, sorry for the missing.
A new PR is submitted at https://github.com/hyperledger/fabric/pull/1673.
Thanks!

orderer/common/follower/follower_chain.go
```diff
@@ -426,6 +426,129 @@ func (c *Chain) pullUpToJoin() error {
...
+		}
+		c.logger.Debugf("Orderer endpoint %s has the biggest ledger height: %d", endpoint, latestHeight)
+
+		if latestHeight <= c.ledgerResources.Height() {
+			c.logger.Debugf("My height: %d, latest height: %d; going to wait %v for latest height to grow",
+				c.ledgerResources.Height(), latestHeight, c.retryInterval)
+			select {
+			case <-c.stopChan:
+				return nil, errors.New("stopped while waiting for latest height to grow")
+			case <-c.timeAfter(c.retryInterval):
```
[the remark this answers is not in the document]
jyellick: The improved interval defaults help.

It's still a bit of an unintuitive pattern to me to adaptively poll for new items.  In the worst case, we will have 10s of blocks to catch up on either with a simple periodic polling or with an adaptive polling.

We're re-using the error-based retry for a polling base retry, which, makes sense from a code perspective.  If it weren't for this code savings, I think I'd advocate a simpler approach.

orderer/common/follower/follower_chain.go
```diff
@@ -426,6 +426,129 @@ func (c *Chain) pullUpToJoin() error {
...
+
+	// Pull until the latest height or a config block
+	for seq := firstBlockToPull; seq < latestHeight; seq++ {
+		select {
+		case <-c.stopChan:
+			c.logger.Warnf("Stopped while pulling blocks: from %d until %d, last pulled %d",
+				firstBlockToPull, latestHeight, nextBlock.Header.Number)
+			return nil, errors.New("stopped while pulling blocks")
+		default:
+			nextBlock = puller.PullBlock(seq)
```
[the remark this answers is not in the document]
jyellick: > The saw-tooth type backoff actually helps here, as we eventually get a larger batch of blocks rather than pull them one by one.

I don't quite follow this?

If blocks are committing 'fast enough', then we should not be hitting the sleep portion of the loop, and in the worst case would always pull one block at a time?

If blocks are committing slowly, then we will sleep, but the first time we see a block, the retry interval resets.  For example, given the 1.5 base for the exponentiation, if we saw 0 blocks in 2 seconds, it seems very likely we will see 0-1 blocks in 3 seconds (assuming the commit rate is relatively uniformly distributed).  In which case we are back to our 'worst case behavior', of 'one block at a time'.

> On the first BlockPull(n), the BlockPuller will connect to some orderer that has at least block n, and will inquire what is its latest block as well. It will then try to fill a buffer with with as many blocks between [n,latest], up to a maximal buffer size. On the next BlockPull(n+1), it will pop from the buffer, until the buffer is empty. When the buffer is empty, it all starts again. For pulling a batch of blocks, it uses a stream.

So if we connect to an orderer who has more blocks than what we are looking for, we're going to pull, and discard those blocks in this case? (Since we break this loop based on target sequence even if the block puller has more blocks)

orderer/common/follower/follower_chain.go
```diff
@@ -426,6 +426,129 @@ func (c *Chain) pullUpToJoin() error {
...
+			if nextBlock == nil {
+				return nil, errors.Wrapf(cluster.ErrRetryCountExhausted, "failed to pull block %d", seq)
+			}
+			reportedPrevHash := nextBlock.Header.PreviousHash
```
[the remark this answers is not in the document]
jyellick: If the other orderer is able to return blocks which pass the block validation policy, but are forged, then I think we have more serious problems.  But it sounds like this is intended as a sanity, and not a security check, and we would never expect this to fail?

orderer/common/follower/follower_chain.go
```diff
@@ -286,146 +255,277 @@ func (c *Chain) run() {
...
+	//assuming this will never overflow int64, as upperLimit cannot be over MaxInt64/2
+	*retryInterval = time.Duration(1.5 * float64(*retryInterval))
+	if *retryInterval > upperLimit {
+		*retryInterval = upperLimit
+	}
+	c.logger.Debugf("retry interval increased to: %v", *retryInterval)
+}
+
+func (c *Chain) resetRetryInterval(retryInterval *time.Duration, lowerLimit time.Duration) {
+	if *retryInterval == lowerLimit {
```
jyellick, line 278: Not the end of the world, but this seems like a weird check, is it to avoid the debugging statement?  If so, I'd rather see it guard the debug, and unconditionally set the retryInterval.

orderer/common/follower/follower_chain.go
```diff
@@ -286,146 +255,277 @@ func (c *Chain) run() {
...
-		}
-	}
-
-	// Trigger creation of a new chain. This will regenerate 'ledgerResources' from the tip of the ledger, and will
-	// start either a follower.Chain or an etcdraft.Chain, depending on whether the orderer is a follower or a
-	// member of the cluster.
-	c.halt()
 
-	if c.joinBlock != nil {
+		// The join block never returns an error. This is checked before the follower is started.
```
jyellick, line 310: I realize we have the join block, but is this not the exact same check that we need to perform at every iteration of pulling after join? Can we not delete this one, and require that we always transition to a following state, and only from there do the chain creation?

orderer/common/follower/follower_chain.go
```diff
@@ -286,146 +255,277 @@ func (c *Chain) run() {
...
-			c.followerCreationCallback()
+			return nil
 		}
-	} else {
-		c.logger.Info("Block pulling finished successfully, invoking chain creation callback")
-		c.chainCreationCallback()
+
+		c.setStatusReport(types.ClusterRelationFollower, types.StatusActive)
+		c.joinBlock = nil
+		err = c.loadLastConfig()
```
jyellick, line 322: Why load the latest config here?

orderer/common/follower/follower_chain.go
```diff
@@ -286,146 +255,277 @@ func (c *Chain) run() {
...
 // pullUpToJoin pulls blocks up to the join-block height without inspecting membership on fetched config blocks.
 // It checks whether the chain was stopped between blocks.
 func (c *Chain) pullUpToJoin() error {
 	targetHeight := c.joinBlock.Header.Number + 1
-	firstBlockToPull := c.ledgerResources.Height()
-	c.logger.Debugf("first block to pull: %d, target height: %d", firstBlockToPull, targetHeight)
-	if firstBlockToPull >= targetHeight {
-		c.logger.Infof("Target height (%d) is <= to our ledger height (%d), skipping pulling", targetHeight, firstBlockToPull)
+	if c.ledgerResources.Height() >= targetHeight {
+		c.logger.Infof("Target height (%d) is <= to our ledger height (%d), skipping pulling", targetHeight, c.ledgerResources.Height())
```
jyellick: Would be nice to clarify that this is "skipping initial join pulling" or similar (that more blocks may still be pulled).

orderer/common/follower/follower_chain.go
```diff
@@ -286,146 +255,277 @@ func (c *Chain) run() {
...
 		return nil
 	}
 
-	puller, err := c.createBlockPullerFunc()
+	var err error
+	c.blockPuller, err = c.createBlockPullerFunc() // TODO replace with a factory that has a join block as input
```
jyellick, line 353: Why create the block puller here?  It's used inside of the `pullUntilLatest`, not in this function? Seems like we could delete it here and in `pullAfterJoin` by moving it into `pullUntilLatestWithRetry`?

Edit: I see below you're using this in `pullAfterJoin` to look up target height, but I still think this can be consolidated.  Will comment below.

PR #1628, opened 2016-05-27
Add a noop chaincode using TxID instead of UUID

gabre:
This PR introduces a trivial chain code, that serves a subset of the fabric_api, namely sendTransaction, getTransaction and listen for transactions.
## Description

A wider reaching change to devops.go is that "invoke" type transactions will be assigned an id derived from their user provided content instead of a random UUID. This is needed, so a client submitting a transaction is able to check if it ends up being in the block. The change makes however following assumptions, that potentially break current use:

1) User provided transaction content of "invoke" type transactions is unique
2) User provided transaction content has a known encoding (currently base64)
## Motivation and Context

Begin to build a chaincode serving fabric_api
## How Has This Been Tested?

Unit tests were run. Behavior tests were run.
## Checklist:

Signed-off-by: Gabor Hosszu (gabor@digitalasset.com)


core/devops.go
```diff
@@ -208,37 +209,55 @@ func (d *Devops) invokeOrQuery(ctx context.Context, chaincodeInvocationSpec *pb.
 	return resp, err
 }
 
-func (d *Devops) createExecTx(spec *pb.ChaincodeInvocationSpec, uuid string, invokeTx bool, sec crypto.Client) (*pb.Transaction, error) {
+func (d *Devops) createExecTx(spec *pb.ChaincodeInvocationSpec, invokeTx bool, sec crypto.Client) (*pb.Transaction, error) {
 	var tx *pb.Transaction
 	var err error
+        var temp_uuid = ""
```
tamasblummer: why using temp_uuid instead of calculated?

gabre: You are right, there is no reason to use that. We can use the calculated as it was before the changes. I will upload a fix.

tamasblummer:
Change to TXID will rather be implemented with an optional parameter to "invoke" type transactions

orderer/consensus/etcdraft/chain.go
```diff
@@ -434,6 +436,14 @@ func (c *Chain) Halt() {
 	}
 }
 
+// HaltExternal stops the chain but does not call the haltCallback function.
+// This is intended to be used by the channel participation API when removing
+// a channel.
+func (c *Chain) HaltExternal() {
```
tock-ibm: The exported `Halt()` is part of the `Chain` interface. This is what will be called from the "outside" when a chain is removed. It is also called from the inside. A better approach IMO is to have an "internal" `halt()` that is called when the internal go-routine isterminating and transferring execution, e.g.:
```go
func halt() {
  Halt()
  if c.haltCallback != nil {
	c.haltCallback()
  }
}
```
whereas the external Halt doesn't call this callback at all. That is, simply move 
```
if c.haltCallback != nil {
		c.haltCallback()
	}
```
from external `Halt` to internal `halt`.

orderer/consensus/etcdraft/chain.go
```diff
@@ -434,6 +436,14 @@ func (c *Chain) Halt() {
...
+	c.haltCallback = nil
```
tock-ibm: It is not thread safe to set nil to this callback because it may be accessed concurrently by internal threads. As indicated above, IMO a better approach it to separate to internal and external Halt/halt.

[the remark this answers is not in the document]
tock-ibm: True, long term the plan is to get rid of the inactive, but the callback will still be used to indicate the transfer of responsibility to whatever will replace it. Without a system channel, that will be a follower.

Issue #1491, opened 2016-05-17
Contributors needed for an "Introduction to Hyperledger fabric" overview doc


joshhus:
## Description

Current fabric/docs doc lacks "big picture" overview of the solution. Proposing some adds for "Introduction to Hyperledger fabric" documentation here. Please feel free to add / change / edit / add new sections as you see fit:

Proposed text for a "Welcome" page (note words in bold will be linked to the glossary definitions):
## Welcome to Hyperledger fabric

The Linux Foundation Hyperledger Project(1) is the result of collaboration between project members(2) who share the common goal of designing a business-ready, open-source standard for blockchain networks. Extensive research, planning and implementation by this community has led to the Linux Foundation adopting Hyperledger fabric as its open source standard for blockchain network architecture.

The Hyperledger fabric blockchain standard is a base architecture, with built-in functionality, security and platform independence for blockchain networks. The open standard facilitates adoption by any public or private organization or enterprise, and in turn by their constituent members. 

References: 
(1) Hyperledger Project: https://www.hyperledger.org/ 
(2) Hyperledger Project members: https://www.hyperledger.org/about/members
### Audience

The audience for this Hyperledger fabric documentation includes members of any public or private group, organization or enterprise who are interested in joining the Linux Foundation Hyperledger Project and adopting its open standard for their business networks. The organizational roles that this audience fulfills are diverse, and include business executives and decision-makers, system  architects, network designers, application developers, security administrators, auditors and  regulators. 

Any potential end user of a blockchain business network would also benefit from this documentation, including professionals at securities and commodities exchanges, manufacturing plants, distribution centers, leasing agencies, business service providers and government agencies.
### Prerequisites

The only prerequisites to employ Hyperledger fabric for blockchain networks are an interest in, and the resources to support, a distributed business network that processes transactions with significant 
decreases in cost, increases in speed and enhancements in security, compared to traditional networks. Though not required, most adopters of Hyperledger fabric will already be using one or more active business networks; integration of these existing networks with Hyperledger fabric is a fully-supported business service.

All sizes of business networks are supported by Hyperledger fabric, from a single node to networks 
containing hundreds of nodes. Typical hardware and software networking components, and resources to develop applications and write chaincode, are basic requirements.
## Describe How to Reproduce

The current fabric/docs are lacking "big picture" descriptions of the Hyperledger Project blockchain solution.

joshhus:
Taking this one ...

ghost:
Josh, 

[...]

PR #1637, opened 2016-05-27
New welcome page for Intro / big picture doc [ci skip] 

joshhus:
## Description

Adding big picture doc: this is the "Welcome" page. 
Note: Not adding TOC link for this yet; TBA.
## Motivation and Context

Users indicate "big picture" HL fabric doc is lacking.
https://github.com/hyperledger/fabric/issues/1491
## How Has This Been Tested?

Viewed on joshhus branch
joshhus branch view / check / tweak = the test case
## Checklist:

Signed-off-by: Josh Horton joshh@us.ibm.com


joshhus:
README changes not intended as part of this; have to check my sync.

orderer/common/multichannel/registrar.go
```diff
@@ -577,8 +577,71 @@ func (r *Registrar) joinAsFollower(ledgerRes *ledgerResources, clusterConsenter
...
+	if err != nil {
+		return err
+	}
+
+	cs.Halt()
+
+	err = r.ledgerFactory.Remove(channelID)
+	if err != nil {
+		return errors.Errorf("error removing ledger for channel %s", channelID)
+	}
```
tock-ibm: I am worried because this operation is performed under the global registrar lock. If this takes even more then a fraction of a second, it will impact the performance of the server on all other channels. This method 
```go

// GetChain retrieves the chain support for a chain if it exists.
func (r *Registrar) GetChain(chainID string) *ChainSupport {
	r.lock.RLock()
	defer r.lock.RUnlock()

	return r.chains[chainID]
}
```
is called by every client request, through `BroadcastChannelSupport` so we can't really hold that lock for too long (for write - exclusive).

What I propose is async deletion: 
- introduce a `toBeRemoved` set (e.g. `map[string]bool`) that holds channels that are being removed. 
- After stopping the channel move the channelID to that set, 
- and only then delete the ledger, but make the delete async with a different go-routine.
- release the lock.
- when the delete go-routine finishes, it removes the channelID from the `toBeRemoved`
- any attempt to Join or Remove should also check the `toBeRemoved`

What do you think? @wlahti @jyellick @yacovm 
We can possibly do that in another Jira that will improve upon this one.

orderer/common/multichannel/registrar.go
```diff
@@ -577,8 +577,71 @@ func (r *Registrar) joinAsFollower(ledgerRes *ledgerResources, clusterConsenter
 // RemoveChannel instructs the orderer to remove a channel.
 // Depending on the removeStorage parameter, the storage resources are either removed or archived.
 func (r *Registrar) RemoveChannel(channelID string, removeStorage bool) error {
+	r.lock.RLock()
+	defer r.lock.RUnlock()
+
```
tock-ibm: This must be a Lock / Unlock as we are writing the maps here, and need exclusive access.
This means that you can't use method `r.SystemChannelID()` since locks in go are not re-entrant.

orderer/common/multichannel/registrar.go
```diff
@@ -577,8 +577,56 @@ func (r *Registrar) joinAsFollower(ledgerRes *ledgerResources, clusterConsenter
...
+	delete(r.chains, channelID)
+
+	logger.Infof("Removed channel: %s", channelID)
+
+	return nil
+}
+
+func (r *Registrar) removeFollower(channelID string, follower *follower.Chain) error {
+	follower.Halt()
+
```
tock-ibm, line 642: If the follower is doing on-boarding, we have to remove the join-block as well. If we don't, after a restart the orderer will rejoin... that's probably just a TODO comment at this stage, but thought I'd mention it so it won't get forgotten.

PR #1513, opened 2016-05-19
Abac

andresgaragiola:
We have been work on this feature with @diegomasini  and @raullaprida  it is a large PR including  support to attributes and Attribute Based Access Control mechanisms.  This PR include the changes of the closed PR #1169. 
## Description

We have added the doc aca.md to describe this feature (this document is included in the doc folder but can be acceded from [this](https://github.com/hyperledger/fabric/wiki/Attribute-Based-Access-Control-%28ABAC%29) wiki page too).
We have added an example called "chaincode_example07" this example describe how ABAC verification could be invoked. The document "ChaincodeAPI.md" was modified to explain it.
## Motivation and Context

This is a required feature as is mentioned in #1139
Fixes #1139
## How Has This Been Tested?
- Current behave tests has been run without errors.
- Current unit tests has been run. The test TestExecuteDeploySysChaincode defined in the file exectransaction_test.go fails but this test fails in the master too (we are working with @muralisrini to understand if this is a problem related with the PR or not)
- We have modified the test TestClientGetAttributesFromTCert defined in crypto_test.go to test the new functionality.
- A new behave test is pending we will work with @jeffgarratt on that.
## Checklist:

- [ ] I have added a [Signed-off-by](https://github.com/hyperledger/fabric/blob/master/CONTRIBUTING.md#legal-stuff) (This is a PR **for review** that is the reason because I haven't added a "Signed-off", I will add a "Signed-off" once the test problem described above will be fixed).
- [X] We added new documentation
- [X] We added new tests
- [X] I have run [goimports](https://godoc.org/golang.org/x/tools/cmd/goimports), [go vet](https://golang.org/cmd/vet/), and [golint](https://github.com/golang/lint). I have cleaned up all valid errors and warnings in code I have added or modified. These tools may generate false positives. Don't be worried about ignoring some errors or warnings. The goal is clean, consistent, and readable code.

Signed-off-by:

PR #1639, opened 2016-05-27
Attribute Base Access Control core functions

diegomasini:
We have been reworking PR #1513 with @andresgaragiola and @raullaprida to split it in smaller PRs. This is the first one and it includes the core functions needed for Attribute Base Access Control. 
## Description

Core functions for Attribute Based Access Control needed by Membership Services, the client and chaincode were factored out in a common package that can be included in any of the three components mentioned. Importing this package does not provide any dependency with the rest of the fabric, just with the utils and the primitives packages. These three packages with common functionality could be extracted in a separated library.
## Motivation and Context

This is a required feature to address #1139
## How Has This Been Tested?

Current behave tests has been run without errors.
Current unit tests has been run without errors.
Additional unit tests where included (attributes_test.go in core/crypto/abac).
A new behave test is pending we will work with @jeffgarratt on that.
## Checklist:

Signed-off-by: Diego Masini damasini@us.ibm.com / diego.masini@gmail.com


core/crypto/abac/attributes.go
```diff
@@ -0,0 +1,271 @@
...
+	return key, err
+}
+
+//GetValueForAttribute derives the K for the attribute "attributeName" and returns the value
+func GetValueForAttribute(attributeName string, preK0 []byte, cert *x509.Certificate) ([]byte, error) {
+	_, value, err := getKAndValueForAttribute(attributeName, preK0, cert)
+	return value, err
+}
+
+func createABACHeaderEntry(preK0 []byte) (*pb.ABACMetadataEntry, error) {
```
srderson: Should `error` be part of the return definition if the function never returns an error?

core/crypto/abac/attributes.go
```diff
@@ -0,0 +1,271 @@
...
+	attKey := getAttributeKey(preK0, HeaderAttributeName)
+	return &pb.ABACMetadataEntry{HeaderAttributeName, attKey}, nil
+}
+
+func createABACMetadataEntry(attributeName string, preK0 []byte) (*pb.ABACMetadataEntry, error) {
```
srderson: Should `error` be part of the return definition if the function never returns an error?

core/crypto/abac/attributes.go
```diff
@@ -0,0 +1,271 @@
...
+		entry, err := createABACMetadataEntry(key, preK0)
+		if err == nil {
+			entries = append(entries, entry)
+		}
+	}
+	headerEntry, err := createABACHeaderEntry(preK0)
+	if err == nil {
+		entries = append(entries, headerEntry)
+	}
+	return &pb.ABACMetadata{metadata, entries}, nil
```
srderson: Should any of the `err` variables be returned here?

srderson:
Thanks @diegomasini, much easier to review a commit of this size. You probably want to squash your commits as there are so many for this change.

core/crypto/abac/attributes.go
diegomasini: Yes, the error comes from the getKAndValueForAttribute call, it could be an error or nil. Are you thinking about a different way to achieve this?

core/crypto/abac/attributes.go
diegomasini: Yes, the error comes from the getKAndValueForAttribute call, it could be an error or nil. Are you thinking about a different way to achieve this?

core/crypto/abac/attributes.go
srderson: My comment is for the `createABACHeaderEntry` function. It never calls `getKAndValueForAttribute`

core/crypto/abac/attributes.go
srderson: My comment is for the `createABACMetadataEntry` function. It never calls `getKAndValueForAttribute`

core/crypto/abac/attributes.go
diegomasini: You are right, let me fix it

diegomasini:
@srderson I just removed the err variables from CreateABACMetadataObjectFromCert() you mentioned and I squashed the changes of this PR.

srderson:
@adecaro please give this a quick review when you have a chance. Thanks

core/crypto/abac/attributes.go
```diff
@@ -0,0 +1,267 @@
...
+var (
+	// TCertEncAttributesBase is the base ASN1 object identifier for attributes.
+	// When generating an extension to include the attribute an index will be
+	// appended to this Object Identifier.
+	TCertEncAttributesBase = asn1.ObjectIdentifier{1, 2, 3, 4, 5, 6}
+
+	// TCertAttributesHeaders is the ASN1 object identifier of attributes header.
+	TCertAttributesHeaders = asn1.ObjectIdentifier{1, 2, 3, 4, 5, 6, 9}
+
+	padding = []byte{255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255}
```
adecaro: What is the purpose of padding and why does it have that specific value?

core/crypto/abac/attributes.go
```diff
@@ -0,0 +1,267 @@
...
+	if err != nil {
+		return nil, encrypted, err
+	}
+	return value, encrypted, nil
+}
+
+//EncryptAttributeValue encrypts "attributeValue" using "attributeKey"
+func EncryptAttributeValue(attributeKey []byte, attributeValue []byte) ([]byte, error) {
+	value := append(attributeValue, padding...)
+	return primitives.CBCPKCS7Encrypt(attributeKey, value)
```
adecaro: Because the attributes are used in multiple places, shall we use GCM mode to protect against malleability attacks? The primitives package offers such an encryption mode.

core/crypto/abac/attributes.go
```diff
@@ -0,0 +1,267 @@
...
+}
+
+//getAttributeKey returns the attributeKey derived from the preK0 to the attributeName.
+func getAttributeKey(preK0 []byte, attributeName string) []byte {
+	return primitives.HMACTruncated(preK0, []byte(attributeName), 32)
```
adecaro: Shall we add a prefix to []byte(attributeName) in order to distinguish the scope of the keys derived from preK0? 

Suppose that in the future we will use preK0 to derive keys wth different scopes. A way to identify the scope is by adding a prefix (associated to that specific scope) to the argument. This offers more modularity and retro compatibility. Of course, if we are pretty sure that preK0 will be used only for one thing, then we don't need this prefixing :)

core/crypto/abac/proto/abac.proto
```diff
@@ -0,0 +1,33 @@
...
+
+package protos;
+
+//ABACMetadataEntry is an entry within the metadata that store an attribute name with its respective key.
+message ABACMetadataEntry {
+    string AttributeName = 1;
+    bytes AttributeKey = 2;
+}
+
+//ABACMetadata hold both the original metadata bytes and the metadata required by the Attribute Based Access Control
```
adecaro: Just a clarification: Is the marshaled version of ABACMetadata what we will have to put in the transaction metadata for access control?

core/crypto/abac/attributes.go
```diff
@@ -0,0 +1,267 @@
...
+
+	return result, nil
+}
+
+//ReadAttributeHeader read the header of the attributes.
+func ReadAttributeHeader(tcert *x509.Certificate, headerKey []byte) (map[string]int, bool, error) {
+	var err error
+	var headerRaw []byte
+	encrypted := false
+	if headerRaw, err = utils.GetCriticalExtension(tcert, TCertAttributesHeaders); err != nil {
```
adecaro: Just a clarification: Does the function assume that the x509 certificate has been already validated? I guess so, given that certificates must be always validated before extracting any information from them.

adecaro:
Hi @diegomasini, a way to avoid to have all these "Merge remote-tracking branch 'upstream/master'" is to rebase rather then merging. This helps reducing the number of commits :)

core/crypto/abac/attributes.go
diegomasini: Yes, at this point the TCert is validated in the stub, that code is not part of this PR since we are splitting a bigger PR in smaller pieces.

core/crypto/abac/proto/abac.proto
diegomasini: Yes, the use is part of the client. And there is also an example of its use but is part of a coming PR as part of the split of the initial PR.

core/crypto/abac/attributes.go
diegomasini: When a value is decrypted the decryption does not fail if someone uses the wrong key. It just produce garbage information. The padding was introduced to allow someone to verify that the decryption was successful. It needs to be a constant known value.

core/crypto/abac/attributes.go
diegomasini: Attributes are associated to a given affiliation. In order to request a batch of TCerts the client needs to provide its ECert. That way attributes are implicitly scoped to that affiliation. It is possible to add a prefix to the attribute name if needed.

core/crypto/abac/attributes.go
diegomasini: Attribute's keys are distributed in the transaction by using the provided confidentiality mechanism. I would like to review this with you since for the asset transfer use case we need to use GCM and potentially modify the transaction structure. But for the scope of this PR I think the confidentiality mechanism is enough. Also, I don't know is this is clear, but attributes are encrypted each time with a different key.

core/crypto/abac/attributes_test.go
```diff
@@ -0,0 +1,229 @@
...
+
+func TestGetKAndValueForAttribute(t *testing.T) {
+	expected := "Software Engineer"
+
+	tcert, prek0, err := loadTCertAndPreK0()
+	if err != nil {
+		t.Error(err)
+	}
+
+	_, attribute, err := getKAndValueForAttribute("position", prek0, tcert)
```
christo4ferris: why are we (only) testing private functions? Seems to me that we should be testing both, but most importantly the public functions. Also, we only have positive tests here. Should have tests that ensure the correct processing of calls that pass incorrect argument values.

core/crypto/abac/attributes_test.go
```diff
@@ -0,0 +1,229 @@
...
+
+func TestGetValueForAttribute(t *testing.T) {
+	expected := "Software Engineer"
+
+	tcert, prek0, err := loadTCertAndPreK0()
+	if err != nil {
+		t.Error(err)
+	}
+
+	value, err := GetValueForAttribute("position", prek0, tcert)
```
christo4ferris: should include cases for incorrect arguments and validate expected errors are returned, etc

core/crypto/abac/attributes_test.go
```diff
@@ -0,0 +1,229 @@
...
+
+func TestReadTCertAttributeByPosition(t *testing.T) {
+	expected := "Software Engineer"
+
+	tcert, prek0, err := loadTCertAndPreK0()
+	if err != nil {
+		t.Error(err)
+	}
+
+	encryptedAttribute, err := ReadTCertAttributeByPosition(tcert, 1)
```
christo4ferris: we should really be including more variability in the test cases. What if the function could only return the 1st position? please add more test cases.

core/crypto/abac/attributes_test.go
```diff
@@ -0,0 +1,229 @@
...
+		t.Error("Empty header should produce a parsing error")
+	}
+}
+
+func TestBuildAndParseAttributesHeader(t *testing.T) {
+	attributes := make(map[string]int)
+	attributes["company"] = 1
+	attributes["position"] = 2
+
+	header := string(BuildAttributesHeader(attributes)[:])
```
christo4ferris: this public function is only being indirectly tested, should have tests that check for things like passing an empty set of attributes, for instance.

core/crypto/abac/attributes.go
```diff
@@ -0,0 +1,267 @@
...
+	cert, err := utils.DERToX509Certificate(raw)
+	if err != nil {
+		return nil, err
+	}
+
+	return CreateABACMetadataFromCert(cert, metadata, preK0, attributeKeys)
+}
+
+//GetABACMetadata object from the original metadata "metadata".
+func GetABACMetadata(metadata []byte) (*pb.ABACMetadata, error) {
```
christo4ferris: Please add tests for this public function

core/crypto/abac/attributes.go
```diff
@@ -0,0 +1,267 @@
...
+
+//CreateABACMetadataFromCert creates the ABACMetadata from the original metadata and certificate "cert".
+func CreateABACMetadataFromCert(cert *x509.Certificate, metadata []byte, preK0 []byte, attributeKeys []string) ([]byte, error) {
+	abacMetadata := CreateABACMetadataObjectFromCert(cert, metadata, preK0, attributeKeys)
+
+	return proto.Marshal(abacMetadata)
+}
+
+//CreateABACMetadata create the ABACMetadata from the original metadata
+func CreateABACMetadata(raw []byte, metadata []byte, preK0 []byte, attributeKeys []string) ([]byte, error) {
```
christo4ferris: Please add tests for this

core/crypto/abac/attributes.go
```diff
@@ -0,0 +1,267 @@
...
+		entry := createABACMetadataEntry(key, preK0)
+		entries = append(entries, entry)
+	}
+	headerEntry := createABACHeaderEntry(preK0)
+	entries = append(entries, headerEntry)
+
+	return &pb.ABACMetadata{metadata, entries}
+}
+
+//CreateABACMetadataFromCert creates the ABACMetadata from the original metadata and certificate "cert".
```
christo4ferris: please add tests for this

core/crypto/abac/attributes.go
```diff
@@ -0,0 +1,267 @@
...
+	return &pb.ABACMetadataEntry{HeaderAttributeName, attKey}
+}
+
+func createABACMetadataEntry(attributeName string, preK0 []byte) *pb.ABACMetadataEntry {
+	attKey := getAttributeKey(preK0, attributeName)
+	return &pb.ABACMetadataEntry{attributeName, attKey}
+}
+
+//CreateABACMetadataObjectFromCert creates an ABACMetadata object from certificate "cert", metadata and the attributes keys.
+func CreateABACMetadataObjectFromCert(cert *x509.Certificate, metadata []byte, preK0 []byte, attributeKeys []string) *pb.ABACMetadata {
```
christo4ferris: please add tests for this

core/crypto/abac/attributes_test.go
andresgaragiola: TestGetKAndValueForAttribute is a test for private function "getKAndValueForAttribute", this private function is called by two public functions "GetValueForAttribute" and "GetKForAttribute". "GetValueForAttribute" is tested in the function TestGetValueForAttribute and we are adding a new test now called "TestGetKForAttribute" to test GetKForAttribute. Other public functions tests currently exists in this file. TestReadTCertAttributeByPosition_InvalidPositions is an example of a negative case test we will be adding more negative cases tests to this file.

christo4ferris:
test coverage is currently 67% for this new component, please use a tool such as go-carpet to flesh out the remaining unit test coverage.

core/crypto/abac/attributes_test.go
diegomasini: I think I don't understand completely here, position is an attribute holding the position inside the company, not a position inside the TCert.

christo4ferris: to clarify my comment: if the function only returned the cert at position 1, this test passes and everything looks fine... but requesting the cert at position 2 would fail - and clearly that would be a bug.

diegomasini:
@srderson @binhn could you take a look at this PR again? We addressed the comments posted here and as @christo4ferris suggested we are using go-carpet to measure the testing coverage. With the last unit tests it is now 86.2% Remember that this PR needs to be processed to allow us to send the rest of the PRs that constitutes the whole ABAC feature.

christo4ferris:
86% is decent thank you... I can't review in detail now but am pleased with the coverage.

binhn:
@diegomasini Thanks Diego for your patience and collaboration on this pull request. 

I looked over this again tonight, and one last question: In `abac.proto`, you define `ABACMetadata`. Why metadata? The content obviously contains an array of attributes, not something that we don't know. Also `ABACMetadata` contains `Metadata` as bytes. It is not clear to me why we need both `Metadata` and `Entries` in this struct. Could you offer some explanation as part of comment of the struct.

andresgaragiola:
thanks for your comment @binhn. ABACMetadata includes both the original metadata and one entry per each attribute included in the transaction. Each ABACMetadataEntry includes the attributeName and the key required to decrypt that attribute. We have added the metadata field to wrap the existing metadata as it is used currently. We also provided an API to read the attributes from it.

binhn:
@diegomasini @andresgaragiola  As discussed, we agreed to change the package name, including files and functions, so that it would be more general since attributes are not only for abac but can be used anyhow the app wishes. We also agreed to defer the resolution to the problem of having the peer knows about the decryption key of the attribute keys, until we complete other parts of attributes so that other work may continue.

andresgaragiola:
@binhn as we discussed today, we recently have pushed a new version of this library renamed to 'attributes' instead of 'abac' and some additional test increasing the coverage to 92%.

binhn:
@andresgaragiola Excellent.  Thank you!

Issue #1609, opened 2016-05-26
"make unit-test" needs to depend on protoc-gen-go


ghaskins:
## Description

The unit-tests depend on protoc-gen-go.  However, the makefile does not properly represent this relationship.  Therefore, it is possibe that the tools might not be compiled ahead of running the tests.
## Describe How to Reproduce

"make clean unit-test" will cause the system to clean out previously compiled tools and then try to run the unit-tests without them present.

@tuand27613

ghaskins:
Workaround: ensure you run "make gotools" prior to running unit-test, such as "make gotools unit-test"

PR #1641, opened 2016-05-27
Fixes for Makefile issues 1534 and 1609

### LINKED ISSUE #1534: "make unit-test" can leave docker-based peer instance running
## Description

The current Makefile will fire up a temporary peer instance during unit-test execution and then shut it down at the conclusion.  However, if the unit-tests fail or are otherwise interrupted, the peer instance may be left running.  This should be cleaned up.
## Describe How to Reproduce

Run "make unit-test" and then cntrl-c the tests mid-way.  Try to run a second session and it will run into errors when it tries to create an instance when one is already running.  You may also observe the instance is still running with "docker ps".

Work-around: use docker ps + docker kill to remove dangling session.

ghaskins:
Makefile fix rollup:  Fixes #1534 and #1609 
## Description

The makefile has been enhanced in two primary ways
1. We now make unit-test depend on gotools so we can ensure that the protoc-gen-go prerequisite is met before running the tests (Issue #1609).
2. We moved unit-test execution to a script so that we can instrument proper cleanup of the docker-based peer (Issue #1534)
## Motivation and Context

Bug squashing
## How Has This Been Tested?

"make unit-test" has been driven through various scenarios, such as making sure "make clean unit-test" properly rebuild tools, and that unit-test failures do not leave the peer running.
## Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -66,7 +66,7 @@ peer: base-image
 membersrvc:
 	cd membersrvc; CGO_CFLAGS=" " CGO_LDFLAGS="$(CGO_LDFLAGS)" go build
 
-unit-test: peer-image
+unit-test: peer-image gotools
 	@echo "Running unit-tests"
 	$(eval CID := $(shell docker run -dit -p 30303:30303 hyperledger-peer peer node start))
 	@go test -timeout=20m $(shell go list $(PKGNAME)/... | grep -v /vendor/ | grep -v /examples/)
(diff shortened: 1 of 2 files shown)


ghaskins:
@srderson ideally fast track this please:  some users have started running into these bugs.

srderson:
LGTM. Tested locally and confirmed both #1534 and #1609 are resolved.

Issue #1567, opened 2016-05-24
'peer.port' variable in 'core.yaml


kchristidis:
## Description

Is the `peer.port` variable in `core.yaml` used anywhere? As far as I can tell, it isn't.
## Action to take

If it's not needed, I suggest removing it from the config file.

bcbrock:
I agree #1401 If you have the time please remove it

kchristidis:
Ah, hadn't seen your #1401 issue - sorry.

I'll go ahead and create a PR tomorrow that takes care of this and #1548.

lehors:
I did some digging around this last night and peer/main.go also defines a flag peer-port on [line 242](https://github.com/hyperledger/fabric/blob/master/peer/main.go#L242) which doesn't seem to be used anywhere and have any actual interaction with the parameter set up in core.yaml, although I think it should if it were to be useful at all.

`flags.Int("peer-port", 30309, "Port this peer listens to for incoming connections")`

bcbrock:
@lehors @kchristidis I don't believe any of those flags actually work. VIPER does not equate command line flags to YAML variables. I had an issue on this that I closed because I couldn't remember what the issue was - now you've reminded me #578

lehors:
Interesting. I was tempted to dive deeper into this Viper setup so I would really understand and kind of decided not to but now you've got me even more interested...

Issue #578, opened 2016-02-01
obc-peer command-line flags do not (can not) work


bcbrock:
The Viper/Cobra option-processing system used by openblockchain
apparently does not support the idea that command-line flags are
equivalent to nested keys from config files. See for example
recent conversations here: https://gitter.im/spf13/viper

This means that many (if not all) of the obc-peer command line
flags do not work, since the flag X_Y_Z is distinct from the
nested key X.Y.Z., and the codes that I've examined only look for
the nested key.

You can see this for yourself by putting lines like the following
at the top of the `serve()` function in main.go:

```
fmt.Println("SERVE VIPER peer_tls_key_file", viper.GetString("peer_tls_key_file"))
fmt.Println("SERVE VIPER peer.tls.key.file", viper.GetString("peer.tls.key.file"))
```

and call the obc-peer as

```
obc-peer peer peer-tls-key-file=foo.bar
```

The output is

```
SERVE VIPER peer_tls_key_file foo.bar
SERVE VIPER peer.tls.key.file testdata/server1.key
```

that is, the nested key is taken from the openchain.yaml, in spite of being
'overridden' on the command line.

bcbrock:
My description of the issue is so vague that I can't remember what I was complaining about any more. Closing.

kchristidis:
I spent the past couple of days looking into the `pflag`, `viper`, and `cobra` packages. I may be missing something, but I don't see anything wrong with the `viper` package itself here. Rather, the issue seems to be in the way we use it.

Specifically:

We bind Viper key `a_b_c` to flag `a-b-c`. We then use/set flag `a-b-c` and expect Viper key `a.b.c` to be modified. (So as to override config value `a.b.c`, which is binded to the `a.b.c` key.)

This is not reasonable.

I _assume_ that part of the confusion stems from the use of these commands:

```
replacer := strings.NewReplacer(".", "_")
viper.SetEnvKeyReplacer(replacer)
```

and an expectation that this changes the `a_b_c` key to `a.b.c`.

However what `viper.SetEnvKeyReplacer()` does here is to simply bind any `A_B_C` **ENV** variables to `a.b.c` keys. I tested this by setting up a MWE: https://github.com/kchristidis/golang-experiments/tree/viper

The solution here is to rename all Viper keys in our code from the `a_b_c` format to the `a.b.c` one.

[...]

PR #1645, opened 2016-05-27
Remove unused config variable and Docker Compose files

### LINKED ISSUE #1548: Logic behind `docker-compose-files` folder
## Description

I'm not sure I get the logic behind the `fabric/consensus/docker-compose-files` directory being how it is today, and the included Markdown file (which should probably be moved to the Wiki) doesn't really clarify things.

This directory currently has nine YAML files.

I see the need for `compose-consensus-4-{batch, classic, sieve}.yml` so we're three files down - good. Based on what I read in the Markdown file, `compose-consensus-4.yml` is actually the, err, true `docker-compose.yml` and it extends `compose-consensus-4-defaults.yml`. So:
- `compose-consensus-4-batch.yml`
- `compose-consensus-4-classic.yml`
- `compose-consensus-4-sieve.yml`
- `compose-consensus-4-defaults.yml`
- `compose-consensus-4.yml`

Are accounted for.
## Questions
1. What role does the _existing_ `docker-compose.yml` (and `compose-defaults.yml`) serve?
2. Is there a reason why we're keeping `compose-consensus-4-links.yml` in a separate file and we don't merge it with `compose-consensus-4-defaults.yml`?
## Expected Behavior

As a developer, I would expect the `docker-compose.yml` file to do what `compose-consensus-4.yml` and `compose-consensus-4-links.yml` currently do.

There's always a chance I'm missing something.

kchristidis:
## Description

Removes the unused `port` setting in `peer/core.yaml` and the now-obsolete Docker Compose files in the `consensus` directory.
## Motivation and Context

Eliminate user confusion. Fixes #1548, #1567.
## How Has This Been Tested?

No new test needed, this PR just removes unused files. Passes all unit and behave tests.
## Checklist:


ghaskins:
Thank you @kchristidis, this is a nice cleanup.  It sounds from the comments in the referenced issues that this is generally agreed to be the right actions.  However, I would feel better if a stakeholder or two could chime in with agreement here before merging.

peer/core.yaml
```diff
@@ -107,8 +107,6 @@ peer:
     # This case is useful for docker containers.
     addressAutoDetect: false
 
-    # Peer port to accept connections on
-    port:    30303
```
ghaskins, line 111: I'm a little apprehensive about dropping this.  It seems we do in fact use port 30303:

```
$ git grep peer.port
peer/main.go:   flags.Int("peer-port", 30303, "Port this peer listens to for incoming connections")
peer/main.go:   viper.BindPFlag("peer_port", flags.Lookup("peer-port"))
```

I am not sure if the viper mechanism automatically binds "peer.port" in YAML to "peer-port" in PFlags, but at the very least we can see that this does seem to be a tunable knob and it might be convenient to allow the knob to be set from YAML.  That means we have a few different scenarios:
1. The YAML and PFlags are already wired together and probably should remain that way.
2. The YAML and PFlags are not wired together but should be, we need to fix.
3. This patch does the right thing.

Please investigate further.  If you would like to submit the compose file deletion as a separate PR, I have no problem merging those.  Thanks @kchristidis

kchristidis:
@ghaskins: Thanks for the note. As you note in #1662, and as a grep for `viper.GetInt("peer.port")` shows, nothing in the code reads this option. For the sake of completeness, I will amend the original PR by removing the lines in `main.go` that reference the `port` variable, so I'm going for Option 3.

Depending on the verdict of #1662, feel free to accept or reject it.

bcbrock:
@ghaskins Viper does not bind command-line flags to YAML/Environment Variables. All of the command-line flags other than --logging-level should be eliminated from the main.go as they don't work. #578

binhn:
@tuand27613 @corecode I am not familiar enough with the consensus docker compose, so could you please review and comment.  In general, I like what @kchristidis did: Removing a lot of stuff :-)

tuand27613:
Agree with @kchristidis  on removal of consensus/docker-compose-files/*
These files originally were for unit testing back in obc-peer days. Not needed as we now have the files in ./bddtests, Bishop's busywork tool and Ramesh's utility.

kchristidis:
@ghaskins is waiting on me, and I'm checking to see if @bcbrock [is right](https://github.com/hyperledger/fabric/pull/1645#issuecomment-222715962).

I have reproduced the behavior that @bcbrock sees in #578, i.e. the flag does not bind to the config values (but the environment variables do). This is not supposed to happen according to the [viper documentation](https://github.com/spf13/viper).

Running some tests to make sure we're configuring this right. Will update.

kchristidis:
1. @ghaskins: Removed the additional `port` references you found. Once, Travis is done, this PR should be good to go.
2. I debugged the issue that @bcbrock brought up in #578. Viper works fine; no need to remove the flags but we do need to issue a fix. Will push a separate PR for this today.

kchristidis:
(@ghaskins: The failure reported by Travis seems to be on test #680, which [according to your statement](https://github.com/hyperledger/fabric/pull/1685#issuecomment-223641237) is "notoriously flaky". Let me know if you need anything else from me.)

christo4ferris:
if 680 is flakey, can we stub it out until it is fixed?

jyellick:
Looking at the logs, I am seeing the failure as caused by:

```
    When I query chaincode "example2" function name "query" with value "a" on peers                                             # steps/peer_basic_impl.py:422
      | vp0 | vp1 | vp2 | vp3 |
      Traceback (most recent call last):
        File "/usr/local/lib/python2.7/dist-packages/behave/model.py", line 1456, in run
          match.run(runner.context)
        File "/usr/local/lib/python2.7/dist-packages/behave/model.py", line 1903, in run
          self.func(context, *args, **kwargs)
        File "steps/peer_basic_impl.py", line 424, in step_impl
          query_common(context, chaincodeName, functionName, value, True)
        File "steps/peer_basic_impl.py", line 450, in query_common
          resp = requests.post(request_url, headers={'Content-type': 'application/json'}, data=json.dumps(chaincodeInvocationSpec), timeout=30, verify=False)
        File "/usr/local/lib/python2.7/dist-packages/requests/api.py", line 94, in post
          return request('post', url, data=data, json=json, **kwargs)
        File "/usr/local/lib/python2.7/dist-packages/requests/api.py", line 49, in request
          return session.request(method=method, url=url, **kwargs)
        File "/usr/local/lib/python2.7/dist-packages/requests/sessions.py", line 457, in request
          resp = self.send(prep, **send_kwargs)
        File "/usr/local/lib/python2.7/dist-packages/requests/sessions.py", line 569, in send
          r = adapter.send(request, **kwargs)
        File "/usr/local/lib/python2.7/dist-packages/requests/adapters.py", line 422, in send
          raise ReadTimeout(e, request=request)
      ReadTimeout: HTTPConnectionPool(host='172.17.0.71', port=5000): Read timed out. (read timeout=30)
```

The last line of vp3's log (172.17.0.71) is:

```
19:53:20.602 [rest] Query -> ERRO f08 {"Error": "Querying Chaincode -- Error:Failed to execute transaction or query(Timeout expired while executing transaction)"}
```

Prior to this timeout in vp3's log I see:

```
19:51:49.807 [consensus/obcpbft] processEvent -> DEBU ce7 Replica 3 batch main thread looping
19:51:49.807 [consensus/obcpbft] processEvent -> DEBU ce8 Replica 3 processing event
19:51:49.807 [consensus/obcpbft] processEvent -> INFO ce9 Replica 3 application caught up via state transfer, lastExec now 42
19:51:49.807 [consensus/obcpbft] moveWatermarks -> DEBU cea Replica 3 updated low watermark to 42
19:51:49.807 [consensus/handler] ValidateState -> DEBU ceb Validating the current state
19:51:49.807 [consensus/obcpbft] executeOutstanding -> DEBU cec Replica 3 attempting to executeOutstanding
19:51:49.807 [consensus/obcpbft] executeOutstanding -> DEBU ced Replica 3 certstore map[]
```

This indicates to me that vp3 is actually in a good state, state transfer has completed, and it is up to date. (Checking the log for vp1 confirms, that indeed, the last seqNo executed should be 42).  I think it's worth investigating what's going on here, because as far as consensus/statetransfer (which is the target of this test) is concerned, this test is completing successfully.  Instead, it is the query to verify this which is timing out.

kchristidis:
@jyellick: Interesting. I pushed an empty commit so as to re-trigger Travis. Let's see what we get this time.

ghaskins:
@jyellick are you suspecting that the unit-test failures are possibly legit and we should hold off this merge?

jyellick:
@ghaskins I have no reason to believe that this particular set of commits is problematic.  I was responding to the suggestion that we disable the issue_680 behave test.  When tests are failing because of inconsistent timing or other problems, I understand disabling them until the test can be redesigned to work more consistently.  In this case though, as best as I can tell, the test is operating as expected, and it's unclear why this query is timing out.

As you can see from the log, vp3 was caught up and should have been in a consistent state as of 19:51:50, and a query timed out after 30s at 19:53:20.  This would indicate to me that we are witnessing a bug in the fabric that should be investigated and fixed, not a bug in the test that should be suppressed.

ghaskins:
@jyellick ah, ok.  Thank you for the explanation.  I just wanted to make sure you weren't concerned that this patch might be causing a real failure.  It sounds like it should be ok to go in.

ghaskins:
LGTM

christo4ferris:
@kchristidis you can re-run travis by clicking on the details link in the PR summary page and there's a small recycle glyph in the Travis UI

christo4ferris:
@jyellick agree, think we should open an issue and have someone investigate

kchristidis:
@christo4ferris: I did not know that and instead went with what the Travis CI README suggested. Thank you!

PR #1650, opened 2016-05-28
Fix go vet warnings

srderson:
## Description

Fix up a number of `go vet` warnings.
## Motivation and Context

All developers are now being asked to run `go vet` on proposed changes. Cleaning up pre-existing issues.
## How Has This Been Tested?

`make checks` has been run. There is no new function in this PR so no new tests are required. This only fixes `go vet` warnings.
## Checklist:

### DIFF
--- a/consensus/obcpbft/mock_network_test.go
+++ b/consensus/obcpbft/mock_network_test.go
@@ -253,11 +253,8 @@ func (net *testnet) process() error {
 			case <-time.After(100 * time.Millisecond):
 				continue
 			}
-			return nil
 		}
 	}
-
-	return nil
 }
 
 func (net *testnet) processContinually() {
--- a/core/crypto/crypto_test.go
+++ b/core/crypto/crypto_test.go
@@ -160,11 +160,11 @@ func TestRegistrationSameEnrollIDDifferentRole(t *testing.T) {
 	}
 
 	if err := RegisterValidator(conf.Name, nil, conf.GetEnrollmentID(), conf.GetEnrollmentPWD()); err == nil {
-		t.Fatalf("Reusing the same enrollment id must be forbidden", err)
+		t.Fatal("Reusing the same enrollment id must be forbidden", err)
 	}
 
 	if err := RegisterPeer(conf.Name, nil, conf.GetEnrollmentID(), conf.GetEnrollmentPWD()); err == nil {
-		t.Fatalf("Reusing the same enrollment id must be forbidden", err)
+		t.Fatal("Reusing the same enrollment id must be forbidden", err)
 	}
 }
 
@@ -628,27 +628,27 @@ func TestPeerVerify(t *testing.T) {
 
 	err = peer.Verify(nil, signature, msg)
 	if err == nil {
-		t.Fatalf("Verify should fail when given an empty id.", err)
+		t.Fatal("Verify should fail when given an empty id.", err)
 	}
 
 	err = peer.Verify(msg, signature, msg)
 	if err == nil {
-		t.Fatalf("Verify should fail when given an invalid id.", err)
+		t.Fatal("Verify should fail when given an invalid id.", err)
 	}
 
 	err = peer.Verify(validator.GetID(), nil, msg)
 	if err == nil {
-		t.Fatalf("Verify should fail when given an invalid signature.", err)
+		t.Fatal("Verify should fail when given an invalid signature.", err)
 	}
 
 	err = peer.Verify(validator.GetID(), msg, msg)
 	if err == nil {
-		t.Fatalf("Verify should fail when given an invalid signature.", err)
+		t.Fatal("Verify should fail when given an invalid signature.", err)
 	}
 
 	err = peer.Verify(validator.GetID(), signature, nil)
 	if err == nil {
-		t.Fatalf("Verify should fail when given an invalid messahe.", err)
+		t.Fatal("Verify should fail when given an invalid messahe.", err)
 	}
 }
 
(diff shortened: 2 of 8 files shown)


christo4ferris:
Hi srderson,

Thanks for submitting this pull request!

I can confirm that the DCO1.1 sign-off has been included. It is okay to process this pull request.

dco-bot

ghaskins:
LGTM

core/ledger/kvledger/kv_ledger.go
```diff
@@ -52,6 +52,9 @@ type kvLedger struct {
 	historyDB                 *history.DB
 	configHistoryRetriever    *confighistory.Retriever
 	snapshotRequestBookkeeper *snapshotRequestBookkeeper
+	events                    chan *event
+	commitProceed             chan struct{}
+	requestResponses          chan *requestResponse
```
manish-sethi: four things to maintain is a good enough to have a separate construct say `snapshotMgr` to abstract from kvledger code.

core/ledger/kvledger/kv_ledger.go
```diff
@@ -158,6 +164,18 @@ func newKVLedger(initializer *lgrInitializer) (*kvLedger, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	// start a goroutine to process commit and snapshot events.
+	go l.processEvents(bcInfo.Height)
+
+	if err = l.recoverSnapshot(bcInfo.Height); err != nil {
+		return nil, err
+	}
```
manish-sethi: - Like the above comment, when you introduce a separate `snapshotMgr` struct. These two functions make better sense on that... Otherwise, `processEvents` appears to be too generic.
- When we launch a never ending gorouting at the start, it's desired to stop it when the ledger is closed. Though, peer does not close ledgers gracefully, so, that should not be a big problem as of now but in a separate PR, we can introduce another signal channel for closing the other channels and getting out of the loop. I suggest opening a Jira for this for now.

core/ledger/kvledger/process_events.go
```diff
@@ -0,0 +1,145 @@
+/*
```
manish-sethi: The code in the file can be moved to snapshot_mgmt.go

core/ledger/kvledger/process_events.go
```diff
@@ -0,0 +1,145 @@
...
+	// snapshotInProgress is set to true before a generateSnapshot is called when processing commitDone or requestAdd event
+	// and set to false when snapshotDone event is received
+	snapshotInProgress := false
+
+	for {
+		e := <-l.events
+		logger.Debugw("event received", "type", e.typ, "height", e.height, "snapshotInProgress=", snapshotInProgress)
+		switch e.typ {
+		case commitStart:
+			// toStartCommitHeight is updated to be the new block height after the block is committed;
```
manish-sethi: `after` -> `before`

core/ledger/kvledger/process_events.go
```diff
@@ -0,0 +1,145 @@
...
+			// therefore, before commitDone event, toStartCommitHeight is equal to committedHeight + 1.
+			// In other words, if toStartCommitHeight is not equal to committedHeight, it means that a commit is in progress.
+			toStartCommitHeight = e.height
+			if !snapshotInProgress {
+				l.commitProceed <- struct{}{}
+				continue
+			}
+			// there is a in-progress snapshot, log a warning since commit will be blocked on commitProceed channel until snapshotDone event
+			logger.Warnw("commit waiting on snapshot to be generated", "snapshotHeight", committedHeight)
```
manish-sethi: This should be an Info message.

core/ledger/kvledger/process_events.go
```diff
@@ -0,0 +1,145 @@
...
+					logger.Errorw("Failed to generate snapshot", "height", requestedHeight, "error", err)
+				}
+				l.events <- &event{snapshotDone, requestedHeight}
+			}()
+
+		case snapshotDone:
+			requestedHeight := e.height
+			logger.Debugw("snapshot is generated", "height", requestedHeight, "toStartCommitHeight", toStartCommitHeight, "committedHeight", committedHeight)
+			if err := l.snapshotRequestBookkeeper.delete(requestedHeight); err != nil {
+				logger.Errorw("Failed to delete snapshot request", "height", requestedHeight, "error", err)
```
manish-sethi: Better to add to the message... the pending snapshot requests (if any) may not be processed.

core/ledger/kvledger/process_events.go
```diff
@@ -0,0 +1,145 @@
...
+				continue
+			}
+			// ready to generate snapshot
+			snapshotInProgress = true
+			go func() {
+				if err := l.generateSnapshot(); err != nil {
+					logger.Errorw("Failed to generate snapshot", "height", requestedHeight, "error", err)
+				}
+				l.events <- &event{snapshotDone, requestedHeight}
+			}()
```
manish-sethi: I feel that if you reverse the condition and move the snapshot generation code in the if block, that would make this more readable (primarily, because that would make the if condition a rare case, like the other `If's`... and the main more general case is the main flow outside the `if's`).

```
if toStartCommitHeight == committedHeight && requestedHeight == committedHeight{
     // launch generateSnapshot...
}
```

core/ledger/kvledger/snapshot_mgmt.go
```diff
@@ -63,6 +67,34 @@ func (l *kvLedger) DeleteSnapshot(height uint64) error {
 	return errors.Errorf("not implemented")
 }
 
+func (l *kvLedger) recoverSnapshot(blockHeight uint64) error {
```
manish-sethi: Nit: like at other places, to be consistent, better to use `height` instead of `blockHeight`

PR #1651, opened 2016-05-28
tweak README to test dco-bot

christo4ferris:
really to test doc-bot


ghaskins:
LGTM

common/ledger/blkstorage/blockstore_provider.go
```diff
@@ -106,11 +106,11 @@ func (p *BlockStoreProvider) Open(ledgerid string) (*BlockStore, error) {
 	return newBlockStore(ledgerid, p.conf, p.indexConfig, indexStoreHandle, p.stats)
 }
 
-// BootstrapFromSnapshottedTxIDs initializes blockstore from a previously generated snapshot
+// ImportFromSnapshot initializes blockstore from a previously generated snapshot
 // Any failure during bootstrapping the blockstore may leave the partial loaded data
 // on disk. The consumer, such as peer is expected to keep track of failures and cleanup the
 // data explicitly.
-func (p *BlockStoreProvider) BootstrapFromSnapshottedTxIDs(
+func (p *BlockStoreProvider) ImportFromSnapshot(
```
cendhu, line 112: Not related to this PR. For better readability, we can pass the `ledgerID` as the first argument to this function rather than using that `snapshotInfo.LedgerID`. I believe such a change could explicitly convey that the ledger data associated with a channel is being imported from the snapshot.

core/ledger/kvledger/kv_ledger.go
```diff
@@ -243,41 +256,49 @@ func (l *kvLedger) syncStateAndHistoryDBWithBlockstore() error {
...
 	recoverables := []recoverable{l.txmgr}
 	if l.historyDB != nil {
 		recoverables = append(recoverables, l.historyDB)
 	}
 	recoverers := []*recoverer{}
 	for _, recoverable := range recoverables {
-		recoverFlag, firstBlockNum, err := recoverable.ShouldRecover(lastAvailableBlockNum)
+		// nextRequiredBlock is nothing but the nextBlockNum expected by the state DB.
+		// In other words, the nextRequiredBlock is nothing but the height of stateDB.
+		recoverFlag, nextRequiredBlock, err := recoverable.ShouldRecover(lastBlockInBlockStore)
```
cendhu, line 295: `firstBlockNum` is used in the `recoverable`. Give that we change `firstBlockNum` to `nextRequiredBlock`, it would be good to change the `recoverable.firstBlockNum` as well but can be done in a separate PR.

core/ledger/kvledger/kv_ledger_provider.go
```diff
@@ -267,59 +257,75 @@ func (p *Provider) initSnapshotDir() error {
...
-	if exists {
-		return nil, ErrLedgerIDExists
-	}
-	if err = p.idStore.setUnderConstructionFlag(ledgerID); err != nil {
+	if err = p.idStore.createLedgerID(
+		ledgerID,
+		&msgs.LedgerMetadata{
+			Status: msgs.Status_UNDER_CONSTRUCTION,
+		},
+	); err != nil {
```
cendhu, line 273: I was expecting to use `handleLedgerCreationError()` here but I understood later. The name mislead me.

core/ledger/kvledger/kv_ledger.go
```diff
@@ -243,41 +256,49 @@ func (l *kvLedger) syncStateAndHistoryDBWithBlockstore() error {
...
+			if nextRequiredBlock <= lastBlockInSnapshot {
+				return errors.Errorf(
+					"recovery for DB [%s] not possible. Ledger [%s] is created from a snapshot. Last block in snapshot = [%d], DB needs block [%d] onward",
+					recoverable.Name(),
+					l.ledgerID,
+					lastBlockInSnapshot,
+					nextRequiredBlock,
+				)
+			}
+		}
```
cendhu, line 311: Okay to have this check but I couldn't come up with a scenario where this could occur other than a code bug or manual changes to a savepoint or failure during the last snapshot import (but an entry in the `idStore` takes care of that anyway).

core/ledger/kvledger/kv_ledger_provider.go
```diff
@@ -408,59 +415,77 @@ func (p *Provider) Close() {
...
+		if err != nil {
+			return errors.WithMessage(err, "error while iterating over ledger for checking for any under construction ledger")
+		}
+		if !hasMore {
+			return nil
+		}
+		ledgerID := ledgerIDFromMetadataKey(itr.Key())
+		metadataBytes := itr.Value()
+		metadata := &msgs.LedgerMetadata{}
+		if err := proto.Unmarshal(metadataBytes, metadata); err != nil {
```
cendhu: Nit. As the `metadataBytes` is not used after this line, we can do a `proto.Unmarshal(itr.Value(), metadata)` instead.

core/ledger/kvledger/kv_ledger_provider.go
```diff
@@ -408,59 +415,77 @@ func (p *Provider) Close() {
...
+			return errors.Wrapf(err, "error while unmarshalling metadata bytes for ledger [%s]", ledgerID)
+		}
+		if metadata.Status == msgs.Status_UNDER_CONSTRUCTION {
```
cendhu, line 442: Nit. Can revert the condition and `continue` earlier. Maybe the current condition improves the readability.

core/ledger/kvledger/ledger_data_remover.go
```diff
@@ -46,9 +46,11 @@ func (r *ledgerDataRemover) Drop(ledgerID string) error {
 		return err
 	}
 
-	if err = r.historydbProvider.Drop(ledgerID); err != nil {
-		logger.Errorw("failed to drop ledger data from historyDB", "channel", ledgerID, "error", err)
-		return err
+	if r.historydbProvider != nil {
```
cendhu, line 49: Good catch. When we move tests from kvledger to ledger-level integeration test, we need to have test with historyDB enabled/disabled so that we don't miss this in the future.

core/ledger/kvledger/snapshot.go
```diff
@@ -7,24 +7,30 @@ SPDX-License-Identifier: Apache-2.0
...
 	"github.com/pkg/errors"
 )
 
 const (
-	snapshotMetadataFileName     = "_snapshot_signable_metadata.json"
-	snapshotMetadataHashFileName = "_snapshot_additional_info.json"
-	jsonFileIndent               = "    "
-	simpleKeyValueDB             = "SimpleKeyValueDB"
+	snapshotSignableMetadataFileName   = "_snapshot_signable_metadata.json"
+	snapshotAdditionalMetadataFileName = "_snapshot_additional_metadata.json"
```
cendhu, line 31: I remember that we discussed and agreed to have two metadata files. Now, I read the code after some time with a fresh mind. Usage of `signable`, `additional`, and `bootSnapshot` metadata adds some difficulties. In some other comment, I have discussed whether it is possible to remove `additional` and `bootSnapshot` metadata altogether.

core/ledger/kvledger/snapshot.go
```diff
@@ -147,19 +185,190 @@ func (l *kvLedger) generateSnapshotMetadataFiles(
...
 		return err
 	}
-	metadataAdditionalInfo, err := json.MarshalIndent(
-		&snapshotAdditionalInfo{
-			SnapshotHashInHex:        hex.EncodeToString(hash.Sum(nil)),
-			LastBlockCommitHashInHex: hex.EncodeToString(l.commitHash),
+
+	additionalMetadata := &snapshotAdditionalMetadata{
+		SnapshotHashInHex:        hex.EncodeToString(hash.Sum(nil)),
+		LastBlockCommitHashInHex: hex.EncodeToString(l.commitHash),
```
cendhu, line 194: Can't we just get rid of additional metadata and leave the hash computation needed for the signing to the admin or the user? If the peer needs to sign, we can do `{{"Metadata": "{xxxx}", "Signature": "yyyyy"} {"Optional": {"BlockCommitHash": "yyyyy"}}` (similar to our `Envelope`, `Payload`, and `Signature` usage). Removing additional metadata could improve readability IMO. As this is an existing code, even if we agree to do it, we can submit a separate PR. 

If the peer does not need to sign it, we can just write a simple metadata file and let the admin handle signature and other things.

core/ledger/kvledger/kv_ledger.go
cendhu: Got it. Anyway, I assume we will return error when user tries to use peer cli to rebuild dbs, etc...

core/ledger/kvledger/kv_ledger_provider_test.go
```diff
@@ -239,151 +253,111 @@ func TestCheckUpgradeEligibilityEmptyDB(t *testing.T) {
...
-	defer provider2.Close()
-	require.NoError(t, err, "Provider failed to recover an underConstructionLedger")
-	flag, err = provider2.idStore.getUnderConstructionFlag()
-	require.NoError(t, err, "Failed to read the underconstruction flag")
-	require.Equal(t, "", flag)
+	verifyLedgerDoesNotExist(t, provider, ledgerID)
 }
 
-func TestRecoveryHistoryDBDisabled(t *testing.T) {
+func TestLedgerCreationFailureDuringLedgerDeletion(t *testing.T) {
```
cendhu, line 343: As this is more of a sub-test of `TestLedgerCreationFailure()`, we could have merged this test with the above. This test name confused me and had to spend some time to realize that it is not a duplicate.

core/ledger/kvledger/snapshot_test.go
```diff
@@ -138,14 +170,44 @@ func TestGenerateSnapshot(t *testing.T) {
 			},
 		},
 	)
+
+	snapshotDir := SnapshotDirForLedgerHeight(snapshotRootDir, kvlgr.ledgerID, 4)
+
+	t.Run("create-ledger-from-snapshot", func(t *testing.T) {
+		createdLedger := testCreateLedgerFromSnapshot(t, snapshotDir)
```
cendhu, line 177: could have directly embedded the function content here because it is called only once and also the expectation that I had was that the function itself should verify whether the ledger is created successful (don't know whether I had an incorrect expectation) and hence, the next line `verifyCreatedLedger()` confused me. Also, I am not sure about returning `kvLedger` from  `testCreateLedgerFromSnapshot()`

core/ledger/kvledger/snapshot_test.go
```diff
@@ -318,6 +381,243 @@ func TestGenerateSnapshotErrors(t *testing.T) {
...
+		verifyLedgerDoesNotExist(t, provider, metadata.ChannelName)
+	})
+
+	t.Run("idStore-returns-error", func(t *testing.T) {
+		init(t)
+		defer cleanup()
+
+		provider.idStore.close()
+		_, err := provider.CreateFromSnapshot(snapshotDirForTest)
+		require.Contains(t, err.Error(), "error while creating ledger id")
```
cendhu, line 567: `verifyLedgerDoesNotExist()` is missing but not that needed. Would be good to have to maintain consistency with other sub-tests.

core/ledger/kvledger/snapshot_test.go
```diff
@@ -373,20 +673,88 @@ func verifySnapshotOutput(
...
+	previousBlockHash []byte
+	lastCommitHash    []byte
+	namespace         string
+	publicState       map[string]string
+	collectionConfig  map[uint64]*peer.CollectionConfigPackage
+}
+
+func verifyCreatedLedger(t *testing.T,
+	p *Provider,
+	l *kvLedger,
```
cendhu, line 710: Instead of using this arguments, this function could have used `Open(ledgerID)` on the provider to increase the code coverage a bit more (such as loading the stored `snapshotMetadata`)

PR #1658, opened 2016-05-30
Introduce an optionally user-defined UUID generation and a noop system chaincode

gabre:
The commits add two major changes:
- a noop system chaincode that is able to look up older transactions using their ID but does nothing when invoked
- an option for the user to use a custom TxID generation algorithm (specified by a hash function plus a decoding)
- support for sha256base64 (SHA256 hashing with base64 decoding)
## Description

The noop system chaincode is loaded at system startup. It implements the following functions:
- Invoke: this part does nothing (noop), it has no internal logic. It has to be called with the function name parameter "execute" plus an argument (the payload of the transaction).
- Query: the only query implemented is "getTran" which needs a transaction ID and searches for the matching transaction.

Transactions have a new possibility: the user can select a custom algorithm for TxID generation.
- if no custom algorithm is given, more precisely if the "txIDgenerationAlg" field of ChaincodeInvocationSpec is left empty, the previously used UUID is stored as an ID
- if using CLI client "peer", one has to add "-t 'somealgorithm'" to the command where somealgorithm can be ''sha256base64"
- if using gRPC, one has to include a 'txIDgenerationAlg' string in the ChaincodeInvocationSpec (the related proto file was modified and the generated parts of source were regenerated)

A TxID generator algorithm:
- has two parts, a hash function to generate the ID and a decoding to decode user given data (arg[0] of the function that was invoked)
- hashing is done using the first argument of the invoked function provided by the user
## Motivation and Context

This modification would make the framework even more generic while not breaking any code based on previous versions of fabric:
- The user is able to create a transaction and generate an ID for it: no need to get the ID from fabric. It is important to emphasize that before this a user who created a transaction (e.g. calling invoke via gRPC) was unable to get its ID and unable to check its status etc. The modification fixes this.
## How Has This Been Tested?

Unit tests were run.
## Checklist:

Signed-off-by: Gabor Hosszu gabor@digitalasset.com


christo4ferris:
I restarted the Travis build... if it fails again, more investigation is needed.

@gabre thanks for submitting this PR. It looks good on the surface, but needs some tests, please.

core/system_chaincode/noop/chaincode.go
```diff
@@ -28,12 +28,6 @@ import (
 
 var logger = shim.NewLogger("noop")
 
-// SystemChaincode UTXO example chaincode contains a single invocation function named execute. This function accepts BASE64
-// encoded transactions from the Bitcoin network. This chaincode will parse the transactions and pass the transaction
-// components to the Bitcoin libconsensus C library for script verification. A table of UTXOs is maintained to ensure
-// each transaction is valid.
-// Documentation can be found at
-// https://github.com/hyperledger/fabric/blob/master/examples/chaincode/go/utxo/README.md
```
muralisrini: @gabre there are other references such as "bitcoin" (e.,g line 50) that should be removed as well.

core/system_chaincode/noop/chaincode.go
```diff
@@ -0,0 +1,97 @@
...
+                        var dataInByteForm, b64err = base64.StdEncoding.DecodeString(data)
+                        if b64err != nil {
+                                return nil, fmt.Errorf("Error in decoding from Base64:  %s", b64err)
+                        }
+                        return dataInByteForm, nil
+                default:
+                        return nil, errors.New("Unsupported operation")
+	}
+
+}
```
muralisrini, line 105: @gabre if TXID is supplied (optionally) via "-t", querying it from Args is incorrect ? I'm not sure we need this query even ?

core/devops.go
```diff
@@ -171,7 +171,14 @@ func (d *Devops) invokeOrQuery(ctx context.Context, chaincodeInvocationSpec *pb.
 	}
 
 	// Now create the Transactions message and send to Peer.
-	uuid := util.GenerateUUID()
+        var userGivenID = chaincodeInvocationSpec.UserGivenID
+        var uuid string
+        if userGivenID != "" {
+                uuid = userGivenID
```
muralisrini: @gabre we would want to check userGivenID's uniqueness right about here I think (under a lock to be safe probably)

protos/chaincode.proto
```diff
@@ -94,7 +94,7 @@ message ChaincodeInvocationSpec {
 
     ChaincodeSpec chaincodeSpec = 1;
     //ChaincodeInput message = 2;
-
+    string userGivenID = 2;
```
muralisrini: seems this field is a good place for a comment explaining rationale for this work. Also perhaps change ID to TID or TransID (perhaps shorten to givenTransID or useTransID or ... )?

core/devops.go
```diff
@@ -171,7 +171,14 @@ func (d *Devops) invokeOrQuery(ctx context.Context, chaincodeInvocationSpec *pb.
 	}
 
 	// Now create the Transactions message and send to Peer.
-	uuid := util.GenerateUUID()
+        var userGivenID = chaincodeInvocationSpec.UserGivenID
```
corecode: Seems you didn't run gofmt and got indent by space instead of tab?

christo4ferris:
build failed again... needs investigation.

--- FAIL: TestSieveNoDecision (4.00s)
    obc-sieve_test.go:134: replica 0 executed 0 requests, expected 2
    obc-sieve_test.go:134: replica 1 executed 0 requests, expected 2
    obc-sieve_test.go:134: replica 2 executed 0 requests, expected 2
    obc-sieve_test.go:134: replica 3 executed 0 requests, expected 2

corecode:
Test timing races - it'll get better soon.

christo4ferris:
Ok, I have restarted again

core/system_chaincode/noop/chaincode.go
gabre, line 105: In this case, a user given TXID (via -t) would be the transaction ID of the query itself (but queries are not stored on the ledger so it is not as useful as TXID of an 'invoke'). With this query, you are able to get the payload for a specific TXID (if you only know the TXID for any purpose).

protos/chaincode.proto
gabre: I would vote for userGivenTxID (I think that is not too long).

core/devops.go
gabre: I will correct that.

core/devops.go
gabre: @muralisrini I think that is not needed.That would imply enormous performance problems (e.g. if one has a big ledger it would need an expensive lookup).

core/system_chaincode/noop/chaincode.go
muralisrini, line 105: @gabre Ok.  I assumed this was remnant of the previous impl. where the TID was stored in Args[0]. But then you are not getting the entire payload... you are only returning Args[0]). Also, this wont work with security on as Payload is encrypted (direct Unmarshal will fail).

core/system_chaincode/noop/chaincode.go
```diff
@@ -0,0 +1,105 @@
...
+	ledgerH ledgerHandler
+}
+
+// Init initailizes the system chaincode
+func (t *SystemChaincode) Init(stub *shim.ChaincodeStub, function string, args []string) ([]byte, error) {
+	var ledger, err = ld.GetLedger()
+	if nil != err {
+		return nil, err
+	}
+	t.ledgerH = ledger
```
srderson, line 48: @muralisrini will this cause an issue? If the peer were to be stopped and started, is Init called again for a system chaincode or only during deployment?

ghost:
Can one of the admins verify this patch?

core/system_chaincode/noop/chaincode.go
```diff
@@ -0,0 +1,105 @@
...
+
+	case "getTran":
+		if len(args) < 1 {
+			return nil, errors.New("getTran operation must include a single argument, the TX hash hex")
+		}
+		logger.Infof("Executing NOOP QUERY")
+		logger.Infof("--> %x", args[0])
+
+		var txHashHex = args[0]
+		var tx, txerr = t.ledgerH.GetTransactionByUUID(txHashHex)
```
tamasblummer, line 86: rename GetTransactionByUUID to GetTransactionByID

tamasblummer, line 86: Better, take first 128bit of content hash to initialize UUID

core/system_chaincode/noop/chaincode.go
srderson, line 48: Confirmed with @muralisrini that this will be an issue in system chaincodes. They behave the same as regular chaincodes in the sense that anything that is not persisted in the state should be considered stateless between invocations.

On deploy `Init` will be called and set the value of `ledgerH`. However, if the peer is stopped and then started again, the protocol says the `Init` will not be called a second time because the chaincode is already deployed. This could result in an error in the `Query` function.

gabre, line 48: For me, it seems that Init is called every time I start the peer. By the way I tried it with only one peer.

gabre, line 48: To be honest, this is just for testing. I wanted to mock the ledger somehow. If I cannot use Init, I can also check if t.ledgerHandler is nil or not. If it is nil, I can get the ledger singleton.

binhn:
@gabre To the extent of my understanding of what you're trying to do with this pull request, may I suggest a different approach: 
1. To look up a transaction using its UUID, let's provide a GRPC API by adding the function to [api.proto](https://github.com/hyperledger/fabric/blob/master/protos/api.proto). Currently there is a REST API, but it doesn't use a GRPC API. You would prefer GRPC than REST in any case.
2. According to the [Client SDK proposal](https://github.com/hyperledger/fabric/wiki/Client-SDK-Specification), the `devops` will go away. And the SDK generates [transaction UUID](https://github.com/hyperledger/fabric/blob/master/sdk/node/src/hlc.ts) when building the transaction object. So you could leverage the same technique to implement this in your Java SDK, and you could also provide any kind of custom UUID generations.

tamasblummer:
@binh we are building an API in the spirit of the predicted SDK we support CLI but also extended grpc interface. Just as described in 2) the transaction ID is generated while building the transaction object on client side. The transaction ID is derived from content of the user transaction. Derivation is identical on client and server side if -tid is used.

tamasblummer:
Split this request into three PRs to better facilitate discussion and decision:
https://github.com/hyperledger/fabric/pull/1719
https://github.com/hyperledger/fabric/pull/1721
https://github.com/hyperledger/fabric/pull/1722

PR #1663, opened 2016-05-31
Periodic viewchange

corecode:
## Description

Implement periodic view changes to actively try to incorporate "lost" replicas back into the network.
## Motivation and Context

Related to #1454 
## How Has This Been Tested?

New tests
## Checklist:

### DIFF
--- a/consensus/obcpbft/pbft-core.go
+++ b/consensus/obcpbft/pbft-core.go
@@ -1240,7 +1240,7 @@ func (instance *pbftCore) updateViewChangeSeqNo() {
 	if instance.viewChangePeriod <= 0 {
 		return
 	}
-	instance.viewChangeSeqNo = instance.seqNo + instance.viewChangePeriod
+	instance.viewChangeSeqNo = instance.seqNo + instance.viewChangePeriod*instance.K
 }
 
 func (instance *pbftCore) startTimerIfOutstandingRequests() {
--- a/consensus/obcpbft/pbft-core.go
+++ b/consensus/obcpbft/pbft-core.go
@@ -131,6 +131,8 @@ type pbftCore struct {
 
 	nullRequestTimer   eventTimer    // timeout triggering a null request
 	nullRequestTimeout time.Duration // duration for this timeout
+	viewChangePeriod   uint64        // period between automatic view changes
+	viewChangeSeqNo    uint64        // next seqNo to perform view change
 
 	missingReqs map[string]bool // for all the assigned, non-checkpointed requests we might be missing during view-change
 
@@ -219,6 +221,7 @@ func newPbftCore(id uint64, config *viper.Viper, consumer innerStack) *pbftCore 
 		panic("Log multiplier must be greater than or equal to 2")
 	}
 	instance.L = instance.logMultiplier * instance.K // log size
+	instance.viewChangePeriod = uint64(config.GetInt("general.viewchangeperiod"))
 
 	instance.byzantine = config.GetBool("general.byzantine")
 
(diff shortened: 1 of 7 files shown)


corecode:
This builds on #1605 and can be merged after or instead of #1605.

corecode:
For now I implemented a dumb version of just cycling after a fixed number of requests from a primary, but we could "optimize" it by cycling after stable checkpoints.  The advantage of cycling after stable checkpoints is that the new primary does not have to repeat any request; the disadvantage is that it may stop the network for considerable time (wait for execution to finish at 2f+1 replicas).

christo4ferris:
restarted Travis build

consensus/obcpbft/pbft-core.go
```diff
@@ -252,6 +255,11 @@ func newPbftCore(id uint64, config *viper.Viper, consumer innerStack) *pbftCore
 	} else {
 		logger.Info("PBFT null requests disabled")
 	}
+	if instance.viewChangePeriod > 0 {
```
jyellick, line 258: I don't think this is a strong enough check. What happens if the sequence number is set to below the checkpoint interval?  In this case, my rough analysis of the code shows that the network would just constantly and forever change view, without ever reaching a new checkpoint.  If we want to go with the 'every n sequence numbers' approach, rather than 'every n checkpoints', I think we need to have better test coverage, making sure we include intervals that are above and below the checkpoint size as well.

Edit: It looks like maybe, during the view change, because the xSet should be non-empty, the seqNo will appropriately be reset, and the network will make progress.  This still isn't obvious though, so I maintain that additional test cases should be added.

consensus/obcpbft/viewchange.go
```diff
@@ -413,6 +413,8 @@ func (instance *pbftCore) processNewView2(nv *NewView) error {
 		instance.persistQSet()
 	}
 
+	instance.updateViewChangeSeqNo()
```
jyellick, line 417: Could you explain the logic here?  In particular, how can you be certain that `instance.seqNo` is consistent across all replicas at this point?

Maybe the `if` statement above isn't as conditional as it seems, and could be clarified instead to be follow the loop as something like `instance.seqNo = max(xSet)`, but it is not immediately obvious to me.

What if a primary is partitioned from the network, advances its own `instance.seqNo` from n to n+3, but the network never receives pre-prepares for n+1, n+2, or n+3, when the view change completes, the highest sequence number in the xSet is n+1.  Now, the old primary will set its next view change mark to be (n+3)+instance.viewChangePeriod, while the rest of the network picks (n+1)+instance.viewChangePeriod.

This may be a more general bug of how we manage instance.seqNo, but since this had only previously been used to assign sequence numbers to pre-prepares for the primary (and the primary was the one computing the xSet) it may not have been a problem that it was potentially out of sync across replicas on view change.

I don't think this is likely to deadlock the network, but it could cause timeout induced view changes, and degrade the network performance.

corecode, line 417: The new-view assigns all sequence numbers that were mentioned in view-change messages.  The primary could have sent a new-view without having received the view-change from the previous primary.  I think we introduced this issue by deleting "trailing" null requests from the new-view message.  We can either re-introduce those null requests (meaning lots of messages on view change), or we could reset `seqNo` to `max(xSet)`.  What do you prefer?

In any case, this is a long standing bug.

consensus/obcpbft/pbft-core.go
corecode, line 258: I am fine converting the period to multiples of the checkpoint interval.

consensus/obcpbft/viewchange.go
jyellick, line 417: The simplest solution seems to me to explicitly set `seqNo` to `max(xSet)`.  The correctness is more clear to me this way, as `seqNo` is used only by the primary to assign new sequence numbers, and the new primary will obviously include its own view change in the new view, I think this solves it somewhat cleanly.

consensus/obcpbft/pbft-core.go
jyellick, line 258: I think this would be superior, though I'm fine if you'd like to do this is a subsequent PR.

PR #1675, opened 2016-06-01
Move event manager out of pbft core

jyellick:
## Description

This changeset removes the instantiation of the event manager from pbft-core and pushes it into the plugin layer.  The tests are correspondingly updated.
## Motivation and Context

This is yet another PR, in a series intended to remove the multi-threading from PBFT.  In the previous changesets, the internal PBFT and plugin threads were merged to utilize only the event manager thread.  However, for testing, sometimes the core PBFT needs to be initialized without a plugin, which lead to an inelegant solution of instantiating an event manager inside of PBFT, then having the plugins override that manager after instantiation.

Ultimately, the state machine should not have references to the event manager, and should communicate new events via return, and this changeset moves in that direction.

As with the preceding changesets, this is not intended to be a final product, but a step in the right direction, there are some still notably hacky things, in particular, sending view changed events is extremely ugly right now, but this will be fixed in a future PR.

Attention to @corecode @tuand27613 @kchristidis 
## How Has This Been Tested?

Tested locally with unit tests.  CI should be sufficient for the remainder.
## Checklist:

### DIFF
--- a/consensus/obcpbft/viewchange.go
+++ b/consensus/obcpbft/viewchange.go
@@ -438,7 +438,7 @@ func (instance *pbftCore) processNewView2(nv *NewView) error {
 
 	logger.Debug("Replica %d done cleaning view change artifacts, calling into consumer", instance.id)
 
-	instance.manager.inject(viewChangedEvent{})
+	instance.sendViewChangedEvent = true
 
 	return nil
 }
--- a/consensus/obcpbft/events.go
+++ b/consensus/obcpbft/events.go
@@ -18,6 +18,9 @@ package obcpbft
 
 import "time"
 
+// Ultimately, the events code should all be converted to use this type to indicate the intent of parameters and returns
+type event interface{}
+
 // eventReceiver is a consumer of events, processEvent will be called serially
 // as events arrive
 type eventReceiver interface {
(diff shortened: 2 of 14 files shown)


tuand27613:
LGTM

consensus/obcpbft/viewchange.go
```diff
@@ -435,7 +435,7 @@ func (instance *pbftCore) processNewView2(nv *NewView) error {
 
 	logger.Debug("Replica %d done cleaning view change artifacts, calling into consumer", instance.id)
 
-	instance.manager.inject(viewChangedEvent{})
+	instance.sendViewChangedEvent = true
```
corecode, line 441: How would this work once refactoring is complete?

jyellick, line 441: Ultimately, I would see the view change path switching to be driven off events.  There would be something like a `viewChangeQuorumEvent` which would be returned from the the `recvViewChange`, which would trigger the `processNewView` path (as would receiving a new view).  Then, the `processNewView` path should return the `viewChangedEvent`.

corecode, line 441: Sounds good.

binhn:
@jyellick could you rebase to resolve the conflicts

PR #1676, opened 2016-06-01
split out security section [ci skip]

joshhus:
## Description

Split out security section from protocol-spec.md into new document.
## Motivation and Context

Non-protocol sections need to be split out from the protocol-spec doc.
## How Has This Been Tested?

Viewed as branch. 
## Checklist:

Signed-off-by: joshhus joshh@us.ibm.com


binhn:
@joshhus why would security section be a separate document? As it is written, this section is part of the protocol as it contributes to the wiring of the message structures. I think it would be better to rewrite the protocol-spec and stay close with the code to avoid stale, and extract useful info out into tech docs.

joshhus:
Okay I thought there was agreement that the protocol spec content went beyond a typical protocol spec, and that it would be helpful / usable / consumable to split up the protocol spec document into smaller pieces / documents. The intent here was to start moving in that direction. I guess the agreement on this does not exist; perhaps a topic for future HL call.

docs/security-spec.md
```diff
@@ -0,0 +1,1080 @@
...
+
+2. The client-application can request the infrastructure sitting on the client-side to encrypt/decrypt information using a specific public encryption key, or that client's long-term decryption key.
+
+3. The transaction format offers the ability to the application to store additional transaction metadata, that can be passed to the client-application after the latter's request. Transaction metadata, as opposed to code-metadata, is not encrypted or provided to the chaincode at execution time. Validators treat these metadata as a list of bytes they are not responsible for checking validity of.
+
+**Application read-access control.**
+For this reason the application may request and obtain access to the public encryption key of the us
```
christo4ferris, line 844: we need to make clear that we don't currently have a wallet service.

docs/security-spec.md
```diff
@@ -0,0 +1,1080 @@
+﻿# Security Specification
+
+#### 4. Security
+   - 4. Security
```
christo4ferris, line 4: can we make the TOC link to the respective sections, please? Also, why are we starting a new document with a section numbered 4?

joshhus, line 4: See the above discussion. Team needs to decide if the protocol spec is good as-is, or if it should be split into multiple documents, such as Security. 4. ... (This was the start of doing that, as an iterative process, but now there are objections ... ) The TOC linking will be resolved by the new Sphinx/Python/ReadtheDocs build, which is coming soon and being worked in another issue.

christo4ferris:
@joshhus going to close this so that we can take the discussion to Gerrit. Thanks

docs/source/write_first_app.rst
```diff
@@ -43,13 +53,14 @@ We’ll go through three principle steps:
   .. image:: images/AppConceptsOverview.png
 
   **2. Explore a sample smart contract.**
-  We’ll inspect the sample Fabcar smart contract to learn about the transactions within them,
+  We’ll inspect the sample assetTransfer (javascript) smart contract to learn about the transactions within them,
```
lehors: within them -> within it (this refers to the smart contract, right?)

docs/source/write_first_app.rst
```diff
@@ -43,13 +53,14 @@ We’ll go through three principle steps:
...
   and how they are used by applications to query and update the ledger.
```
lehors: applications -> an application
I think it's better to stick to singular because it matches the title of the tutorial.

docs/source/write_first_app.rst
```diff
@@ -43,13 +53,14 @@ We’ll go through three principle steps:
...
 
   **3. Interact with the smart contract with a sample application.** Our application will
-  use the FabCar smart contract to query and update car assets on the ledger.
-  We'll get into the code of the apps and the transactions they create,
-  including querying a car, querying a range of cars, and creating a new car.
+  use the assetTransfer smart contract to create, query, and update assets on the ledger.
+  We'll get into the code of the apps and the transactions they create, including initializing
```
lehors: apps -> app

docs/source/write_first_app.rst
```diff
@@ -90,55 +101,68 @@ bringing up a new one.
...
+
+Launch the Fabric test network using the ``network.sh`` shell script.
+
+.. code:: bash
+
+  ./network.sh up createChannel -c mychannel -ca
+
+This command will deploy the Fabric test network with two peers, three certificate authorities (Orderer, Org1, Org2),
+and an ordering service. Instead of using the cryptogen tool, we bring up the test network using Certificate Authorities, 
+hence the ``-ca`` flag. Additionally, the org admin user registration was bootstrapped when the Certificate Authority was started.
```
lehors: was -> is  twice

docs/source/write_first_app.rst
```diff
@@ -90,55 +101,68 @@ bringing up a new one.
...
-Launch your network using the ``startFabric.sh`` shell script.
+Next, let's deploy the chaincode by calling the ``./network.sh`` script with the chaincode name and language options.
 
 .. code:: bash
 
-  ./startFabric.sh javascript
+  ./network.sh deployCC -ccn basic -ccl javascript
+
+.. note:: Behind the scenes, this script uses the chaincode lifecycle to package, install,
+          query installed, approve for my org, and commit chaincode.
```
sijocherian:
```suggestion
          query installed chaincode, approve chaincode for both Org1 & Org2, and finally commit the chaincode.
```

docs/source/write_first_app.rst
```diff
@@ -164,329 +188,673 @@ You should see the following:
...
+            // 1. register & enroll admin user with CA, stores admin identity in local wallet
+            await enrollAdmin.EnrollAdminUser();
 
-Here is a simplified representation of how a query works:
+This command stores the CA administrator's credentials in the ``wallet`` directory.
+You can find administrator's certificate and private key in the ``wallet/admin.id``
+file.
 
-.. image:: tutorial/write_first_app.diagram.1.png
+.. note:: If you decide to start over by taking down the network and bringing it back up again, you will
```
sijocherian, line 259: Great ! thanks for adding this note

docs/source/write_first_app.rst
```diff
@@ -164,329 +188,673 @@ You should see the following:
...
+
+Third, how the sample application interacts with the chaincode
+-----------------------------------------------------------
+
+In the prior steps, the application generated the admin and app user credentials and placed them in the wallet.
+If the credentials exist and have the correct permissions attributes associated with them, the sample application user
+will be able to call chaincode functionality from the sample application after getting reference to the channel name and contract name.
+
+You will notice that in the following lines of application code, the application is getting reference
+to the channel on the network and the contract name:
```
sijocherian:
```suggestion
to the Contract using the  contract name and channel name via Gateway:
```

docs/source/write_first_app.rst
```diff
@@ -164,329 +188,673 @@ You should see the following:
...
```
sijocherian, line 320: add a blurb about discovery & asLocalhost

"Our connection configuration specifies only the peer from your own Org. We tell node client sdk to use the service discovery (running on the peer), which fetches other peers that are currently online, metadata like relevant endorsement policies and any static information it would have otherwise needed to communicate with the rest of the node.
 
The asLocalhost set to true, tells it to connect as localhost, since our client is running on same network as the other fabric nodes."

PR #1680, opened 2016-06-02
use a separate lock for grpc send operatons

### LINKED ISSUE #1627: System hangs under stress testing
## Description

I setup a simple stress testing environment, and use jmeter to perform the stress testing.  When I set the number of clients to a relative large number, the system hung and would not accept any request. Even if I stopped jmeter and wait for a long time, the system could still not recover.
### Setup the environment

Start up a single peer node

```
docker run --rm -it --name peer1 -e CORE_LOGGING_LEVEL=debug  -e CORE_VM_ENDPOINT=http://172.17.0.1:25 -e CORE_PEER_ID=vp0 -e CORE_PEER_ADDRESSAUTODETECT=true hyperledger-peer peer node start
```

Deploy a chain code

```
{
  "jsonrpc": "2.0",
  "method": "deploy",
  "params": {
    "type": 1,
    "chaincodeID":{
        "path":"github.com/hyperledger/fabric/examples/chaincode/go/chaincode_example02"
    },
    "ctorMsg": {
        "function":"init",
        "args":["a", "1000", "b", "2000"]
    }
  },
  "id": 1
}
```

Verify the chaincode could work correctly:

```
{
  "jsonrpc": "2.0",
  "method": "query",
  "params": {
      "type": 1,
      "chaincodeID":{
          "name":"7b07c59e9b9405c1aef33493b63b9a766d9bb836989ded1730052de650aa8ce5654274d148ceff96a4e5bd43bca26aba099f55c400e4befdc8b2ee4c0a94e30b",
          "path":"github.com/hyperledger/fabric/examples/chaincode/go/chaincode_example02"
      },
      "ctorMsg": {
         "function":"query",
         "args":["a"]
      }
  },
  "id": 5
}
```
### Setup Jmeter

Use http post to send the query json data as above.  So it's a very simple stress test. 
### Benchmark Result

When the number of Jmeter clients is set to 1000, the throughput of that single peer is relative ok: 2100 TPS, although not that high.
When the number of Jmeter clients is enlarged to 2000, the system will hang.

muralisrini:
Use a separate lock for protecting grpc sends instead of reusing global handler lock.
## Description

Chaincode and shim use locks for two purposes - protect handler state and protecting grpc sends. Using the same lock for both creates a problem when the grpc channel is full and the send blocks on the receiver to drain. Overloading a lock could cause a deadlock across the grpc connection where the receiver in turn is blocked waiting for the lock taken by the blocked sender.

To emphasize, this would happen only when the channel buffers are full - such as when doing 1000's of queries concurrently. In the reported test (and the subsequent recreate) this occurs when sending when 2000+ concurrent queries to a chaincode.
## Motivation and Context

Use of a separate lock for protecting grpc sends allows concurrent requests to be processed without fear of deadlock. 

Fixes #1627
## How Has This Been Tested?

Modified TestExecuteQuery to do upto 10000 concurrent queries. While even 2000+ queries failed without the fix 10000 queries succeeded in every attempt.

A test case was NOT created for the following reasons
- The problem manifests only in high stress situation.  Currently there is no process for adding long running stress unit or behave tests
- Current tests provide code coverage for the changes
## Checklist:

### DIFF
--- a/core/chaincode/handler.go
+++ b/core/chaincode/handler.go
@@ -71,6 +71,8 @@ type nextStateInfo struct {
 // Handler responsbile for management of Peer's side of chaincode stream
 type Handler struct {
 	sync.RWMutex
+	//peer to shim grpc serializer. User only in serialSend
+	serialLock  sync.RWMutex
 	ChatStream  ccintf.ChaincodeStream
 	FSM         *fsm.FSM
 	ChaincodeID *pb.ChaincodeID
@@ -102,8 +104,8 @@ func shortuuid(uuid string) string {
 }
 
 func (handler *Handler) serialSend(msg *pb.ChaincodeMessage) error {
-	handler.Lock()
-	defer handler.Unlock()
+	handler.serialLock.Lock()
+	defer handler.serialLock.Unlock()
 	if err := handler.ChatStream.Send(msg); err != nil {
 		chaincodeLogger.Error(fmt.Sprintf("Error sending %s: %s", msg.Type.String(), err))
 		return fmt.Errorf("Error sending %s: %s", msg.Type.String(), err)
--- a/core/chaincode/handler.go
+++ b/core/chaincode/handler.go
@@ -72,7 +72,7 @@ type nextStateInfo struct {
 type Handler struct {
 	sync.RWMutex
 	//peer to shim grpc serializer. User only in serialSend
-	serialLock  sync.RWMutex
+	serialLock  sync.Mutex
 	ChatStream  ccintf.ChaincodeStream
 	FSM         *fsm.FSM
 	ChaincodeID *pb.ChaincodeID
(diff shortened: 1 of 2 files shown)


core/chaincode/handler.go
```diff
@@ -71,6 +71,8 @@ type nextStateInfo struct {
 // Handler responsbile for management of Peer's side of chaincode stream
 type Handler struct {
 	sync.RWMutex
+	//peer to shim grpc serializer. User only in serialSend
+	serialLock  sync.RWMutex
```
corecode: Why RWMutex and not just Mutex?

muralisrini: @corecode  No particular reason really. Makes sense to change it to mutex. Will do that. Thanks!

muralisrini:
also confirmed by @yingfeng in  a comment in #1627.

BTW, noticed travis failed for the first check in.... not sure why. I did run it through behave and unit-test successfully in my env.

PR #1684, opened 2016-06-02
code coverage for go unit-tests using -cover flag

rameshthoomu:
Updated go unit-test command with `-cover` flag (this provides coverage report to all the unit-tests performing with `go test -cover -timeout=20m $PKGS` command)
## Description

Provides code coverage report to all the fabric unit-tests.
## Motivation and Context

Fixes #
## How Has This Been Tested?

Tested this change in inside vagrant and Travis environments.
## Checklist:

Signed-off-by: thoomu@us.ibm.com

### DIFF
--- a/scripts/goUnitTests.sh
+++ b/scripts/goUnitTests.sh
@@ -16,4 +16,4 @@ trap cleanup 0
 echo "DONE!"
 
 echo "Running tests..."
-go test -timeout=20m $PKGS
+go test -cover -timeout=20m $PKGS


bcbrock:
@rameshthoomu How much does this increase the run time for the test suite? I assume that this is not free? Why do we need to run coverage analysis every time we run the test suite?

ghaskins:
To @bcbrock's point, if there is indeed a non-negligible cost, what we could do is create some kind of mode flag/target that is off by default, on for CI, and can try to be smart about not running the tests twice when coverage analysis is desired.

binhn:
@rameshthoomu I would prefer to have a different script called by `make -unit-test-cover` option

rameshthoomu:
@bcbrock: I didn't observe unit-test time increment in build process with this change. The just added a flag to our go test command and I assume it's free.

christo4ferris:
I just ran a test, with -cover on it is 16s wall time, and 2s system time longer than without. Hardly a concern for now. We can always reverse if we feel it is becoming a problem. I think that before that we would actually move to a more smoke-test followed by review followed by full test pipeline.

bcbrock:
@rameshthoomu @ghaskins I stand corrected - I've done a few quick tests myself and I agree, it does seem to be free for the things I tried. So I withdraw my objection.

This almost can't be true in general however (that -cover is free). If I believe what I read on the Internet, they do coverage by rewriting the source code to add counting statements. I think the reason that we see it as free for Hyperledger fabric may be that the code does not contain computational "kernels" where the counting would be a major factor. If I'm ever feeling bored I may try coding up a matrix multiply in Go and testing it with and without -cover.

ghaskins:
Seems there are no remaining objections, and this would be trivial to revert if it is later found to be problematic.  Therefore, LGTM.

integration/nwo/network.go
```diff
@@ -1023,11 +1026,25 @@ func (n *Network) OrdererGroupRunner() ifrit.Runner {
...
+	cmd.Env = append(cmd.Env, env...)
+	return ginkgomon.New(ginkgomon.Config{
+		AnsiColorCode:     n.nextColor(),
+		Name:              "chaincode",
+		Command:           cmd,
+		StartCheck:        `starting up ...`,
+		StartCheckTimeout: 15 * time.Second,
+	})
+}
+
```
sykesm: I don't believe we need a function but, if we do, why is it scoped to "chaincode."  It's looking for a magic string from _one_ implementation of chaincode.

The one bit that I think we're getting from this is colorized output; everything else is just the `ginkgomon.Config` literal.

integration/nwo/network.go
```diff
@@ -1065,7 +1082,9 @@ func (n *Network) peerCommand(command Command, env ...string) *exec.Cmd {
 	cmd := NewCommand(n.Components.Peer(), command)
 	cmd.Env = append(cmd.Env, env...)
 	if ConnectsToOrderer(command) {
-		cmd.Args = append(cmd.Args, "--tls")
+		if n.TLSEnabled {
+			cmd.Args = append(cmd.Args, "--tls")
+		}
```
sykesm: Why nested checks? When would we want `--cafile` when TLS isn't enabled?

Seems this should be a single compound test with `&&`.

integration/nwo/commands/peer.go
```diff
@@ -19,6 +23,7 @@ func (n NodeStart) SessionName() string {
 func (n NodeStart) Args() []string {
 	return []string{
 		"node", "start",
+		fmt.Sprintf("--peer-chaincodedev=%s", strconv.FormatBool(n.DevMode)),
```
sykesm: I'm pretty sure we can do this without strconv and fmt.

```go
args := []string{"node", "start"}
if n.DevMode {
    args = append(args, "--peer-chaincode-dev")
}
```

integration/devmode/devmode_suite_test.go
```diff
@@ -0,0 +1,45 @@
...
+	Expect(err).NotTo(HaveOccurred())
+})
+
+var _ = SynchronizedAfterSuite(func() {
+}, func() {
+	components.Cleanup()
+})
+
+func BasePort() int {
+	return 39000 + 1000*GinkgoParallelNode()
```
sykesm, line 44: Ports should be in ports.go, iirc.

integration/devmode/devmode_test.go
```diff
@@ -0,0 +1,171 @@
...
+		By("Querying and invoking the chaincode")
+		RunQueryInvokeQuery(network, orderer, org1peer0, channelName, 90)
+	})
+})
+
+func buildPeerEnv(peerChaincodeAddress, chaincodeID string) []string {
+	envs := []string{"CORE_PEER_TLS_ENABLED=false"}
+	envs = append(envs, fmt.Sprintf("CORE_PEER_ADDRESS=%s", peerChaincodeAddress))
+	envs = append(envs, fmt.Sprintf("CORE_CHAINCODE_ID_NAME=%s", chaincodeID))
+	return envs
```
sykesm: Why are we disabling TLS here *and* in `core.yaml`?
Why is the `peer.address` getting set to the chaincode address?

Looking at the call, this does not seem to the the address for the peer but for the chaincode. Am I misreading?

PR #1686, opened 2016-06-02
Multi rootnode support

yacovm:

Original discussion was [here](https://github.com/hyperledger/fabric/pull/1482)
## Description

All peers can now start with an array of root nodes instead of a single one, for example (from core.yaml): 

```
    discovery:

        # The root nodes are used for bootstrapping purposes, and generally
        # supplied through ENV variables
        # It can be either a single host or a comma separated list of hosts.
        rootnode: 172.17.0.2,172.17.0.3
```

This removes the SPOF of having a peer depend on a single root node for joining the cluster.
The original GetRootNode() method is not present anymore and instead is replaced by GetRandomNode() which returns a random root node. 
- In addition, NVP nodes now query a random validating peer out of the root node list instead of the same one each time
- In terms of discovery process, validating peers now connect to all root nodes at startup, and non-validating peers seek a servicing peer to connect to.
## How Has This Been Tested?

- Ran the unit tests (go test -timeout=20m $(go list github.com/hyperledger/fabric/... | grep -v /vendor/ | grep -v /examples/))
- Behave tests ran by Travis
## Checklist:

### DIFF
--- a/core/discovery.go
+++ b/core/discovery.go
@@ -16,9 +16,34 @@ limitations under the License.
 
 package core
 
-import "github.com/spf13/viper"
+import (
+	"math/rand"
+	"strings"
+	"time"
+)
 
-// GetRootNode place holder function for providing a boostrap address for a Validating Peer.
-func GetRootNode() (string, error) {
-	return viper.GetString("peer.discovery.rootnode"), nil
+// StaticDiscovery is an implementation of Discovery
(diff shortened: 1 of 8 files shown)


yacovm:
I wish you could re-open a closed PR

binhn:
Based on commends from #1482 , this is good to go.

PR #1688, opened 2016-06-02
Client-sdk support for TLS connection to the member services

adecaro:
## Description

This PR allows the client-sdk to connect to the member services using TLS.
## Motivation and Context

The communication channel between the client-sdk and the member services must be protected to avoid eavesdropping of confidential information.
## How Has This Been Tested?

The PR has been tested by setting up the fabric in dev mode as described in https://github.com/hyperledger/fabric/blob/master/docs/API/SandboxSetup.md, activating TLS on the 
the member services by setting the following properties in _membersrvc.yaml_ as follows:

```
server:
        tls:
              certfile: "/var/hyperledger/production/.membersrvc/tlsca.cert"
              keyfile: "/var/hyperledger/production/.membersrvc/tlsca.priv"
```

And by setting the following properties in _core.yaml_ as follows:

```
    pki:
        eca:
            paddr: localhost:50051
        tca:
            paddr: localhost:50051
        tlsca:
            paddr: localhost:50051
        tls:
            enabled: true
            rootcert:
                file: "/var/hyperledger/production/.membersrvc/tlsca.cert"
            # The server name use to verify the hostname returned by TLS handshake
            serverhostoverride:
```

The test _chain-tests.js_ shows how to setup TLS. 
## Checklist:

This is a joint effort with: Thorsten Kramp, Thomas Eirich and Elli Androulaki.


ghaskins:
Is this covered by the unit-tests?  If so, is there a way I can run it locally to verify since CI doesn't currently trigger sdk tests?

adecaro:
Hi @ghaskins, please, find in the README the instructions on how to test TLS connection in dev mode.  Please, let me know if they are clear enough.

binhn:
@ghaskins currently unit test requires manual set up per readme until deploy is implemented.

integration/nwo/deploy.go
```diff
@@ -220,13 +220,12 @@ func InstallChaincodeLegacy(n *Network, chaincode Chaincode, peers ...*Peer) {
...
 
 func ApproveChaincodeForMyOrg(n *Network, channel string, orderer *Orderer, chaincode Chaincode, peers ...*Peer) {
-	if chaincode.PackageID == "" {
-		chaincode.SetPackageIDFromPackageFile()
-	}
-
 	// used to ensure we only approve once per org
 	approvedOrgs := map[string]bool{}
 	for _, p := range peers {
+		if chaincode.PackageID == "" && !p.DevMode {
```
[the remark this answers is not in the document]
DereckLuo: fixed.

integration/devmode/devmode_test.go
```diff
@@ -0,0 +1,305 @@
...
+		Expect(err).NotTo(HaveOccurred())
+
+		soloConfig := nwo.BasicSolo()
+		soloConfig.RemoveOrg("Org2")
+		network = nwo.New(soloConfig, testDir, client, StartPort(), components)
+		network.TLSEnabled = false
+		// turn on dev mode for all peers
+		for _, p := range network.Peers {
+			p.DevMode = true
+		}
```
sykesm: There should be only one peer. Why not just change it?

`network.Peer("Org1", "peer0").DevMode = true`

integration/devmode/devmode_test.go
```diff
@@ -0,0 +1,305 @@
...
+		var err error
+		channelName = "testchannel"
+		testDir, err = ioutil.TempDir("", "devmode")
+		Expect(err).NotTo(HaveOccurred())
+
+		client, err = docker.NewClientFromEnv()
+		Expect(err).NotTo(HaveOccurred())
+
+		soloConfig := nwo.BasicSolo()
+		soloConfig.RemoveOrg("Org2")
```
sykesm: Please don't add to the config structure. They are meant to be literals. (There's one other method on config and it really needs to be removed.)

Instead, you can create a network configuration for this package that does exactly what you want and nothing else. This one works for me:

```go
var devNetwork = &nwo.Config{
	Organizations: []*nwo.Organization{{
		Name:   "OrdererOrg",
		MSPID:  "OrdererMSP",
		Domain: "example.com",
		CA:     &nwo.CA{Hostname: "ca"},
	}, {
		Name:          "Org1",
		MSPID:         "Org1MSP",
		Domain:        "org1.example.com",
		EnableNodeOUs: true,
		Users:         1,
		CA:            &nwo.CA{Hostname: "ca"},
	}},
	Consortiums: []*nwo.Consortium{
		{Name: "DevConsortium", Organizations: []string{"Org1"}},
	},
	Consensus: &nwo.Consensus{
		Type:            "solo",
		BootstrapMethod: "file",
	},
	SystemChannel: &nwo.SystemChannel{
		Name:    "systemchannel",
		Profile: "OrderingProfile",
	},
	Channels: []*nwo.Channel{
		{Name: "devchannel", Profile: "DevProfile"},
	},
	Profiles: []*nwo.Profile{
		{Name: "OrderingProfile", Orderers: []string{"orderer"}},
		{Name: "DevProfile", Consortium: "DevConsortium", Organizations: []string{"Org1"}},
	},
	Orderers: []*nwo.Orderer{
		{Name: "orderer", Organization: "OrdererOrg"},
	},
	Peers: []*nwo.Peer{{
		Name:         "peer0",
		Organization: "Org1",
		Channels:     []*nwo.PeerChannel{{Name: "devchannel"}},
	}},
}
```

integration/devmode/devmode_test.go
```diff
@@ -0,0 +1,305 @@
...
+		network.CreateAndJoinChannel(orderer, channelName)
+
+		By("building chaincode")
+		chaincodeExecutePath := components.Build(legacyChaincode.Path)
+
+		By("running the chaincode")
+		peerChaincodeAddress := network.PeerAddress(org1peer0, nwo.ChaincodePort)
+		envs := []string{
+			"CORE_PEER_TLS_ENABLED=false",
+			fmt.Sprintf("CORE_CHAINCODE_ID_NAME=%s", legacyChaincodeID),
```
sykesm: Appending is a good choice here: `"CORE_CHAINCODE_ID_NAME=" + legacyChaincodeID`

integration/devmode/devmode_test.go
```diff
@@ -0,0 +1,305 @@
...
+		cmd.Env = append(cmd.Env, envs...)
+		chaincodeRunner = ginkgomon.New(ginkgomon.Config{
+			Name:              "chaincode",
+			Command:           cmd,
+			StartCheckTimeout: 15 * time.Second,
+		})
+		chaincodeProcess = ifrit.Invoke(chaincodeRunner)
+		Eventually(chaincodeProcess.Ready(), network.EventuallyTimeout).Should(BeClosed())
+
+		docContainers, err := client.ListContainers(docker.ListContainersOptions{All: false})
```
sykesm: nit: 'doc' implies documentation, not docker.

non-nit: Why are we playing with Docker at all? If this is dev mode, the chaincode should be started and registered by the developer. If it goes down, the chaincode should not be started by the peer - which appears to be what is being tested.

At the end of the day, the behavior from the new chaincode is what I'd expect here: the peer *does not* start (non-system) chaincode when we're in dev mode. That means all of the docker related enumerations and validation can be removed.

integration/devmode/devmode_test.go
```diff
@@ -0,0 +1,305 @@
...
+		envs := []string{
+			"CORE_PEER_TLS_ENABLED=false",
+			fmt.Sprintf("CORE_CHAINCODE_ID_NAME=%s", legacyChaincodeID),
+		}
+		cmd := exec.Command(chaincodeExecutePath, []string{"-peer.address", peerChaincodeAddress}...)
+		cmd.Env = append(cmd.Env, envs...)
+		chaincodeRunner = ginkgomon.New(ginkgomon.Config{
+			Name:              "chaincode",
+			Command:           cmd,
+			StartCheckTimeout: 15 * time.Second,
```
sykesm, line 112: If there's no start check so the timeout doesn't matter and the `Eventually(chaincodeProcess.Ready())` assertions are useless. When the start check is not specified, it becomes immediately satisfied.

integration/devmode/devmode_test.go
```diff
@@ -0,0 +1,305 @@
...
+		})
+		chaincodeProcess = ifrit.Invoke(chaincodeRunner)
+		Eventually(chaincodeProcess.Ready(), network.EventuallyTimeout).Should(BeClosed())
+
+		docContainers, err := client.ListContainers(docker.ListContainersOptions{All: false})
+		Expect(err).NotTo(HaveOccurred())
+		Expect(docContainers).Should(HaveLen(0))
```
sykesm: Expect/To Eventually/Should (here and elsewhere)

integration/devmode/devmode_test.go
```diff
@@ -0,0 +1,305 @@
...
+
+		docContainers, err = client.ListContainers(docker.ListContainersOptions{All: false})
+		Expect(docContainers).Should(HaveLen(0))
+
+		By("querying and invoking the chaincode")
+		RunQueryInvokeQuery(network, orderer, org1peer0, channelName, 90)
+
+		docContainers, err = client.ListContainers(docker.ListContainersOptions{All: false})
+		Expect(docContainers).Should(HaveLen(1))
+		Expect(docContainers[0].Names[0]).To(ContainSubstring("Org1.peer0-mycc-0.0"))
```
sykesm: Please see above: I believe this is asserting incorrect behavior that we do not want to preserve in v2

integration/devmode/devmode_test.go
```diff
@@ -0,0 +1,305 @@
...
+
+		org1peer0 := network.Peer("Org1", "peer0")
+		peers := network.PeersWithChannel(channelName)
+		orderer := network.Orderer("orderer")
+
+		By("setting up the channel")
+		network.CreateAndJoinChannel(orderer, channelName)
+
+		By("enabling V2_0 application capabilities")
+		nwo.EnableCapabilities(network, "testchannel", "Application", "V2_0", orderer, org1peer0)
```
sykesm: You're using `channelName` everywhere else. I had to change this when I used the hand-crafted network.

integration/devmode/devmode_test.go
```diff
@@ -0,0 +1,305 @@
...
+			Name:    "mycc",
+			Version: "0.0",
+			Path:    "github.com/hyperledger/fabric/integration/chaincode/simple/cmd",
+			Ctor:    `{"Args":["init","a","100","b","200"]}`,
+			Policy:  `OR ('Org1MSP.member')`,
+		}
+		legacyChaincodeID := fmt.Sprintf("%s:%s", legacyChaincode.Name, legacyChaincode.Version)
+
+		org1peer0 := network.Peer("Org1", "peer0")
+		peers := network.PeersWithChannel(channelName)
```
sykesm: This confused me since we have exactly one peer. Again, with one peer, I'd expect to use the explicit peer instance that we're testing.

Later on this is used in the variadic list for no effect as we already explicitly use org1peer0.

integration/devmode/devmode_test.go
```diff
@@ -0,0 +1,305 @@
...
+		By("building chaincode")
+		chaincodeExecutePath := components.Build(legacyChaincode.Path)
+
+		By("running the chaincode")
+		peerChaincodeAddress := network.PeerAddress(org1peer0, nwo.ChaincodePort)
+		envs := []string{
+			"CORE_PEER_TLS_ENABLED=false",
+			fmt.Sprintf("CORE_CHAINCODE_ID_NAME=%s", legacyChaincodeID),
+		}
+		cmd := exec.Command(chaincodeExecutePath, []string{"-peer.address", peerChaincodeAddress}...)
```
sykesm: Why the complicated slice and expansion instead of just providing the two arguments?

integration/devmode/devmode_test.go
```diff
@@ -0,0 +1,305 @@
...
+			PackageFile:     filepath.Join(testDir, "simplecc.tar.gz"),
+			Ctor:            `{"Args":["init","a","100","b","200"]}`,
+			SignaturePolicy: `OR ('Org1MSP.member')`,
+			Sequence:        "1",
+			InitRequired:    true,
+			Label:           "my_prebuilt_chaincode",
+		}
+
+		org1peer0 := network.Peer("Org1", "peer0")
+		peers := network.PeersWithChannel(channelName)
```
sykesm: Same as the other test: We have one peer so I don't understand why we need this list and the expansions later on.

integration/devmode/devmode_test.go
```diff
@@ -0,0 +1,305 @@
...
+		By("enabling V2_0 application capabilities")
+		nwo.EnableCapabilities(network, "testchannel", "Application", "V2_0", orderer, org1peer0)
+
+		By("running the chaincode")
+		chaincodeID := fmt.Sprintf("%s:%s", chaincode.Name, chaincode.Version)
+		peerChaincodeAddress := network.PeerAddress(org1peer0, nwo.ChaincodePort)
+		envs := []string{
+			"CORE_PEER_TLS_ENABLED=false",
+			fmt.Sprintf("CORE_PEER_ADDRESS=%s", peerChaincodeAddress),
+			fmt.Sprintf("CORE_CHAINCODE_ID_NAME=%s", chaincodeID),
```
sykesm: No need to use `Sprintf` and when we're just concatenating strings.

integration/devmode/devmode_test.go
```diff
@@ -0,0 +1,305 @@
...
+		orderer := network.Orderer("orderer")
+
+		By("setting up the channel")
+		network.CreateAndJoinChannel(orderer, channelName)
+
+		By("enabling V2_0 application capabilities")
+		nwo.EnableCapabilities(network, "testchannel", "Application", "V2_0", orderer, org1peer0)
+
+		By("running the chaincode")
+		chaincodeID := fmt.Sprintf("%s:%s", chaincode.Name, chaincode.Version)
```
sykesm: No need to use `Sprintf` when we're concatenating three strings.

`chaincode.Name + ":" + chaincodeVersion`

integration/devmode/devmode_test.go
```diff
@@ -0,0 +1,305 @@
...
+
+	It("executes chaincode in dev mode using legacy lifecycle", func() {
+		legacyChaincode = nwo.Chaincode{
+			Name:    "mycc",
+			Version: "0.0",
+			Path:    "github.com/hyperledger/fabric/integration/chaincode/simple/cmd",
+			Ctor:    `{"Args":["init","a","100","b","200"]}`,
+			Policy:  `OR ('Org1MSP.member')`,
+		}
+		legacyChaincodeID := fmt.Sprintf("%s:%s", legacyChaincode.Name, legacyChaincode.Version)
```
sykesm: No need to use `Sprintf` to concatenate three strings.

`legacyChaincode.Name + ":" + legacyChaincode.Version` ?

Also, it's way up here but not used until the environment is built.

integration/devmode/devmode_test.go
```diff
@@ -0,0 +1,305 @@
...
+		By("running the chaincode")
+		chaincodeID := fmt.Sprintf("%s:%s", chaincode.Name, chaincode.Version)
+		peerChaincodeAddress := network.PeerAddress(org1peer0, nwo.ChaincodePort)
+		envs := []string{
+			"CORE_PEER_TLS_ENABLED=false",
+			fmt.Sprintf("CORE_PEER_ADDRESS=%s", peerChaincodeAddress),
+			fmt.Sprintf("CORE_CHAINCODE_ID_NAME=%s", chaincodeID),
+		}
+
+		cmd := exec.Command(chaincode.Path, []string{"-peer.address", peerChaincodeAddress}...)
```
sykesm: Why the complicated slice + expansion instead of just providing the two arguments?

integration/devmode/devmode_test.go
```diff
@@ -0,0 +1,305 @@
...
+		cmd.Env = append(cmd.Env, envs...)
+		chaincodeRunner = ginkgomon.New(ginkgomon.Config{
+			Name:              "chaincode",
+			Command:           cmd,
+			StartCheckTimeout: 15 * time.Second,
```
sykesm, line 167: Same comment about start check. Timeout is irrelevant if you don't have a check.

integration/nwo/config.go
```diff
@@ -29,3 +29,41 @@ func (c *Config) RemovePeer(orgName, peerName string) {
...
+	for _, pro := range c.Profiles {
+		orgs := []string{}
+		for _, o := range pro.Organizations {
+			if o != orgName {
+				orgs = append(orgs, o)
+			}
+		}
+		pro.Organizations = orgs
+	}
+}
```
sykesm: Please remove this in favor of defining a fit-for-purpose network.

integration/devmode/devmode_test.go
wlahti: We added this because we couldn't think of another way to truly prove that we're using our devmode chaincode process and not just a docker container started by the peer. With that said, this exact check is at a point that doesn't make sense. If anything, it should be after the instantiation call, since i'm pretty sure the init baked into instantiate is when the container is started with legacy chaincodes. 

In general though, we'll defer to your judgment on whether to include any docker checks or not.

sykesm: Issue a message when the chaincode starts and on every request received when a flag (not normally provided by the peer) or an environment variable (not normally provided by the peer) is set. Then make assertions on the output of the chaincode.

No docker. Low complexity. Easy to implement and addresses the readiness check comments.

wlahti: Ah right, similar to what we did for the _lifecycle test to prove we actually stopped chaincodes when they were no longer referenced. Sounds good. We'll handle this particular comment, @DereckLuo.

integration/devmode/devmode_test.go
DereckLuo: fixed.

integration/devmode/devmode_test.go
DereckLuo: fixed.

integration/devmode/devmode_test.go
DereckLuo: fixed

integration/devmode/devmode_test.go
DereckLuo: fixed. and else where

integration/devmode/devmode_test.go
DereckLuo: fixed

integration/devmode/devmode_test.go
DereckLuo: fixed

integration/devmode/devmode_test.go
DereckLuo: fixed

integration/devmode/devmode_test.go
DereckLuo: fixed

integration/devmode/devmode_test.go
DereckLuo: fixed

integration/devmode/devmode_test.go
DereckLuo: fixed

integration/devmode/devmode_test.go
DereckLuo: fixed

integration/devmode/devmode_test.go
DereckLuo: fixed

integration/nwo/config.go
DereckLuo: fixed.

integration/nwo/standard_networks.go
```diff
@@ -219,3 +219,59 @@ func MultiNodeEtcdRaft() *Config {
...
+			Organization: "Org1",
+			Channels: []*PeerChannel{
+				{Name: "testchannel", Anchor: true},
+			},
+		}},
+		Profiles: []*Profile{{
+			Name:     "TwoOrgsOrdererGenesis",
+			Orderers: []string{"orderer"},
+		}, {
+			Name:          "TwoOrgsChannel",
```
sykesm: No longer TwoOrgs but one.

integration/devmode/devmode_test.go
```diff
@@ -0,0 +1,340 @@
...
+
+		By("installing the chaincode")
+		nwo.InstallChaincodeLegacy(network, legacyChaincode, org1peer0)
+
+		By("instantiating the chaincode")
+		nwo.InstantiateChaincodeLegacy(network, channelName, orderer, legacyChaincode, org1peer0, org1peer0)
+
+		By("querying and invoking the chaincode")
+		RunQueryInvokeQuery(network, orderer, org1peer0, channelName, 100)
+		Eventually(chaincodeRunner.Buffer()).Should(gbytes.Say("invoking in devmode"))
```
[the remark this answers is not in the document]
wlahti: Done

PR #1690, opened 2016-06-02
Bump null request timeout to fix sporadic CI failures

jyellick:
## Description

This is a very simple change which increases the timeout for the `TestNetworkNullRequestMissing` test which has been periodically failing via CI.
## Motivation and Context

There periodic CI failures appear to be simply because the test has not executed quickly enough.  There is no harm (other than lengthening the test runs) of waiting longer for this, so this simple fix adds 50% more time (1 second) to the wait.

Ultimately, as we transition the PBFT tests to use deterministic timers, these sleeps will be removed, but in an effort to help CI now, this fix is needed.
## Checklist:

### DIFF
--- a/consensus/obcpbft/obc-sieve.go
+++ b/consensus/obcpbft/obc-sieve.go
@@ -352,7 +352,7 @@ func (op *obcSieve) processRequest() {
 
 func (op *obcSieve) recvExecute(exec *Execute) {
 	if !(exec.View >= op.epoch && exec.BlockNumber > op.blockNumber && op.pbft.primary(exec.View) == exec.ReplicaId) {
-		logger.Debug("Invalid execute from %d", exec.ReplicaId)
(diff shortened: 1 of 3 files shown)


ghaskins:
LGTM, thanks.  I'll be prudent and wait for CI confirmation that nothing else broke, but should be able to merge this right away.

Thanks!

jyellick:
The fix above was for the `TestNetworkNullRequestMissing` test, but the failure observed in CI this time was the `TestSieveNoDecision` test.  It's not entirely obvious what went wrong from the log, but, under the theory that it's just decreased Travis performance, I've tuned the timeouts in that test up by 50% as well, and enhanced the logging statement that looked suspicious to be a little more expressive so if it fails again, hopefully we'll get more info.

ghaskins:
Thank you for the rapid response @jyellick!

LGTM

bramwelt:
reverify

bramwelt:
(testing jenkins)

PR #1705, opened 2016-06-03
Write only successful TXs to the ledger

gabre:
This PR changes the system so it adds only successful
TXs to the ledger. The TXs where chaincode execution was erroneous
are not persisted.
## Description

See the summary.
## Motivation and Context

Writing erroneous TXs (where the chaincode execution resulted
in an exception etc.) to the ledger makes it possible to flood the
ledger easily. On the other hand, Bitcoin follows the same approach
as this commit, it does not store unsuccessful transactions.
## How Has This Been Tested?

Unit tests were run.
## Checklist:

Signed-off-by: Gabor Hosszu gabor@digitalasset.com


tamasblummer:
This PR is much more than it says in the comment.

gabre:
I created a new PR containing only the last commit that adds the specified functionality.

PR #1708, opened 2016-06-03
Make enhancements

ghaskins:
Roll up of numerous improvements to the build system
## Description

This patch series includes numerous operational and performance improvements to the build system, including:
- cc9d82d Use repo clone as opposed to raw copy in travis
- b45b481 Make WriteGopathSrc() more precise
- d027736 Use new build-area for certain artifacts of our build
- 9eb0211 Rename our peer/ca images to hyperledger/fabric-xxx
- e04afb7 Rename the "ca*" target in the Makefile to membersrvc
- 10056ef Build peer and membersvcs docker images natively
- f338921 Add a src-image so that we can derive peer/membersrvc and chaincodes from it
- 00afef1 Add hyperledger/fabric-ccenv image
- 47fbc6b Optimize peer/membersrvc -image generation

The result is that the system now employs greater intelligence for automatically detecting the minimum set of work in order to produce a valid build for various targets such as application binaries and docker images.  This means unnecessary recompilation is avoided whenever possible, resulting in faster overall builds, particularly for subsequent rebuilds typically found in iterative development patterns.

Prior to this patch, any rebuild of peer regardless of the size of the change would consume 32s (on my 2009 MacPro).  After this patch, subsequent builds can be as fast as 3s, though this will grow depending on how many files need to be recompiled.  A typical cycle seems to be about 9s, but even this is over 70% reduction. 

I am pleased to announce that the unit-test time has also been improved due to the necessary work to improve the way we package up chaincode applications (see b45b481).  Prior to this patch, the github.com/hyperledger/fabric/core/chaincode test would often require over 400s to complete, and it now finishes in about 220s.  Overall, the TravisCI cycle dropped from about 44 minutes to 38 minutes, buying us a little more headroom against the 50 min limit of the free TravisCI environment.

This patch also has the side effect of greatly optimizing the size of the deploy transaction, since we are now much more selective in the contents of the deploy package.  This will likely solve or at least improve some of the peripheral issues we have faced surrounding this, such as very large deployment transactions and peer OOM issues as it tried to create a tarball from inadvertent artifacts in the tree.

Perhaps more important than speed, we also improved the correctness of the system.  We now build all resources that are targetted to run within docker using docker, which means we are ensuring we are generating object code that is actually compatible with the targetted runtime environment, rather than just banking on the fact that vagrant happens to be ABI compatible.  This compatibility was never ensured, particularly as some environments do not use vagrant.
## How Has This Been Tested?

"make checks" has passed locally, TravisCI has been confirmed to work as well.

I also performed numerous testing against the build system to look for things like:
- inclusive correctness: does the system build the necessary pre-requisites when asked
- exclusive correctness: does the build system elide unnecessary steps from recompilation when the option is available, and does this make the process faster and/or experience better
- usability: does the system seem to make logical sense to use, are the outputs clear
- etc
## Checklist:


ghaskins:
@bcbrock @corecode @muralisrini @rameshthoomu

bcbrock:
@ghaskins - If you think it's good that's OK with me. I have 2 comments:
1. You've changed a lot of names and locations. What would be the brief operational summary you would send to the mailing list to describe to people how their daily life has changed? (I know you documented it all.)
2. I see in devnet-setup.md some words about "To create a Docker image for the `hyperledger/fabric`, first clean out any active containers...". Is there a make target that does all this automatically?

ghaskins:
screen shot 2016-06-03 at 1 43 27 pm

tools/busywork/bin/Makefile
```diff
@@ -156,12 +156,10 @@ images: docker_fabric_clean
 	$(MAKE) membersrvc_image
 
 membersrvc_image:
-	-docker rmi -f hyperledger/fabric-membersrvc
-	go test github.com/hyperledger/fabric/core/container -run=BuildImage_Obcca
+	cd $(FABRIC); $(MAKE) ca-image-clean ca-image
```
ghaskins, line 164: Need to follow up and make sure I updated these to s/ca/membersrvc

tools/busywork/bin/Makefile
```diff
@@ -156,12 +156,10 @@ images: docker_fabric_clean
 	$(MAKE) membersrvc_image
 
 membersrvc_image:
-	-docker rmi -f membersrvc
-	go test github.com/hyperledger/fabric/core/container -run=BuildImage_Obcca
+	cd $(FABRIC); $(MAKE) ca-image-clean ca-image
```
ghaskins, line 164: Looks like I neglected to rename s/ca/membersrvc.  I will fix this.

tools/busywork/tcl/fabric.tcl
```diff
@@ -357,7 +357,7 @@ proc ::fabric::caLogin {i_peer i_user i_secret} {
     http::cleanup $token
 }
 
-
```
ghaskins, line 360: @bcbrock I apologize for all the whitespace churn here, I didn't notice it happened in my preliminary reviews.  I think emacs got a little overzealous on me.  If this was not the right thing to do, I can revert the patch and then judiciously reapply the meaningful hunks.

ghaskins:
@bcbrock both good points:

Regarding (1), I'll put something together and post it here.

Regarding (2), there is "make images-clean" but it's probably not fully prescriptive with the intent of that section of the doc, as Make is only managing the hyperledger/fabric-\* images.  There are still all those "jdoe" and "vpx" type images that seem to creep in over time.  I'll review the doc in this context and see if there are any opportunities to re-tool the "make *clean" or otherwise clarify the doc.

tools/busywork/tcl/fabric.tcl
bcbrock, line 360: @ghaskins I'm not worried about the white space. Is there an Emacs setting that does this automatically (whitespace cleanup)? I wasn't aware of one.

ghaskins, line 360: I'm not aware of one either, but I recently started using [Emacs Prelude](http://batsov.com/prelude/) and it loads some funky defaults for certain things.  It's quite plausible the default configuration for Tcl mode has something like that turned on.

bcbrock, line 360: Prelude does whitespace-cleanup by default 
http://batsov.com/prelude/ 
Look for "whitespace-mode", tells how to disable it if you want to
I suppose being "whitespace clean" could also be another requirement

ghaskins:
@bcbrock 

Regarding operational summary

"As of commit $commit id, artifacts such as the peer and membersrvc binaries are now written to a common subdirectory ./build/bin instead of within the source tree (e.g. ./build/bin/peer instead of ./peer/peer).  If you are using a recent vagrant, ./build/bin should already be in your $PATH and therefore, you can simply type "peer" rather than a fully-qualified or relative path.  If you are _NOT_ using vagrant, you may choose to either a) add ./build/bin to your $PATH manually, or b) run the tools using the new path (e.g. "./build/bin/peer node start")"

binhn:
LGTM -- I pulled and did a `make all`. Below is the status. The failures seem to do with timeout:

```
Exception: Max time exceeded waiting for transactions with current response map = {'bddtests_vp0_1': 200, 'bddtests_vp1_1': 404, 'bddtests_vp2_1': 0}
```

However, I don't think those related to your PR.  Other things that I should investigate before merging?

```
Failing scenarios:
  peer_basic.feature:747  4 peers and 1 membersrvc, consensus still works if 1 peer (vp3) is byzantine -- @1.1 Consensus Options
  peer_basic.feature:748  4 peers and 1 membersrvc, consensus still works if 1 peer (vp3) is byzantine -- @1.2 Consensus Options

0 features passed, 1 failed, 1 skipped
25 scenarios passed, 2 failed, 4 skipped
429 steps passed, 2 failed, 191 skipped, 0 undefined
Took 21m4.411s
make: *** [behave] Error 1
rm build/docker/bin/membersrvc build/docker/bin/peer
```

bcbrock:
@ghaskins The changes in the paths make sense, thanks. What about the new images, anything particular to be aware of with them?

ghaskins:
@binhn If you'd like, you can try a few things like this:

```
$ make clean
$ make peer #should rebuild all prereqs
$ make peer #should do nothing
$ touch peer/main.go
$ time make peer #should do a very small amount of work that includes rebuilding fabric-src, fabric-ccenv and peer
```

ghaskins:
@bcbrock I can't think of any impact to the images themselves.....I guess one thing is that the system more or less assumes you are using the makefile to build now...previously we did not (e.g. you could still do "cd peer; go build" in certain circumstances).

Therefore, perhaps we should also mention that it is now mandatory to use the makefile, or something along those lines.

ghaskins:
@binhn before we merge, we should decide what to do about PR #1720 and also I have to fix the mistake I noted in busywork/bin/Makefile.....my thought on PR 1720 is, @angrbrd's impl is superior to my impl in this PR...I _think_ the errors we are seeing there are similar to errors we _would_ see with mine too, except I have other patches in this PR which mask them.  Namely, I no longer rely on WriteGopathSrc() to generate the peer/ca images since I do them in the makefile now.

I placed my impl early in the series because what I found was that prior to this PR, the chaincode unit-test took about 450 seconds....when I first wrote this PR, it ballooned up to 900s.  I then root caused it to be the image size was being impacted by the ./build/\* directory.  Therefore, I wrote the filtercode, reducing the chaincode time to 200s, and put it early in the series so that we would prevent a bisect from landing on the 900s whopper.

I suspect my patch would be equally broken (if not worse) than #1720 if taken alone.

What I propose is that we
1. Drop my WriteGopathSrc() patch currently in the beginning of this PR
2. Cherry pick @angrbrd's patches from #1720, but add them to the _end_ of this PR
3. Check that everything passes, and merge them both together as one unit.

As opposed to trying to deal with any problems introduced by WriteGopathSrc() filtering in a context which we know will be cleaned up shortly.  The only downside to doing this is that if someone bisects between the patches in this PR, they'll see a slowdown in the UTs but nothing more.

Thoughts?

binhn:
@ghaskins @angrbrd 
I agreed with your proposed steps above to cherry pick #1720 and add to this PR or keep them separate, whichever is easier.

I ran into the timeouts problem on Friday trying to add more unit tests to the chaincode.

Could you and Anya work on that.  I will merge as soon as I have your changes and can run through the make all.

Thanks.

ghaskins:
@bcbrock Thinking about it some more, and perhaps to your point, I should also mention the docker image name changes minimally.

ghaskins:
Replaced by PR #1774

internal/peer/snapshot/snapshot.go
```diff
@@ -0,0 +1,91 @@
...
+var (
+	channelID       string
+	blockNumber     uint64
+	peerAddress     string
+	tlsRootCertFile string
+)
+
+var snapshotCmd = &cobra.Command{
+	Use:   "snapshot",
+	Short: "Manage snapshot requests: submitrequest/cancelrequest/listpending",
```
manish-sethi: `submitrequest|cancelrequest|listpending`, to be consistent with the existing commands.

internal/peer/snapshot/client_connection.go
```diff
@@ -0,0 +1,60 @@
...
+
+	if !viper.GetBool("peer.tls.enabled") {
+		tlsRootCertFile = ""
+		return nil
+	}
+
+	if tlsRootCertFile == "" {
+		return errors.New("the required parameter 'tlsRootCertFile' is empty. Rerun the command with --tlsRootCertFile flag")
+	}
+
```
manish-sethi: Nit: I guess that a `switch-case` on `viper.GetBool("peer.tls.enabled")` will make it more consumable

internal/peer/snapshot/client_connection.go
```diff
@@ -0,0 +1,60 @@
...
+)
+
+// ClientConnection holds client side dependency for the snapshot commands
+type ClientConnection struct {
+	SnapshotClient pb.SnapshotClient
+	Signer         common.Signer
+}
+
+// NewClientConnection creates new client connections based on the input parameters.
+func NewClientConnection(cmdName string, cryptoProvider bccsp.BCCSP) (*ClientConnection, error) {
```
manish-sethi: As this is a generic client for snapshot grpc, irrespective of the actual `cmd`, the parameter `cmdName` does not seem fit. It's is just used to be included in the error message. I suggest that caller of this function should add this info, if desired.

internal/peer/snapshot/client_connection.go
```diff
@@ -0,0 +1,60 @@
...
+import (
+	pb "github.com/hyperledger/fabric-protos-go/peer"
+	"github.com/hyperledger/fabric/bccsp"
+	"github.com/hyperledger/fabric/internal/peer/common"
+	"github.com/pkg/errors"
+	"github.com/spf13/viper"
+)
+
+// ClientConnection holds client side dependency for the snapshot commands
+type ClientConnection struct {
```
manish-sethi: The naming confused me for a minute... This is the client itself.. no? Can we simply call this `Client`?

internal/peer/snapshot/submitrequest_test.go
```diff
@@ -0,0 +1,42 @@
...
+	mockSigner.SignReturns([]byte("snapshot-request-signature"), nil)
+	mockSnapshotClient := &mock.SnapshotClient{}
+	mockSnapshotClient.GenerateReturns(&empty.Empty{}, nil)
+	mockCC := &ClientConnection{mockSnapshotClient, mockSigner}
+
+	resetFlags()
+	cmd := submitRequestCmd(mockCC, nil)
+	cmd.SetArgs([]string{"-C", "mychannel"})
+	require.NoError(t, cmd.Execute())
+	cmd.SetArgs([]string{"-C", "mychannel", "-b", "100"})
```
manish-sethi, line 38: Need to call resetFlags() before this.. else, it would pass even if you don't supply `-C`, because of the flag set in previous line

internal/peer/snapshot/submitrequest_test.go
```diff
@@ -0,0 +1,42 @@
...
+	mockSigner := &mock.Signer{}
+	mockSigner.SignReturns([]byte("snapshot-request-signature"), nil)
+	mockSnapshotClient := &mock.SnapshotClient{}
+	mockSnapshotClient.GenerateReturns(&empty.Empty{}, nil)
+	mockCC := &ClientConnection{mockSnapshotClient, mockSigner}
+
+	resetFlags()
+	cmd := submitRequestCmd(mockCC, nil)
+	cmd.SetArgs([]string{"-C", "mychannel"})
+	require.NoError(t, cmd.Execute())
```
manish-sethi, line 30: I guess that since it's a user facing command, it's better to test the output message as well. In the actual function, instead of hard-coding `fmt.Println`, you can take an `io.writer` and for the test, you can pass a buffer.

internal/peer/snapshot/submitrequest.go
```diff
@@ -0,0 +1,81 @@
...
+		"blockNumber",
+		"peerAddress",
+		"tlsRootCertFile",
+	}
+	attachFlags(snapshotSubmitRequestCmd, flagList)
+
+	return snapshotSubmitRequestCmd
+}
+
+func submitRequest(cmd *cobra.Command, args []string, cc *ClientConnection, cryptoProvider bccsp.BCCSP) error {
```
manish-sethi: `args` -> unused param.. these are already set in the flags

internal/peer/snapshot/client.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/hyperledger/fabric/internal/peer/common"
+	"github.com/pkg/errors"
+	"github.com/spf13/viper"
+)
+
+// Client holds client side dependency for the snapshot commands
+type Client struct {
+	SnapshotClient pb.SnapshotClient
+	Signer         common.Signer
+	Writer         io.Writer
```
manish-sethi, line 24: do they need to be exported?

PR #1713, opened 2016-06-03
Adding chaincode API range query test and skip main

binhn:
## Description

Adding unit test on range query API. Adding unit test to skip the main module.
## Motivation and Context

Providing more unit test coverage
## How Has This Been Tested?

run `go test`
## Checklist:

Signed-off-by: binhn@us.ibm.com

### DIFF
--- a/core/chaincode/exectransaction_test.go
+++ b/core/chaincode/exectransaction_test.go
@@ -1312,6 +1312,77 @@ func TestChaincodeQueryChaincodeWithSec(t *testing.T) {
 	finitPeer(peerLis)
 }
 
+// Test the invocation of a transaction.
+func TestRangeQuery(t *testing.T) {
+	var opts []grpc.ServerOption
+	if viper.GetBool("peer.tls.enabled") {
+		creds, err := credentials.NewServerTLSFromFile(viper.GetString("peer.tls.cert.file"), viper.GetString("peer.tls.key.file"))
+		if err != nil {
+			grpclog.Fatalf("Failed to generate credentials %v", err)
+		}
+		opts = []grpc.ServerOption{grpc.Creds(creds)}
+	}
+	grpcServer := grpc.NewServer(opts...)
+	viper.Set("peer.fileSystemPath", "/var/hyperledger/test/tmpdb")
+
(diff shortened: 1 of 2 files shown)


peer/main_test.go
```diff
@@ -0,0 +1,25 @@
...
+
+package main
+
+import (
+	"testing"
+)
+
+func TestMain(t *testing.T) {
+	t.Skip("No needs to test main since it is covered by other tests")
+}
```
ghaskins, line 25: Hi Binh,

I realize from your PR Description comment that the goal of skipping the tests was intentional, and thus justifies why this file is here.  What's not clear is why this is needed.  I am more curious than anything, but perhaps a comment in this file is a good idea as I suspect I won't be the only one wondering what this accomplishes.

core/chaincode/exectransaction_test.go
```diff
@@ -1312,6 +1312,77 @@ func TestChaincodeQueryChaincodeWithSec(t *testing.T) {
...
+		t.Fail()
+		t.Logf("Error invoking <%s>: %s", chaincodeID, err)
+		GetChain(DefaultChain).Stop(ctxt, &pb.ChaincodeDeploymentSpec{ChaincodeSpec: spec})
+		closeListenerAndSleep(lis)
+		return
+	}
+	GetChain(DefaultChain).Stop(ctxt, &pb.ChaincodeDeploymentSpec{ChaincodeSpec: spec})
+	closeListenerAndSleep(lis)
+}
+
```
muralisrini, line 1385: TestRangeQuery LGTM, thanks @binhn !

binhn:
@ghaskins I'd like to have an indication that we don't need to test this directory/module, so that the coverage report doesn't look like this `?    github.com/hyperledger/fabric/peer  [no test files]`  but more like `ok     github.com/hyperledger/fabric/peer  0.008s  coverage: 0.0% of statements`.

It seems to mean the same thing, but if someone looks into that directory, he would see the test file and the intent. Let me known if you're aware of a better method.

ghaskins:
Thanks Binh,

LGTM,

PR #1717, opened 2016-06-04
Fixed issue #1702 and added tests in crypto_test.go in order to test with different settings

### LINKED ISSUE #1702: Is not possible to test transaction using ConfidentialityProtocolVersion 1.1
## Description

ConfidentialityProtocolVersion is set to 1.2 [here](https://github.com/hyperledger/fabric/blob/master/core/crypto/client_tx.go#L64) and it is not possible to modify this value. This situation produces that all the code developed for ConfidentialityProtocolVersion 1.1 as [this](https://github.com/hyperledger/fabric/blob/master/core/crypto/client_confidentiality.go#L47#L85) could not be tested using a transaction.  
In my opinion if ConfidentialityProtocolVersion 1.1 is not required anymore it should be deleted, on the other hand if it is still being required it should be configurable.
## Describe How to Reproduce

Execute transactions always use ConfidentialityProtocolVersion 1.2

andresgaragiola:
## Description

This PR includes the fix to the issue #1702 that allow configure the ConfidentialityProtocolVersion to the value 1.1. 
Additionally a way to test crypto_test.go using different settings was added in this PR. This improve the coverage of the package to %64. 
## Motivation and Context

ConfidentialityProtocolVersion 1.1 couldn't be used because the value 1.2 was set in the code. - -- 
This PR provides a way to configure the confidentiality protocol version to the value 1.1 in the core.yaml file.
In order to improve the coverage of the core/crypto coverage several tests were included. 

Fixes #1702
## How Has This Been Tested?

This PR includes the following test scenarios to test "crypto" packages:
- Using multithread enabled.
- Using SHA3-384.
- Using SHA2-256.
- Using SHA2-384
## Checklist:

- [X]  No new documentation is required by this change,
- [X]  I added new tests
- [X]  I have run [goimports](https://godoc.org/golang.org/x/tools/cmd/goimports), [go vet](https://golang.org/cmd/vet/), and [golint](https://github.com/golang/lint). I have cleaned up all valid errors and warnings in code I have added or modified. These tools may generate false positives. Don't be worried about ignoring some errors or warnings. The goal is clean, consistent, and readable code.

Signed-off-by: Andres Garagiola andresgaragiolal@gmail.com / garagiol@ar.ibm.com


peer/core.yaml
```diff
@@ -422,6 +422,8 @@ security:
         company: IBM
         position: "Software Engineer"
```
binhn, line 422: Use a fictitious company "ACME"

binhn:
@andresgaragiola except my comment above, LGTM otherwise.  Thanks.

peer/core.yaml
andresgaragiola, line 422: Fixed in 25e7520

andresgaragiola:
Thanks @binhn  I have fixed the company issue.

binhn:
I cancelled the CI build b/c it took too long and did the build and ran tests locally.

PR #1719, opened 2016-06-04
Do not store failed invoke transactions into the ledger

tamasblummer:
Do not store invoke transaction that fails to execute into the ledger
## Description

Chain code will likely implement checks of a transaction. A transaction that does not pass the checks without error should not be forwarded to consensus on order.
## Motivation and Context

Storing invalid transactions turns the ledger into an error log and can hurt performance by a single misbehaving client that pushes junk.
## How Has This Been Tested?

unit tests were run
## Checklist:

This commit changes transaction creation in a way that only
successful transactions (if the chaincode execution is
successful) are added to the ledger.


ghaskins:
This seems like a good idea to me:  Other networks like bitcoin do similar things where lack of confirmation is considered an implicit NACK as a protection against a DOS type attack.  Has there been any discussion of whether others agree?  @binhn @jeffgarratt @srderson?

@gabre note that the CI failure looks like it might be legit, please investigate.  Also, this patch has merge conflicts with master that will need to be resolved.

gabre:
@ghaskins I will check that CI failure ASAP. BTW any other comments, suggestions are welcome. :)
Edit: okay I got the errors. I will fix them ASAP.
Edit2: I have this: https://github.com/gabre/fabric/commits/successful-tx-only-unit-tests where I removed TxResults from unit tests (it is not stored in blocks anymore).

protos/fabric.proto
```diff
@@ -107,10 +107,8 @@ message BlockchainInfo {
 // the block hash when verifying the blockchain.
 // localLedgerCommitTimestamp - The time at which the block was added
 // to the ledger on the local peer.
-// transactionResults - The results of transactions.
 message NonHashData {
     google.protobuf.Timestamp localLedgerCommitTimestamp = 1;
-    repeated TransactionResult transactionResults = 2;
```
srderson, line 116: If the TransactionResult is removed, is there a reason to keep the TransactionResult message at line 67 or can that be removed also?

core/ledger/ledger.go
```diff
@@ -177,6 +177,9 @@ func (ledger *Ledger) CommitTxBatch(id interface{}, transactions []*protos.Trans
 	ledger.blockchain.blockPersistenceStatus(true)
 
 	sendProducerBlockEvent(block)
+	if len(transactionResults) != 0 {
+		ledgerLogger.Debug("There were some erroneous transactions. We need to send a 'TX rejected' message here.")
```
srderson, line 184: Is there an existing open issue? This is an important feature if the TransactionResult is to be removed. Assume it would use the events mechanism?

srderson:
The motivation behind this PR makes sense. It LGTM once the above comments are addressed. Also, it needs to be rebased as there is a merge conflict.

core/ledger/ledger.go
gabre, line 184: @srderson Do you mean something like this? https://github.com/hyperledger/fabric/pull/1754

protos/fabric.proto
gabre, line 116: Do you mean the type? See core/ledger/ledger.go:180. Btw https://github.com/hyperledger/fabric/pull/1754 (events) removes it as far as I remember because it sends events per Tx.

core/ledger/ledger.go
srderson, line 184: Yes, thanks. Hadn't seen that PR yet.

bcbrock:
@srderson @ghaskins @tamasblummer One of the issues that @jyellick fixed recently has to do with the blockchain containing empty blocks, that is, otherwise valid blocks that contained no transactions. This suggested change seems related. What if the failed transaction(s) is(are) the only TX in a block? Is the block formed? If so, then what if a group of TX fail on 1 peer for some reason and not on another? If we don't create the block, or if we have a different number of TX on one peer than the other, haven't we just created a fork in the chain? Even if some other mechanism eventually resolves the fork it seems we have violated one of the strongest conditions of our blockchain -- never a fork.

tamasblummer:
@bcbrock @jyellick an empty block should be avoided to save bandwidth, but it should remain a valid block.

gabre:
@srderson @ghaskins CI fails at TestSieveNoDecision. Do you have any idea why this happens? I made no modifications in that part of the code.

christo4ferris:
reverify Jenkins

christo4ferris:
@gabre I restarted jenkins build, but this also needs to be rebased and have the merge conflicts resolved. Thanks

gabre:
We have the following problem: so-called Chain Code Events are sent out which need transaction results to be in the block. I have removed them as these TX results are the same as storing erroneous TXs (and e.g. someone can easily flood the system and fill the blockchain with erroneous results, on the other hand, successful results are redundant as we just stored a zero) but they are definitely needed e.g. when PutRawBlock is called (I guess at synchronization). PutRawBlock (and other functions) need those TX results in order to be able to send out the mentioned chaincode events.
However, this also means that in case of synchronization, a node sends out events for some older blocks (the ones that are being synchronized). @christo4ferris @tamasblummer

muralisrini:
@gabre every chaincode invocation can send one event which needs to be store on the ledger if successful. It was convenient to store it in NonHashData->[ ]TransactionResult

We could create a new field called NonHashData->[ ]Event which is an array of events for each transaction in the block (non-nil in an element if the transaction generated an Event). I think @srderson mentioned that... 

That should work out for both uses.

@pmullaney, FYI

srderson:
@muralisrini I think that's a good idea. As far as I can tell, that's the only bit we need to keep from the TransactionResult

binhn:
If we change the structure effecting the current data, please post an update [here](https://github.com/hyperledger/fabric/wiki/Sprint-Release-Notes)

tamasblummer:
@muralisrini why would we store any event for successful invocation? The transaction being stored is already the proof of successful invocation. Unsuccessful invocation should not be stored at all to avoid DoS.

muralisrini:
@tamasblummer the event is a arbitrary custom event generated by the chaincode (it not a "system" event that says if a transaction is successful or not). It is stored only if the transaction is successful.

tamasblummer:
@muralisrini I wonder if there is a use case that required the feature 'chain code generated persistent event'. Is worldstate not enough for any chaincode if it wants to communicate?

muralisrini:
@tamasblummer good point. As you point out, user's could store the same "event" in world state and have users poll for it. By separating the two we enable proper "events" semantics to be tied to chaincode - such as listening vs polling (more information on https://github.com/hyperledger/fabric/wiki/Custom-Events-High-level-specification). It was made persistent for auditing and (in future) be part of consensus.

tamasblummer:
@muralisrini I think that this is a superfluous feature that should be removed except someone proves it can not be done with world state. An adjective like 'proper' is not a justification to complicate a core infrastructure and introduce DoS surface.

tamasblummer:
@muralisrini listening to chain code execution is addressed by our other PR also under review: https://github.com/hyperledger/fabric/pull/1754

binhn:
We need to keep the event with the tx, but it is a hack at this point until the next architecture, it is needed for playback during sync, so at this point, I would discourage doing things that changes the API or data compatibility, until we move to the next architecture, to reduce the churns in the user community.

My vote is to hold this PR/idea until we implement the Fabric next, which should be soon (starting next month?) and just focus on stabilizing the code base.

tamasblummer:
@binhn you introduced a feature that creates the requirement in the first place. I am questioning if the feature is needed and not its handling. It is BTW not an option to keep the feature in the new architecture as the events could only be delivered to endorsers without disclosing information about the transaction. There is also no reason to deliver to anyone else then endorsers since it is a result of chaincode execution. Delivering event back from chaincode execution would only happen in the submitting peer and well served with the mechanism we suggest in the other PR #1754

binhn:
@tamasblummer I think you completely misunderstood the purpose of the event mechanism in the Fabric. It has little to do with the new architecture. It is there for the applications who are interested in certain events and allowed to listen to those events.

tamasblummer:
@binhn I understand what you propose them for, but I am questioning if persistent events are a necessary component of the technology we build. The blockchain already creates a shared infrastructure that applications may listen to receive invoke transactions. Events you propose are result of chain code execution. Submitting peer could receive the events through the channel we proposed in PR #1754 as by-product of execution. Endorsing peers could also return them with the endorsed execution result.

A distribution of persistent events through shared infrastructure to all would either leak privacy or be worthless for other than submitter.

ghaskins:
@tamasblummer and I discussed in depth on slack.  I will try to summarize the conversation here
1. Events are critically important (to me)
2. Having them persist in the ledger is not important (to me)
3. Tamas is strongly against persisting them in the ledger, I don't care that much either way.
4. We can probably solve most of the problems by noting that events can still participate in consensus and collect endorsement signatures without committing the event object to a persistent layer.

@tamasblummer please correct any of the above if I misstated your position.

@muralisrini As a side note to the discussion, I would like to propose that chaincodes should ideally be able to emit more than one event per invoke (though there should be a practical limit, perhaps based on aggregate message sizes).  Consider something like an asset manager moving tokens:  It might want to emit an event for both the src and dst accounts.

ghaskins:
@muralisrini An alternate to multiple events might be to permit multiple topic keys per event, e.g. event.BalanceTransfer{ topics:{wallet-a, wallet-b} src:wallet-a, dst:wallet-b, amount:10}

binhn:
@ghaskins thanks for the summary.

Regarding persisting events with the trans, as I mentioned above, this is required for playback. Consider a scenario: Some applications listen on their peer for events, and the peer missed some blocks (network problems, whatever). It request sync. If events were not available, the applications would miss all the events they would have received otherwise.

I would not accept removing the events unless someone can propose a better alternative to support the above scenario.

tamasblummer:
@binhn @ghaskins 

I agree that it is a core function of the blockchain to agree on the order of events and distribute them to peers, be the peer online or catching up later. 

The difference in our view is that for me transaction = event, for you transaction = command and command implies events through chain code execution.

Could we establish common ground by implementing events such that:
1) error events are not stored - transactions that create them also not, this is this PR about.
2) an event that simply indicates successful execution will not be stored, since the transaction being there is the same information.
3) other events generated by chaincode will be stored.

ghaskins:
@binhn @tamasblummer I think I now agree with both you.  As I indicated on slack earlier today, after further consideration I do not see how chaincode events can be reliably implemented without storing them, so I reverse my position of not caring.  Or, more specifically I simply care that events work and are reasonably reliable, but I don't see how that can work without the persistence concept proposed.

I also agree with Tamas that what I would call "peer-level" events (errors, rejections, and other ephemeral states of the peer such as blocks being committed) should not persist.  Therefore, I think Tamas' items 1-3 capture my current position.  The only point of clarity is whether Binh was speaking about all (peer and chaincode) events or just chaincode events?  I suspect Binh meant only the latter and we are therefore all in agreement at this point.  However, I didn't want to assume.

Also, one additional point of clarity to add to (3) as we dicussed on slack: The event data should be considered as in the same confidentiality scope as the chaincode state and carry similar provisions for access/encryption/etc.

srderson:
Discussed this PR this morning during the Fabric maintainers meeting. The primary issue raised by @bcbrock was that the PR does not contain an update to the documentation explaining the new behavior. As the behavior may not be well documented in the first place, we cannot ask the submitter to add new docs. @muralisrini agreed to take a look using issue #2031

christo4ferris:
@hyperledger/fabric-maintainers - as I understand this is ready to merge once the merge conflicts have been resolved? If so, @tamasblummer please rebase and resolve conflicts.

gabre:
@christo4ferris @srderson @tamasblummer I have rebased it. Is it OK now?

gabre:
I fixed goimports (https://travis-ci.org/hyperledger/fabric/builds/141096428, consensus/obcpbft/mock_ledger_test.go) error and amended it to the last commit.

srderson:
LGTM. Also had @muralisrini review and he opened issue #2118 which needs to be implemented as part of this PR.

PR #1720, opened 2016-06-04
Adding filtering on file extension during deploy

angrbrd:
## Description

This change is being made to add control as to which files end up inside the .tar.gz package upon deploy. 
## Motivation and Context

The original .tar.gz  package was going through $GOPATH/src and packaging everything, which resulted in the .tar.gz. being around 50Mb. This was subsequently included as the payload in the chaincode deployment transaction, which made it prohibitively large. This change allows you to control which file extensions you want to package into the .tar.gz and avoid unnecessary files (executable files, files under /sdk/, files under /docs/). This provides for a successful deployment and reduces the .tar.gz size to around 3Mb.
## How Has This Been Tested?

I have tested this change by:

(a) Running the original UTs and making sure they pass.
(b) Performing manual chaincode deployments and observing the size of the resulting .tar.gz, which decreased from 50Mb --> 3Mb.

I am not adding any additional tests as this is not introducing any new features. 
## Checklist:


core/container/util/writer.go
```diff
@@ -45,11 +57,14 @@ func WriteGopathSrc(tw *tar.Writer, excludeDir string) error {
 		excludeDir = excludeDir + "/"
 	}
 
+	// Determine which file extensions should be packaged
+	fileExt := viper.GetStringSlice("chaincode.filetypes")
+
```
muralisrini, line 62: @angrbrd suggest getting fileExt once in init(). Also may be use a map[string]bool to store the file types so you don't have to walk the array for each file. ie, implement stringInSlice with allowedExts[filepath.Ext(path)]) ?

muralisrini:
@angrbrd @ghaskins 

This will include "github.com/src/hyperledger/fabric/bddtests/chaincode/go/table/table.go" because ALL go files will be included. But we don't need ANY files under bddtests... we could filter out entire folders that are unnecessary. 

Perhaps a "chaincode.excludefolders" as well ? This will contain "github.com/src/hyperledger/fabric/bddtests, github.com/src/hyperledger/fabric/sdk, github.com/src/hyperledger/fabric/images, github.com/src/hyperledger/fabric/devenv, github.com/src/hyperledger/fabric/docs, github.com/src/hyperledger/fabric/membersrvc, github.com/src/hyperledger/fabric/events".

This will also save some time traversing directory paths.

What do you think ?

binhn:
@muralisrini +1 on exclude folders, and we could exclude more folders, like consensus, peer, examples, scripts, tools, and more inside core/\* 

However, @angrbrd,  the file ext list seems to be golang centric, so should we define them under language?

ghaskins:
@muralisrini @binhn My thought is: Keep this mechanism simple as I view it as a stop-gap.

We can put a far more precise mechanism in place using the "go list" tool.  See [this blog](http://dave.cheney.net/2014/09/14/go-list-your-swiss-army-knife) for more details.  Once we do this, we can just simply ask the chaincode what it needs and get an exact list.  I plan on implementing this in chaintool, and it seems it would make sense to have the SDK/NVP support it too for GOLANG type

I whipped up a simple shell script to demonstrate the concepts presented in the blog.
screen shot 2016-06-04 at 7 56 42 pm

ghaskins:
Run [go-files.sh](https://github.com/hyperledger/fabric/files/299108/go-files.txt) like

```
$ ./go-files.sh github.com/hyperledger/fabric/examples/chaincode/go/chaincode_example01
```

And it will dump all the source files that chaincode_example01 requires to be packaged up.

ghaskins:
@binhn agree with you that this is golang specific and therefore should be scoped appropriately in the yaml

muralisrini:
@ghaskins I like that approach to include just the dependencies a lot! However I think excluding directories is simple enough to do  and worth doing even for a stop gap measure ? When we remove the "file ext" from the yaml, we can also remove the "exclude dir" list.

ghaskins:
@angrbrd I guess my thought was that the fileext gets us 99% of the way there and the go-list solution gets us to 100%, so why bother with the 99.5% solution of custom filter expressions.  That said, I don't have a strong opinion on it so if you and others think it's worth a cycle to add the filtering in, I can support that.

muralisrini:
@ghaskins if its really 99% vs 99.5% I don't have a strong opinion on going with the PR as is either....but it if the difference is more significant, it might be worth it.

angrbrd:
@muralisrini I agree with @ghaskins. I feel like we are over-optimizing this for a temporary solution. I just looked through the directories below and there are hardly any .go, .yaml, .c, or .h files in them. Note, that we are only including table.go under /fabric/bddtests as the other extensions are .yml, .py, and .feature. From this standpoint I don't think this change needs any additional filtering added.

github.com/src/hyperledger/fabric/bddtests, github.com/src/hyperledger/fabric/sdk, github.com/src/hyperledger/fabric/images, github.com/src/hyperledger/fabric/devenv, github.com/src/hyperledger/fabric/docs, github.com/src/hyperledger/fabric/membersrvc, github.com/src/hyperledger/fabric/events

What worries me more is that Travis CI seems to fail on deploy with the output below and it's not clear to me why. All my tests pass locally and I can't reproduce this failure. @ghaskins could you re-run the UTs with my change please?

2016/06/04 14:44:41 deploying chaincode abd6f800a8dc79e881ed9c09379bc4bb8f78ca433ada42a6026c338b202da7e9577134b8783b72fa402f7eac8688b3ba89ee5dc9a9178031ac405212c02dc256(networkid:dev,peerid:jdoe)

2016/06/04 14:44:41 waiting for container(dev-jdoe-abd6f800a8dc79e881ed9c09379bc4bb8f78ca433ada42a6026c338b202da7e9577134b8783b72fa402f7eac8688b3ba89ee5dc9a9178031ac405212c02dc256) lock

2016/06/04 14:44:41 got container (dev-jdoe-abd6f800a8dc79e881ed9c09379bc4bb8f78ca433ada42a6026c338b202da7e9577134b8783b72fa402f7eac8688b3ba89ee5dc9a9178031ac405212c02dc256) lock

2016/06/04 14:44:42 Error building Peer container: The command '/bin/sh -c go install github.com/hyperledger/fabric/examples/chaincode/go/chaincode_example01 && cp src/github.com/hyperledger/fabric/peer/core.yaml $GOPATH/bin && mv $GOPATH/bin/chaincode_example01 $GOPATH/bin/abd6f800a8dc79e881ed9c09379bc4bb8f78ca433ada42a6026c338b202da7e9577134b8783b72fa402f7eac8688b3ba89ee5dc9a9178031ac405212c02dc256' returned a non-zero code: 1

angrbrd:
All the failing tests in Travis CI are under /fabric/core/chaincode. All these tests pass on my machine. Can someone else please give it a shot? Thanks.

--- FAIL: TestExecuteDeployTransaction (2.71s)
--- FAIL: TestGopathExecuteDeployTransaction (2.54s)
--- FAIL: TestHTTPExecuteDeployTransaction (18.08s)
--- FAIL: TestExecuteInvokeTransaction (2.81s)
--- FAIL: TestExecuteQuery (2.67s)
--- FAIL: TestExecuteInvalidQuery (3.16s)
--- FAIL: TestExecuteInvalidQuery (3.16s)
...

JonathanLevi:
@angrbrd I believe there's still some underlying CI pipeline work (most PRs are showing a failure to build/test, while tests pass locally). Yesterday, I saw on Slack that the CI/testing team are planning to address this today.

muralisrini:
@rameshthoomu travis builds should still complete correct ? Only Jenkins  failures are expected ?

rameshthoomu:
@muralisrini @JonathanLevi: Please ignore Jenkins build results for now. LF team and myself are working on behave issue fix. 

Travis build results are still correct. Someone has to look into unit-test failures.

ghaskins:
@muralisrini to be clear, I made the 99/99.5% numbers up to make my point ;).  I don't know what the actual difference is but my gut says that this patch does the vast majority of the optimization.  I think we are all on the same page here, however.

Regarding the unit-test failures, I'll try to take a look.

As an aside, these errors would likely just go away completely if we put this patch in after/with #1708 since that patch gets rid of the peer/ca generation.  So, we should decide if it's worth messing around with investigating the failure or just realign with the other work.

angrbrd:
@ghaskins I am OK with this going in after #1708 if you think that would be easier.

ghaskins:
@angrbrd I'll push an update to 1708 that drops my similar work, and we can layer yours on top...unless @muralisrini @binhn think 1708 will take a while to review and/or be rejected and this fix is urgent.

muralisrini:
@ghaskins I think it'll take a bit of time to review 1708 for me. 

More importantly though, this seems like a great opportunity to enhance package.go so we get more information than just **returned a non-zero code: 1** in
`
2016/06/04 14:44:42 Error building Peer container: The command '/bin/sh -c go install github.com/hyperledger/fabric/examples/chaincode/go/chaincode_example01 && cp src/github.com/hyperledger/fabric/peer/core.yaml $GOPATH/bin && mv $GOPATH/bin/chaincode_example01 $GOPATH/bin/abd6f800a8dc79e881ed9c09379bc4bb8f78ca433ada42a6026c338b202da7e9577134b8783b72fa402f7eac8688b3ba89ee5dc9a9178031ac405212c02dc256' returned a non-zero code: 1
`

In addition to debugging this failure, it will also help users debug chaincode build issues and such.

I don't know how to do that yet though. What do you think ?

ghaskins:
@angrbrd Shall we close this now that #1774 has been merged?

core/ledger/kvledger/txmgmt/privacyenabledstate/db.go
```diff
@@ -46,9 +46,10 @@ type StateDBConfig struct {
 // DBProvider encapsulates other providers such as VersionedDBProvider and
 // BookeepingProvider which are required to create DB for a channel
 type DBProvider struct {
-	VersionedDBProvider statedb.VersionedDBProvider
-	HealthCheckRegistry ledger.HealthCheckRegistry
-	bookkeepingProvider *bookkeeping.Provider
+	VersionedDBProvider      statedb.VersionedDBProvider
+	HealthCheckRegistry      ledger.HealthCheckRegistry
+	bookkeepingProvider      *bookkeeping.Provider
+	versionFromSnapshotValue func(snapshotValue []byte) ([]byte, error)
```
cendhu: as snapshot.go introduces `versionFromSnapshotValueFunc` type, we can use that type here. Maybe we can move that type definition to this file and use it at both places.

core/ledger/kvledger/txmgmt/privacyenabledstate/snapshot.go
```diff
@@ -419,3 +473,8 @@ func (c *cursor) canMove() bool {
 func (c *cursor) currentNamespace() string {
 	return c.metadata[c.currentRow].namespace
 }
+
+type SnapshotPvtdataHashesConsumer interface {
+	ConsumeSnapshotData(namespace, coll string, keyHash []byte, version []byte) error
```
cendhu, line 472: instead of `version []byte`, we could have directly passed the `*version`. Otherwise, both the pvtdatastore and purgemgr have to unmarshal it.

core/ledger/kvledger/txmgmt/privacyenabledstate/snapshot.go
```diff
@@ -419,3 +473,8 @@ func (c *cursor) canMove() bool {
...
+	Done() error
+}
```
cendhu: IMO, this is one among the correct usages of an interface. There are multiple consumers and the contract is that each of them should have implemented these methods (similar to io.Writer).

core/ledger/kvledger/txmgmt/pvtstatepurgemgmt/snapshot_purge_mgr_builder.go
```diff
@@ -0,0 +1,69 @@
...
+	ver, _, err := version.NewHeightFromBytes(versionBytes)
+	if err != nil {
+		return errors.WithMessage(err, "invalid version bytes")
+	}
+
+	committingBlk := ver.BlockNum
+	expiringBlock, err := b.btlPolicy.GetExpiringBlock(namespace, coll, committingBlk)
+	if err != nil {
+		return errors.WithMessage(err, "error from btlpolicy")
+	}
```
cendhu, line 45: After this, shouldn't we do
```
if neverExpires(blkNum) {
    return nil
}
```
Maybe it is better to make `GetExpiringBlock()` return a `bool` so that we would not miss it in the future. Not for this PR anyway.

core/ledger/kvledger/txmgmt/privacyenabledstate/db.go
manish-sethi: done

core/ledger/kvledger/txmgmt/privacyenabledstate/snapshot_test.go
```diff
@@ -643,3 +645,98 @@ func TestSnapshotImportErrorPropagation(t *testing.T) {
...
+}
+
+func TestSnapshotImportPvtdataHashesConsumer(t *testing.T) {
+	var dbEnv *LevelDBTestEnv
+	var snapshotDir string
+	var cleanup func()
+	var err error
+
+	init := func() {
+		dbEnv = &LevelDBTestEnv{}
```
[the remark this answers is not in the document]
manish-sethi: There are separate unit tests for `VersionFromSnapshotValue` in each of the packages, in `stateleveldb` and `statecouchdb`. In this package, the goal of testing the code in this package is achieved with one of these.

core/ledger/kvledger/txmgmt/privacyenabledstate/snapshot_test.go
```diff
@@ -643,3 +645,98 @@ func TestSnapshotImportErrorPropagation(t *testing.T) {
...
+		dbEnv.Init(t)
+		db := dbEnv.GetDBHandle(generateLedgerID(t))
+		updateBatch := NewUpdateBatch()
+		updateBatch.PubUpdates.Put("ns-1", "key-1", []byte("value-1"), version.NewHeight(1, 1))
+		updateBatch.HashUpdates.Put("ns-1", "coll-1", []byte("key-hash-1"), []byte("value-hash-1"), version.NewHeight(1, 1))
+		require.NoError(t, db.ApplyPrivacyAwareUpdates(updateBatch, version.NewHeight(1, 1)))
+		snapshotDir, err = ioutil.TempDir("", "testsnapshot")
+		require.NoError(t, err)
+		_, err := db.ExportPubStateAndPvtStateHashes(snapshotDir, testNewHashFunc)
+		require.NoError(t, err)
```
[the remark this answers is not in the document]
manish-sethi: `testify` has a better way of dealing with this now. Added that here.

core/ledger/kvledger/txmgmt/pvtstatepurgemgmt/snapshot_purge_mgr_builder_test.go
```diff
@@ -0,0 +1,182 @@
...
+	require.NoError(t,
+		purgeMgrBuilder.ConsumeSnapshotData("ns1", "coll1", []byte("key-hash-2"), version.NewHeight(5, 1).ToBytes()),
+	)
+	// add data that expires at block-8
+	require.NoError(t,
+		purgeMgrBuilder.ConsumeSnapshotData("ns2", "coll2", []byte("key-hash-2"), version.NewHeight(5, 1).ToBytes()),
+	)
+	// add data that never expires
+	require.NoError(t,
+		purgeMgrBuilder.ConsumeSnapshotData("ns1", "no-btl-collection", []byte("key-hash-3"), version.NewHeight(5, 1).ToBytes()),
```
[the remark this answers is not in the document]
manish-sethi: The sample BTL mock returns 0 for not explicitly mentioned collections. So, it assumes a BTL 0 for everything except explicitly mentioned. I guess that it may be better to avoid this confusion and instead return the error for the not registered names.

core/ledger/kvledger/txmgmt/pvtstatepurgemgmt/snapshot_purge_mgr_builder_test.go
```diff
@@ -0,0 +1,182 @@
...
+func TestPurgeMgrBuilder(t *testing.T) {
+	bookkeepingEnv := bookkeeping.NewTestEnv(t)
+	defer bookkeepingEnv.Cleanup()
+
+	ledgerID := "test-ledger"
+	bookkeepingProvider := bookkeepingEnv.TestProvider
+	btlPolicy := btltestutil.SampleBTLPolicy(
+		map[[2]string]uint64{
+			{"ns1", "coll1"}: 1,
+			{"ns2", "coll2"}: 2,
```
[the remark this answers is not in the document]
manish-sethi: See the related comment below

core/ledger/kvledger/txmgmt/pvtstatepurgemgmt/snapshot_purge_mgr_builder.go
```diff
@@ -0,0 +1,69 @@
...
+	if err := b.expKeeper.update([]*expiryInfo{expInfo}, nil); err != nil {
+		return err
+	}
+	return nil
+}
+
+// Done implements the function in the interface privacuenabledstate.SnapshotPvtdataHashesConsumer.
+func (b *PurgeMgrBuilder) Done() error {
+	return nil
+}
```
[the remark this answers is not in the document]
manish-sethi: I don't mind removing it for now. However, one subtle difference is that I like to bring is that for the purge mgr the optimization may not be that important because for most of the data items that do not explicitly specify the BTL, no expiry entries gets created. For the one that does specify, I would assume that the entries will be handful (assuming the BTL value is reasonably small). So, the impact may not be much. I feel that we will perhaps need to do some form of batching for the pvtdata store, where the above is no longer true.

core/ledger/kvledger/txmgmt/pvtstatepurgemgmt/snapshot_purge_mgr_builder.go
manish-sethi: Good catch! Yes, let me change the function signature for this one in separate PR, so not to have a miss on this.

core/ledger/kvledger/txmgmt/privacyenabledstate/snapshot.go
manish-sethi: What I had in mind was to introduce an optimization by extracting just the block number. Both the consumer just need the block number, in fact. Let me change the function signature to return just the block number in a next PR

PR #1721, opened 2016-06-04
Optionally generate invoke transaction ID from transaction content

tamasblummer:
Optionally generate invoke transaction ID  from transaction content
## Description

Adds a -tid options to CLI that determines the transaction id derivation algorithm. 
Adds uuidGenerationAlg to ChaincodeInvocationSpec with same meaning and content option. 

Currently supported derivation algorithm is `sha256base64`,  that means transaction is assumed to be encoded as a base64 string and will be hashed with sha256, then first 128 bits will be formatted like an RFC 4122 UUID to identify the transaction
## Motivation and Context

If UUID is derived from transaction content, then it is already known by the ledger client application, that is able to track it from construction to storage into the ledger. If UUID was assigned by the ledger (as is now) then a client would have to find the transaction in the block chain by comparing its content with content that it sent in. Deriving transaction id from content with a cryptographic hash practically ensures that ID will be unique, no other transaction with any different content will have the same ID. A limit of the approach is that identical invoke would receive same ID, which is however easy to avoid by the creator of the transaction. It is questionable if identical invokes should be processed at all since if they are allowed then trivial transaction replay attack can be executed on the ledger.
## How Has This Been Tested?

unit tests were run
## Checklist:

### DIFF
--- a/core/devops.go
+++ b/core/devops.go
@@ -19,6 +19,7 @@ package core
 import (
 	"errors"
 	"fmt"
+	"strings"
 
 	"github.com/op/go-logging"
 	"github.com/spf13/viper"
@@ -209,7 +210,18 @@ func (d *Devops) invokeOrQuery(ctx context.Context, chaincodeInvocationSpec *pb.
 	}
 
 	// Now create the Transactions message and send to Peer.
-	uuid := util.GenerateUUID()
+	var customUUIDgenAlg = strings.ToLower(chaincodeInvocationSpec.UuidGenerationAlg)
+	var uuid string
+	var generr error
+	if customUUIDgenAlg != "" {
+		uuid, generr = util.GenerateUUIDWithAlg(customUUIDgenAlg, chaincodeInvocationSpec.ChaincodeSpec.CtorMsg.Args[0])
+		if generr != nil {
+			return nil, generr
+		}
+	} else {
+		uuid = util.GenerateUUID()
+	}
+	devopsLogger.Info("Transaction ID: %v", uuid)
 	var transaction *pb.Transaction
 	var err error
 	var sec crypto.Client
--- a/core/devops.go
+++ b/core/devops.go
@@ -210,18 +210,18 @@ func (d *Devops) invokeOrQuery(ctx context.Context, chaincodeInvocationSpec *pb.
 	}
 
 	// Now create the Transactions message and send to Peer.
-	var customUUIDgenAlg = strings.ToLower(chaincodeInvocationSpec.UuidGenerationAlg)
-	var uuid string
+	var customIDgenAlg = strings.ToLower(chaincodeInvocationSpec.IdGenerationAlg)
+	var id string
 	var generr error
-	if customUUIDgenAlg != "" {
-		uuid, generr = util.GenerateUUIDWithAlg(customUUIDgenAlg, chaincodeInvocationSpec.ChaincodeSpec.CtorMsg.Args[0])
+	if customIDgenAlg != "" {
+		id, generr = util.GenerateIDWithAlg(customIDgenAlg, chaincodeInvocationSpec.ChaincodeSpec.CtorMsg.Args[0])
 		if generr != nil {
 			return nil, generr
 		}
 	} else {
-		uuid = util.GenerateUUID()
+		id = util.GenerateUUID()
 	}
-	devopsLogger.Info("Transaction ID: %v", uuid)
+	devopsLogger.Info("Transaction ID: %v", id)
 	var transaction *pb.Transaction
 	var err error
 	var sec crypto.Client
@@ -238,7 +238,7 @@ func (d *Devops) invokeOrQuery(ctx context.Context, chaincodeInvocationSpec *pb.
 		}
 	}
 
-	transaction, err = d.createExecTx(chaincodeInvocationSpec, attributes, uuid, invoke, sec)
+	transaction, err = d.createExecTx(chaincodeInvocationSpec, attributes, id, invoke, sec)
 	if err != nil {
 		return nil, err
 	}
--- a/bddtests/peer_basic.feature
+++ b/bddtests/peer_basic.feature
@@ -315,6 +315,27 @@ Feature: lanching 3 peers
 
 #    @doNotDecompose
 #    @wip
+#    Arg[0] = a, base64 = 'YQ=='
+#    sha256 = 'ca978112ca1bbdcafac231b39a23dc4da786eff8147c4e72b9807785afee48bb'
+	Scenario: chaincode map single peer content generated UUID
+	    Given we compose "docker-compose-1.yml"
+	    When requesting "/chain" from "vp0"
+	    Then I should get a JSON response with "height" = "1"
+	    When I deploy chaincode "github.com/hyperledger/fabric/examples/chaincode/go/map" with ctor "init" to "vp0"
+	      ||
+              ||
+	    Then I should have received a chaincode name
+	    Then I wait up to "60" seconds for transaction to be committed to all peers
+
+        When I invoke chaincode "map" function name "put" on "vp0" with "sha256base64"
+	    | arg1  |arg2|
+            | YQ==  | 10 |
+	    Then I should have received a transactionID
+	    Then I wait up to "25" seconds for transaction to be committed to all peers
+	    Then I check the transaction ID if it is "ca978112-ca1b-bdca-fac2-31b39a23dc4d"
+
+#    @doNotDecompose
+#    @wip
 	Scenario: chaincode example 02 single peer
 	    Given we compose "docker-compose-1.yml"
 	    When requesting "/chain" from "vp0"
--- a/bddtests/peer_basic.feature
+++ b/bddtests/peer_basic.feature
@@ -317,7 +317,7 @@ Feature: lanching 3 peers
 #    @wip
 #    Arg[0] = a, base64 = 'YQ=='
 #    sha256 = 'ca978112ca1bbdcafac231b39a23dc4da786eff8147c4e72b9807785afee48bb'
-	Scenario: chaincode map single peer content generated UUID
+	Scenario: chaincode map single peer content generated ID
 	    Given we compose "docker-compose-1.yml"
 	    When requesting "/chain" from "vp0"
 	    Then I should get a JSON response with "height" = "1"
--- a/bddtests/steps/peer_basic_impl.py
+++ b/bddtests/steps/peer_basic_impl.py
@@ -204,10 +204,10 @@ def step_impl(context):
     else:
         fail('chaincodeSpec not in context')
 
-@when(u'I invoke chaincode "{chaincodeName}" function name "{functionName}" on "{containerName}" with "{uuidGenAlg}"')
-def step_impl(context, chaincodeName, functionName, containerName, uuidGenAlg):
+@when(u'I invoke chaincode "{chaincodeName}" function name "{functionName}" on "{containerName}" with "{idGenAlg}"')
+def step_impl(context, chaincodeName, functionName, containerName, idGenAlg):
     assert 'chaincodeSpec' in context, "chaincodeSpec not found in context"
-    invokeChaincode(context, "invoke", functionName, containerName, uuidGenAlg)
+    invokeChaincode(context, "invoke", functionName, containerName, idGenAlg)
 
 @when(u'I invoke chaincode "{chaincodeName}" function name "{functionName}" on "{containerName}" "{times}" times')
 def step_impl(context, chaincodeName, functionName, containerName, times):
@@ -234,7 +234,7 @@ def step_impl(context, chaincodeName, functionName, containerName):
 def step_impl(context, chaincodeName, functionName, containerName):
     invokeChaincode(context, "query", functionName, containerName)
 
-def invokeChaincode(context, devopsFunc, functionName, containerName, uuidGenAlg=None):
+def invokeChaincode(context, devopsFunc, functionName, containerName, idGenAlg=None):
     assert 'chaincodeSpec' in context, "chaincodeSpec not found in context"
     # Update hte chaincodeSpec ctorMsg for invoke
     args = []
@@ -248,8 +248,8 @@ def invokeChaincode(context, devopsFunc, functionName, containerName, uuidGenAlg
         "chaincodeSpec" : context.chaincodeSpec
     }
     ipAddress = bdd_test_util.ipFromContainerNamePart(containerName, context.compose_containers)
-    if uuidGenAlg is not None:
-	    chaincodeInvocationSpec['uuidGenerationAlg'] = uuidGenAlg
+    if idGenAlg is not None:
+	    chaincodeInvocationSpec['idGenerationAlg'] = idGenAlg
     request_url = buildUrl(context, ipAddress, "/devops/{0}".format(devopsFunc))
     print("{0} POSTing path = {1}".format(currentTime(), request_url))
 
--- a/bddtests/steps/peer_basic_impl.py
+++ b/bddtests/steps/peer_basic_impl.py
@@ -204,6 +204,11 @@ def step_impl(context):
     else:
         fail('chaincodeSpec not in context')
 
+@when(u'I invoke chaincode "{chaincodeName}" function name "{functionName}" on "{containerName}" with "{uuidGenAlg}"')
+def step_impl(context, chaincodeName, functionName, containerName, uuidGenAlg):
+    assert 'chaincodeSpec' in context, "chaincodeSpec not found in context"
+    invokeChaincode(context, "invoke", functionName, containerName, uuidGenAlg)
+
 @when(u'I invoke chaincode "{chaincodeName}" function name "{functionName}" on "{containerName}" "{times}" times')
 def step_impl(context, chaincodeName, functionName, containerName, times):
     assert 'chaincodeSpec' in context, "chaincodeSpec not found in context"
@@ -229,7 +234,7 @@ def step_impl(context, chaincodeName, functionName, containerName):
 def step_impl(context, chaincodeName, functionName, containerName):
     invokeChaincode(context, "query", functionName, containerName)
 
-def invokeChaincode(context, devopsFunc, functionName, containerName):
+def invokeChaincode(context, devopsFunc, functionName, containerName, uuidGenAlg=None):
     assert 'chaincodeSpec' in context, "chaincodeSpec not found in context"
     # Update hte chaincodeSpec ctorMsg for invoke
     args = []
@@ -243,6 +248,8 @@ def invokeChaincode(context, devopsFunc, functionName, containerName):
         "chaincodeSpec" : context.chaincodeSpec
     }
     ipAddress = bdd_test_util.ipFromContainerNamePart(containerName, context.compose_containers)
+    if uuidGenAlg is not None:
+	    chaincodeInvocationSpec['uuidGenerationAlg'] = uuidGenAlg
     request_url = buildUrl(context, ipAddress, "/devops/{0}".format(devopsFunc))
     print("{0} POSTing path = {1}".format(currentTime(), request_url))
 
@@ -328,6 +335,22 @@ def step_impl(context, seconds):
     print("Result of request to all peers = {0}".format(respMap))
     print("")
 
+@then(u'I check the transaction ID if it is "{tUUID}"')
+def step_impl(context, tUUID):
+    assert 'transactionID' in context, "transactionID not found in context"
+    assert context.transactionID == tUUID, "transactionID is not tUUID"
+
+def getContainerDataValuesFromContext(context, aliases, callback):
+    """Returns the IPAddress based upon a name part of the full container name"""
+    assert 'compose_containers' in context, "compose_containers not found in context"
+    values = []
+    containerNamePrefix = os.path.basename(os.getcwd()) + "_"
+    for namePart in aliases:
+        for containerData in context.compose_containers:
+            if containerData.containerName.startswith(containerNamePrefix + namePart):
+                values.append(callback(containerData))
+                break
+    return values
 
 @then(u'I wait up to "{seconds}" seconds for transaction to be committed to peers')
 def step_impl(context, seconds):
@@ -475,7 +498,7 @@ def step_impl(context, userName, secret):
         assert resp.status_code == 200, "Failed to POST to %s:  %s" %(request_url, resp.text)
         context.response = resp
         print("message = {0}".format(resp.json()))
-    
+
         # Create new User entry
         bdd_test_util.registerUser(context, secretMsg, containerData.composeService)
 
--- a/core/util/utils.go
+++ b/core/util/utils.go
@@ -18,6 +18,8 @@ package util
 
 import (
 	"crypto/rand"
+	"crypto/sha256"
+	"encoding/base64"
 	"fmt"
 	"io"
 	"math/big"
@@ -27,6 +29,15 @@ import (
 
 	"golang.org/x/crypto/sha3"
 )
+
+type alg struct {
+	hashFun func([]byte) string
+	decoder func(string) ([]byte, error)
+}
+
+var availableUUIDgenAlgs = map[string]alg{
+	"sha256base64": alg{GenerateUUIDfromTxSHAHash, base64.StdEncoding.DecodeString},
+}
 
 // ComputeCryptoHash should be used in openchain code so that we can change the actual algo used for crypto-hash at one place
 func ComputeCryptoHash(data []byte) (hash []byte) {
@@ -62,7 +73,7 @@ func GenerateIntUUID() *big.Int {
 // GenerateUUID returns a UUID based on RFC 4122
 func GenerateUUID() string {
 	uuid := GenerateBytesUUID()
-	return fmt.Sprintf("%x-%x-%x-%x-%x", uuid[0:4], uuid[4:6], uuid[6:8], uuid[8:10], uuid[10:])
+	return uuidBytesToStr(uuid)
 }
 
 // CreateUtcTimestamp returns a google/protobuf/Timestamp in UTC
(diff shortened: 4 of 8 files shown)


adecaro:
Hi @tamasblummer, I have some concerns related to the security aspects of this proposal that I would like to share. If the fabric allows a client to decide which algorithm to use to generate the uuid, then the uuid itself can be used as a side channel to link transactions that are using the same algorithm. This combined with other auxiliary information could potentially reveal too much about certain chaincodes.

I would say then that a chain should speak only one _language_ (in terms of protocols, cryptographic primitives and so on) defined when the chain is bootstrapped and all the parties playing on that chain should speak that same language as well. Changing the language of a chain already bootstrapped should pass through the invocation of a system chaincode. 

What do you think? Am I missing something?

tamasblummer:
@adecaro  I would welcome a restriction to a single algorithm for content derived transaction ID. Actually I'd love to see random UUID go away entirely for invoke.

Unfortunately, the current protobuf transaction format defines payload as a string, so it already raises the question of encoding binary content, for which there are many choices.
We also saw hashing algorithms being broken.

I chose to address these contradicting requirements by defining only one option that is sha256base64. Maybe we can ease your concern by simply not defining further alternatives unless there is some very strong security argument for it.

adecaro:
@tamasblummer Fair enough. Thanks for you reply :)

adecaro:
@tamasblummer How does this PR affect the client-sdk? Is it necessary to align it?

Also, shall the fabric enforce that the uuid has the proper format? The fabric could also enforce that a certain algorithm is used and only that among the available ones. What do you think?

tamasblummer:
@adecaro this is not a dependency for client sdk, but it should consider using the same content derived id.

I think it is a weakness that transaction id is represented as string but there is no strict check of its format, that is assumed to be RFC 4122 compliant. I think transaction id should rather be binary for best performance and should be serialised as an uuid like string only on UI and in logs.

tamasblummer:
@binhn @srderson @ghaskins May I have an opinion on this?

srderson:
I agree with this change and think the concept makes sense. I see it as a helper for the CLI and REST interface. I think most clients should be using the `ProcessTransaction` service in `fabric.proto` which allows the client to specify the transaction ID.

JonathanLevi:
A few words:
- +1 for the `binary` representation (especially for storing/loading/comparing/processing/sending/receiving. Totally!)
- +1 for `SHA-256` at this point (rather than the 'faster' `SHA-1`).
- Yes, I think it's right that we should def. design it so that later on one/we can replace the `algo` (like a specified/supported "cypersuite" specification) for whatever unforeseeable future reason/event.

Frankly, as/while so much is changing...imo, if it's extensible/specifiable, then we can probably live happily with one good option now (and revisit later if/when needed).

christo4ferris:
@tamasblummer please rebase to resolve merge conflicts otherwise LGTM thanks

bcbrock:
My concern about this suggestion is only that it makes an implicit statement that transactions are not guaranteed to have unique IDs - the correctness of a robust system can not depend on the correctness of its inputs. I don't know if non-uniqueness is real problem or not, however I think it's important for everyone to realize that this PR weakens the specification by removing the requirement that every transaction has a unique ID.

tamasblummer:
@bcbrock the transaction ID would be guaranteed unique provided the transaction content is unique; since it would be generated with a crypto hash that is assumed not to have collisions. Identical content invoke transactions would open the door for replay attacks, therefore should not be part of a safe design of a chain code use.

bcbrock:
@tamasblummer The TX IDs are only unique if clients "play by the rules". If there is a hard requirement for IDs to be unique then either 1) they can not be created by the clients, or 2) the peer will need to check every submitted TX ID against all previous TX IDs to enforce uniqueness. Option 2) is probably unworkable in a real system in real time. Again, I don't know if there really _is_ a requirement for unique IDs - but if someone thinks there should be such a requirement, they should be aware that this approach can not formally guarantee that requirement.

tamasblummer:
@bcbrock It is not the ID created by the client, but transaction content. ID is a function of content. Therefore the requirement is not that client has to create distinct id, but that client has to send unique transaction content. If client would not send unique transaction content and duplicate transaction content was valid to process, then any man-in-the-middle attacker would be able to replay the transaction. Therefore the design of chain code was critically flawed.

tamasblummer:
@bcbrock using a global lookup of ID, that I agree was expensive, is only needed if there is no other way for chain code to ensure content is unique.

srderson:
@bcbrock - @tamasblummer is not changing any functionality here. The transaction ID was generated by the client in the past and is still generated by the client after his change. `devops.go` is essentially part of the client as it calls gRPC on the peer. Any client can go around `devops.go` and call the peer directly. `devops.go` just supplies a REST and CLI.

@tamasblummer It sounds to me you may be suggesting that we move the ID generation completely to the server, based on the content of the transaction. Is that correct?

tamasblummer:
@srderson if ID is derived by content then it is derivable both on client and server, and this is the point. The client knows transaction ID in advance. The implication is the requirement of unique content, but that is as explained above is already a requirement in my opinion to avoid trivial replay attack.

adecaro:
@tamasblummer @bcbrock @srderson In the protocol-spec, we have a sketched design for replay attack resistance (Section 4.3.3) that leverage tCerts validity period in order to reduce the amount of memory needed to check for replay attacks. Of course, it can be improved further, any feedback on that would be really appreciated :)

Therefore, I think it should be fine to let the client computing the transaction id. What I think is more important is that the fabric enforces a single algorithm to compute the id to avoid linkability attacks.

rjones-lf:
reverify jenkins

tamasblummer:
@adecaro I think that replay attack protection needs to be present without using the particular certificates you advertise, best if it is implied by the design of the chain code, so replayed transaction can not be valid. 

As I wrote above, 16 days ago, I'd love to fix id generation algorithm and binary encoding.

binhn:
@tamasblummer @adecaro I agreed that the tx id should be generated by the client, and the peer can recompute if needed. However, shouldn't we be worry about hash collisions, especially when use only the first 128 bits? 

On the replay attacks, the next iteration of Fabric would implicitly prevent that with MVCC, but coupling with content-based tx id would be better.

About this PR, we could either merge now (given code reviews addressed) and change it later as the current plan for the SDK is to absorb the creation of the transactions (as implemented in the Nodejs SDK) and eventually deprecate/remove NVP (devops, REST), specified here https://github.com/hyperledger/fabric/wiki/Client-SDK-Specification. I'd like to build this same functionality into Fabric-API for Java SDK.

Thoughts?

core/devops.go
```diff
@@ -209,7 +210,18 @@ func (d *Devops) invokeOrQuery(ctx context.Context, chaincodeInvocationSpec *pb.
...
+	var generr error
+	if customUUIDgenAlg != "" {
+		uuid, generr = util.GenerateUUIDWithAlg(customUUIDgenAlg, chaincodeInvocationSpec.ChaincodeSpec.CtorMsg.Args[0])
+		if generr != nil {
+			return nil, generr
+		}
+	} else {
+		uuid = util.GenerateUUID()
+	}
+	devopsLogger.Info("Transaction ID: %v", uuid)
```
muralisrini: @tamasblummer @gabre just noticed on line 217 we are using just Args[0]... is that the intent ? Doesn't the uuid need to be computed on the whole content (and even the whole transaction bytes) ?

srderson:
@binhn I'm in favor of merging because
1) It's a step in the right direction
2) It ensures @tamasblummer's proposal is in the Fabric and that we'll be sure to keep this requirement given future modifications as he has added a test case.

I have three questions that may not need to be addressed in this PR, but I want to make sure I understand

@adecaro Do we really need both UUID and nonce? Could we just use nonce? I'm in favor of @tamasblummer's suggestion to replace the string with a binary encoding.

Should we consider placing the `uuidGenerationAlg` in the `Transaction` message instead of the `ChaincodeInvocationSpec` message? Currently we're saying it's not applicable to deploy transactions.

@tamasblummer Could we rename this to `idGenerationAlg`? Ideally I'd like to rename `uuid` to `id` if it's not going to be a uuid.

core/devops.go
tamasblummer: This assumes that the entire transaction is encoded base64 in the first parameter. We should split processing of CLI and rpc later, so rpc can be all binary.

gabre:
@srderson I think, of course, we can rename uuidGen... to idGen. As for uuid vs id, there are thousands of occurrences of uuid - I think that renaming should be another PR.

gabre:
@srderson Please check it now.

tamasblummer:
@srderson lets merge as is to make the step into the right direction. Subsequent PRs will clean up naming, address binary id, binary tx transfer in rpc and remove random uuid altogether.

srderson:
I missed @muralisrini's comment at https://github.com/hyperledger/fabric/pull/1721#discussion_r67946972 where you respond and state that "This assumes that the entire transaction is encoded base64 in the first parameter". The problem is that not everyone makes that assumption.

Could `GenerateIDWithAlg` take the `chaincodeInvocationSpec` instead of just `chaincodeInvocationSpec.ChaincodeSpec.CtorMsg.Args[0]`. That would allow the hashing algorithm to decide what actually makes up a transaction.

tamasblummer:
@srderson the problem is that the invoke rpc interface is directly linked with the CLI implementation. CLI could use text interface in any number of words. I see no reason why on a lower level, which should be the interface of invoke, we could not assume that transaction is a single binary array. The solution now is not pure, but again goes a step into the right direction as you also recognise. It forces to use single binary representation of transaction even in CLI, but only if you want to use the new content derived id. Before this gets mandatory (which I hope) we will clean up the rest. I am losing my patience here. Is it possible to do incremental changes into the right direction or do you expect DA to rewrite fabric in an atomic PR which is rebased daily? @christo4ferris

srderson:
@tamasblummer My goal is certainly not to frustrate anyone, though I can certainly understand why you feel that way at this point. 

I will merge this PR as I want you to be successful in your use of Fabric, though I want to clearly outline where I agree and don't agree and we can continue further discussion.

I agree with
1. The ID of a transaction should be based on its hash
2. The client should generate the ID

I disagree with or don't understand the use case
1. The ID generation algorithm is stored on the blockchain for every transaction

This would make sense if the different clients were using different hashing algorithms, but is this really the case? Why could the hashing algorithm not be stored in the `core.yaml` or the chaincode which is verifying the ID would know the algorithm? Maybe another case is if the algorithm changes overtime, but I would think this would be an exceptionally rare case that could warrant the peers upgrading their software. 

Maybe we need a version number in the transaction, like Bitcoin, to help indicate future changes to the hashing algorithm. That would give us space savings over a string.

Again, I am hesitant to merge because this is an API change (changes a proto file) AND adds additional data to the blockchain for every transaction, but I will merge so we can move forward in subsequent PRs.

Please let me know your thoughts.

core/ledger/pvtdatastorage/store.go
```diff
@@ -140,6 +143,7 @@ func (p *Provider) OpenStore(ledgerid string) (*Store, error) {
 		return nil, err
 	}
 	s.launchCollElgProc()
+	s.deprioritizedMissingDataPeriodicity = 1000
```
manish-sethi: Nit: Better to define a constant, with a comment

core/ledger/pvtdatastorage/kv_encoding.go
```diff
@@ -107,47 +108,45 @@ func decodeDataValue(datavalueBytes []byte) (*rwset.CollectionPvtReadWriteSet, e
...
-		keyBytes = append(keyBytes, nilByte)
-		return append(keyBytes, []byte(key.coll)...)
-	}
-
-	keyBytes := append(ineligibleMissingDataKeyPrefix, []byte(key.ns)...)
-	keyBytes = append(keyBytes, nilByte)
-	keyBytes = append(keyBytes, []byte(key.coll)...)
-	keyBytes = append(keyBytes, nilByte)
-	return append(keyBytes, []byte(encodeReverseOrderVarUint64(key.blkNum))...)
+func encodeElgMissingDataKey(group []byte, key *missingDataKey) []byte {
```
manish-sethi: The pattern that was followed intentionally in this package was that encoding/decoding various types of keys are contained in one function and everywhere else in the code, we work with the high level structs in order to keep the binary representation separated from high level code - just like you deal with the proto objects. Passing the `group` which is a `[]byte` from outside seem taking a step back where everywhere in the code we need to take care of passing the right prefix - which is part of the encoding knowledge.

Further, In the current mixed form, its very misleading (e.g., see function `decodeElgMissingDataKey` below, this is being returned with default value of `false`. I have highlighted one more usage below). 

One option could be to remove the field [isEligible](https://github.com/hyperledger/fabric/blob/ceb56c7caf50758c31f76f25c39335b093ae1edc/core/ledger/pvtdatastorage/store.go#L102) entirely and instead have explicit three functions for keys encoding/decoding. The functions that needs to operate on either priortized or depriortized data can take explicit bool flag instead of taking this `[]byte`.

core/ledger/pvtdatastorage/kv_encoding.go
```diff
@@ -184,44 +183,45 @@ func decodeCollElgVal(b []byte) (*CollElgInfo, error) {
...
 }
 
-func createRangeScanKeysForIneligibleMissingData(maxBlkNum uint64, ns, coll string) (startKey, endKey []byte) {
-	startKey = encodeMissingDataKey(
+func createRangeScanKeysForInelgMissingData(maxBlkNum uint64, ns, coll string) ([]byte, []byte) {
+	startKey := encodeInelgMissingDataKey(
 		&missingDataKey{
 			nsCollBlk:  nsCollBlk{ns: ns, coll: coll, blkNum: maxBlkNum},
 			isEligible: false,
 		},
```
manish-sethi: Just to highlight the redundancy at another place... calling the function `encodeInelgMissingDataKey` and still pass `isEligible: false` could be a source of confusion.

core/ledger/pvtdatastorage/kv_encoding_test.go
```diff
@@ -51,27 +51,39 @@ func TestDatakeyRange(t *testing.T) {
 	}
 }
 
-func TestEligibleMissingdataRange(t *testing.T) {
+func TestEligibleMissingDataRange(t *testing.T) {
```
manish-sethi, line 54: This change does not bother me much but it was not a camelcasing mistake... I had it deliberately (here and other places) to distinguish that it's  `Missingdata-Range` and not the `Missing-DataRange`. In other words, treating Missingdata a noun.

core/ledger/pvtdatastorage/reconcile_missing_pvtdata.go
```diff
@@ -21,91 +22,169 @@ import (
 //     from the above created data entries
 // (2) create a db update batch from the update entries
 // (3) commit the update batch to the pvtStore
-func (s *Store) CommitPvtDataOfOldBlocks(blocksPvtData map[uint64][]*ledger.TxPvtData) error {
+func (s *Store) CommitPvtDataOfOldBlocks(
+	blocksPvtData map[uint64][]*ledger.TxPvtData,
+	deprioritizedMissingData ledger.MissingPvtDataInfo,
```
manish-sethi: Typically, the name of the parameters should be meaningful from the caller point of view than the other way around. Here, `deprioritizedMissingData` does not make sense for the reconciler

core/ledger/pvtdatastorage/reconcile_missing_pvtdata.go
```diff
@@ -21,91 +22,169 @@ import (
...
+) error {
+	s.purgerLock.Lock()
+	defer s.purgerLock.Unlock()
+
```
manish-sethi, line 38: Though, it does not harm but isn't this a noop as per the current code? The data in this function is written as a single batch which is an atomic action in the leveldb and so is the iterator that is obtained by the purger routine. So, either this batch will be visible to the purger in entirety or not at all...no?

core/ledger/pvtdatastorage/reconcile_missing_pvtdata.go
```diff
@@ -114,115 +193,140 @@ func (p *oldBlockDataProcessor) constructExpiryKeyFromDataEntry(dataEntry *dataE
 	}, nil
 }
 
-func (p *oldBlockDataProcessor) getExpiryDataFromEntriesOrStore(entries *entriesForPvtDataOfOldBlocks, expiryKey expiryKey) (*ExpiryData, error) {
-	if expiryData, ok := entries.expiryEntries[expiryKey]; ok {
+func (p *oldBlockDataProcessor) getExpiryData(expKey expiryKey) (*ExpiryData, error) {
+	if expiryData, ok := p.entries.expiryEntries[expKey]; ok {
```
manish-sethi: This rename makes the function name short but is misleading. When I was reviewing the portion of the code where this gets invoked (at a single place), to clear the doubt that we are not over ridding the existing key in the map, I had to visit this function. Separate tiny functions help if they save distraction which was served earlier by the longer name of the function. In the current form, it rather creates a distraction. My suggestion would be to either have a full name or this can be made in-line.

core/ledger/pvtdatastorage/reconcile_missing_pvtdata.go
```diff
@@ -114,115 +193,140 @@ func (p *oldBlockDataProcessor) constructExpiryKeyFromDataEntry(dataEntry *dataE
...
 
-	expiryData, err := p.getExpiryDataOfExpiryKey(&expiryKey)
+	expData, err := p.db.Get(encodeExpiryKey(&expKey))
 	if err != nil {
 		return nil, err
 	}
-	return expiryData, nil
+	if expData == nil {
+		return nil, errors.Wrap(err, "error while getting expiry data from the store")
+	}
```
manish-sethi: either this..
```
	if err != nil || expData == nil{
		return nil, errors.Wrap(err, "error while getting expiry data from the store")
	}
```
or this...
```
	if err != nil {
		return nil, errors.Wrap(err, "error while getting expiry data from the store")
	}
	if expData == nil {
		return nil, nil
	}
```

core/ledger/pvtdatastorage/reconcile_missing_pvtdata.go
```diff
@@ -21,91 +22,169 @@ import (
...
+			// the pruge scheduler removed these entries and the
+			// associated data entry is no longer needed. Note
+			// that the associated missingData entry would also
+			// be not present. Hence, we can skip this data entry.
+			continue
+		}
+		expData.addPresentData(nsCollBlk.ns, nsCollBlk.coll, txNum)
+
+		p.entries.dataEntries[*dataEntry.key] = dataEntry.value
+		p.entries.expiryEntries[expKey] = expData
```
manish-sethi, line 116: Typically, if you override the existing key in a map, the lookup and checking for existence is close to the code. See related comment on function `getExpiryData`.

core/ledger/pvtdatastorage/reconcile_missing_pvtdata.go
```diff
@@ -21,91 +22,169 @@ import (
...
-		entries.add(dataEntry, expKey, expData, missingData)
+		deprioMissingData, err := p.getMissingDataFromDeprioritizedList(key)
+		if err != nil {
+			return err
+		}
+		if deprioMissingData != nil && deprioMissingData.Test(txNum) {
+			p.entries.deprioritizedMissingDataEntries[key] = deprioMissingData.Clear(txNum)
+		}
+		// if the missing data entry is already purged by the purge scheduler, we would
+		// get nil missingData from both prioritized and deprioritized list
```
manish-sethi: How does this happen? The data entries only contain the data that has not been yet expired. Given that the condition at line 130 should always be true.

core/ledger/pvtdatastorage/reconcile_missing_pvtdata.go
```diff
@@ -21,91 +22,169 @@ import (
...
-			// data entry is already expired
-			// and purged (a rare scenario)
+		if prioMissingData != nil && prioMissingData.Test(txNum) {
+			p.entries.prioritizedMissingDataEntries[key] = prioMissingData.Clear(txNum)
 			continue
 		}
-		missingData.Clear(uint(txNum))
 
-		entries.add(dataEntry, expKey, expData, missingData)
+		deprioMissingData, err := p.getMissingDataFromDeprioritizedList(key)
```
manish-sethi: Just to take a note, as we ask the data exclusively for either of the group in one cycle, when we refactor the reconciliation code to make it more of a pull model than a push, this can be made better by checking with one list only.

core/ledger/pvtdatastorage/reconcile_missing_pvtdata.go
```diff
@@ -21,91 +22,169 @@ import (
...
+			p.entries.deprioritizedMissingDataEntries[key] = deprioMissingData.Clear(txNum)
+		}
+		// if the missing data entry is already purged by the purge scheduler, we would
+		// get nil missingData from both prioritized and deprioritized list
+	}
+
+	return nil
+}
+
+func (p *oldBlockDataProcessor) prepareMissingDataEntriesToReflectPriority(deprioritizedList ledger.MissingPvtDataInfo) error {
```
manish-sethi, line 147: `deprioritizedList`  -> `toDeprioritize` carries a better meaning.

core/ledger/pvtdatastorage/helper.go
```diff
@@ -48,13 +50,29 @@ func prepareDataEntries(blockNum uint64, pvtData []*ledger.TxPvtData) []*dataEnt
...
+					ns:     nsColl.Namespace,
+					coll:   nsColl.Collection,
+					blkNum: committingBlk,
+				},
+			}
+
+			missingDataEntries = elgMissingDataEntries
+			if !nsColl.IsEligible {
+				missingDataEntries = inelgMissingDataEntries
+			}
```
manish-sethi: Though I prefer the approach of assigning the map to a common var based on type and then have a common flow (as, I have suggested at another place), however, here I find it too involved because the assignment in not one time. I believe that having a separate function `getOrCreateBitSet` will improve the readability.

core/ledger/pvtdatastorage/reconcile_missing_pvtdata.go
```diff
@@ -114,115 +200,144 @@ func (p *oldBlockDataProcessor) constructExpiryKeyFromDataEntry(dataEntry *dataE
...
+		return nil, nil
+	}
+
+	return decodeExpiryValue(expData)
 }
 
-func (p *oldBlockDataProcessor) getMissingDataFromEntriesOrStore(entries *entriesForPvtDataOfOldBlocks, nsCollBlk nsCollBlk) (*bitset.BitSet, error) {
-	if missingData, ok := entries.missingDataEntries[nsCollBlk]; ok {
-		return missingData, nil
+func (p *oldBlockDataProcessor) getMissingDataFromEntriesOrStore(group elgMissingDataGroup, nsCollBlk nsCollBlk) (*bitset.BitSet, error) {
```
manish-sethi: I feel that the code in this function can be made better readable if you have a single switch case...that initialize variables but then the rest of the flow is common and simple.

```
	var entries map[nsCollBlk]*bitset.BitSet
	var dbKey []byte

	switch group {
	case prioritized:
		entries = p.entries.prioritizedMissingDataEntries
		dbKey = encodeElgPrioMissingDataKey(&missingDataKey{
			nsCollBlk: nsCollBlk,
		})

	case deprioritized:
		entries = p.entries.deprioritizedMissingDataEntries
		dbKey = encodeElgPrioMissingDataKey(&missingDataKey{
			nsCollBlk: nsCollBlk,
		})
	}
```

core/ledger/pvtdatastorage/reconcile_missing_pvtdata.go
```diff
@@ -114,115 +200,144 @@ func (p *oldBlockDataProcessor) constructExpiryKeyFromDataEntry(dataEntry *dataE
...
+		for nsCollBlk, missingData := range missingDataList {
+			missingKey := &missingDataKey{
+				nsCollBlk: nsCollBlk,
+			}
+			switch group {
+			case "prioritized":
+				key = encodeElgPrioMissingDataKey(missingKey)
+			case "deprioritized":
+				key = encodeElgDeprioMissingDataKey(missingKey)
+			}
```
manish-sethi: Just like at the other place, I think that try to reuse the code in-place has caused compromising the readability. I believe that a s separate function with parameters `map[nsCollBlk]*bitset.BitSet` and the `encoding funciton` may make it better readable... otherwise, if you prefer to duplicate the code, even that would be fine.

core/ledger/pvtdatastorage/store_test.go
```diff
@@ -191,6 +191,72 @@ func TestStoreIteratorError(t *testing.T) {
...
+			1: {
+				{
+					Namespace:  "ns-1",
+					Collection: "coll-2",
+				},
+			},
+		},
+	}
+
+	deprioritizedMissingDataPeriodicity = 10
```
manish-sethi, line 249: since this is now a package level variable, better to move it to the TestMain. Otherwise, better to reset it in a defer statement.

core/ledger/pvtdatastorage/reconcile_missing_pvtdata_test.go
```diff
@@ -7,196 +7,569 @@ SPDX-License-Identifier: Apache-2.0
...
-	require.NoError(t, err)
+	for _, b := range blocksPvtData {
+		for _, dkey := range b.dataKeys {
+			require.False(t, testDataKeyExists(t, store, dkey))
+		}
+	}
+	// deprioritized list should not be present
+	p := &oldBlockDataProcessor{
+		Store: store,
+	}
```
manish-sethi: unnecessarily created. Can get `db` directly from store.

core/ledger/pvtdatastorage/reconcile_missing_pvtdata_test.go
```diff
@@ -7,196 +7,569 @@ SPDX-License-Identifier: Apache-2.0
...
+			blkNum: 1,
+		},
+	}
 
-	ns1Coll2Blk1Tx1 = &dataKey{nsCollBlk: nsCollBlk{ns: "ns-1", coll: "coll-2", blkNum: 1}, txNum: 1}
-	ns2Coll2Blk1Tx1 = &dataKey{nsCollBlk: nsCollBlk{ns: "ns-2", coll: "coll-2", blkNum: 1}, txNum: 1}
-	ns1Coll2Blk1Tx2 = &dataKey{nsCollBlk: nsCollBlk{ns: "ns-1", coll: "coll-2", blkNum: 1}, txNum: 2}
-	ns3Coll2Blk1Tx2 = &dataKey{nsCollBlk: nsCollBlk{ns: "ns-3", coll: "coll-2", blkNum: 1}, txNum: 2}
+	for _, k := range keys {
+		encKey := encodeElgPrioMissingDataKey(&missingDataKey{k})
```
manish-sethi: `encodeElgDeprioMissingDataKey`

core/ledger/pvtdatastorage/reconcile_missing_pvtdata_test.go
```diff
@@ -7,196 +7,569 @@ SPDX-License-Identifier: Apache-2.0
...
+		require.Nil(t, missingData)
+	}
+}
 
-	require.False(t, testDataKeyExists(t, store, ns1Coll2Blk1Tx1)) // purged
-	require.False(t, testDataKeyExists(t, store, ns2Coll2Blk1Tx1)) // purged
-	require.False(t, testDataKeyExists(t, store, ns1Coll2Blk1Tx2)) // purged
-	require.True(t, testDataKeyExists(t, store, ns3Coll2Blk1Tx2))  // never expires
+func TestCommitPvtDataOfOldBlocksWithDeprioritization(t *testing.T) {
+	sampleDataForTest := func() (map[uint64]*pvtDataForTest, ledger.MissingPvtDataInfo) {
```
manish-sethi: What is the value of this function? It just returns the static data.

PR #1722, opened 2016-06-04
Add noop system chain code. 

tamasblummer:
Add noop system chain code
## Description

Noop system chain code simply stores invoke transactions without validation. This chaincode is useful for testing. The noop system chaincode is prepared to work with fabric-api
## Motivation and Context

Preparation for performance tests.
## How Has This Been Tested?

run unit tests
## Checklist:


core/system_chaincode/noop/chaincode.go
```diff
@@ -0,0 +1,105 @@
...
+		var data = newCCIS.ChaincodeSpec.CtorMsg.Args[0]
+		var dataInByteForm, b64err = base64.StdEncoding.DecodeString(data)
+		if b64err != nil {
+			return nil, fmt.Errorf("Error in decoding from Base64:  %s", b64err)
+		}
+		return dataInByteForm, nil
+	default:
+		return nil, errors.New("Unsupported operation")
+	}
+
```
muralisrini: @adecaro : would Unmarshal on tx.Payload cause issues with security.privacy enabled ?

ghaskins: It would be good to have the answer in general. However, one thing I will point out is my impression is that this is an optional/testing harness so it doesn't necessarily need to be applicable to all situations to be useful.  @tamasblummer would you agree?

peer/core.yaml
```diff
@@ -335,6 +335,13 @@ ledger:
       # Deploy chaincodes into the genesis block
       chaincodes:
 
+        noop_syscc:
+          path: github.com/hyperledger/fabric/core/system_chaincode/noop
+          type: GOLANG
+          constructor:
+            args:
+
+
```
ghaskins: Do we want to have this enabled by default?  It seems that could be construed as a security risk since any transaction could be queried.  Or would all transactions committed to the ledger be considered public info anyway?

Either way, i think it would be more palatable if this was disabled by default.  Thoughts @muralisrini @tamasblummer

muralisrini: @ghaskins @tamasblummer I like the idea of disabling by default. Couple of other suggestions
- if we just test for successful return of GetTransactionByUUID (ie, not unmarshal etc) we don't have to worry about failures from unmarshal (such as questions dealing with decryption)
-  also perhaps not use a global var for ledger (the ledgerH var) but do GetLedger() every time ?

tamasblummer: @ghaskins data stored in ledger is available to any replica on disk and could be retrieved by any operator of a replica bypassing any authorisation. 
Yes, the noop chain code should not be enabled by default.

tamasblummer: @muralisrini this chain code does not mean to decrypt, it is plain access to data as is on the block chain.

ghaskins: @tamasblummer understood the data is available within the node environment and thus to many entities with access to the peer itself.  My concern is this seems to expose that data to remote clients as well without any obvious ACL considerations.  But since we are in agreement this is an optional and off by default component, I don't see the harm.

tamasblummer: @ghaskins it was a fallacy to assume any access protection of data on blockchain. If protection is needed for a data then it should not be on block chain or should be encrypted.  Relying on access control is not a part of this technology.

binhn:
@tamasblummer Since this scc is for testing, I'd suggest putting the code in bddtests/chaincode/go or bddtests/syschaincode/go, not in the base code, and default build should not pick up this code.

We need to think about build options for various needs, so that one may build with certain scc's or configuration features.

ghaskins:
@tamasblummer @binhn Any conclusion regarding Binh's comments above about the code location?

bramwelt:
reverify jenkins

tamasblummer:
@ghaskins @binhn could not find suitable means of passing in test specific configuration to behave test. @gabre is working on an other pull request that makes easy to specify configuration for tests, so this chain code can be moved to bddtests and also tested.

gabre:
@tamasblummer @ghaskins @binhn I have this "draft": https://github.com/gabre/fabric/commit/00b1a68a16a47a5d91e062324abf1c666b120224

gabre:
The check fails because the CI server starts behave tests not from bddtests (as I have done on my machine) but from some other location so it is unable to find the alternative core.yaml which would deploy noop syscc. The questions is how we could work this around in an elegant way. E.g. we could use absolute path for the config location path but... my opinion is that this wouldn't be the best... Any suggestions?

binhn:
Let's hold off merging this until @muralisrini provides a fix for #1741, which will change the genesis block spec in the yaml file, effecting this PR.  Thanks.

gabre:
There is a failing test: peer_basic.feature:516 which is a known bug as far as I know: https://github.com/hyperledger/fabric/issues/1820

christo4ferris:
@tamasblummer would you mind rebasing to resolve merge conflicts? Apologies for the delay in getting this merged.

core/system_chaincode/importsysccs.go
```diff
@@ -19,10 +19,18 @@ package system_chaincode
 import (
 	"github.com/hyperledger/fabric/core/system_chaincode/api"
 	//import system chain codes here
+	"github.com/hyperledger/fabric/bddtests/syschaincode/noop"
 )
 
 //see systemchaincode_test.go for an example using "sample_syscc"
-var systemChaincodes = []*api.SystemChaincode{}
+var systemChaincodes = []*api.SystemChaincode{
```
gabre, line 26: @tamasblummer As you can see, deployed system chaincodes are determined at build time, again. (this is a change from master)

peer/main.go
```diff
@@ -253,15 +253,22 @@ func main() {
 	viper.BindPFlag("peer.gomaxprocs", flags.Lookup("peer-gomaxprocs"))
 	viper.BindPFlag("peer.discovery.enabled", flags.Lookup("peer-discovery-enabled"))
 
+	var alternativeCfgPath = os.Getenv("PEER_CFG_PATH")
```
gabre, line 247: This part can be removed if we do not need modifiable configuration file locations.

gabre:
@christo4ferris I rebased it. But this introduces some problems which were already solved but now need some other solution. System chain codes need to be switchable/configurable but the current implementation uses chain codes wired into the source so it is determined at build time which ones will be deployed. This makes it very hard to have our noop chain code switched off by default and also test it using bdd tests (because a bdd test can't build its own fabric, it has to use a prebuilt one).

christo4ferris:
@binhn @muralisrini @tamasblummer @gabre please get together on this and develop an approach that allows syscc to be used in the manner that Tamas and team require. Maybe we need to have a third class e.g. inlinecc or something. Whatever. We need to work together on this. @mastersingh24

gabre:
I see the following possibilities:
- add a section to the configuration where system chain code can be enabled/disabled (as it was before)
- enable noop syscc by default (this would be not so good as it is not needed by everyone)

It is nearly sure that the first option is better. But why was system chain code configuration removed (from core.yaml)? @binhn @muralisrini @christo4ferris

mastersingh24:
Sorry for joining the party late here.  Which of the several problems do we want to solve first?
I've probably missed variations of this pull request due to changes in both the PR as well as the fabric.

Here are my observations:

1) I am no expert on Behave, but as far as I can tell it seems that all Behave tests use the various _compose_ files to start up one or more peers

2) All of those use the **_hyperledger/fabric-peer**_ as their Docker image

3) **_hyperledger/fabric-peer**_ appears to be built before ever running Behave tests so unless code / config is enabled / available by default in the src tree, I don't see how within the current structure of tests you would ever be able to test optional system chaincode within Behave unless the code is compiled as part of building the **_hyperledger/fabric-peer**_ image

ghaskins:
I still need to understand more about the options for syscc configuration in the latest code...but one comment I have related to @mastersingh24's last post is that if the issue is purely one of available docker image, we could trivially add a new one to the roster that is configured however this test needs it.  I could work with @gabre @tamasblummer if that is deemed the right direction.

muralisrini:
@gabre to first answer your question as to why sys chaincode was removed from the genesis block in the yaml, please see #1854. 

Given that syscc have to be "baked into" the system, registering programmatically makes it hardest to change peers and prevent divergence. However it does cause problems for creating special peers (such as for testing).

I'd second @ghaskins suggestion to create a special image with noops baked into it. Is that an option ?

Would it further help if we move the following from importsyscc.go into its own file (say "systemchaincodes.go") ?

```
//see systemchaincode_test.go for an example using "sample_syscc"
var systemChaincodes = []*api.SystemChaincode{}
```

It would separate the defined sys cc's into its own file so you can simply replace systemchaincodes.go file before image generation without modifying importsyscc.go itself. 

(As an aside, I miss "class loading" at times....)

ghaskins:
@muralisrini I understand the motivation to get rid of the ctor from the yaml and to make things more flexible than requiring genesis inclusion...but I am wondering if, given golang's relative inflexibility in this regard, we would be best served to still retain a notion of something like a whitelist in the yaml.  Whereby syscc can be compiled in and registered, but it will only be loaded if the whitelist includes it.  That way, local admins can decide what they want to include and we can opt to disable certain testing-related things by default.  Thoughts?

gabre:
@ghaskins @muralisrini "a whitelist in the yaml" From my point of view (without a total understanding of the details) this would be the same as it was before: importsyscc (syscc baked into the peer binary) + yaml **together** decide, which code will be deployed. This way, noop syscc could be included in importsyscc but it would not be included in that yaml/whitelist. This would make it a lot easier to test (This is very similar to the way it worked before @mastersingh24 - this is how I tested it) as I have already implemented some functionality to make the config yaml replacable if it is needed (I used an ENV for that as that was the easiest way to solve this problem).

mastersingh24:
OK - this is going in circles for no real reason.  Totally agree that we should figure out some better options for system chaincode inclusion and deployment moving forward, but given that we have to stop making tactical changes, is there any reason why we really have to have BDD tests for this at all?  If it passes the unit tests, what purpose do the BDD tests serve in this case?

christo4ferris:
agree with @mastersingh24 on principle, but the problem is that it will break all subsequent builds. Is there a specific behave test(s) we can disable to get this merged?

tamasblummer:
@mastersingh24 @christo4ferris We should not include this or any other system chain code without an explicit whitelisting in configuration. Whitelisting would also solve the problem of testing, provided configuration can be read from an argument instead the current rather complicated on-the-fly-edit process.

mastersingh24:
@tamasblummer   -  just to doublecheck - so you want to make sure that even if the system chaincode is compiled in the peer executable that it is not actually "registered" with the system unless explicitly in a whitelist?

That seems okay to me.  This should be outside the previous list which was used as part of the genesis  block (which is I think why it was removed)

@gabre  - is this something you can do?  sounds pretty straightforward and looks like you know how to do it

tamasblummer:
@mastersingh24 yes that is what I mean. User should not be exposed to a chaincode just because it is compliled in (that is a technical necessity) but only if it is explicitly whitelisted.

The config needs rework, such that it is easy to start a peer with different config (as argument) and not do some on-the-fly edit of a single config as is now. This however is stuff for yet another pull request. For now please fix however you feel fit temporarily, provided we get this merged.

christo4ferris:
Agree config and so forth can be separate PR; but we need to resolve the build issues... can we disable the behave tests that are failed by this so that post merge we don't have everyone with a failing build? Also, we need an issue to track so we can be sure to restore those skipped tests before we cut another release.

muralisrini:
I'm good with whitelist too. For production envs it seemed registering via code provided strictest way to control changes but it does make it hard for dev/test. Also when we enable system chaincode for security (#1964) we could factor in the whitelist.

@cbf just took a quick look at the failures.... while agree we should disable the behave tests while working on whitelist, none of the failures above are related to that I think.

The travis seems to have stopped during build time at

```
g++ -DHAVE_CONFIG_H -I. -I..  -I./../gmock/gtest/include -I./../gmock/include  -pthread -DHAVE_PTHREAD=1  -Wall -Wwrite-strings -Woverloaded-virtual -Wno-sign-compare -g -DNDEBUG -MT google/protobuf/stubs/protobuf_test-stringpiece_unittest.o -MD -MP -MF google/protobuf/stubs/.deps/protobuf_test-stringpiece_unittest.Tpo -c -o google/protobuf/stubs/protobuf_test-stringpiece_unittest.o `test -f 'google/protobuf/stubs/stringpiece_unittest.cc' || echo './'`google/protobuf/stubs/stringpiece_unittest.cc
In file included from ./../gmock/gtest/include/gtest/gtest.h:1929:0,
                 from google/protobuf/stubs/stringpiece_unittest.cc:40:
google/protobuf/stubs/stringpiece_unittest.cc: In member function 'virtual void google::protobuf::{anonymous}::StringPiece_ComparisonOperators_Test::TestBody()':

```

The jenkins failures have the same signature we saw when running unit tests in parallel (which should have been fixed in PR #1930) 

```

07:29:52 panic: Could not open openchain db error = [IO error: lock /var/hyperledger/test/tmpdb/db/LOCK: Resource temporarily unavailable] [recovered]
07:29:52    panic: Could not open openchain db error = [IO error: lock /var/hyperledger/test/tmpdb/db/LOCK: Resource temporarily unavailable]
07:29:52 
```

gabre:
I have temporarily removed the test and added https://github.com/hyperledger/fabric/issues/1967

gregmisiorek:
Are Jenkins check errors to be expected?

srderson:
Yes, Jenkins is still a work in progress. It's best to rely on Travis results for now. I'm not sure why Travis is not showing up in this PR. @rameshthoomu do you have any ideas?

rameshthoomu:
@srderson : I am working on this. Will let you know the reason.

srderson:
Discussed PR this morning in the Fabric maintainers issue. Two issues
1. No documentation
2. How is the system chaincode enabled/disabled? Is this part of the build process or is this a whitelist? @muralisrini and @ghaskins are investigating. Do we need a new issue opened?

gabre:
@srderson @muralisrini @tamasblummer I have rebased the change and added a whitelist functionality. Please check the md file I added (and the directory structure). Any comments, suggestions are welcome.

srderson:
Waiting for @muralisrini to review whitelist functionality.

peer/main.go
```diff
@@ -244,15 +244,22 @@ func main() {
 	flags := nodeStartCmd.Flags()
 	flags.BoolVarP(&chaincodeDevMode, "peer-chaincodedev", "", false, "Whether peer in chaincode development mode")
 
+	var alternativeCfgPath = os.Getenv("PEER_CFG_PATH")
+	if alternativeCfgPath != "" {
```
muralisrini, line 248: @gabre is PEER_CFG_PATH  set anywhere ?

gabre, line 248: @muralisrini When I added https://github.com/hyperledger/fabric/issues/1967, I removed the behave test of noop that used PEER_CFG_PATH (in order to load a custom config with noop syscc enabled, but this functionality was removed in master, see the comment history). If the whitelist functionality is OK, I think I can mark the issue as solved and readd the noop bdd test. What is your opinion? If you agree, I readd the test (where PEER_CFG_PATH will be set and a custom config will be used with noop enabled) and the PR can be merged.

gabre:
reverify jenkins

peer/main.go
muralisrini, line 248: @gabre I think the whitelist function is ok for now.  Please do readd the noop bdd test and mark the issue resolved.

I think the other option for the whitelist is to keep a separate file in the system chaincode folder just for the whitelist. Will add a comment and an issue for that. That can be treated separate from this PR.

muralisrini:
@gabre LGTM with a proviso.

As mentioned in the inline comment, it might make sense to create a separate file for the whiltelist to be read in at startup as opposed to making it a part of core.yaml along with other properties. File content could be used in implementation of #1964.

srderson:
Thanks @gabre and @muralisrini! LGTM

PR #1726, opened 2016-06-06
create a chaincode image if not found

### LINKED ISSUE #1603: chaincode image needs to be created after state transfer of new deploy transaction
## Description

When state transfer obtains a new block that contains a deploy transaction, the chaincode subsystem needs to create a corresponding docker image.
## Describe How to Reproduce

See #1472

@muralisrini @jyellick

muralisrini:
Pass the chaincode docker targz extracted from deploy transaction to launch of chaincode. The targz will be used to create the image if not found.
## Description

We have the targz at hand already quite conveniently. The "Start" interface is modified to take this is as a parameter. If the start of a container fails with ErrNoSuchImage, the targz is ued to create the image and retry start. As this happens under the aegis of "Start" it is atomic to the "Start" operation.
## Motivation and Context

There are few legitimate situations where a deployed chaincode's image is not found
- a peer that is catching up
- docker error conditions

The image is created on demand only when an invoke or query is issued against the chaincode. Note that this has no adverse performance effect under "normal" conditions.

Fixes #1603 
## How Has This Been Tested?

Introduced a "Destroy" command to the VM just so we could destroy an image for testing. Used this in an existing "ExecuteInvokeTransaction" test case as opposed to introducing a new test case (the chaincode test suite is quite long already). The test case deploys a chaincode, stops the container, deletes the image and then attempts to invoke. The invoke will cause the image to be created first.
## Checklist:

### DIFF
--- a/core/chaincode/chaincode_support.go
+++ b/core/chaincode/chaincode_support.go
@@ -259,8 +259,8 @@ func (chaincodeSupport *ChaincodeSupport) getArgsAndEnv(cID *pb.ChaincodeID) (ar
 	return args, envs, nil
 }
 
-// launchAndWaitForRegister will launch container if not already running
-func (chaincodeSupport *ChaincodeSupport) launchAndWaitForRegister(ctxt context.Context, cds *pb.ChaincodeDeploymentSpec, cID *pb.ChaincodeID, uuid string) (bool, error) {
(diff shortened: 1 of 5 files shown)


binhn:
LGTM - thanks.

muralisrini:
Opened a new PR #1747 after resolving conflicts.

Issue #1701, opened 2016-06-03
Inconsistent state after writing transactions to multiple peers concurrently


yingfeng:
## Description

According to the definition of consensus, each fabric peer should have a consistent state at all time, or else, the decentralized ledger could not work at all. However, after I setup a fabric network, and perform the stress test, I've found the state of all fabric peer nodes are inconsistent at all.
## Describe How to Reproduce
- Setup N fabric peer nodes, and I've tried N=2, N=3, N=4, each of the peer adopts the default consensus plugin---classic PBFT
- Deploy a chaincode `chaincode_example02`:

```
{
  "jsonrpc": "2.0",
  "method": "deploy",
  "params": {
    "type": 1,
    "chaincodeID":{
        "path":"github.com/hyperledger/fabric/examples/chaincode/go/chaincode_example02"
    },
    "ctorMsg": {
        "function":"init",
        "args":["a", "1000", "b", "2000"]
    }
  },
  "id": 1
}
```
- The initial state of each peer is consistent:

```
{
  "jsonrpc": "2.0",
  "method": "query",
  "params": {
    "type": 1,
    "chaincodeID":
    {
        "name":"7b07c59e9b9405c1aef33493b63b9a766d9bb836989ded1730052de650aa8ce5654274d148ceff96a4e5bd43bca26aba099f55c400e4befdc8b2ee4c0a94e30b",
        "path":"github.com/hyperledger/fabric/examples/chaincode/go/chaincode_example02"
    },
    "ctorMsg": {
        "function":"query",
        "args":["a"]
    }
  },
  "id": 5
}
```

The initial status for all peer nodes is:

```
{"jsonrpc":"2.0","result":{"status":"OK","message":"1000"}
```
- Perform stress invoking tests to two peer nodes concurrently, say peer2 and peer4:

```
{
  "jsonrpc": "2.0",
  "method": "invoke",
  "params": {
    "type": 1,
    "chaincodeID":
    {
        "name":"7b07c59e9b9405c1aef33493b63b9a766d9bb836989ded1730052de650aa8ce5654274d148ceff96a4e5bd43bca26aba099f55c400e4befdc8b2ee4c0a94e30b",
        "path":"github.com/hyperledger/fabric/examples/chaincode/go/chaincode_example02"
    },
    "ctorMsg": {
        "function":"approach",
        "args":["a", "1000", "b", "2000"]
    }
  },
  "id": 3
}
```
[...]

PR #1689, opened 2016-06-02
Remove hacky sendViewChangedEvent

jyellick:
## Description

This is a follow up to PR #1675.  It removes the hacky mechanism for generating `viewChangedEvent` events introduced, and causes the view change code to pass back the events directly.  It moves the old `innerStack` `viewChange` method into the legacy structures.
## Motivation and Context

This is one in the long series of PRs trying to turn PBFT into more of a simple state machine, by removing the threading, and causing all state changes to be event driven.

The old code was pushing execution into the plugin layer directly by sending the thread into the plugin code.  This changeset removes this tight coupling via the `viewChange()` method by removing it from the `innerStack` interface and into the `legacyInnerStack` interface.  It also removes the `viewChange` empty implementation from `obc-batch.go` as it now consumes only the event.
## How Has This Been Tested?

This has been tested locally via the unit tests.  The remainder should be covered by CI. As this is code restructuring, with no new function, no new tests are required.
## Checklist:
- [X] I have added a [Signed-off-by](https://github.com/hyperledger/fabric/blob/master/CONTRIBUTING.md#legal-stuff)
- [X] Either no new documentation is required by this change, OR I added new documentation
- [X] Either no new tests are required by this change, OR I added new tests
- [X] I have run [goimports](https://godoc.org/golang.org/x/tools/cmd/goimports), [go vet](https://golang.org/cmd/vet/), and [golint](https://github.com/golang/lint). I have cleaned up all valid errors and warnings in code I have added or modified. These tools may generate false positives. Don't be worried about ignoring some errors or warnings. The goal is clean, consistent, and readable code.

Signed-off-by: Jason Yellick jyellick@us.ibm.com

Issue #752, opened 2016-02-29
Error handling and panics


tuand27613:
This started out as an email thread but I'm creating an issue so we can discuss some of the design questions that came up.

tuand27613:
comments from @corecode 

We could just rename it to "ASSERT()".

It seems we're lacking a user friendly way of aborting if something is wrong (e.g. config file is inconsistent).

On a broader level, the project overall should define what constitutes an error, and how errors need to be handled.  Right now most functions return an error, even though for some of those function no error can ever happen, or the error would be fatal.  Ideally we would define for every function which conditions cause an error, and whether the error can be transient and the operation can be retried later.

For example:  if GetLedger() fails, what are we supposed to do?  To me, that's a fatal condition, if you can't get the ledger.  Something must be seriously wrong.

Or, what does it mean that Broadcast() failed?  Did one peer Send() not succeed?  Are there no peers?  Did no Send() succeed?  Right now, consensus tends to return these errors, when the called function is the last function in the consensus function.  But essentially, during operation, consensus returning an error does not make any sense.

tuand27613:
more from @corecode 

what we should do if:
1. there is a configuration error, which prevents us from operating (e.g. number of nodes is negative)
2. there is a programming error (e.g. invalid argument or return value, which should have been prevented by the interfacing code).

Typically I would handle (1) by printing an error and exiting the application, or using a generic function somewhere which does exactly that.
I would handle (2) by adding ASSERT()s, which then turn into panic() or some other form of abort.
I guess this would apply across OBC

kchristidis:
@binhn Can we have your take on this one? Specifically @corecode's comments ([one](https://github.com/openblockchain/obc-peer/issues/752#issuecomment-190377184), [two](https://github.com/openblockchain/obc-peer/issues/752#issuecomment-190377674)).

binhn:
For recoverable conditions, obviously we don't want to abort the peer. 

For irrecoverable conditions,  we should recycle the peer.  We know some conditions, such as the peer being compromised, we may not want to recycle, but kill it and wait for service. Obviously we need ability to notify someone about this situation.

To recycle, I found this [Graceful Restart in Golang](http://grisha.org/blog/2014/06/03/graceful-restart-in-golang) that could help.

bmos299:
@movee2016 please see @kchristidis on this issue.

tuand27613:
will wait until @jyellick gets his arms around #751.

[...]

PR #1734, opened 2016-06-06
Move pbft batch execution off main thread

jyellick:
## Description

This changeset moves the actual fabric execution off of the main event thread and onto its own thread so that PBFT can continue to process messages while execution is being performed.

Note that this is only a single commit stacked on top of PR #1689 so please review only the diff for the latest commit when considering this PR (assuming #1689 is likely to be merged in the near future).
## Motivation and Context

The execution of certain transactions, in particular chaincode deploy transactions, can take relatively long amounts of time (multiple seconds).  If the main event thread which is responsible for processing new messages blocks while executing transactions, the message queues can fill up, and messages might be discarded.  Although it's unclear if this is the root of the problem witnessed in #1701, discarding messages unnecessarily can cause thrashing in the system, and seriously degrade the PBFT's performance.

As a brief history of this code path for context in those wishing to understand this changeset:
1. The slowness of deploy transactions introduced the necessity that they execute on their own thread, so executions were performed on their own go routine.
2. In order to detect when requests were being censored, a complaints mechanism was introduced, which required hooking into the execution path to remove transactions from custody as they were processed.
3. Race conditions were encountered between the execution go routine and the event routine as they both attempted to concurrently modify the same data structures.
4. To eliminate the race, the execution go routine queued an event for the main event go routine to perform the execution, which returns us to the original problem from (1).

The solution was to eliminate the execution go routine from PBFT core, and have it call into the plugin synchronously.  This allows the plugin to manipulate the custody data structure safely in a non-concurrent way, then spawn the go routine itself which performs the actual execution, and queues an execution completed event once done.
## How Has This Been Tested?

This changeset changes only the execution flow, not the actual executed code, so existing tests should be sufficient.  This has been tested locally with unit tests, and CI will cover the remainder.
## Checklist:

### DIFF
--- a/consensus/obcpbft/obc-classic.go
+++ b/consensus/obcpbft/obc-classic.go
@@ -129,23 +129,25 @@ func (op *obcClassic) validate(txRaw []byte) error {
 
 // execute an opaque request which corresponds to an OBC Transaction
 func (op *obcClassic) execute(seqNo uint64, txRaw []byte) {
-	tx := &pb.Transaction{}
-	err := proto.Unmarshal(txRaw, tx)
-	if err != nil {
-		logger.Error("Unable to unmarshal transaction: %v", err)
-		return
-	}
+	go func() {
+		tx := &pb.Transaction{}
+		err := proto.Unmarshal(txRaw, tx)
+		if err != nil {
+			logger.Error("Unable to unmarshal transaction: %v", err)
+			return
+		}
 
-	meta, _ := proto.Marshal(&Metadata{seqNo})
+		meta, _ := proto.Marshal(&Metadata{seqNo})
 
-	id := []byte("foo")
-	op.stack.BeginTxBatch(id)
-	result, err := op.stack.ExecTxs(id, []*pb.Transaction{tx})
-	_ = err    // XXX what to do on error?
-	_ = result // XXX what to do with the result?
-	_, err = op.stack.CommitTxBatch(id, meta)
+		id := []byte("foo")
+		op.stack.BeginTxBatch(id)
+		result, err := op.stack.ExecTxs(id, []*pb.Transaction{tx})
+		_ = err    // XXX what to do on error?
+		_ = result // XXX what to do with the result?
+		_, err = op.stack.CommitTxBatch(id, meta)
 
-	op.pbft.execDone()
+		op.pbft.execDone()
+	}()
 }
 
 // called when a view-change happened in the underlying PBFT
--- a/consensus/obcpbft/events.go
+++ b/consensus/obcpbft/events.go
@@ -303,9 +303,6 @@ type viewChangedEvent struct{}
 // complaintEvent is sent when custody has a complaint
 type complaintEvent custodyInfo
 
-// batchExecEvent is sent when a batch execution should take place
-type batchExecEvent execInfo
-
 // returnRequestEvent is sent by pbft when we are forwarded a request
 type returnRequestEvent *Request
 
--- a/consensus/obcpbft/obc-batch.go
+++ b/consensus/obcpbft/obc-batch.go
@@ -194,13 +194,6 @@ func (op *obcBatch) validate(txRaw []byte) error {
 
 // execute an opaque request which corresponds to an OBC Transaction
 func (op *obcBatch) execute(seqNo uint64, raw []byte) {
-	op.manager.queue() <- batchExecEvent{
-		seqNo: seqNo,
-		raw:   raw,
-	}
-}
-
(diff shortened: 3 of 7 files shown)


corecode:
Great!

tuand27613:
LGTM

consensus/obcpbft/obc-classic.go
```diff
@@ -129,23 +129,25 @@ func (op *obcClassic) validate(txRaw []byte) error {
...
+			logger.Error("Unable to unmarshal transaction: %v", err)
+			return
+		}
+
+		meta, _ := proto.Marshal(&Metadata{seqNo})
+
+		id := []byte("foo")
+		op.stack.BeginTxBatch(id)
+		result, err := op.stack.ExecTxs(id, []*pb.Transaction{tx})
+		_ = err    // XXX what to do on error?
```
binhn, line 145: if err,  RollbackTxBatch ?

jyellick, line 145: Why would we roll back the entire batch of transactions if we encountered a single error?

There has been some substantial debate about this, but today, the path is that we record all the transactions, and accompanying errors in the block.  There has been the argument, that we should not recorded errored transactions, but this should be implemented away from consensus.  (It is currently in helper, which is generally an odd place to be making decisions about the ledger contents).

In general, I agree that we need a stronger contract from the consensus API, "when an error is received, it means X and you should do Y", but today, it's not clear that rolling back is correct, and if we were to rollback, what should be the next step? Should we retry?  Should we discard? Should we panic? There's no clear recovery path, which is why this is flagged in the code as a trouble spot.

Please note, that this is not new logic for this PR.

binhn:
@jyellick I don't think that err is a single error from a tx, but I'd like @muralisrini to take a look as well.

If you look at the chaincode.ExecuteTransactions, the err is actually fatal from Ledger.  Also look at helper.ExecTxs to see how it accum's tx's errors on to the block.

jyellick:
@binhn You're correct, I may have been thinking of some older logic or just been wrong.  Regardless, I'm still not sure what the right response to this error is, if the chaincode cannot access the ledger, then what is the recourse from consensus?  We could try to rollback, but as you say, this seems fatal with no clear resolution.

jyellick:
The CI failure also looks like there was a problem in bootstrapping, not finding some rocksdb lib.  Is there a way for the submitter to re-run the CI without pushing a bogus commit?

binhn:
@manish-sethi @srderson @muralisrini 
What should we do when `ExecuteTransactions` in `exectransaction.go` returns `err` due to `ledger.GetLedger()` returns error ?

manish-sethi:
@binhn I think that the error (`ledger.GetLedger()` in the method `ExecuteTransactions` in `exectransaction.go`) should raise a panic. The panic can be raised within ledger module - this was discussed in some other PR (where it was being discussed to revisit and have a standard for all the error handlings). Similarly, `lgr.GetTempStateHash()` (the next step in the same method) could also cause an error, again - mostly likely due to an error received from the db and should have a similar handling. Because, these are fundamentally related to functioning of a peer so, the situation is as bad as peer is not up. 

If you think that `panic` is too "harsh", then at least, we can maintain some peer level `state` that can be marked as `unavailable` with the `error` causing this state.

jyellick:
@manish-sethi @binhn It sounds like we should open an issue to eliminate error feedback from this call or otherwise handle this fatal condition?

As this behavior is not new to this PR, I'd request that this be merged in the interim.

muralisrini:
There are 3 actors

```
chaincode.ExecuteTransactions (provides txerr array, system error) ----called by ---->

(ignores txerr) consensus (pbft, noops) (returns above system error) ----called by --->

(ignore system error) consensus.helper.ExecTxs 
```

The "system error" is ledger error. It could be reasonable to raise a panic there as @manish-sethi points out but it does not do away with the problem of what to do with the returned error (there could be others in future besides ledger errors).... unless we don't return an error at all.

Seems to me error handling has to be decided at the consensus level.

For now we could have chaincode.ExecuteTransactions return an ErrLedger which consensus tests for and raises a panic there in helper.ExecTxs. At least then we are "handling" error there and will _have_ to deal with other errors if they arise.

binhn:
@manish-sethi I have suggested a while back to have an ability to recycle a peer, but I guess that has not been implemented, so we should solve it separately from this PR. 
@muralisrini Let's not solve this here with ErrLedger but move to another issue to address fabric error handling.
@jyellick I agreed to merge and have another issue to handle errors for fabric in general.

Thanks.

kchristidis:
The issue to handle errors for fabric in general is: https://github.com/hyperledger/fabric/issues/752

PR #1737, opened 2016-06-07
Fix #1660 to use sync.once to protect db singleton

### LINKED ISSUE #1660: Create OpenchainDB not safe when invocation in multiple goroutines
## Description

In db.go the singleton openchainDB just protect by a package variable isOpen, it should be not safe when invocation by multiple goroutines.
## Describe How to Reproduce

In chaincode, invoke persist.go in different goroutines.

GrapeBaBa:
## Description

Use sync.once to promise db singleton safely and clean up db code.
## Motivation and Context

Fixes #1660 
1.Make db singleton safely.
2.Clean up db create and open code.
## How Has This Been Tested?

make unit-test successfully 
## Checklist:

### DIFF
--- a/core/db/db.go
+++ b/core/db/db.go
@@ -22,6 +22,7 @@ import (
 	"os"
 	"path"
 	"strings"
+	"sync"
 
 	"github.com/op/go-logging"
 	"github.com/spf13/viper"
@@ -44,6 +45,13 @@ var columnfamilies = []string{
 	persistCF,    // persistent per-peer state (consensus)
 }
 
+type dbState int32
+
+const (
+	closed dbState = iota
+	opened
+)
+
 // OpenchainDB encapsulates rocksdb's structures
 type OpenchainDB struct {
 	DB           *gorocksdb.DB
@@ -52,59 +60,27 @@ type OpenchainDB struct {
 	StateDeltaCF *gorocksdb.ColumnFamilyHandle
 	IndexesCF    *gorocksdb.ColumnFamilyHandle
 	PersistCF    *gorocksdb.ColumnFamilyHandle
+	dbState      dbState
+	mux          sync.Mutex
 }
 
 var openchainDB *OpenchainDB
-var isOpen bool
+var once sync.Once
 
-// CreateDB creates a rocks db database
-func CreateDB() error {
+// Create create an openchainDB instance
+func Create() *OpenchainDB {
 	dbPath := getDBPath()
-	dbLogger.Debugf("Creating DB at [%s]", dbPath)
-	missing, err := dirMissingOrEmpty(dbPath)
-	if err != nil {
-		return err
-	}
-
-	if !missing {
-		return fmt.Errorf("db dir [%s] already exists", dbPath)
-	}
-	err = os.MkdirAll(path.Dir(dbPath), 0755)
-	if err != nil {
-		dbLogger.Errorf("Error calling  os.MkdirAll for directory path [%s]: %s", dbPath, err)
-		return fmt.Errorf("Error making directory path [%s]: %s", dbPath, err)
-	}
-	opts := gorocksdb.NewDefaultOptions()
-	defer opts.Destroy()
-	opts.SetCreateIfMissing(true)
-
-	db, err := gorocksdb.OpenDb(opts, dbPath)
-	if err != nil {
-		return err
-	}
-
-	defer db.Close()
-
-	dbLogger.Debugf("DB created at [%s]", dbPath)
-	return nil
+	mkDirIfNotExist(dbPath)
+	return &OpenchainDB{dbState: closed}
 }
 
-// GetDBHandle returns a handle to OpenchainDB
+// GetDBHandle get an opened openchainDB singleton
 func GetDBHandle() *OpenchainDB {
-	var err error
-	if isOpen {
-		return openchainDB
-	}
+	once.Do(func() {
+		openchainDB = Create()
+	})
+	openchainDB.Open()
 
-	err = createDBIfDBPathEmpty()
-	if err != nil {
-		panic(fmt.Sprintf("Error while trying to create DB: %s", err))
-	}
-
-	openchainDB, err = openDB()
-	if err != nil {
-		panic(fmt.Sprintf("Could not open openchain db error = [%s]", err))
-	}
 	return openchainDB
 }
 
@@ -172,32 +148,23 @@ func getDBPath() string {
 	return dbPath + "db"
 }
 
-func createDBIfDBPathEmpty() error {
-	dbPath := getDBPath()
-	missing, err := dirMissingOrEmpty(dbPath)
-	if err != nil {
-		return err
-	}
-	dbLogger.Debugf("Is db path [%s] empty [%t]", dbPath, missing)
-	if missing {
-		err := CreateDB()
-		if err != nil {
-			return nil
-		}
-	}
-	return nil
-}
-
-func openDB() (*OpenchainDB, error) {
-	if isOpen {
-		return openchainDB, nil
+// Open open underlying rocksdb
+func (openchainDB *OpenchainDB) Open() {
+	openchainDB.mux.Lock()
+	if openchainDB.dbState == opened {
+		openchainDB.mux.Unlock()
+		return
 	}
 
+	defer openchainDB.mux.Unlock()
+
 	dbPath := getDBPath()
+	mkDirIfNotExist(dbPath)
+
 	opts := gorocksdb.NewDefaultOptions()
 	defer opts.Destroy()
 
-	opts.SetCreateIfMissing(false)
+	opts.SetCreateIfMissing(true)
 	opts.SetCreateIfMissingColumnFamilies(true)
 
 	cfNames := []string{"default"}
@@ -210,23 +177,34 @@ func openDB() (*OpenchainDB, error) {
 	db, cfHandlers, err := gorocksdb.OpenDbColumnFamilies(opts, dbPath, cfNames, cfOpts)
 
 	if err != nil {
-		fmt.Println("Error opening DB", err)
-		return nil, err
+		panic(fmt.Sprintf("Error opening DB: %s", err))
 	}
-	isOpen = true
-	// XXX should we close cfHandlers[0]?
-	return &OpenchainDB{db, cfHandlers[1], cfHandlers[2], cfHandlers[3], cfHandlers[4], cfHandlers[5]}, nil
+
+	openchainDB.DB = db
+	openchainDB.BlockchainCF = cfHandlers[1]
+	openchainDB.StateCF = cfHandlers[2]
+	openchainDB.StateDeltaCF = cfHandlers[3]
+	openchainDB.IndexesCF = cfHandlers[4]
+	openchainDB.PersistCF = cfHandlers[5]
+	openchainDB.dbState = opened
 }
 
-// CloseDB releases all column family handles and closes rocksdb
-func (openchainDB *OpenchainDB) CloseDB() {
+// Close releases all column family handles and closes rocksdb
+func (openchainDB *OpenchainDB) Close() {
+	openchainDB.mux.Lock()
+	if openchainDB.dbState == closed {
+		openchainDB.mux.Unlock()
+		return
+	}
+
+	defer openchainDB.mux.Unlock()
 	openchainDB.BlockchainCF.Destroy()
 	openchainDB.StateCF.Destroy()
 	openchainDB.StateDeltaCF.Destroy()
 	openchainDB.IndexesCF.Destroy()
 	openchainDB.PersistCF.Destroy()
 	openchainDB.DB.Close()
-	isOpen = false
+	openchainDB.dbState = closed
 }
 
 // DeleteState delets ALL state keys/values from the DB. This is generally
@@ -377,4 +355,19 @@ func makeCopy(src []byte) []byte {
 	dest := make([]byte, len(src))
 	copy(dest, src)
 	return dest
+}
+
+func mkDirIfNotExist(dbPath string) {
+	missing, err := dirMissingOrEmpty(dbPath)
+	if err != nil {
+		panic(fmt.Sprintf("Error while trying to open DB: %s", err))
+	}
+	dbLogger.Debugf("Is db path [%s] empty [%t]", dbPath, missing)
+
+	if missing {
+		err = os.MkdirAll(path.Dir(dbPath), 0755)
+		if err != nil {
+			panic(fmt.Sprintf("Error making directory path [%s]: %s", dbPath, err))
+		}
+	}
 }
--- a/core/db/db.go
+++ b/core/db/db.go
@@ -157,12 +157,23 @@ func (openchainDB *OpenchainDB) Open() {
 	defer openchainDB.mux.Unlock()
 
 	dbPath := getDBPath()
-	mkDirIfNotExist(dbPath)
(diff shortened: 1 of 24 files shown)


vendor/google.golang.org/grpc/stream.go
```diff
@@ -149,7 +149,7 @@ type clientStream struct {
 
 	tracing bool // set to EnableTracing when the clientStream is created.
 
-	mu sync.Mutex
```
corecode, line 152: Probably should modify vendor files

examples/chaincode/chaintool/example02/src/chaincode/chaincode_example02.go
```diff
@@ -117,7 +117,7 @@ func (t *ChaincodeExample) CheckBalance(stub *shim.ChaincodeStub, param *example
 
 func main() {
 	self := &ChaincodeExample{}
-	interfaces := ccs.Interfaces {
```
corecode: I have a feeling this will break some other chaincode

GrapeBaBa: @corecode this is a go fmt change, what do you mean break other chaincode?

vendor/google.golang.org/grpc/stream.go
GrapeBaBa, line 152: @corecode this is go fmt plugin change, i just notice it is in vendor catelog, did it need change?

binhn:
@manish-sethi Could you review. Thanks.

core/db/db.go
```diff
@@ -52,59 +60,25 @@ type OpenchainDB struct {
...
-	}
-
-	err = createDBIfDBPathEmpty()
-	if err != nil {
-		panic(fmt.Sprintf("Error while trying to create DB: %s", err))
-	}
+	once.Do(func() {
+		openchainDB = Create()
+	})
+	openchainDB.Open()
```
manish-sethi, line 76: - A call to `GetDBHandle()`would acquire two locks. Can you make it one instead. `Create` alone is not of any use.
- Can you check where else in the code, this method has been used and the db handle can be  cached. Till now, this method was not synchronized, hence caching was not meaningful.

@binhn this is yet another example where a singleton can be created in `init` function and would not require any locks. (Previously, we saw the similar example where it would have been better to reading configurations via `Viper` and cache in the `init`). However, the issue is that `init(s)` are invoked before `main` (and in the current code, `main` function loads the yaml file). Would it make sense to have a separate module for config and in the init of that module load yaml. This module would need wrapper apis that other modules would use instead of viper apis for getting configurations (in order to make other modules dependent on this configuration loading module so that it's init will be invoked before others. The benefit would be that we would not require locking for singletons but the downside is having a wrapper over viper APIs.

core/db/db_test.go
```diff
@@ -31,246 +31,104 @@ func TestMain(m *testing.M) {
...
-		t.Fatalf("Expected value from db snapshot [%s], found [%s]", "value1", v2)
-	}
-
-	if !bytes.Equal(v3, nil) {
-		t.Fatalf("Expected value from db snapshot is 'nil', found [%s]", v3)
-	}
-}
+	openchainDB = GetDBHandle()
 
-func TestDBIteratorAndSnapshotIterator(t *testing.T) {
```
manish-sethi, line 188: Why do you want to delete this test? I had added this unit test for increasing the code coverage during unit tests.

core/db/db_test.go
```diff
@@ -31,246 +31,104 @@ func TestMain(m *testing.M) {
...
-	value2, err := openchainDB.GetFromStateCF([]byte("key2"))
-	if err != nil {
-		t.Fatalf("Error getting in value: %s", err)
-	}
-	if value2 != nil {
-		t.Fatalf("A nil value expected. Found [%s]", value2)
-	}
-}
-
-func TestDBSnapshot(t *testing.T) {
```
manish-sethi, line 135: Why do you want to delete this test? I had added this unit test for increasing the code coverage during unit tests.

core/db/db_test.go
```diff
@@ -31,246 +31,104 @@ func TestMain(m *testing.M) {
...
 		columnfamilies = oldcfs
 	}()
-	db, err = openDB()
-	if nil != err {
-		t.Fatalf("Error re-opening DB with upgraded columnFamilies")
-	}
-	db.CloseDB()
-}
-
-func TestDeleteState(t *testing.T) {
```
manish-sethi, line 110: Why do you want to delete this test?

core/db/db_test.go
```diff
@@ -31,246 +31,104 @@ func TestMain(m *testing.M) {
...
-	originalSetting := viper.GetString("peer.fileSystemPath")
-	viper.Set("peer.fileSystemPath", "")
+func TestCreateDB(t *testing.T) {
+	openchainDB := Create()
+	openchainDB.Open()
+	defer deleteTestDBPath()
+	defer openchainDB.Close()
+}
+
+func TestOpenDB_DirDoesNotExist(t *testing.T) {
```
manish-sethi, line 54: Why do you want to rename test `TestCreateDB_DirDoesNotExist` to `TestOpenDB_DirDoesNotExist`? This test actually creates a db in an empty folder not just opens an existing db.

core/db/db_test.go
GrapeBaBa, line 188: this should be a merge error, i will move it back

core/db/db.go
GrapeBaBa, line 76: Cool, actually i thought a long time for this two locks. Previous I used two mutex, for a further search I found Once instead. I have not  know the init function yet.

By the way, I also thought why make db a singleton. Actually make db as an property for high level object is enough, not need expose db to all the invoker and just expose high level object such as ledger.

core/db/db_test.go
GrapeBaBa, line 54: as I change the logic for CreateDB method, now it only return a pointer

core/db/db_test.go
GrapeBaBa, line 110: this should also a merge error, i will fix

core/db/db_test.go
GrapeBaBa, line 135: merge error two

core/db/db.go
```diff
@@ -172,32 +146,23 @@ func getDBPath() string {
...
+	defer openchainDB.mux.Unlock()
 
 	dbPath := getDBPath()
+	mkDirIfNotExist(dbPath)
+
 	opts := gorocksdb.NewDefaultOptions()
 	defer opts.Destroy()
 
-	opts.SetCreateIfMissing(false)
+	opts.SetCreateIfMissing(true)
```
manish-sethi: Earlier, I had two separate functions. 1) `createDB` and 2) `openDB`. I think that I should have documented the purpose. Now, you have merged the two into `open` function.
The purpose of separation was to handle a special case - suppose a user by mistake specifies the path of an existing 'non-empty' folder (as the path to be used for the db). In the earlier approach, he would have received an error (and we would have prevented rocksDB to create it's files and polluting the folder). However, now, rocksDB would go ahead and start using the folder as db. Apart from polluting the existing user folder, I am not sure what other long term implication could be (e.g., does user continues to access other files he had? does rocksDB compaction mechanics ignore these files or does something special?...). I know that combining the two, makes the code crisper but thinking about above unusual scenario's, I chose to be on a safer note and not to allow user to specify a folder for db that already has files/folders.

core/db/db_test.go
manish-sethi, line 54: irrespective of the return type, this test actually creates rocksDB files in an empty directory so, the previous name was more meaningful.

core/db/db.go
```diff
@@ -52,59 +60,25 @@ type OpenchainDB struct {
...
+	dbState      dbState
+	mux          sync.Mutex
 }
 
 var openchainDB *OpenchainDB
-var isOpen bool
+var once sync.Once
 
-// CreateDB creates a rocks db database
-func CreateDB() error {
```
manish-sethi, line 61: see comments at line 165 in the new code of the PR

core/db/db.go
GrapeBaBa: Got it. Actually I have a previous version for this change. Create will panic when path exist, I also think it more make sense. But some test failed I am not quite sure, I can push that version for further discussion.

By the way, in previous CreateDB method, it also has a logic to open db, then close db. what's the purpose for this logic?

manish-sethi: Once you reverse the above mentioned change, I would go through the remaining changes - because, they would also get affected.

core/db/db.go
GrapeBaBa, line 61: I saw it. A question for line 81 to line 86. It seems open a rocksdb on the dbpath, then close it.I cannot catch it

GrapeBaBa, line 61: I think create a dir is enough for createDB method.

core/db/db.go
manish-sethi: Yes, panic is fine instead of throwing error in DB creation. However, there were some discussions of revising the errors and their handling so I am not sure whether we should take things one by one or do we want a holistic approach. The main doubt in mind is what are the different handling we want is it only a choice between throw error and panic?

The reason of closing the db in the create method was to handle the columnfamily related stuff at one place i.e., `openDB` method.

GrapeBaBa: i agree you and will reverse some codes. But as I understand it does not need open db at all in createDB method, just create an empty dir and leave the open db logic to OpenDB method? Correct me if i understand wrong.

manish-sethi: then you cannot call `opts.SetCreateIfMissing(false)` in openDB. And if you call `opts.SetCreateIfMissing(true)` in openDB, it leads to the same problem that we discussed above - unless there is a way of checking that if a non-empty dir actually has contents that belong to rocksDB.

GrapeBaBa: @manish-sethi  I just looked the old code, I found though you have a separate two create DB and open DB methods, but in GetDBHandle method actually called createDBIfDBpathempty method. When dbpath exists, it will skip createDB method, no error will raise. It actually same behaviour with my current version.

manish-sethi: Alternatively, you can check in the openDB if dir empty then call `opts.SetCreateIfMissing(true)` else `opts.SetCreateIfMissing(false)`. I am fine with this as well.

manish-sethi: @GrapeBaBa I think that you are discounting `opts.SetCreateIfMissing(false)` in the openDB. It will raise an error in the case directory has some non-relevant contents.

manish-sethi: The whole discussion is about not using `opts.SetCreateIfMissing(true)` blindly.

GrapeBaBa: yes, i need investigate this method further. thanks. let me understand this method first

examples/chaincode/chaintool/example02/src/chaincode/chaincode_example02.go
ghaskins: I think this change itself will not break anything.

core/db/db_test.go
GrapeBaBa, line 188: reversed this test

core/db/db_test.go
GrapeBaBa, line 110: reversed this test

core/db/db.go
GrapeBaBa: @manish-sethi I follow your alternative suggestion, it seems small change

core/db/db.go
GrapeBaBa, line 76: For item1.Do you mean take open method in Do func? If yes, when i done this, some unit test failed, i need investigate.
For item2, i cannot catch your mean for cache. It seems singleton already cached, what is cache you refer?

``` go
func GetDBHandle() *OpenchainDB {
    once.Do(func() {
        openchainDB = Create()
                openchainDB.Open()
    })

    return openchainDB
}
```

manish-sethi, line 76: @GrapeBaBa what I meant in item2 is - caching the dbHandler in other modules; rather than they calling `GetDBHandle` function every time.

GrapeBaBa, line 76: @manish-sethi I looked at my code again, it seems already no need sync.once, what your thoughts?

GrapeBaBa, line 76: @manish-sethi I am not quite sure we can cache the function call because I found some codes such as test will close db and open again. So the open method should be invoked every time.Is that right?

GrapeBaBa:
@manish-sethi Codes updated, would you like review again?

christo4ferris:
@GrapeBaBa would you please rebase to resolve merge conflicts? thanks

GrapeBaBa:
@christo4ferris Will do.

GrapeBaBa:
@christo4ferris Rebased, hehave test will fail because behave #680

GrapeBaBa:
reverify Jenkins

GrapeBaBa:
This one should canfix parallel test.

vendor/google.golang.org/grpc/stream.go
srderson, line 152: Please do not modify folders under the vendor path. If you run `make linter` it will run `goimports` which runs `go fmt` and the /vendor directory will be excluded.

These will need to be removed before the PR can be merged. If we do ever want to modify files under the vendor folder, we will want to make PRs in their respective repositories.

christo4ferris:
@binhn @manish-sethi this has been rebased and travis passes. Is it ready to merge?

core/db/db_test.go
```diff
@@ -44,72 +44,96 @@ func TestGetDBPathEmptyPath(t *testing.T) {
...
+	openchainDB := GetDBHandle()
+
+	defer openchainDB.Close()
 }
 
-func TestOpenDB_DirDoesNotExist(t *testing.T) {
-	deleteTestDBPath()
-	defer deleteTestDB()
-	performBasicReadWrite(t)
+func TestGetDB_NonEmptyDirExists(t *testing.T) {
```
manish-sethi: This test is almost identical to the one above `TestOpenDB_NonEmptyDirExists`. Better to delete one of these. Also, please call `deleteTestDBPath()` (as done in the above test) before calling `createNonEmptyTestDBPath` to make sure that test begins with the folder contents that we want.

core/db/db_test.go
```diff
@@ -44,72 +44,96 @@ func TestGetDBPathEmptyPath(t *testing.T) {
...
+		}
+	}()
+	openchainDB.Open()
 }
 
-func TestWriteAndRead(t *testing.T) {
-	createTestDB()
-	defer deleteTestDB()
-	performBasicReadWrite(t)
+func TestGetDB_DirDoesNotExist(t *testing.T) {
```
manish-sethi: This test is almost identical to `TestOpenDB_DirDoesNotExist`. Better to delete one of these. Also, please call `deleteTestDBPath` this explicitly to make sure that test runs in the intended condition. I just noticed that this was an issue previously as well - can you please fix this here.

manish-sethi:
@christo4ferris - @srderson had commented to remove the vendor go files from this PR.
I have added a couple of minor comments just now. 
Other changes LGTM.

GrapeBaBa:
@manish-sethi @srderson Fixed.

GrapeBaBa:
reverify jenkins

GrapeBaBa:
make checks successful in local, not sure why jenkins/x86_64 failed

GrapeBaBa:
reverify jenkins

GrapeBaBa:
@christo4ferris @manish-sethi @srderson  Can this PR be merged or still need some change?

srderson:
LGTM. Thanks

PR #1749, opened 2016-06-08
Correctly order PBFT requests on resubmission

jyellick:
## Description

This changeset causes the resubmission of requests to occur in order by timestamp, rather than randomly in map order.

This changeset also includes a couple minor lint fixes and some enhanced debugging.
## Motivation and Context

Once PBFT runs out of sequence numbers, it simply starts queueing requests into a map.  After more sequence numbers become available as the watermarks move, the requests are resubmitted for ordering.  The previous implementation naively resubmitted them in the order in which they emerged from iterating over the map.  This caused an interaction with the custody deduplication mechanism in batch which expects for requests to arrive for execution in the same order that the batch primary submits them to PBFT for ordering (by timestamp).

This would result in later requests being executed before earlier, and consequently, when earlier requests would be scheduled for execution, they would be pruned as 'stale', producing empty blocks.

Fixes the empty block problem from #1091 (which is only half of #1091)
## How Has This Been Tested?

A new test has been added and verified to fail under the old behavior, but to pass under this PR.
## Checklist:

### DIFF
--- a/consensus/helper/helper.go
+++ b/consensus/helper/helper.go
@@ -215,13 +215,17 @@ func (h *Helper) CommitTxBatch(id interface{}, metadata []byte) (*pb.Block, erro
 	}
 
 	size := ledger.GetBlockchainSize()
-	h.curBatch = nil     // TODO, remove after issue 579
-	h.curBatchErrs = nil // TODO, remove after issue 579
+	defer func() {
+		h.curBatch = nil     // TODO, remove after issue 579
+		h.curBatchErrs = nil // TODO, remove after issue 579
+	}()
 
 	block, err := ledger.GetBlockByNumber(size - 1)
 	if err != nil {
 		return nil, fmt.Errorf("Failed to get the block at the head of the chain: %v", err)
 	}
+
+	logger.Debug("Committed block with %d transactions, intended to include %d", len(block.Transactions), len(h.curBatch))
 
 	return block, nil
 }
--- a/consensus/obcpbft/pbft-core.go
+++ b/consensus/obcpbft/pbft-core.go
@@ -658,7 +658,10 @@ func (instance *pbftCore) resubmitRequests() {
 		return
 	}
 
+	var submissionOrder []*Request
+
 outer:
+
 	for d, req := range instance.outstandingReqs {
 		for _, cert := range instance.certStore {
 			if cert.digest == d {
@@ -668,6 +671,17 @@ outer:
 		}
 		logger.Debug("Replica %d has detected request %s must be resubmitted", instance.id, d)
 
+		submissionOrder = append(submissionOrder, req)
+	}
+
+	if len(submissionOrder) == 0 {
+		return
+	}
+
+	// We sort the outstanding reqs by their timestamp for resubmission
+	sort.Sort(requestSlice(submissionOrder))
+
+	for _, req := range submissionOrder {
 		// This is a request that has not been pre-prepared yet
 		// Trigger request processing again.
 		instance.recvRequest(req)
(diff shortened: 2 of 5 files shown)


consensus/obcpbft/pbft-core.go
```diff
@@ -1266,3 +1280,28 @@ func (instance *pbftCore) stopTimer() {
...
+		// a[i] has no timestamp, handle it later, TODO, eventually this should be an error
+		return false
+	}
+
+	if a[j].Timestamp == nil {
+		// a[j] has no timestamp, handle it later, TODO, eventually this should be an error
+		return true
+	}
+
+	iTime := time.Unix(a[i].Timestamp.Seconds, int64(a[i].Timestamp.Nanos))
```
corecode, line 1303: maybe we should move away from the protobufs timestamp structure and plainly use nanos. (not in this PR tho)

consensus/obcpbft/pbft-core_test.go
```diff
@@ -793,6 +793,61 @@ func TestSendQueueThrottling(t *testing.T) {
 	}
 }
 
+// Test for issue #1091
+// Once the primary ran out of sequence numbers, it would queue requests into a map, and resubmit them in arbitrary order
+// This is incorrect, they need to be resubmitted in the order of their timestamps
```
corecode, line 798: Not incorrect, but not optimal.

jyellick:
@rameshthoomu Looks like CI simply timed out above.  When I attempt to click the 'restart' button, Travis gives me the message "This build can't be restarted" in a red drop down banner across the top of the page.

consensus/obcpbft/pbft-core_test.go
jyellick, line 798: Incorrect is perhaps a strong word, but the arbitrary ordering resulted in rather undesirable in behavior in the context of batch, which expects for requests to be ordered by the PBFT primary in the same order in which it submitted them to the PBFT primary.

consensus/obcpbft/pbft-core.go
jyellick, line 1303: At the very least perhaps some utility functions for converting to and from the protobuf Timestamp structure.

consensus/helper/helper.go
```diff
@@ -215,14 +215,18 @@ func (h *Helper) CommitTxBatch(id interface{}, metadata []byte) (*pb.Block, erro
...
+		h.curBatch = nil     // TODO, remove after issue 579
+		h.curBatchErrs = nil // TODO, remove after issue 579
+	}()
 
 	block, err := ledger.GetBlockByNumber(size - 1)
 	if err != nil {
 		return nil, fmt.Errorf("Failed to get the block at the head of the chain: %v", err)
 	}
 
+	logger.Debug("Committed block with %d transactions, intended to include %d", len(block.Transactions), len(h.curBatch))
```
binhn, line 228: Debug log without guard just produces garbage and spends cycles. I would suggest `if logger.IsEnabledFor(logging.DEBUG)` especially in high frequency paths.

consensus/obcpbft/pbft-core.go
```diff
@@ -668,6 +671,17 @@ outer:
 		}
 		logger.Debug("Replica %d has detected request %s must be resubmitted", instance.id, d)
```
binhn, line 672: Debug log without guard just produces garbage and spends cycles. I would suggest if logger.IsEnabledFor(logging.DEBUG) especially in high frequency paths.

consensus/helper/helper.go
corecode, line 228: The logger checks itself: https://github.com/hyperledger/fabric/blob/master/vendor/github.com/op/go-logging/logger.go#L130

binhn, line 228: it does after having created the garbage and spent cycles, so we want to prevent that in critical paths.

corecode, line 228: What garbage?  The test is exactly the same.  We've been removing these verbose conditionals everywhere in the code base.

jyellick, line 228: I've definitely seen a mix of this in the code, but adding 2 lines of code for a fairly insignificant performance enhancement does not seem like a worthwhile trade-off at this stage.

consensus/obcpbft/pbft-core.go
jyellick, line 672: As there is currently no log level checking at all in the obcpbft package, I've opened #1770 to track this.

rameshthoomu:
@jyellick: Travis CI has specific time limits for each job. Limitation for public repositories is 50mins. I could see 24 behave tests are executed and passed out of 25. Please ignore this error message for now. Soon, we are coming up with PR's to reduce overall build time. Thanks

PR #1754, opened 2016-06-08
Add transaction rejection events

gabre:
This change adds a new type of events: transaction rejection. These events are sent out to listeners in case the corresponding chain code of a transaction exited with error.
## Description

See the summary. Note that this commit does **not** introduce changes such as the removal of erroneous transactions (to which belongs a non-zero exit status) from the ledger.
The change also adds a behave test that uses event-listener for testing. Cause of this, some changes were made in event-listener code and event-listener binary was added to the peer image (but not to the PATH).
## Motivation and Context

A mechanism was needed for the clients to be able to get the results of their transactions without constantly querying them (and without receiving all the completed blocks with all the transactions that had a zero exit status).
## How Has This Been Tested?

BDD tests were run.
## Checklist:

Signed-off-by:  Gabor Hosszu gabor@digitalasset.com


core/chaincode/exectransaction.go
```diff
@@ -139,6 +140,9 @@ func ExecuteTransactions(ctxt context.Context, cname ChainName, xacts []*pb.Tran
 	txerrs = make([]error, len(xacts))
 	for i, t := range xacts {
 		_, txerrs[i] = Execute(ctxt, chain, t)
+		if txerrs[i] != nil {
+			sendTxRejectedEvent(xacts[i].Uuid, txerrs[i].Error())
```
corecode: I don't think this is the right place to emit the event.  You should only emit the event when the errored transaction is (not) being committed to a block.  Until the block is committed, anything is possible.

gabre: Do you mean somewhere in CommitTxBatch? (e.g. a for loop over all the erroneous TXs calling Send?) By the way this commit does not remove the storage of errored TXs.

corecode: Yes, somewhere in CommitTxBatch.

gabre: @tamasblummer what is your opinion about this idea? We agreed that the event should be sent after the chaincode fails.

christo4ferris:
@gabre - restarted travis but if it fails again, then the change needs further investigation

gabre:
@christo4ferris 
Error opening DB IO error: lock /var/hyperledger/test/ledger_test/db/LOCK: Resource temporarily unavailable
Isn't this some kind of environment problem?

ghaskins:
@gabre the DB error I believe was fixed as part of #1877.  I'll respin your CI run since that was committed

gabre:
@ghaskins It is OK now. Any other comments are welcome, please check the code/functionality if you have some time. It would be good to merge it soon.

christo4ferris:
@gabre sadly this needs to be rebased to resolve conflicts. @ghaskins or @srderson  could you please review?

gabre:
@christo4ferris I have rebased it and run the tests locally.

protos/events.proto
```diff
@@ -55,6 +56,14 @@ message Register {
     repeated Interest events = 1;
 }
 
+//Rejection is sent by consumers for erroneous transaction rejection events
+//string type - "rejection"
+message Rejection {
+    string Uuid = 1;
```
srderson: Just a nitpick, but I'd encourage changing this to just `id` as it's clear from PR #1721 that this will not always be a UUID in the future

srderson:
This will need to be rebased due to the merge of PR #1721 (the BDD feature files have a conflict) but I'd suggesting waiting until review is complete in case there are other conflicts.

I think these changes make sense in conjunction with with PR #1719.

@muralisrini Can you please review as you have been the primary contributor to the events module?

srderson:
Discussed in the Fabric maintainers meeting this morning. @muralisrini said he will be reviewing early this week.

gabre:
Rebased it, waiting for review.

examples/events/block-listener/block-listener.go
```diff
@@ -29,24 +29,28 @@ import (
...
-	notfy chan *pb.Event_Block
+	notfy              chan *pb.Event_Block
+	rejected           chan *pb.Event_Rejection
+	listenToRejections bool
 }
 
 //GetInterestedEvents implements consumer.EventAdapter interface for registering interested events
 func (a *adapter) GetInterestedEvents() ([]*pb.Interest, error) {
-	return []*pb.Interest{{EventType: pb.EventType_BLOCK}}, nil
+	return []*pb.Interest{{EventType: pb.EventType_BLOCK}, {EventType: pb.EventType_REJECTION}}, nil
```
muralisrini, line 39: I'd add interest for pb.EventType_REJECTION only if listenToRejections == true

protos/events.proto
muralisrini: +1

core/chaincode/exectransaction.go
muralisrini: @gabre this PR is to be taken in conjunction with #1719 ? If so the event _has_ to be emitted in execute and cannot be in commit correct (only successful txs are available in commit) ?

muralisrini:
There are two sets of changes (1) changes related to sending transaction rejection event and (2) changes to `behave` test framework for testing rejection event. 

Would request @jeffgarratt to review (2) (in particular, perhaps there's a way to directly use gRPC with behave now ?)

For (1), I've added a few inline comments but looks good for most part. 

However, there's a larger question that needs consensus : 

 This PR is a replacement for sending rejection events via Block's "NonHashData" structure (per PR #1719 which removes failed transactions). "Failed transactions" data sent via a block event (coarse grained event) are now broken up into individual "transaction rejection" events (fine grained events). Clients listening for "block" events will have to listen for transaction "reject" events as well.  And unlike registering interest for specific chaincode events, the client will have to register interest for _all_ the failed transactions and filter out uninteresting ones.

@binhn @ghaskins @pmullaney @srderson what do you think of the above granularity change ? On the one hand, this certainly would be valuable information (there's no other direct way to get transaction failure notifications) and on the other we have the above fine grained, "get all individual transaction failed events" that clients need to work with.

tamasblummer:
@muralisrini your analysis is correct as the rest of fabric currently works. This however will change as we move towards next consensus algorithm proposal. Reject messages will be the result of either:
1. of chain code execution by submitting peer
2. submitting peer timeout while trying to collect endorsements
3. consensus service reject relayed back to submitting peer

in all cases the submitting peer will relay the reject message to its clients only for transactions submitted through the same submitting peer.

Please consider merging this as the described behaviour, that is I agree suboptimal but not critical will disappear with later changes.

@binhn @ghaskins @pmullaney @srderson

gabre:
reverify jenkins

srderson:
I think the change is a move in the right direction and agree that it can be improved in later changes. LGTM.

There was a merge conflict after I merged #1719. @gabre please rebase on master again so we can merge.

gabre:
reverify jenkins

srderson:
Discussed latest change with @gabre and @esgott where rejection event returns and entire TX instead of just the ID. Opened issue at https://github.com/hyperledger/fabric/issues/2124 for future consideration as to whether this should only be an ID.

PR #1755, opened 2016-06-08
Statetransfer full state block hole fix

jyellick:
## Description

This PR builds on top of PR #1749 so please consider only the latest commit for code review.

This changeset causes state transfer to check for missing blocks after state transfer skips to a point in time via a state snapshot.
## Motivation and Context

Via #1091, the ledger reported `null` blocks on occasion.  After investigation, these seems to occur when there is a missing block in the ledger (this should be handled more cleanly by the REST API, but that is a separate issue).  These 'missing blocks' turn out to be caused by state transfer falling sufficiently behind, that is moves to a point in time in the future.  When this completes, state transfer was not filling in the missing blocks in the background (despite its design to do so).

This was caused by a relatively simple bug, where `break` was used inside a select statement, when it was intended to break from the enclosing for loop.  This bug has been fixed.

Additionally, there was some small logical inefficiencies which are addressed in this changeset.  Rather than retrieve blocks, then see if there are sufficient deltas to recover, this change now checks before the block retrieval.  Similarly the state is detected as invalid when comparing against the known state hash from a valid block, rather than explicitly invalidating it (as the update does).

Does not fix, but is relevant to #1091 
## How Has This Been Tested?

An existing test case has been enhanced to fail under the old behavior, and verifies that the entire chain is recovered after one of these 'skip' operations.
## Checklist:

### DIFF
--- a/consensus/helper/helper.go
+++ b/consensus/helper/helper.go
@@ -215,13 +215,17 @@ func (h *Helper) CommitTxBatch(id interface{}, metadata []byte) (*pb.Block, erro
 	}
 
 	size := ledger.GetBlockchainSize()
-	h.curBatch = nil     // TODO, remove after issue 579
-	h.curBatchErrs = nil // TODO, remove after issue 579
+	defer func() {
+		h.curBatch = nil     // TODO, remove after issue 579
+		h.curBatchErrs = nil // TODO, remove after issue 579
+	}()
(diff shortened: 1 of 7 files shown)


corecode:
nice!

binhn:
LGTM

PR #1758, opened 2016-06-08
Remove state transfer state thread

jyellick:
## Description

This changeset drastically simplifies the statetransfer service, and eliminates one of its go routines.

It adds a shim in the consensus helper to maintain compatibility until the pending executor service can take advantage of its APIs directly.

This changeset builds on top of PR #1755 and contains only one commit beyond, so please review only the last commit in this PR, and refer to the preceeding PRs to comment on the others.

@jeffgarratt Although you have not worked much in this code, I think you may be the most qualified to review based on your understanding of the lower level underpinnings of state transfer.
## Motivation and Context

There exists a possible race condition between transaction execution, and state transfer which could under some conditions corrupt a blockchain.  Although this does not fix that problem, it lays the groundwork for an execution service which coordinates execution and state transfer which will eliminate this race.

_History_

When the state transfer service was initially created, it was done so under the requirement that it must run asynchronously from the consensus service, as a long running process like state transfer could not safely be invoked in a blocking manner (not to mention it relied on the message handling thread which was at the time, being consumed by consensus).

State transfer also needed a way to recover blocks in the background, both while state transfer was active or inactive. This led to a block go routine, and a state go routine inside of the state transfer service.  In order to give outside services like consensus insight into the progress of state transfer, listener callbacks were introduced.

_Problem_

The fabric needs to ensure that executions and state transfer do not overlap.  Although it may have been possible to use state transfer as it existed today for this purpose, the indirection caused by the callback structures would have complicated the implementation and ultimately increased its bug surface.  Instead, by eliminating the state go routine, and consuming the calling thread for state transfer, it is easily possible to ensure that executions and state transfer do not occur concurrently, simply by performing them on the same go routine.

_Future_
Once this PR has (hopefully) been accepted, a new executor service which consensus can give transactions, or state targets to, will be introduced.  The helper backwards compatible hacks added in this PR will be removed.
## How Has This Been Tested?

All of the state transfer tests have been updated to utilize the new API.

Though somewhat large, this PR generally only removed unneeded function and some other cruft such as some pointer indirection which had accumulated over many patches of development.  The remaining logical flow remains the same, so the existing test suite of CI should cover it well.
## Checklist:

### DIFF
--- a/consensus/helper/helper.go
+++ b/consensus/helper/helper.go
@@ -333,10 +333,11 @@ func (h *Helper) SkipTo(tag uint64, id []byte, peers []*pb.PeerID) {
 				}
 				if err != nil {
 					h.Completed(info.Height-1, info.CurrentBlockHash, peers, tag)
+					break
 				}
 			}
-
 			h.stateTransfering = false
+
 		}()
 	}
 }
--- a/consensus/helper/helper.go
+++ b/consensus/helper/helper.go
@@ -43,6 +43,7 @@ type Helper struct {
 	curBatch     []*pb.Transaction       // TODO, remove after issue 579
 	curBatchErrs []*pb.TransactionResult // TODO, remove after issue 579
 	persist.Helper
+	stateTransfering bool // Whether state transfer is active
 
 	sts *statetransfer.StateTransferState
 }
@@ -56,7 +57,6 @@ func NewHelper(mhc peer.MessageHandlerCoordinator) *Helper {
 		valid:       true, // Assume our state is consistent until we are told otherwise, TODO: revisit
 	}
 	h.sts = statetransfer.NewStateTransferState(mhc)
-	h.sts.RegisterListener(h)
 	return h
 }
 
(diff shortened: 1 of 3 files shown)


consensus/helper/helper.go
```diff
@@ -313,9 +313,32 @@ func (h *Helper) SkipTo(tag uint64, id []byte, peers []*pb.PeerID) {
...
+	// This looks racey, but this should always be called in a serial fashion so it should not be, also this will be removed when the executor is introduced
+	// This is just a temporary hack to enable legacy-like behavior until the executor is finished
+	if !h.stateTransfering {
+		h.stateTransfering = true
+		info := &pb.BlockchainInfo{}
+		proto.Unmarshal(id, info)
+		go func() {
+			h.Initiated(info.Height-1, info.CurrentBlockHash, peers, tag)
+
+			for {
```
corecode, line 326: How does this loop exit when a transfer was successful?

corecode:
All this code gone - nice.

consensus/helper/helper.go
jyellick, line 326: Ah! Good catch. State transfer will benignly ignore transfer to targets it already has, but this will leak a go routine and needlessly waste cycles, I'll push a fix shortly.  This entire codepath should be going away shortly though, which is why there is nothing like an accompanying test case to catch this.

binhn:
LGTM
-- waiting for @jeffgarratt 's review to merge

jeffgarratt:
@jyellick @corecode @binhn First, removing this much code always gets my vote!! Second, I do not see any obvious issues, but I assume if any arise it will be simpler to deal with as there is less code.  Kudos!!    LGTM

srderson:
Travis failure is just 680 again

srderson:
Thanks for the reviews @corecode and @jeffgarratt. LGTM.

PR #1765, opened 2016-06-09
Added chaincode and client part of the attribute feature.

andresgaragiola:
## Description

This is the last PR of PR #1513 split into smaller PR's we were doing it with @diegomasini, @jeronimoirazabal, @raullaprida and @gromeroar. It includes the client and chaincode functions needed for Attribute Management.
## Motivation and Context

This is a required feature to address #1139
## How Has This Been Tested?

Current behave tests has been run without errors.
Current unit tests has been run without errors.
Additional unit tests where included (attr_support_test.go and in file crypto_test.go).
A new behave test is pending we will work with @jeffgarratt on that.
## Checklist:

- [X] I have added a [Signed-off-by]
- [X] I added new documentation (update to attributes.md)
- [X] I added new tests

Signed-off-by: Andres Garagiola andresgaragiola@gmail.com / garagiol@ar.ibm.com


andresgaragiola:
@binhn @smithbk this is the changes we were talking yesterday.

core/crypto/client_tx.go
```diff
@@ -42,7 +46,11 @@ func (client *clientImpl) createDeployTx(chaincodeDeploymentSpec *obc.ChaincodeD
 	}
 
 	// Copy metadata from ChaincodeSpec
-	tx.Metadata = chaincodeDeploymentSpec.ChaincodeSpec.Metadata
+	tx.Metadata, err = getMetadata(chaincodeDeploymentSpec.GetChaincodeSpec(), tCert, attrs...)
```
adecaro, line 45: Why do we impose a structure on the metadata so tightened to the attributes?

Moreover, this will be break compatibility with chaincodes that are already using metadata (asset_management.go) under the contract that the fabric does not impose any special formatting.

If we want the fabric to impose a formatting on the metadata this should be in the direction of a separation between user metadata and transaction metadata. Then, the attributes metadata will be considered as transaction metadata, meaning that they are managed by the fabric. This approach will also offer retro-compatibility.

core/crypto/crypto.go
```diff
@@ -49,13 +49,13 @@ type Client interface {
 	Node
 
 	// NewChaincodeDeployTransaction is used to deploy chaincode.
-	NewChaincodeDeployTransaction(chaincodeDeploymentSpec *obc.ChaincodeDeploymentSpec, uuid string) (*obc.Transaction, error)
+	NewChaincodeDeployTransaction(chaincodeDeploymentSpec *obc.ChaincodeDeploymentSpec, attributes []string, uuid string) (*obc.Transaction, error)
```
adecaro: Shall we move attributes to the spec in order to preserve retro-compatibility, given also that this PR added an _attributes_ field to _ChaincodeSpec_? This comment extends to all the methods that have been changed in the same sense.

core/crypto/crypto.go
```diff
@@ -64,16 +64,13 @@ type Client interface {
...
 	// GetTCertHandlerFromDER returns a CertificateHandler whose certificate is the one passed
-	GetTCertificateHandlerFromDER(der []byte) (CertificateHandler, error)
-
-	// ReadAttribute reads the attribute with name 'attributeName' from the der encoded x509.Certificate 'tcertder'.
-	ReadAttribute(attributeName string, tcertder []byte) ([]byte, error)
+	GetTCertificateHandlerFromDER(tCertDER []byte) (CertificateHandler, error)
 
 	// GetNextTCert gets next available (not yet used) transaction certificate.
-	GetNextTCert() (tCert, error)
+	GetNextTCert(attributes []string) (tCert, error)
```
adecaro: Shall we avoid to expose internal interfaces like tCert which is anyway unusable outside the crypto package? The equivalent public interface is CertificateHandler.

JonathanLevi: Changing the return value of this function is not in the scope of this PR (un?)fortunately. This PR  simply/mainly adds the `attributes` parameter to the function, not changing the returned value/type. Given that the existing interface returns a `tCert` already, it might be quicker/more efficient to review/focus on the current addition/diff and discuss other proposed changes to the API separately.

adecaro: @JonathanLevi thanks for the clarification :)

core/crypto/client_tx.go
andresgaragiola, line 45: thanks for your comment @adecaro. This is a temporary solution, the correct solution implies modifying the transaction structure, we didn't want to make that change as part of the PR to not break any compatibility with existing transactions. Anyway we are not breaking retro-compatibility since this feature is disabled by default.

core/crypto/crypto.go
andresgaragiola: @adecaro transaction can be deployed, invoked and queried in the old way too, in other words transaction can be generated without attributes preserving retro-compatibility.

binhn:
Pending further changes. Do not merge.

core/crypto/crypto.go
adecaro: @andresgaragiola, I see that with the new commit 8d5cb820429613de0638c5edd72f8726001da31f  you moved to more compatible methods'  signatures. 

Still, I don't understand why you need to change the signature at all given that you have an _attributes_ in _ChaincodeSpec_. May you clarify this? Just to understand :)

core/crypto/client_tx.go
adecaro, line 45: @andresgaragiola, May I suggest the following approach: Rather than changing the confidential transaction processing in the way is proposed in this PR, you could expose the internal method _tCert#GetKForAttribute_ via the public _CertificateHandler_ interface with a new method called something like _GetProofOfAttributeOwnership(attrName) ([]byte, error)_. This will give back the freedom to the application to decide how to embed that proof of ownership in the transaction metadata. This does not change the transaction structure and preserve the contract on the transaction metadata field. Thoughts?

andresgaragiola:
@binhn I have sent changes we have talked.

binhn:
@andresgaragiola Thanks.

Issue #1589, opened 2016-05-25
TDD and BDD frameworks for Fabric


JonathanLevi:
## Description

Fabric currently has a set of integration/Behavior-Driven/Feature harnesses that work quite nicely for testing REST API and JSON (JavaScript Object Notation) output testing and manipulation, using [Behave](https://behave.github.io/behave.example), the "[cucumber](https://cucumber.io/docs)-clone for Python" (as their web-site suggests).
## Background

Behavior-driven development (or BDD) is great, as it "encourages collaboration between developers, QA and non-technical or business participants in a software project".

Keeping the ability to describe scenarios in high-level, using a human-readable and parseable grammar (aka, "natural words") makes it easy for non-technical or business participants to also take a more active part, etc.
## Context

The question is whether we should use Behave which is Python based/backed, rather than other frameworks (e.g., something that is written in Golang, which will allow us to work with the objects/API natively) - I am thinking beyond (or actually, below) the REST API.

I am genuinely asking, as I haven't been privy to earlier discussions that lead to Behave, and I wonder what other frameworks we could look into, that will make it easy to test Fabric's REST API as we do today, but will offer more (with/without the need for other to learn other programming/scripting languages).

While keeping the good BDD-features and the Cucumber-style syntax for the acceptance tests from the angle of the developers/QA people: shall we try other frameworks (for example, some that are written in Golang?
## Alternatives

Of course this is not to imply/suggest that it is mandatory/necessary to use Golang for testing the orchestration/external-api layer - the current Behave test using Python's `requests` and `json` packages are doing great for the REST API, and the Python code/script there is not that complex.

Thinking out loud here (issue came up when discussing Python vs Golang, even though I'm a Python developer myself). Mainly in terms of barrier to entry for other Golang devs - discussion/suggestions welcome!

Trying to gauge interest: what, for example, are the current/latest thoughts on [Ginkgo](https://github.com/onsi/ginkgo) / [Gomega](https://onsi.github.io/gomega) ?

srderson:
For background, @jeffgarratt picked Behave and wrote the initial BDD testing implementation that Fabric is using. I'm guessing he picked Behave because that's what he had used in the past and it allowed him to quickly get something working. I'll let @jeffgarratt comment as to other reasons why he may have picked Behave and whether there are any technical differences.

jeffgarratt:
[...]

PR #1772, opened 2016-06-09
Issue 1652

### LINKED ISSUE #1652: use mainstream package for asserts in ledger tests
## Description

the core/ledger package should use one of the more popular and supported assertion packages
## Describe How to Reproduce

N/A

I'll be sending a PR this weekend

christo4ferris:
## Description

Rewrote the core/ledger/ledger_test.go using Ginkgo and Gomega. Tried to add in more comprehensive validation to the existing tests. Also refactored some of the foriegn helper functions to be more localized to the test suite.
## Motivation and Context

The ledger tests used some bespoke assertion capability. There are many well traveled golang packages that provide more robust capability. I used this as an example of a possible direction for the remainder of the testing.

Fixes #1652 
## How Has This Been Tested?

Added test directory with new test suite using Ginko/Gomega - 72/73 tests pass. One test is left in Pending state (PIt) that reveals a possible bug in the database or ledger that leaves state from a prior database. This needs some further investigation.
## Checklist:


christo4ferris:
ah, I may need to add ginkgo and gomega to the vendored packages in the development environment

JonathanLevi:
@christo4ferris OMG, Please do! We'll be all over it !!!

christo4ferris:
pushed commit which has triggered rebuild

christo4ferris:
reverify jenkins

JonathanLevi:
Really appreciate you adding [Ginkgo](https://github.com/onsi/ginkgo) @christo4ferris!

LGTM

core/ledger/test/test.yaml
```diff
@@ -0,0 +1,17 @@
+###############################################################################
+#
+#    Peer section
+#
+###############################################################################
+peer:
+    # Path on the file system where peer will store data
+    fileSystemPath: /var/hyperledger/test/ledger_test
```
JonathanLevi, line 8: As a side note: I wish we had a way of being able to use relative paths in the YAMLs/tests. While your `fileSystemPath` is configured correctly, I checked after looking at this issue's changes, and I found that at the project levels not all the absolute paths are. Relative paths or concatenated ones, at least, would have made it better.

JonathanLevi, line 8: Clarification: No action required at this point. LGTM.

corecode, line 8: I've suggested before to use temp databases, instead of something in `/var`

core/ledger/test/ledger_test.go
```diff
@@ -0,0 +1,1128 @@
...
+		It("rolls backwards once without error", func() {
+			delta2, err := ledgerPtr.GetStateDelta(2)
+			Expect(err).To(BeNil())
+			delta2.RollBackwards = true
+			err = ledgerPtr.ApplyStateDelta(1, delta2)
+			Expect(err).To(BeNil())
+			err = ledgerPtr.CommitStateDelta(1)
+			Expect(err).To(BeNil())
+		})
+		PIt("should return committed state from batch 2 after rollback", func() {
```
christo4ferris, line 635: Note that this is the test that is currently skipped, which reveals that there is a leak from a previous database (which was deleted, no less) that can find its way into a subsequent database. Specifically, "chaincode4", "key4" returns a value of "value4" which AFAIK was put into a previous database at either line 205 or 258.

christo4ferris:
I should also probably mention that once this is merged, and with @srderson concurrence I would follow this up by fleshing out the remaining tests in the ledger package as ginkgo tests, and once those are merged and demonstrating at least equivalent coverage with  what is there now, I would replace the current tests with the ginkgo based tests so that we can formally measure coverage.

corecode:
I'd love to see examples of mocking interfaces with the new test framework.

core/ledger/test/test.yaml
JonathanLevi, line 8: I would like us to be able to move stuff around/refactor/clear temp files, not worry about name-, file-, dir- collisions, etc. `tmp` and `temp` is the new permanent! Easier also in terms of more consistent guidelines for new tests/etc...

ghaskins:
This LGTM.  @christo4ferris this appears to be a candidate for a squash merge.  Would you be ok with that?

@binhn @srderson @jeffgarratt you guys have any opinion on the addition (or possible future wholesale replacement) of the unit test framework?

JonathanLevi:
Some more context/earlier discussions around Ginkgo/BDD can be found also here: #1589

ghaskins:
It seems from the conversation that @jeffgarratt isn't opposed to alternate frameworks per se, so thats a postive in favor of accepting this.  Anyone else have a comment?

JonathanLevi:
BTW: I'm happy to go ahead with this, also because Ginkgo plays nice with the existing Go's `testing` framework/infrastructure... going forward, if we ever choose/consider another testing framework, we can re-evaluate where we stand there and then. At this point, echo'ing what I read around it from everyone, it is already a big improvement.

srderson:
Thanks! LGTM

PR #1777, opened 2016-06-09
Fix reset for event timers in consensus

kchristidis:
## Description

For the event timers in consensus, `Reset()` should reset an existing timer to a new timeout and clear out any pending events, while `SoftReset()` should reset the timer only if it's not already running. Right now, their functionality is the opposite.
## Motivation and Context

This change is needed to deliver the expected functionality for the two methods.
## How Has This Been Tested?

I have modified the existing unit tests in the `events` package by adding a field to the `mockEvent` struct. This makes detection of such errors possible.
## Checklist:

### DIFF
--- a/consensus/obcpbft/events/events.go
+++ b/consensus/obcpbft/events/events.go
@@ -254,12 +254,12 @@ func (et *timerImpl) loop() {
 			eventDestChan = nil
 		case <-et.stopChan:
 			if et.timerChan == nil && eventDestChan == nil {
-				logger.Warning("Attempting to stop an unfired idle timer")
+				logger.Debug("Attempting to stop an unfired idle timer")
 			}
 			et.timerChan = nil
-			logger.Debug("Stopping timer for")
+			logger.Debug("Stopping timer")
 			if eventDestChan != nil {
-				logger.Debug("Timer for cleared pending event")
+				logger.Debug("Timer cleared pending event")
 			}
 			eventDestChan = nil
 			event = nil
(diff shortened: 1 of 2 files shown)


jyellick:
Good catch! Looks good to me.

corecode:
How do the changes in tests cover this?

jyellick:
@corecode This was a subtle one, here's a [playground example](https://play.golang.org/p/oHJjtrvN_r).  It appears that the golang compiler treats the empty struct specially.  As they both referred to an empty struct, which could not be modified, I believe the compiler made the optimization of making both of those literal references refer to the same chunk of memory (or special reserved pointer maybe?).  Even by simply embedding a string in the struct, but not modifying it, you can see the behavior go away.  Still, I think modifying the string makes @kchristidis's intent clearer.

kchristidis:
@corecode: What @jyellick said -- and exactly the reason why I modified the string (even though it was redundant).

See [here](http://dave.cheney.net/2014/03/25/the-empty-struct): "Interestingly, the address of two struct{} values may be the same."

christo4ferris:
reverify Jenkins

corecode:
So this fixes the test and implementation?

kchristidis:
@corecode: Fixes, as in works around the compiler optimization and exposes the problem.

Notice how the current test passes even though it shouldn't. Then add an `info string` field to the `mockEvent` struct definition (no need to set it), re-run `events_test.go`, and notice how the test fails.

If you have a better approach in mind, let me know and I'll be happy to submit a patch.

corecode:
Right, I just was confused because the test seemingly didn't change.  +1

bramwelt:
reverify jenkins

srderson:
LGTM. Thanks

PR #1783, opened 2016-06-10
fixes #1299 by removing instance of multiplication by time.Millisecond

### LINKED ISSUE #1299: Timeout multiplied by Millisecond at 2 places
The timeout is multiplied by time.Millisecond at 2 places as mentioned below, resulting in a huge timeout

https://github.com/hyperledger/fabric/blob/master/peer/main.go#L712

https://github.com/hyperledger/fabric/blob/189e0db0e377ef8f40160fc0eac9740f11134716/core/chaincode/chaincode_support.go#L123

mastersingh24:
## Description

Removed redundant multiplication of timeout setting 
## Motivation and Context

There is a parameter in core.yaml ( `chaincode.startuptimeout` ) which ends up being accidentally multiplied by `time.Millisecond` in two places (once in the peer on startup and later in the chaincode_support module).  This results in a startup timeout of 277h46m40s rather than the intended 1s

Fixes #1299
## How Has This Been Tested?

No new test case added as the current tests still pass.  We have also been patching the code we use to run the Bluemix service for months as well and know this change works.
## Checklist:

Signed-off-by: Gari Singh gari.r.singh@gmail.com

### DIFF
--- a/core/chaincode/chaincode_support.go
+++ b/core/chaincode/chaincode_support.go
@@ -114,7 +114,7 @@ func NewChaincodeSupport(chainname ChainName, getPeerEndpoint func() (*pb.PeerEn
 
 	s.userRunsCC = userrunsCC
 
-	s.ccStartupTimeout = ccstartuptimeout * time.Millisecond
+	s.ccStartupTimeout = ccstartuptimeout
 
 	//TODO I'm not sure if this needs to be on a per chain basis... too lowel and just needs to be a global default ?
 	s.chaincodeInstallPath = viper.GetString("chaincode.installpath")


christo4ferris:
reverify Jenkins

christo4ferris:
@mastersingh24 there is still a bug in here, and I am not thrilled with the approach of having the caller set the timeout. Seems to me that NewChaincodeSupport should be reading and caching the "chaincode.startuptimeout" and handling the default (which is still broken at line 720 of peer/main.go) rather than which ever module is calling NewChaincodeSupport. That way the logic of converting from the string to a time.Duration is encapsulated and the same mistake cannot be made.

mastersingh24:
@christo4ferris  - what's the bug?  I understand the other part about fixing how this whole thing actually works, but not sure what bug you are referring to?

I can submit a bigger fix and/or open another issue on this for re-implementing the current logic.  Anyone who actually tries to run this like we do in a multi-tenant environment (or a Docker env with a lot of containers will run into the current issue).  My worry is that making a big change at this point might have unknown consequences given the current test coverage.  We can withdraw this change and continue to patch ourselves as well.  Please advise

christo4ferris:
@mastersingh24 the default timeout is set at 5000 at line 720 of peer/main.go and then multiplied by time.Millisecond

christo4ferris:
note, I would at this point just fix the remaining nit and if you like, I can refactor.

mastersingh24:
That multiplication should be ok.  I removed the second one in chaincode_support.NewChaincodeSupport which was causing the issue.  Are you suggesting I remove it in both places or that I should remove the one in NewChaincodeSupport and keep the one in peer?  Sorry - I think I am missing something here.  We have done it both ways

christo4ferris:
@mastersingh24 if you want the timeout to be 5 seconds, then the default value set on line 720 needs to be 5, because line 722 will change it to 5000 seconds unless I am missing something.

mastersingh24:
@christo4ferris  - ah.  yeah - I am no expert on the Golang time package, but basically the way you actually change the unit of measure for any duration is to multiply it by the appropriate unit representation, e.g. `time.Millisecond`.  The default "duration" is in nanoseconds.  So `5000*time.Millisecond` actually means `5000ms`

christo4ferris:
ah, ok that makes sense - you learn something every day. thanks. LGTM aside from the convoluted approach.

srderson:
LGTM. Thanks

PR #1785, opened 2016-06-10
Add executor service

jyellick:
## Description

This changeset introduces the new executor service advertised in PR #1758.  It adds an API for asynchronously invoking executions, commits, and state transfer.

This changeset tries to minimally change the existing consensus plugins to adapt to the new API, and leaves the legacy execution infrastructure in place to attempt to minimize the changeset.
## Motivation and Context

The consensus API has always exposed assorted executor methods such as `ExecTxs` or `CommitTxBatch` as synchronous calls independent of state transfer.  This poses multiple problems.  Firstly, particularly with chaincode deploy transactions, but also for larger batches of transactions, these calls may block for long periods of time (seconds), which causes problems for liveliness tests, such as in PBFT.  Secondly, these calls do not coordinate with state transfer, and it is possible to incidentally execute transactions while state transfer is in progress.  These two problems compound eachother, because as state transfer and execution can both be long running calls, a caller must execute them both asynchronously, and coordinating between the two is bug prone and dangerous.

The executor interface is intended to serialize write access to the ledger, either via state transfer or via chaincode execution.  It is a simply contained testable state machine which is intended to accommodate one and only one asynchronous transaction at a time.  Once a particular transaction completes, a callback is invoked and the caller is informed that it is safe to inject a new request into the executor service.  In the event that a caller mistakenly calls in while another request is in progress, the executor will serialize the call, or in the event that the serialized calls are not consistent (for instance calling `Commit` without calling `Execute`) the executor will safely ignore the call to avoid putting the system into an inconsistent state.

As mentioned in the description, this changeset is intended to minimize the diff, so the legacy interfaces are intact, and code which is not obc-batch still uses these legacy methods to a large extent.  In the future, the consensus API should remove the obsolete and unsafe ledger access methods, the remainder of the consensus plugins should be migrated to use the new API, and the executor should be migrated to interact more directly with the ledger.

Although this PR is unfortunately on the larger side, please note that 522 of the added lines are new test cases for the introduced executor service.
## How Has This Been Tested?

The new executor service comes complete with a new suite of tests.  The consensus code which has been modified to execute using the new API passes unit tests, plus of course the remainder of CI.
## Checklist:

### DIFF
--- a/consensus/executor/executor.go
+++ b/consensus/executor/executor.go
@@ -1,0 +1,220 @@
+/*
+Copyright IBM Corp. 2016 All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+		 http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package executor
+
+import (
+	"fmt"
+
+	"github.com/hyperledger/fabric/consensus"
+	"github.com/hyperledger/fabric/consensus/obcpbft/events"
+	"github.com/hyperledger/fabric/core/peer/statetransfer"
+	pb "github.com/hyperledger/fabric/protos"
+
+	"github.com/op/go-logging"
+)
+
+var logger *logging.Logger // package-level logger
+
+func init() {
+	logger = logging.MustGetLogger("consensus/executor")
+}
+
+// PartialStack contains the ledger features required by the executor.Coordinator
+type PartialStack interface {
+	consensus.LegacyExecutor
+	GetBlockchainInfo() *pb.BlockchainInfo
+}
+
+type coordinatorImpl struct {
+	manager         events.Manager              // Maintains event thread and sends events to the coordinator
+	rawExecutor     PartialStack                // Does the real interaction with the ledger
+	consumer        consensus.ExecutionConsumer // The consumer of this coordinator which receives the callbacks
+	stc             statetransfer.Coordinator   // State transfer instance
+	batchInProgress bool                        // Are we mid execution batch
+	skipInProgress  bool                        // Are we mid state transfer
+}
+
+// NewCoordinatorImpl creates a new executor.Coordinator
+func NewImpl(consumer consensus.ExecutionConsumer, rawExecutor PartialStack, stps statetransfer.PartialStack) consensus.Executor {
+	co := &coordinatorImpl{
+		rawExecutor: rawExecutor,
+		consumer:    consumer,
+		stc:         statetransfer.NewCoordinatorImpl(stps),
+		manager:     events.NewManagerImpl(),
+	}
+	co.manager.SetReceiver(co)
+	return co
+}
+
+// ProcessEvent is the main event loop for the executor.Coordinator
+func (co *coordinatorImpl) ProcessEvent(event events.Event) events.Event {
+	switch et := event.(type) {
+	case executeEvent:
+		if logger.IsEnabledFor(logging.DEBUG) {
+			if logger.IsEnabledFor(logging.DEBUG) {
+				logger.Debug("Executor is processing an executeEvent")
+			}
+		}
+		if co.skipInProgress {
+			logger.Error("FATAL programming error, attempted to execute a transaction during state transfer")
+			return nil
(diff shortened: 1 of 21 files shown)


consensus/executor/executor.go
```diff
@@ -0,0 +1,220 @@
...
+	co.manager.SetReceiver(co)
+	return co
+}
+
+// ProcessEvent is the main event loop for the executor.Coordinator
+func (co *coordinatorImpl) ProcessEvent(event events.Event) events.Event {
+	switch et := event.(type) {
+	case executeEvent:
+		if logger.IsEnabledFor(logging.DEBUG) {
+			if logger.IsEnabledFor(logging.DEBUG) {
```
corecode: double?

Also, absolutely not required when no extra arguments are logged.

consensus/executor/executor.go
```diff
@@ -0,0 +1,220 @@
...
+				logger.Debug("Executor is processing an executeEvent")
+			}
+		}
+		if co.skipInProgress {
+			logger.Error("FATAL programming error, attempted to execute a transaction during state transfer")
+			return nil
+		}
+
+		if !co.batchInProgress {
+			if logger.IsEnabledFor(logging.DEBUG) {
```
corecode: This makes the code just less readable

consensus/executor/executor.go
```diff
@@ -0,0 +1,220 @@
...
+
+		co.rawExecutor.ExecTxs(co, et.txs)
+
+		co.consumer.Executed(et.tag)
+	case commitEvent:
+		if logger.IsEnabledFor(logging.DEBUG) {
+			logger.Debug("Executor is processing an commitEvent")
+		}
+		if co.skipInProgress {
+			logger.Error("Likely FATAL programming error, attempted to commit a transaction batch during state transfer")
```
corecode, line 86: we should panic instead of just logging a message

consensus/executor/executor.go
```diff
@@ -0,0 +1,220 @@
...
+
+		info := co.rawExecutor.GetBlockchainInfo()
+
+		co.consumer.Committed(et.tag, info)
+	case rollbackEvent:
+		if logger.IsEnabledFor(logging.DEBUG) {
+			logger.Debug("Executor is processing an rollbackEvent")
+		}
+		if co.skipInProgress {
+			logger.Error("Programming error, attempted to rollback a transaction batch during state transfer")
```
corecode, line 106: panic

consensus/executor/executor.go
```diff
@@ -0,0 +1,220 @@
...
+		info := et.blockchainInfo
+		for {
+			err, recoverable := co.stc.SyncToTarget(info.Height-1, info.CurrentBlockHash, et.peers)
+			if err == nil {
+				co.skipInProgress = false
+				co.consumer.StateUpdated(et.tag, info)
+				return nil
+			}
+			if !recoverable {
+				if logger.IsEnabledFor(logging.DEBUG) {
```
corecode: Warning or Error

consensus/executor/executor.go
jyellick: Good catch, I'll cleanup

consensus/executor/executor.go
jyellick: I considered making this a more severe message, but the `!recoverable` is less severe than it sounds.  It simply means that the state transfer target supplied is no longer reachable (usually because the target was sufficiently newer than our state that we could not recover via deltas, so we had to recover via a state snapshot, and that snapshot was for a block in the future, which we can't verify yet).

Edit: Bumped this up to "Warning" in latest commit

consensus/executor/executor.go
jyellick, line 106: I'm on the fence about panic-ing in these situations.  This goes back to #752 about what exactly we should do when we encounter 'irrecoverable type errors', it would be nice if we had a cleaner 'shutdown because of error' than panic.  Anyone else want to weigh in here?

consensus/executor/executor.go
jyellick: Yes, on string literals, there is no garbage or performance bottleneck, I threw these on at the last minute because of the comments from #1749 and should have thought about them a little more.

I've pushed a commit fixing these things.

bramwelt:
reverify jenkins

kchristidis:
This is good to go. @srderson

srderson:
@kchristidis Thanks for the review! @jyellick This PR currently has a merge conflict and must be rebased on master.

christo4ferris:
@srderson this looks to be GTG

srderson:
Thanks, LGTM

denyeart:
@pamandrejko I didn't write the comment to copy/paste in, but good enough :-)

denyeart:
@Mergifyio backport release-2.2

PR #1793, opened 2016-06-10
consensus: non-blocking broadcast

### LINKED ISSUE #1056: Investigate blocking network operations in fabric
Per a discussion with @chetmurthy we should audit the fabric code to ensure that we never expect network operations to only succeed/fail immediately, but instead ensure that we also handle the timeout case.

This can happen when a network host ungracefully loses network connectivity (unplugged ethernet cable) or when TCP behavior degrades to its timeout mechanisms (for instance failed connections might not be rejected with a TCP RST packet, perhaps due to a firewall).

When this behavior is observed, operations which normally do not appear to be blocking (sends, opens, and receives) can block for long periods of time.  If this blocking occurs on a thread/goroutine which has other work to be performed, the single network failure can cause a total network stall.  The code should be audited to ensure this does not happen.

Preliminary investigation shows that the fabric code is most likely not vulnerable to the more insidious half of these errors, as reads are done via a dedicated goroutine per connection.  Documentation in `openchain.proto` says

```
// Accepts a stream of OpenchainMessage during chat session, while receiving
// other OpenchainMessage (e.g. from other peers).
```

This seems to be in parity with the gRPC doc which indicates that:

```
Serve accepts incoming connections on the listener lis, creating a new
ServerTransport and service goroutine for each. The service goroutines
read gRPC requests and then call the registered handlers to reply to 
them. Service returns when lis.Accept fails. 
```

On the sending side, the possibility of a bug of this type seems more likely.  The generated code creates a `Peer_ChatServer` which has a `Send` method which is a wrapper around `ServerStream.SendMsg`, which is documented as:

```
    // SendMsg blocks until it sends m, the stream is done or the stream
    // breaks.
    // On error, it aborts the stream and returns an RPC status on client
    // side. On server side, it simply returns the error to the caller.
    // SendMsg is called by generated code. Also Users can call SendMsg
    // directly when it is really needed in their use cases.
```

Since our broadcast serially invokes `Send` across our peers in a for loop, it seems possible a non-closed TCP state could end up causing the send to block until a timeout occurs.

This is purely preliminary doc research, we should verify this behavior on both the send and receive sides.

corecode:
## Description

Introduces a broadcaster service, which will block only for N-f replicas, so that a byzantine replica cannot halt the network.
## Motivation and Context

Fixes #1056 
## How Has This Been Tested?

new unit tests, new bddtest
## Checklist:

Signed-off-by:
Simon Schubert sis@zurich.ibm.com

### DIFF
--- a/consensus/obcpbft/broadcast.go
+++ b/consensus/obcpbft/broadcast.go
@@ -1,0 +1,141 @@
+/*
+Copyright IBM Corp. 2016 All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+		 http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package obcpbft
+
+import (
+	"fmt"
+	"sync"
+	"time"
+
+	"github.com/hyperledger/fabric/consensus"
+	pb "github.com/hyperledger/fabric/protos"
+)
+
+type communicator interface {
+	consensus.Communicator
+	consensus.Inquirer
+}
+
+type broadcaster struct {
+	comm communicator
+
+	f        int
+	msgChans map[uint64]chan *pb.Message
+	closed   sync.WaitGroup
+	closedCh chan struct{}
+}
+
+func newBroadcaster(self uint64, N int, f int, c communicator) *broadcaster {
+	queueSize := 10 // XXX increase after testing
+
+	chans := make(map[uint64]chan *pb.Message)
+	for i := 0; i < N; i++ {
+		if uint64(i) == self {
+			continue
+		}
+		chans[uint64(i)] = make(chan *pb.Message, queueSize)
+	}
+	b := &broadcaster{
+		comm:     c,
+		f:        f,
+		msgChans: chans,
+		closedCh: make(chan struct{}),
+	}
+	return b
+}
+
+func (b *broadcaster) Close() {
+	close(b.closedCh)
+	b.closed.Wait()
+}
+
+func (b *broadcaster) Wait() {
+	b.closed.Wait()
+}
+
+func (b *broadcaster) unicastOne(msg *pb.Message, dest uint64, wait chan bool) {
+	defer func() {
+		b.closed.Done()
+		wait <- true
+	}()
+
+	select {
+	case <-b.closedCh:
+		return
+	default:
+	}
+
+	h, err := getValidatorHandle(dest)
+	if err != nil {
+		logger.Warningf("could not get handle for replica %d", dest)
+	} else {
+		err = b.comm.Unicast(msg, h)
+	}
+	if err != nil {
+		logger.Debugf("could not send to replica %d: %v", dest, err)
+		select {
+		case <-b.closedCh:
+			return
+		case <-time.After(1 * time.Second):
+			return
+		}
+	}
+}
+
+func (b *broadcaster) send(msg *pb.Message, dest *uint64) error {
+	select {
+	case <-b.closedCh:
+		return fmt.Errorf("broadcaster closed")
+	default:
+	}
+
+	var destCount int
+	var required int
+	if dest != nil {
+		destCount = 1
+		required = 1
+	} else {
+		destCount = len(b.msgChans)
+		required = destCount - b.f
+	}
+
+	wait := make(chan bool, destCount)
+
+	if dest != nil {
+		b.closed.Add(1)
+		go b.unicastOne(msg, *dest, wait)
+	} else {
+		b.closed.Add(len(b.msgChans))
+		for i := range b.msgChans {
+			go b.unicastOne(msg, i, wait)
+		}
+	}
+
+	for i := 0; i < required; i++ {
+		<-wait
+	}
+
+	return nil
+}
+
+func (b *broadcaster) Unicast(msg *pb.Message, dest uint64) error {
+	return b.send(msg, &dest)
+}
+
+func (b *broadcaster) Broadcast(msg *pb.Message) error {
+	return b.send(msg, nil)
+}
--- a/bddtests/environment.py
+++ b/bddtests/environment.py
@@ -36,6 +36,8 @@ def after_scenario(context, scenario):
 
             print("Decomposing with yaml '{0}' after scenario {1}, ".format(context.compose_yaml, scenario.name))
             context.compose_output, context.compose_error, context.compose_returncode = \
+                cli_call(context, ["docker-compose"] + fileArgsToDockerCompose + ["unpause"], expect_success=True)
+            context.compose_output, context.compose_error, context.compose_returncode = \
                 cli_call(context, ["docker-compose"] + fileArgsToDockerCompose + ["kill"], expect_success=True)
             context.compose_output, context.compose_error, context.compose_returncode = \
                 cli_call(context, ["docker-compose"] + fileArgsToDockerCompose + ["rm","-f"], expect_success=True)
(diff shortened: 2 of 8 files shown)


bramwelt:
reverify jenkins

christo4ferris:
Both Jenkins and Travis have same failure at:

20:10:46 2016/06/10 20:10:46 Batch replica 2 received new consensus request: UFxXvOFHN3o1kuFDEbipDZ9Pvrw4wr07/p0aPXMsvnqshRMtdNICIEPunHEQfeOFaKByQcpEy47W3YQAA8Bv6Q==
20:10:46 --- FAIL: TestNetworkBatch (0.12s)
20:10:46    obc-batch_test.go:64: 0 messages expected in primary's batchStore, found 1

consensus/obcpbft/broadcast.go
```diff
@@ -0,0 +1,147 @@
...
+	comm communicator
+
+	f        int
+	msgChans map[uint64]chan *pb.Message
+	closed   sync.WaitGroup
+	closedCh chan struct{}
+}
+
+func newBroadcaster(self uint64, N int, f int, c communicator) *broadcaster {
+	queueSize := 10 // XXX increase after testing
```
jyellick, line 43: Presumably this will be configurable?

consensus/obcpbft/broadcast.go
```diff
@@ -0,0 +1,147 @@
...
+				logger.Warning("could not get handle for replica %d", n)
+				continue
+			}
+			err = b.comm.Unicast(msg, h)
+			if err != nil {
+				logger.Debug("could not send to replica %d: %v", n, err)
+				size := len(msg.Payload)
+				speed := 10000000
+				delay := int(time.Second) * size / speed
+				time.Sleep(time.Duration(delay))
```
jyellick: This looks like we're throttling sending the next message based on the size of the message which failed to send? It's not clear what problem this is trying to solve, a comment might help.

consensus/obcpbft/broadcast.go
```diff
@@ -0,0 +1,147 @@
...
+		required = destCount - b.f
+	}
+
+	wait := make(chan bool, destCount)
+
+	if dest != nil {
+		go b.unicastOne(msg, *dest, wait)
+	} else {
+		for i := range b.msgChans {
+			go b.unicastOne(msg, i, wait)
```
jyellick, line 124: This concerns me a bit as a race.  Is it possible that in a system with heavy load that this line might execute multiple times for the same target replica, and then be scheduled in differing orders? (Essentially arbitrarily re-ordering messages)

consensus/obcpbft/broadcast.go
```diff
@@ -0,0 +1,147 @@
...
+	var required int
+	if dest != nil {
+		destCount = 1
+		required = 1
+	} else {
+		destCount = len(b.msgChans)
+		required = destCount - b.f
+	}
+
+	wait := make(chan bool, destCount)
```
jyellick, line 116: Why is this a channel of `bool`s? It seems like the return value is not used so it should be of `struct{}`s.

consensus/obcpbft/broadcast.go
corecode: It's a hack to prevent a defective connection to appear as the fastest connection.  Alternatively we could just put a fixed delay, or keep on retrying.

consensus/obcpbft/broadcast.go
corecode, line 124: What is racy?  The caller waits for unicast to signal success on the wait channel.

consensus/obcpbft/broadcast.go
corecode, line 116: I couldn't figure out how to send a struct{}.  I don't think it matters much.

consensus/obcpbft/broadcast.go
jyellick: I'm still don't understand, what is the problem with a defective connection appearing to be the fastest? Why should we throttle based on the message size?

consensus/obcpbft/broadcast.go
jyellick, line 116: From a performance perspective, no, but from a clarity perspective I think it's important.  If a channel returns a value, intuitively, that value should mean something.

You can send an empty struct literal as:

```
theChannel <- struct{}{}
```

consensus/obcpbft/broadcast.go
jyellick, line 124: A calling go routine calls:

```
b.Broadcast(msg)
```

Which calls

```
b.send(msg,nil)
```

Which then starts `destCount` go routines each invoking:

```
b.unicastOne(msg, i, wait)
```

It then waits for `destCount - b.f` of them to finish.

At this point, the call returns, and the caller may make another call into broadcast.  There are `b.f` go routines which may not have invoked `unicastOne` at this point.  If the caller makes another call into broadcast, it is perfectly possible that the new go routines for those `b.f` destinations will be scheduled before the first batch ever is, and you will get the second message arriving before the first.

corecode, line 124: Any suggestions how to address this?  I don't think it is a big problem.

bramwelt:
reverify jenkins

consensus/obcpbft/broadcast.go
jyellick, line 124: Why not have one persistent go routine and channel per target?

I imagine you considered this, but my naive implementation would have been to have a buffered channel per target, and a go routine per channel.  Attempting to queue a message when a channel is full would result in an error and dropping the message.  You could still make broadcast block until all but f of them have succeeded.

consensus/obcpbft/broadcast.go
corecode: This is so that eventually we can pace our request ingestion rate to match the network processing rate.  Imagine F replicas being disconnected, and their Unicast returns immediately.  Now we have several large requests being sent.  Because F return quickly, we will just wait for another F+1, and the remaining F (correct) replicas will never get their messages, because their chans will be full and will never receive messages.

consensus/obcpbft/broadcast.go
corecode, line 124: There is a persistent goroutine, which is the send goroutine.  These goroutines here are to enqueue to the chans the send goroutines read from.

consensus/obcpbft/broadcast.go
jyellick: Ah, makes sense, maybe paste that into a comment in the code there for posterity? Is there a test case for this?

consensus/obcpbft/broadcast.go
jyellick, line 124: Why the extra go routine? Why not simply do a non-blocking channel write, flagging a message as errored if there is no room in the channel?

corecode, line 124: Because then we will discard messages as soon as the channel is full.

jyellick, line 124: Do we never discard messages then? Could a byzantine replica essentially crash us by refusing messages long enough and we would buffer them until out of memory? It seems like discard must happen eventually.

corecode, line 124: That's why there is a small timeout (1s), and waiting for 2f+1 replicas.

jyellick, line 124: I see.  I'd like to see this reworked, but this is not show stopping to me.

jyellick:
As this fixes a critical issue, and the theoretical bad behavior is still better than the current observed bad behavior, this has my vote to merge.

binhn:
@corecode @jyellick 
could you investigate why CI failed. The raw log is here https://s3.amazonaws.com/archive.travis-ci.org/jobs/137263203/log.txt

```
FAIL    github.com/hyperledger/fabric/consensus/obcpbft 12.759s

2016/06/13 14:52:02 Registering WQaYsCMYBhr1uR7O2TplqaQvUOsgyriAav3HhPIuu+cSgsQEgiotxabPo1Tm/+nes3v+unQ/7wIC7EJv4lpdyg== into custody with timeout 2016-06-13 14:52:02.964685189 +0000 UTC
2016/06/13 14:52:02 Batch replica 1 received new consensus request: WQaYsCMYBhr1uR7O2TplqaQvUOsgyriAav3HhPIuu+cSgsQEgiotxabPo1Tm/+nes3v+unQ/7wIC7EJv4lpdyg==
--- FAIL: TestBatchStaleCustody (0.00s)
panic: runtime error: index out of range [recovered]
    panic: runtime error: index out of range
```

binhn:
@corecode Thanks.

PR #1795, opened 2016-06-10
Fixhash#issue1789

### LINKED ISSUE #1789: chaincode ID computation ignores chaincode content
## Description

In chaincode/platforms/golang/hash.go, files are collected from the chaincode path and read in sorted order in preparation for chaincode ID computation. However, the bytes from the files are not copied to the allocated byte buffer and so the computation was using a buffer with null bytes, thus ignoring file content completely. This results in chaincode ID remaining unchanged if content is modified while retaining the file size.
## Describe How to Reproduce

Modify a byte in the chaincode "go" file (don't add or delete, just modify). The chaincode ID will remain unchanged.

Thanks to @angrbrd for discovering this bug.

muralisrini:
Copy the file contents to the allocated byte buffer before computing hash.
## Description

Hash computation code prepares for computation correctly (fetch files/folders in sorted order) but missed copying the contents of the file into the allocated buffer. This code fixes that with a "copy" command.

The fix also adds a set of dummy files and a dummy folder to the well-used chaincode_example02 to make sure to catch changes to chaincode ID (mainly test dependence on sorting order). chaincode_example04 hard-codes chaincode_example02's ID thus verifying the invariance via unit tests. A README.txt explains the intent of the dummy files.
## Motivation and Context

The intent was to compute chaincode ID based on the hash of the contents of the files in the chaincode path. 
Fixes #1789
## How Has This Been Tested?

Enhanced chaincode_example02 with dummy files to test ordering. Existing unit test covers the fix (especially TestChaincodeInvokeChaincode).
## Checklist:


muralisrini:
thanks @angrbrd

binhn:
LGTM

rameshthoomu:
reverify jenkins

binhn:
@rameshthoomu CI is still failing.  Can we just go with Jenkins now?

rameshthoomu:
@binhn: Jenkins is configured and builds are triggering successfully but have some issues on parallel builds on same machine. @bramwelt and me are working on these issues. Travis is failing because of timeout issues. Specific to this build error, `No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.` Not sure why input is not receiving for 10 mins.

christo4ferris:
@muralisrini please add tests that validate the correct functioning

muralisrini:
@christo4ferris : I used existing test TestChaincodeInvokeChaincode in exectransaction_test.go to test for this fix as well. The hardcoded example02's ID in example04 should never change.  I added dummy files and folder to example02 chaincode folder to strengthen the test further so as to make sure sorted ordering is tested as well. This way the fix is tested without increasing the chaincode test time. 

What do you think ? If you feel a new test is better, I can add one.

christo4ferris:
@muralisrini if that test was effective, it would have caught the bug. It didn't, 'nuf said. You need to add a test that will catch that the file(s) have changed. Further, you could add a unit test that hashed a file's contents and check the result against the expected result.

muralisrini:
@christo4ferris the test is effective to the extent that it tests for id remaining unchanged when code does not change (so when example02 is changed, the test will catch that). But you are right - it does not test the condition that ID should change when file's content is changed.

I will add add a unit test to test hash ID is changed when content changes.

muralisrini:
@christo4ferris added unit test for hash computation

christo4ferris:
restarted Travis, thanks for adding the tests @muralisrini

muralisrini:
@christo4ferris one more change (pretty) please ..... should have checked random quantities are always different.

bramwelt:
reverify jenkins

Issue #718, opened 2016-02-23
database or disk is full


dongmingh:
The database or disk is full when I ran an overnight stress test with invoke.

From the obcca-server window:
ERROR: 2016/02/23 11:53:31 ca.go:205: database or disk is full
ERROR: 2016/02/23 11:53:31 tca.go:252: database or disk is full
ERROR: 2016/02/23 11:53:31 ca.go:205: database or disk is full
ERROR: 2016/02/23 11:53:31 tca.go:252: database or disk is full

From the obc-peer window:
11:53:31.228 [crypto.client.lukas] NewChaincodeExecute -> ERRO 2fb626 Failed getting next transaction certificate [Failed loading TCerts from TCA].
11:53:31.228 [rest] Invoke -> ERRO 2fb627 {"Error": "Invoking Chaincode -- Failed loading TCerts from TCA"}
11:53:31.232 [crypto.client.lukas] callTCACreateCertificateSet -> ERRO 2fb628 Failed requesting tca create certificate set [rpc error: code = 2 desc = "database or disk is full"].
11:53:31.232 [crypto.client.lukas] NewChaincodeExecute -> ERRO 2fb629 Failed getting next transaction certificate [Failed loading TCerts from TCA].
11:53:31.232 [rest] Invoke -> ERRO 2fb62a {"Error": "Invoking Chaincode -- Failed loading TCerts from TCA"}
11:53:31.236 [crypto.client.lukas] callTCACreateCertificateSet -> ERRO 2fb62b Failed requesting tca create certificate set [rpc error: code = 2 desc = "database or disk is full"].
11:53:31.237 [crypto.client.lukas] NewChaincodeExecute -> ERRO 2fb62c Failed getting next transaction certificate [Failed loading TCerts from TCA].
11:53:31.237 [rest] Invoke -> ERRO 2fb62d {"Error": "Invoking Chaincode -- Failed loading TCerts from TCA"}
11:53:31.243 [crypto.client.lukas] callTCACreateCertificateSet -> ERRO 2fb62e Failed requesting tca create certificate set [rpc error: code = 2 desc = "database or disk is full"].
11:53:31.243 [crypto.client.lukas] NewChaincodeExecute -> ERRO 2fb62f Failed getting next transaction certificate [Failed loading TCerts from TCA].
11:53:31.244 [rest] Invoke -> ERRO 2fb630 {"Error": "Invoking Chaincode -- Failed loading TCerts from TCA"}

[...]

PR #1801, opened 2016-06-11
Mutex introduced in Membership Services

### LINKED ISSUE #1800: Missing synchronization around DB operations in Membership Services
## Description

A synchronization mechanism is required for DB operations in order to avoid inconsistencies due to multithreading.
## Describe How to Reproduce

Registering users in parallel using node hlc-tester.js the following error appears in member services
ERROR: 2016/06/07 22:27:53 eca.go:252: database is locked

jeroiraz:
## Description

Provide mutual exclusion in user registration, ecert generation and attributes population.
## Motivation and Context

This change prevents inconsistencies by taking into account multithreading.
Fix  #1800
## How Has This Been Tested?

Unit and behave tests passed
No new test cases have been created as part of this work
## Checklist:
- [x] No new documentation is required by this change
- [x] No new tests are required by this change
- [x] I have run [goimports](https://godoc.org/golang.org/x/tools/cmd/goimports), [go vet](https://golang.org/cmd/vet/), and [golint](https://github.com/golang/lint). I have cleaned up all valid errors and warnings in code I have added or modified. The goal is clean, consistent, and readable code.

Signed-off-by: jeronimo.irazabal jeronimo.irazabal@gmail.com/iraza@ar.ibm.com

### DIFF
--- a/membersrvc/ca/ca.go
+++ b/membersrvc/ca/ca.go
@@ -409,7 +409,7 @@ func (ca *CA) newCertificateFromSpec(spec *CertificateSpec) ([]byte, error) {
 	return raw, err
 }
 
-func (ca *CA) readCertificate(id string, usage x509.KeyUsage) ([]byte, error) {
+func (ca *CA) readCertificateByKeyUsage(id string, usage x509.KeyUsage) ([]byte, error) {
 	Trace.Println("Reading certificate for " + id + ".")
 
 	var raw []byte
@@ -418,7 +418,7 @@ func (ca *CA) readCertificate(id string, usage x509.KeyUsage) ([]byte, error) {
 	return raw, err
 }
 
-func (ca *CA) readCertificate1(id string, ts int64) ([]byte, error) {
+func (ca *CA) readCertificateByTimestamp(id string, ts int64) ([]byte, error) {
 	Trace.Println("Reading certificate for " + id + ".")
 
 	var raw []byte
--- a/membersrvc/ca/ca.go
+++ b/membersrvc/ca/ca.go
@@ -30,6 +30,7 @@ import (
 	"os"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/hyperledger/fabric/core/crypto/primitives"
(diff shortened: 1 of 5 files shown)


membersrvc/ca/ca.go
```diff
@@ -340,6 +345,9 @@ func (ca *CA) createCertificate(id string, pub interface{}, usage x509.KeyUsage,
 }
 
 func (ca *CA) createCertificateFromSpec(spec *CertificateSpec, timestamp int64, kdfKey []byte) ([]byte, error) {
+	mutex.Lock()
```
adecaro, line 348: @jeronimoirazabal Just a clarification, can't we remove this lock by having a db transaction wrapping the sql insert? SQLite3 should support synchronization automatically, no?

jeroiraz, line 348: @adecaro Transactions in SQLite3 are serialized but would that ensure concurrent transactions are serialized entirely or could them be interleaved? Would that depend on the underlying database driver? If so I think the advantage of this approach (synchronization handled uniformly in go) would be being independent of how concurrent transactions are handled in the db and being able to have mutex logic at application level instead. Does it make sense?

adecaro, line 348: @jeronimoirazabal Yes, it makes sense :) I commented because I'm in general worried about locks. When they are not really needed they just reduce performances.

bramwelt:
reverify jenkins

bramwelt:
reverify jenkins

JonathanLevi:
Yes, aside from the fact that the same locking mechanism is currently in use when requesting TCerts..., I like @jeronimoirazabal 's approach as it also does not introduce any DB/vendor "lock-in". See #718's earlier suggestion to potentially move away from SQLite3?

So, all in all (regardless of Jenkins), this change looks good to me.

JonathanLevi:
@srderson can you please take a quick look at this? The "rather be safe than sorry" approach...

srderson:
LGTM. Thanks for the review @adecaro and thanks for the fix @jeronimoirazabal!

PR #1710, opened 2016-06-03
Membership Services Attribute Management

jeroiraz:
We have been reworking PR #1513 with @andresgaragiola, @diegomasini, @raullaprida and @gromeroar  to split it in smaller PRs. This is the second one and it includes the membership services functions needed for Attribute Management.
## Description

Core functions for Membership Services Attribute Management.  A new CA was introduced (Attribute Certificate Authority) to provide a trusted source for attribute values. TCert generation was modified to include encrypted attributes.
## Motivation and Context

This is a required feature to address #1139
## How Has This Been Tested?

Current behave tests has been run without errors.
Current unit tests has been run without errors.
Additional unit tests where included (aca_test.go and util_test.go in membershipsrvc/ca).
A new behave test is pending we will work with @jeffgarratt on that.
## Checklist:

- [x] I have added a [Signed-off-by](https://github.com/hyperledger/fabric/blob/master/CONTRIBUTING.md#legal-stuff)
- [x] I added new documentation
- [x] I added new tests
- [x] I have run [goimports](https://godoc.org/golang.org/x/tools/cmd/goimports), [go vet](https://golang.org/cmd/vet/), and [golint](https://github.com/golang/lint). I have cleaned up all valid errors and warnings in code I have added or modified.

Signed-off-by: Jeronimo Irazabal jeronimo.irazabal@gmail.com / iraza@ar.ibm.com

PR #1802, opened 2016-06-11
Attribute based asset transfer sample code

diegomasini:
## Description

Modifications to the existing attribute based asset transfer sample code to use the features introduced by PRs #1639 #1710 and #1765  
## Motivation and Context

Attribute based asset transfer was introduced at the early stages of the attributes feature implementation. The sample code was updated to use the features introduced to address issue #1513 
## How Has This Been Tested?

A test is provided to verify that the sample chaincode works correctly. The chaincode file can be found in 'fabric/examples/chaincode/go/asset_management_with_roles/asset_management_with_roles.go'. 
## Checklist:

- [x] No new documentation is required by this change.
- [x] I added new tests (asset_management_with_roles_test.go')

Signed-off-by: Diego Masini diego.masini@gmail.com / damasini@us.ibm.com


adecaro:
@diegomasini It looks like the PR modifies also the shim and not just the example chaincode. Why is this? The description says that the modification leverage preexisting code, no?

diegomasini:
@adecaro it was a change needed as a consequence of removing the use of metadata to send attribute decryption keys.
@binhn could you review this one? This PRs contains the changes needed to make the asset transfer with roles sample code work. It contains additional changes needed after removing the use of the transaction's metadata field to deliver decryption keys. The sample code in PR 1765 (now part of the master) will not work without this changes. I explained this to @smithbk today.

bramwelt:
reverify jenkins

bramwelt:
reverify jenkins

binhn:
@diegomasini The `attr_support.go` is confusing to me:  As we said that attributes are not encrypted (for now), but functions in this module require keys. The keys are also stored in `chaincodeHolderImpl.Metadata`, which causes more confusing with function `GetCallerMetadata` that returns the keys.

I think we should not mess around with key at this point, and as @smithbk suggested, put these functions in a lib to be imported by the chaincode, and submit that as a separate PR. Then fix up this PR to only include the example chaincode as described.

andresgaragiola:
@binhn this code is not using keys because encryption support was disabled in this commit #bd7534c as part of PR #1765 merged last Friday. To support asset transfer we had to add a chaincode holder implementation in order of read attributes from an another TCert (different to shim caller certificate). This implementation has a function GetMetadata because the interface called chaincodeHolder implemented in #1765 has this function, but as I mentioned before all key support was disabled. when we have redefined transaction to support keys we will redefine this code supporting the new way of pass keys and attributes encryption will be activated again. But until that this code is required otherwise the code will not compile. Summarizing we have not added any code related with metadata, keys or attributes encryption in this PR, we are just adding the minimal required code in order to keep compatibility with the merged code.

binhn:
@diegomasini Chat with @andresgaragiola to redo this PR. Andres will either submit a new PR and close this one or fix this one up.

binhn:
closed per @andresgaragiola

andresgaragiola:
Thanks @binhn I have added PR #1819 including this changes plus we have talked today.

PR #1819, opened 2016-06-13
Removed attributes encryption code from shim API public functions and asset transfer sample code

andresgaragiola:
## Description

This PR replaces #1802 and additionally removes key encryption code from shim API public functions until transaction will be modified to support pass attributes keys.
## Motivation and Context

- Currently the implementation uses metadata field to pass encryption keys, this will be modified. Temporarily attributes encryption is not enabled and the involved code was removed from shim API public function in order to avoid confusion in API docs.
- Attribute based asset transfer was introduced at the early stages of the attributes feature implementation. The sample code was updated to use the features introduced to address issue #1513
## How Has This Been Tested?

- attr_support_test.go were fixed in order of support new API.
- asset_management_with_roles tests were fixed in order of support new API.
## Checklist:

- [X] I have added a [Signed-off-by]
- [X] No new documentation is required by this change.
- [X]  I added new tests (asset_management_with_roles_test.go')

Signed-off-by: Andres Garagiola andresgaragiola@gmail.com / garagiol@ar.ibm.com


andresgaragiola:
@binhn this is the PR we have talked.

examples/chaincode/go/asset_management_with_roles/asset_management_with_roles.go
```diff
@@ -43,19 +45,23 @@ func (t *AssetManagementChaincode) Init(stub *shim.ChaincodeStub, function strin
...
 	if err != nil {
-		return nil, errors.New("Failed creating AssetsOnwership table.")
+		return nil, fmt.Errorf("Failed creating AssetsOnwership table, [%v]", err)
 	}
 
 	// Set the role of the users that are allowed to assign assets
 	// The metadata will contain the role of the users that are allowed to assign assets
+
+	// Set the role of the users that are allowed to assign assets
+	// The metadata will contain the role of the users that are allowed to assign assets
```
binhn: Duplicate comments

examples/chaincode/go/asset_management_with_roles/asset_management_with_roles.go
```diff
@@ -17,10 +17,12 @@ limitations under the License.
 package main
```
binhn, line 17: @andresgaragiola This comment on the code pre-date this PR but it is in the sample chaincode file that I think it is important we fix it before merging.

Since this chaincode is a sample, it is worth explaining what is being done, especially the strategy to leverage the attributes for asset assignment and transferring -- why should we use attributes?

The comment block about the chaincode doesn't seem to go with this chaincode, but it looks like a left over from the previous copy. The referenced URL is also confusing.

andresgaragiola, line 17: @binhn I have pushed this comment changes. The link was removed temporarily we could add this again once the content will be fixed and consistent with the implementation. 

I have removed the duplicated comment and added a description of the sample functionality.

binhn:
@andresgaragiola CI build failed but doesn't look like related to this PR, so I am going to merge. Thanks

PR #1822, opened 2016-06-13
CC Tables: Fix buildKeyString() to pay attention to Key cols.

tsariounov:
Fixes buildKeyString() to pay attention to Key flag in table column definition.
## Description

This patch fixes a problem with DeleteRow(), specifically the buildKeyString() function which uses non-Key columns to build the key string.  The patch adds buildKeyString to the ChaincodeStub struct (for access to stub) and uses the same logic as getKeyAndVerifyRow() to inspect the table column definitions and not use the column if the Key bool is false.
## Motivation and Context

This fixes a problem with DeleteRow(), specifically the buildKeyString() function which uses non-Key columns to build the key.  Thus, when deleting a row where not all columns are keys, DeleteRow() fails silently and the row still exists.

No open issue to be fixed.  Combination PR and issue.
## How Has This Been Tested?

Ran unit-tests and behave tests.  

Behave test has call for DeleteRow(), so it tests the affected path.
## Checklist:

When DeleteRow() calls buildKeyString(), that function does not pay
attention to the Key field in the column definition.  Thus, the key
created includes non-Key column values.  Then DeleteRow() tries to
delete this nonexistent row and silently fails.  You think the row is
deleted, but in reality it is not.  This patch fixes buildKeyString()
to be the same as getKeyAndVerifyRow() which is used by InsertRow()
and does pay attention to the Key field in the column definition when
inserting or updating a row.

Signed-off-by: Alex Tsariounov tsariounov@gmail.com

### DIFF
--- a/core/chaincode/shim/chaincode.go
+++ b/core/chaincode/shim/chaincode.go
@@ -510,7 +510,7 @@ func (stub *ChaincodeStub) GetRow(tableName string, key []Column) (Row, error) {
 
 	var row Row
 
-	keyString, err := buildKeyString(tableName, key)
+	keyString, err := stub.buildKeyString(tableName, key)
 	if err != nil {
 		return row, err
 	}
@@ -537,7 +537,7 @@ func (stub *ChaincodeStub) GetRow(tableName string, key []Column) (Row, error) {
 // for C and D as their key.
 func (stub *ChaincodeStub) GetRows(tableName string, key []Column) (<-chan Row, error) {
 
-	keyString, err := buildKeyString(tableName, key)
+	keyString, err := stub.buildKeyString(tableName, key)
 	if err != nil {
 		return nil, err
 	}
@@ -596,7 +596,7 @@ func (stub *ChaincodeStub) GetRows(tableName string, key []Column) (<-chan Row, 
 // DeleteRow deletes the row for the given key from the specified table.
 func (stub *ChaincodeStub) DeleteRow(tableName string, key []Column) error {
 
-	keyString, err := buildKeyString(tableName, key)
+	keyString, err := stub.buildKeyString(tableName, key)
 	if err != nil {
 		return err
 	}
@@ -687,7 +687,7 @@ func getTableNameKey(name string) (string, error) {
 	return strconv.Itoa(len(name)) + name, nil
 }
 
-func buildKeyString(tableName string, keys []Column) (string, error) {
+func (stub *ChaincodeStub) buildKeyString(tableName string, keys []Column) (string, error) {
 
 	var keyBuffer bytes.Buffer
 
@@ -697,10 +697,18 @@ func buildKeyString(tableName string, keys []Column) (string, error) {
 	}
 
 	keyBuffer.WriteString(tableNameKey)
+	table, err := stub.getTable(tableName)
+	if err != nil {
+		return "", err
+	}
 
-	for _, key := range keys {
+	for i, key := range keys {
 
 		var keyString string
+		if ! table.ColumnDefinitions[i].Key {
+			// Only use columns that are marked as being a key
+			continue
+		}
 		switch key.Value.(type) {
 		case *Column_String_:
 			keyString = key.GetString_()
@@ -774,7 +782,7 @@ func getKeyAndVerifyRow(table Table, row Row) ([]Column, error) {
 }
 
 func (stub *ChaincodeStub) isRowPrsent(tableName string, key []Column) (bool, error) {
-	keyString, err := buildKeyString(tableName, key)
+	keyString, err := stub.buildKeyString(tableName, key)
 	if err != nil {
 		return false, err
 	}
@@ -819,7 +827,7 @@ func (stub *ChaincodeStub) insertRowInternal(tableName string, row Row, update b
 		return false, fmt.Errorf("Error marshalling row: %s", err)
 	}
 
-	keyString, err := buildKeyString(tableName, key)
+	keyString, err := stub.buildKeyString(tableName, key)
 	if err != nil {
 		return false, err
 	}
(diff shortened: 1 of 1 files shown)


srderson:
@tsariounov I need to look at the change more carefully, but based on the description, I'm not quite sure how the following test was passing.

```
When I invoke chaincode "table_test" function name "deleteRowTableOne" on "vp0"
        | arg1 |
        | test1|
      Then I should have received a transactionID
      Then I wait up to "25" seconds for transaction to be committed to all peers
      When requesting "/chain" from "vp0"
      Then I should get a JSON response with "height" = "7"

      When I query chaincode "table_test" function name "getRowTableOne" on "vp0":
        | arg1 |
        | test1|
      Then I should get a JSON response with "OK" = "{[]}"
```

Would that not have encountered this issue? Is there a test case we can write to check for the bug? Thanks.

tsariounov:
@srderson The test passes because it only passes one parameter (ie | arg1 | test1 |), which is a Key, to the DeleteRow call.  In this case, the old code will work just fine.  However, if you pass more than one arg, for example (| arg1 | test1 | arg2 | 10 |), then the test will fail.  The patched code will correctly ignore the second (ie arg2 | 10) argument since that column is not a Key.   I'll add a test that shows this and re-send.

srderson:
So if I understand correctly, this creates an API change for `DeleteRow`, `GetRow`, and `GetRows`. Each of these functions takes `keys []Column` as a parameter. Currently, this only contains the columns that are keys. With this change, it requires all columns (if you have a table where non-key columns precede key columns), though the non-key columns can be empty or any value.

I think this could be more confusing is some cases. If I have a table with a column A and B where only B is a key, I now always need to pass an array of size two for each of these functions where before I could just pass an array with only B.

Perhaps we could consider counting the number of columns passed in the `keys []Column` parameter. If it is less than the number of columns in the table, keep the same behavior as today. If it matches the number of columns in the table, use the behavior that you're implementing. My concern is that this is trying to make an intelligent guess as to what the user wants instead of just implementing a single standard behavior.

What do you think?

tsariounov:
Actually, there is NO change in API.  This patch merely fixes the current implementation.  

In the current implementation, if all columns are passed to DeleteRow, GetRow, and GetRows, then the function buildKeyString() will use ALL the columns passed in, no matter if they are marked as being a Key or not!  This is incorrect since non-key columns should not be used to create the key.   After applying the patch, the function buildKeyString now correctly ignores any columns passed to it which are not marked as Key=true.

So again: no change in API and current behavior gets fixed.

I'm still working building a behave test for this.   Incidentally, current master has issue_477 disabled in .behaverc.

Thanks.

srderson:
But what if you pass only the keys instead of ALL columns? That's how it is intended to work today. There was no intention that ALL columns would be passed to `DeleteRow`, `GetRow`, and `GetRows`.

In a table like https://github.com/hyperledger/fabric/blob/master/bddtests/chaincode/go/table/table.go#L488 I think your patch will be ignoring the third key if I only pass keys.

With your patch, when I change https://github.com/hyperledger/fabric/blob/master/bddtests/peer_basic.feature#L130 to pass all columns

```
| arg1 | arg2 | arg3 | arg3 |
| foo2 | 34   | 65   | bar8 |
```

it doesn't find the row which I believe is because when the insert row happened, it only used the first two keys in the key string, but the GetRow attempts to use all three keys.

tsariounov:
If you pass only key columns, the patched buildKeyString will use them as long as the column boolean flag Key is set to true.  If it is set to false, then it will ignore any such columns.   Thus, if all the columns passed are Key columns, both patched and unpatched code will get the same result.  If only some of those columns passed in are Key columns, then the results will be different but that is exactly what one would expect.

Now to your examples:  First one is this table https://github.com/hyperledger/fabric/blob/master/bddtests/chaincode/go/table/table.go#L488 where you say that the "third column would be ignored if I only pass keys."  This is not true.  In that table definition, the SECOND column has it's Key=false, the others are Key=true.  Thus, the patched code will ignore the second column and construct a key composed of columns 1, 3, and 4.  The UNpatched code will use all the columns.   Now, I argue that using ALL columns is incorrect because the second column has Key=false, and the correct way is to ignore the non-key column in generating the final "key".

The second example is here https://github.com/hyperledger/fabric/blob/master/bddtests/peer_basic.feature#L130 where you want to pass four columns: [arg1:foo2, arg2:34, arg3:65, arg4:bar8].  The UNpatched code will use only the FIRST three of these columns to generate the final key because the function getRowTableTwo (https://github.com/hyperledger/fabric/blob/master/bddtests/chaincode/go/table/table.go#L323) only uses the first three arguments.  

Now as, the table definition (https://github.com/hyperledger/fabric/blob/master/bddtests/chaincode/go/table/table.go#L488) marks the SECOND column as a non-key (ie Key=false), then the patched code will use ONLY the first and the third column to generate the key because the second column is not a key and the getRowTableTwo function ignores the fourth argument.

Now which of those is "correct"?  My argument is the latter (ie patched code) because it does NOT use the non-key column whilst the unpatched code uses it (incorrectly IMHO).

I suppose that if the Key=[true|false] variable in the table definition does not make any difference in what columns to select, then why is it even in there?

Do you see my point?
Thanks.

tsariounov:
@srderson: After enabling the behave tests for issue_477 and playing around with this, it looks like I was originally confused on the API of the tables interface.

For GetRow(), all columns passed in must be Key=true columns and all are required to find the row since implicit in the "key" name is that the value for that column is unique.

For GetRows(), all columns passed in must still be Key=true columns; however, not all of those are required and in that case, all matching rows for the partial case of columns are returned.

I was fixated on the order of columns passed in to GetRow() and didn't realize that non-key columns are not allowed to be passed in and the uniqueness of keys makes the order of passed in columns irrelevant.

So, please disregard this PR.  I will close it momentarily.

Thanks.

PR #1826, opened 2016-06-14
Remove syscc from genesis#issue1741

### LINKED ISSUE #1741: PBFT goes to viewchange after a few transactions
## Description

I have a PBFT network a 4 nodes, and trying to measure the performance I could get out of it. Unfortunately, the network always go to view-change, and never (I waited about a minute) succeeds to go to a new view.. First I thought, that it may happen because of the big stress of the transactions sent in, so I put a 1 sec timer between transactions, but I get the same problem. Then I thought, maybe because of short timers, so I changed the obcpbft config to have 10 sec instead of 2 sec, but that did not help either. I attach the logs of the peers.
[pbft_problem.zip](https://github.com/hyperledger/fabric/files/302562/pbft_problem.zip)
## Describe How to Reproduce

4 t2.large nodes on EC2, running hyperledger-peer docker image.
I use [PR 1722](https://github.com/hyperledger/fabric/pull/1722) and [PR 1721](https://github.com/hyperledger/fabric/pull/1721) on top of master, my assumption is, that it should not affect PBFT consensus.
I run [this](https://github.com/esgott/fabric-api/blob/perf-test/api/src/test/java/org/hyperledger/connector/PerfTest.java) for performance testing.
In configuration, I change the following things from default:
- `peer.id:` vp0, vp1, vp2 and vp3
- `peer.address:` EC2_instance_private_address:30303
- `peer.validator.consensus.plugin:` pbft
- `peer.discovery.rootnode:` vp0_private_address:30303

muralisrini:
Remove system chaincodes from genesis block and make them completely non-declarative.
# Description

System chaincodes should be treated as part of the fabric with all policies and rules applicable to upgrading/modifying the fabric applicable to the system chaincodes. It does not make sense to have them committed to the genesis block as user chaincodes would. This PR removes system chaincodes from genesis block
- remove system chaincode declaration from peer/core.yaml
- the system chaincodes are embedded into the fabric programmatically (see importsysccs.go and systemchaincode_test.go)

The registration of system chaincodes is currently disabled if security/privacy mode is on. A approach different from user chaincode treatment of security context is needed to address this correctly (the "user" is the "peer"). Working with @adecaro on this. Commit 3ad5fb1 commit will be removed with the PR for the fix.
## Motivation and Context

As highlighted by #1741, having system chaincodes  - whose deployment don't participate in consensus - committed to genesis block can give raise to problems if the transactions are not the same on all peers (even if the chaincode's context and states are identical). Furthermore, treating them as proper on-block transactions would complicated scenarios such as upgrade. These problems really show that system chaincodes should be considered as part of fabric code.

Fixes #1741
## How Has This Been Tested?

Removed a system chaincode unit test from chaincode folder and made it part of the system_chaincode package. This is the right approach now that system chaincodes are registered programmatically. The test case also serves as an example of registering system chaincodes (ie, the sample needs to be only part of the unit test and does not have to be imported/registered in importsysccs.go).
## Checklist:


ghaskins:
This is great Murali, thank you.  FYI, the CI failure looks related to this patch so please investigate.

corecode:
Doesn't that mean that if peers have system chaincode configured differently, they will return different results for transactions accessing the system chaincode?

binhn:
@corecode SCC operates similar to CC that it must be the same on all peers, both code and configuration that effecting SCC. Also upgrade will require backward operational compatibility, so we must be careful when we get there.

muralisrini:
@ghaskins looks like a timeout somewhere  (it did work on my env though), will take a look soon,

muralisrini:
@ghaskins I was computing timeout in nanosecs in the test case. Fixed it with millis. 

Will take a closer look at this crash though. Perhaps a new PR and testcase will come out of this. Opened an issue #1830 to track this.

muralisrini:
Jenkins failed with 

image

Assuming the only listener for 40303 are chaincode/ tests, I cannot understand why the listener is still up (unless these package tests run in parallel .... ) ?  

In any case its probably good to have for system_chaincode test on a different port (will help future parallel test effort).

Issue #1831, opened 2016-06-14
stub. DeleteTable(tablename) doesnot works...


nitesh7sid:
the stub.DeleteTable(tablename) is not working..
I get error on chaincode terminal as :  21:57:47.240 [shim] DEBU : [mycc]Sending RANGE_QUERY_STATE
2016/06/14 21:57:47 transport: http2Client.notifyError got notified that the client transport was broken EOF.
21:57:47.392 [shim] ERRO : Received error from server: rpc error: code = 13 desc = "transport is closing", ending chaincode stream

And error on the peer terminal as : panic: runtime error: invalid memory address or nil pointer dereference
    panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xb code=0x1 addr=0x0 pc=0x62c861]
goroutine 30 [running]:
panic(0xd5afc0, 0xc820010100)
    /opt/go/src/runtime/panic.go:464 +0x3e6
github.com/hyperledger/fabric/core/chaincode.(_Handler).handleRangeQueryState.func1.1(0xc820284080, 0xc82012b8b0, 0xc820043d80)
    /opt/gopath/src/github.com/hyperledger/fabric/core/chaincode/handler.go:651 +0x261
panic(0xd5afc0, 0xc820010100)
    /opt/go/src/runtime/panic.go:426 +0x4e9
github.com/hyperledger/fabric/core/chaincode.(_Handler).handleRangeQueryState.func1(0xc820284080, 0xc82012b8b0)
    /opt/gopath/src/github.com/hyperledger/fabric/core/chaincode/handler.go:700 +0xbef
created by github.com/hyperledger/fabric/core/chaincode.(*Handler).handleRangeQueryState
    /opt/gopath/src/github.com/hyperledger/fabric/core/chaincode/handler.go:736 +0x3f
hyperledger@hj-ctocto2126:/opt/gopath/src/github.com/hyperledger/fabric/peer$

srderson:
Could you include your chaincode or test case? I just checked the existing test case at https://github.com/hyperledger/fabric/blob/master/bddtests/chaincode/go/table/table.go#L268 and did not see the same failure. Thanks.

srderson:
I have created PR #1833 to fix the nil pointer panic.

@adecaro The root problem is that there is an error decrypting the data when confidentiality is enabled. I found https://github.com/hyperledger/fabric/issues/1049. Is the correct way to resolve this to merge PR #1373?

[...]

PR #1833, opened 2016-06-14
Fix nil pointer panic. Issue #1831

srderson:
## Description

Fixes a typo that was resulting in a nil pointer panic.
## Motivation and Context

Partial fix for issue #1831. Still need to work with @adecaro to determine cause of error now that panic is fixed.
## How Has This Been Tested?

As noted above, this is a partial fix so that the error message will be properly logged without a panic. Once the error message is revealed a second fix will be needed in addition to a test case that tests tables with confidentiality enabled.
## Checklist:


christo4ferris:
LGTM

PR #1836, opened 2016-06-14
Run docker builds as the host user

ghost:
DO NOT MERGE

PR #1834, opened 2016-06-14
Run docker builds as the host user

ghaskins:
Modify build process to lower the privilege level of docker-based build steps so that we avoid the problem introduced in Issue #1790
## Description

We utilize the --user feature of docker-run to lower the priviledge level of our build
## Motivation and Context

Current methodology results in certain build artifacts to be owned by root, which is both dangerous and problematic in that non-root users cannot perform operations such as "make clean" 
## How Has This Been Tested?

Ran various scenarios and checked that build artifacts are now owned by vagrant/vagrant.  Also ran "make checks" at the conclusion of the directed testing to ensure nothing else broke.
## Checklist:

- [x] I have added a [Signed-off-by](https://github.com/hyperledger/fabric/blob/master/CONTRIBUTING.md#legal-stuff)
- [x] Either no new documentation is required by this change, OR I added new documentation
- [x] Either no new tests are required by this change, OR I added new tests
- [] I have run [goimports](https://godoc.org/golang.org/x/tools/cmd/goimports), [go vet](https://golang.org/cmd/vet/), and [golint](https://github.com/golang/lint). I have cleaned up all valid errors and warnings in code I have added or modified. These tools may generate false positives. Don't be worried about ignoring some errors or warnings. The goal is clean, consistent, and readable code.

Signed-off-by: Greg Haskins gregory.haskins@gmail.com

Fixes #1790

Signed-off-by: Greg Haskins gregory.haskins@gmail.com

PR #1840, opened 2016-06-15
Revert "Optimize peer/membersrvc -image generation"

ghaskins:
DO NOT MERGE

This reverts commit 85837442c2dfb92cf0be0455581387bfbe0c2f48.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -121,17 +121,6 @@ $(GOPATH)/bin/%:
 	@echo "Installing chaintool"
 	@cp devenv/tools/chaintool $@
 
-# We (re)build a package within a docker context but persist the $GOPATH/pkg
-# directory so that subsequent builds are faster
-build/docker/bin/%: build/image/src/.dummy $(PROJECT_FILES)
-	$(eval TARGET = ${patsubst build/docker/bin/%,%,${@}})
-	@echo "Building $@"
-	@mkdir -p build/docker/bin build/docker/pkg
-	@docker run -i \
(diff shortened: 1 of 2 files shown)


rjones-lf:
reverify jenkins

rjones-lf:
reverify jenkins

rjones-lf:
reverify jenkins

ghost:
well, it passed on Z

ghaskins:
Interesting...checking on the other PR now...

ghaskins:
#1836 / #1834 would be the preferred solution here, but those still seem to be a mess w.r.t. CI.  Do we know what the deal is there?  E.g. are those CI processes failing because 1836/1834 do not fix the bug, or is there something else going on?

ghaskins:
One more thing to note: I haven't looked closely, but I do not think _this_ PR completely removes the aspects of UID 0 (there are two docker targets that do the -v mapping, and this reverts only one of them).  

So....if this direction starts to look like the only viable path forward, we would need to address that second leak.

ghost:
I think #1834 is the way to go - #1836 was my experiment in putting #1834 at the tip of master to see if it helped. I can abandon #1836.

Looking at the most recent failures, it didn't look like it was part of your change that caused the issue.

rjones-lf:
reverify jenkins

ghaskins:
Closing this for now because I am advocating #1834 for inclusion

PR #1855, opened 2016-06-15
Fix a bunch of go vet issues

srderson:
## Description

Fixes go vet issues reported at https://goreportcard.com/report/github.com/hyperledger/fabric#go_vet
## Motivation and Context

Fix issues reported at https://goreportcard.com/report/github.com/hyperledger/fabric#go_vet
## How Has This Been Tested?

Ran `make checks`. No new tests as there is no new functionality.
## Checklist:

- [x] I have added a [Signed-off-by](https://github.com/hyperledger/fabric/blob/master/CONTRIBUTING.md#legal-stuff)
- [x] Either no new documentation is required by this change, OR I added new documentation
- [x] Either no new tests are required by this change, OR I added new tests
- [x] I have run [goimports](https://godoc.org/golang.org/x/tools/cmd/goimports), [go vet](https://golang.org/cmd/vet/), and [golint](https://github.com/golang/lint). I have cleaned up all valid errors and warnings in code I have added or modified. These tools may generate false positives. Don't be worried about ignoring some errors or warnings. The goal is clean, consistent, and readable code.

Signed-off-by: Sheehan Anderson sheehan@us.ibm.com

PR #1866, opened 2016-06-16
Run go vet as part of make linter. Resolves #1859

### LINKED ISSUE #1859: Add go vet command to make linter
## Description

Add the `go vet` to the `make linter` command in the make file.
## Describe How to Reproduce

This will add `go vet` to the `make linter` command. @genggjh has helped to fix a huge chunk of the `go vet` issues and PR #1855 should finish it off. We should therefore run `go vet` along with `make checks`.

srderson:
## Description

This will run the `go vet` command as part of `make linter`
## Motivation and Context

Fixes #1859 
Now that most of the `go vet` errors are resolved as of PR #1855, we should be running `go vet` as part of `make linter` so developers can identify errors.
## How Has This Been Tested?

Ran `make checks`. There are no new tests as this only modifies the make file.
## Checklist:


ghaskins:
Awesome, @srderson thanks.  Does "go vet" return a non-zero $? when it finds a problem such that make flags the build as failed?

ghaskins:
Hmm, looks like it doesn't fail the build.  Here's a suggestion: perhaps pipe the current output through "sort" and then to .govet-exceptions.  For runs, generate a new output and diff them.  Any non-zero diff should be considered a failure of "make linter".  We can accept modifications to .govet-exceptions as part of the PR process.

srderson:
Yes, but I need to modify the command to be `go vet ./... | grep -v '^vendor\/' |`. I was waiting to confirm that PR #1855 resolved all `go vet` issues. Maybe merge that one first, then I can update

srderson:
@ghaskins I've submitted PR #1870 to fix the remaining go vet errors. Once CI finishes, please have a look and merge if it looks good. At that point I'll come back to this PR and update the `go vet` command in the make file so returns the correct exit status on error.

Makefile
```diff
@@ -62,7 +62,7 @@ go.fqp.gomega := github.com/onsi/gomega
 
 all: peer membersrvc checks
 
-checks: unit-test behave linter
+checks: linter unit-test behave
```
ghaskins, line 65: I think this makes a lot of sense.  Note that the TravisCI config explicitly calls each of these so if you wanted to change the order for that environment, it will need a tweak there too.

srderson:
@ghaskins I've updated the make file with a couple of changes.
1. I run `linter` first in `make checks`. I thought this may be useful because it's the quickest task so the developer can see any issues without waiting for tests to complete.
2. I've decided to run `go vet` for each folder with code. This seemed to be the most straight forward and easily readable way to exclude the vendor folder and have `go vet` return a non-zero status on error.

There are no `go vet` issues after PR #1870

Makefile
srderson, line 65: I'll look into changing the order in Travis also

srderson, line 65: I've updated Travis to run `linter` before the tests.

ghaskins:
Very nice.. I'll just wait for CI confirmation and then this can go in.

ghaskins:
This LGTM,

@rameshthoomu any comments?

rameshthoomu:
@srderson : LGTM

Issue #662, opened 2016-02-12
Capture stacktrace on error


binhn:
We need to log the stacktrace with error to tell where in the code.  This package might help https://github.com/go-errors/errors

christo4ferris:
You're suggesting to dump the stack trace on every error? https://golang.org/pkg/runtime/#Stack

genggjh:
Do not understand this requirement clearly. Do we need a new log function to trace the error stack trace like log.Trace? Or need we dump the stack trace into a log file on peer panic?

christo4ferris:
@binhn we need some clarity on the ask - what specifically are you looking for in a PR?

ibmmark:
@binhn can you assign this to @genggjh

christo4ferris:
@ibmmark it needs some clarity. What specifically is needed? Do we want this on every error? Do we want it governed by an environment variable/config property?

genggjh:
@ibmmark, agree with @christo4ferris. We need know more detail requirement before do it. :-)

binhn:
@christo4ferris @genggjh Somehow I didn't see this until hackathon today. Apologize for the lated reply.

In production, logging level is normally set to ERROR, so calls to logger.Error() will output the data, which in Golang, only an error message without the stacktrace. For these instances, we need to capture the stacktrace to help in serviceability as we don't have the debug log to assist in the detail.

PR #1871, opened 2016-06-16
Fixes #653 : [Several serious container errors are being reported at …

### LINKED ISSUE #653: Several serious container errors are being reported at the DEBUG level
There are several log messages related to container handling that I think should be promoted from DEBUG level to a higher level, as these represent errors from the user's perspective. This will make it easier for users (like me) to figure out what's going on, as these messages are normally not printed.

Here are a few examples:

```
12:38:25.785 [container] build -> DEBU 029 Error building Peer container: Error: image bcbrock/golang:latest not found
12:38:25.800 [container] start -> DEBU 031 start-container does not exist, attempting to create API error (404): no such id: dev-vp1-60caf3632e0f43a5a7f14f27115bfaf84f76ee59d3490bbb337703e098ada77fc7b7bddd093f354e8ef89f644330131e5799c966503070b94eb1a83cbc9dc10c
12:38:25.803 [container] start -> DEBU 033 start-could not recreate container no such image
12:57:11.180 [chaincode] LaunchChaincode -> DEBU 09c launchAndWaitForRegister failed Error starting container: no such image
```

ezhilvendhan:
## Description

Errors written on DEBUG Level are moved to ERROR level
## Motivation and Context

This fix will ensure that error messages are not overlooked as debug messages
Fixes #653 
## How Has This Been Tested?

Yes, Ran the Unit tests successfully

## Checklist:

Signed-off-by: Ezhil Vendhan uibuddha@gmail.com

…the DEBUG level]


binhn:
@ezhilvendhan Hi, thanks for the PR; however, I think the change may not be entirely accurate as in some instances, the errors are propagated up to the chain, and so we wouldn't want to output the errors but leave that to the caller to handle. We really need the ability to capture the stacktrace at the time the error occurs #662 so that when the error is logged, we know where it happened in the code.

srderson:
@ezhilvendhan Please do not modify files in the /vendor folder. These are external packages and should only be updated via [govendor](https://github.com/hyperledger/fabric/blob/master/CONTRIBUTING.md#adding-or-updating-go-packages-) and PRs to their respective repositories if necessary. Thanks.

ezhilvendhan:
@binhn @srderson Thanks for your comments. Will reverting the changes on vendor folder be sufficient to get this pull request accepted?

binhn:
@ezhilvendhan I think we should not log the error if the caller is also logging the same error, so really have to walk up the chain otherwise we would end up with the same error messages in the log, difficult to support.

I agreed we should not modify the vendor packages here.

ezhilvendhan:
@binh I have updated as per your suggestion. Please let me know if it works.

bcbrock:
This PR does not fix **any** of the issues first raised in issue #653! I am concerned by the fact that numerous fatal errors that occur during chaincode deployment are being squashed.

Issue #1857, opened 2016-06-16
State Transfer Errors Running a **busywork** target with multiple clients


bcbrock:
## Description

@jyellick Jason, When running busywork PBFT Batch with 2 clients, one of the peers seems to get stuck in state transfer. In this example it is VP2.

[logs.tar.gz](https://github.com/hyperledger/fabric/files/317403/logs.tar.gz)
## Describe How to Reproduce

This is the target I'm trying to run, it does not make it past 2 clients. It fails even w/o the PBFT configs of the network shown below. One client tests seem to work OK.

If you run this you will see a "...state may be inconsistent ..." error. However if you "curl" the "/chain" API of the peer that gives this error you will see that it is not making progress. It is stuck at a block height less than the expected 16387 shown by the other peers.

```.PHONY: s2b 
s2b:
    for clients in 2 4 8 16 32; do \
    CORE_PBFT_GENERAL_TIMEOUT_REQUEST=5s \
    CORE_PBFT_GENERAL_TIMEOUT_VIEWCHANGE=5s \
    CORE_PBFT_GENERAL_VIEWCHANGEPERIOD=0 \
    CORE_PBFT_GENERAL_NULLREQUEST=1s \
    $(NETWORK) -batch 4; \
    set -e; ./driver \
        -clients $$clients \
        -transactions $$((16384 / $$clients)) \
        -peerBurst 100 \
        -interlock \
        -keepLog \
        -checkAgreement -dupCheck; \
    done

```
```

**Comment by bcbrock** [comment]
It is not necessary to use multiple clients, it seems to be sufficient for the network to be receiving transactions on multiple peers simultaneously. If in the above you run with 1 client and change 

```
-peerBurst 100
```

 to 

```
-peerBurst 10 -netBurst 10
```

 it will fail the same way. (`-peerBurst 100` sends 100 TX to a random peer then waits for them to all complete before moving on. `-peerBurst 10 -netBurst 10` sends 10 bursts of 10 TX to random peers then waits.)

UPDATE: This is bizarre - The multi-client runs started working this morning. But the -peerBurst/-netBurst runs consistently fail with 1 client.

ANOTHER UPDATE: There must be a subtle timing dependency here. I noticed that the "driver" processes were taking a lot of CPU time. I assumed this was happening when they polled for transaction completion, so added some code to try to delay the polling loops. This somehow makes the runs work. 

The bottom line is that it may be difficult to reproduce this issue.

jyellick:
[...]

PR #1877, opened 2016-06-16
Stabilize PBFT under stress with periodic viewchange

jyellick:
## Description

This changeset contains three new commits, to accomplish three things.
1. If a view change occurs and selects an initial checkpoint which is higher than our lastExec, but which can be reached via commit certificates, we delay processing of the new view until execution completes.
2. On view change, resubmit the outstanding requests, and enhance the per transaction logging by printing transaction UUID when debugging is enabled.
3. Changes the ledger's `VerifyChain` function to consistently return the lowest validated block number consistently. @srderson If you especially could take a look at the last commit to this change.
## Motivation and Context

This changeset is motivated by the busywork stress2b test.  The problems were observed with the following environmental configuration:

```
CORE_PBFT_GENERAL_TIMEOUT_REQUEST=10s
CORE_PBFT_GENERAL_TIMEOUT_VIEWCHANGE=10s
CORE_PBFT_GENERAL_VIEWCHANGEPERIOD=1
CORE_PBFT_GENERAL_TIMEOUT_NULLREQUEST=1s
```

This causes the network to change views at every checkpoint, resulting in over 1000 view changes during the execution.

This causes frequent small state transfers, and revealed a bug in the ledger's `VerifyChain` function.  This function would typically return the lowest block number in the range whose hash could be validated, but under other error conditions, like missing blocks, failing hashes, etc., the function would return the block number which had the error, sometimes leading state transfer to believe that a particular block was valid, when in fact it was otherwise corrupt.
## How Has This Been Tested?

These changes were tested against the stress2b busywork test as described above.  As issues were revealed, new unit tests were added for each.
## Checklist:

### DIFF
--- a/consensus/obcpbft/obc-batch.go
+++ b/consensus/obcpbft/obc-batch.go
@@ -129,11 +129,11 @@ func (op *obcBatch) submitToLeader(req *Request) events.Event {
 	leader := op.pbft.primary(op.pbft.view)
 	if leader == op.pbft.id && op.pbft.activeView {
 		return op.leaderProcReq(req)
-	} else {
-		logger.Debugf("Replica %d add request %v to its outstanding store", op.pbft.id, req)
-		op.outstandingReqs[req] = struct{}{}
-		op.startTimerIfOutstandingRequests()
 	}
+
+	logger.Debugf("Replica %d add request %v to its outstanding store", op.pbft.id, req)
+	op.outstandingReqs[req] = struct{}{}
+	op.startTimerIfOutstandingRequests()
 
 	return nil
 }
--- a/consensus/obcpbft/obc-batch.go
+++ b/consensus/obcpbft/obc-batch.go
@@ -18,7 +18,6 @@ package obcpbft
 
 import (
 	"fmt"
-	"reflect"
 	"time"
 
 	"github.com/hyperledger/fabric/consensus"
@@ -49,7 +48,7 @@ type obcBatch struct {
 	incomingChan chan *batchMessage // Queues messages for processing by main thread
 	idleChan     chan struct{}      // Idle channel, to be removed
 
-	outstandingReqs map[*Request]struct{}
+	reqStore *requestStore // Holds the outstanding and pending requests
 
 	persistForward
 }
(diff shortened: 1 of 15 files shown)


jyellick:
This PR should fix issue #1857

jyellick:
~~Travis log size exceeded out in what seem to be unrelated tests if someone could rerun.~~

Looks like `VerifyChain` is used outside of ledger tests, fixing. (And done, see ef66b00 below)

binhn:
@jyellick I saw a coupe of failures on CI.  The 2nd was just the limit. @rameshthoomu not sure if we can increase the limit.

```
2016/06/16 21:05:47 Replica 0 received an unknown message type <nil>
--- FAIL: TestOutstandingReqsResubmission (0.00s)
    obc-batch_test.go:220: All requests should have been resubmitted after view change
2016/06/16 21:05:47 Replica 1 obtaining startup information
```

```
The log length has exceeded the limit of 4 MB (this usually means that the test suite is raising the same exception over and over).
```

rameshthoomu:
@binhn : Travis is supports max of 10000 lines and 4 MB size in total. Also, we don't have any option to increase the size.

corecode:
This PR carries a number of commits that have been squash committed before.

consensus/obcpbft/obc-batch.go
```diff
@@ -121,16 +122,19 @@ func (op *obcBatch) submitToLeader(req *Request) events.Event {
...
 	leader := op.pbft.primary(op.pbft.view)
 	if leader == op.pbft.id && op.pbft.activeView {
 		return op.leaderProcReq(req)
-	} else {
-		logger.Debugf("Replica %d add request %v to its outstanding store", op.pbft.id, req)
-		op.outstandingReqs[req] = struct{}{}
-		op.startTimerIfOutstandingRequests()
 	}
 
+	logger.Debugf("Replica %d add request %v to its outstanding store", op.pbft.id, req)
```
corecode: please use hashReq(req) to avoid logspam

jyellick: Good note, I'll address

jyellick:
@corecode I tried to do the 'rebase onto', but after about 20 minutes of trying to resolve merge conflicts in places like chaincode which had nothing to do with my PR, I've pushed as is.  I'll try to clean it up later tonight.  The diff should remain the same.

Edit: Was able to get it to work manually picking the commits, down to the real 6 now

jyellick:
My mistake, I fixed the ledger test to honor the new `VerifyChain` behavior, but did not fix the state transfer mock ledger framework to do so, causing the unit test failures, pending other local test verification, I'll be pushing a fix shortly.

Edit: Looks like CI is good now, I'll be checking this weekend for any feedback which needs to be incorporated in.

binhn:
Fix issue #1907

PR #1878, opened 2016-06-16
Remove the unneeded obc-classic code

jyellick:
## Description

This changeset removes the obc-classic code. 
## Motivation and Context

The obc classic plugin is deprecated and no longer supported, so leaving its code in tree is of little to no value.

Since obc-classic can be easily emulated by setting batchsize=1, the config and tests are left in place as is, to minimize the diff, and to allow other consumers to transition to using batch.

(This is not a critical merge, but removing code is always good)
## How Has This Been Tested?

This introduces no new function, so CI should be sufficient.
## Checklist:
- [X] Either no new documentation is required by this change, OR I added new documentation. 


consensus/obcpbft/obc-classic_test.go
```diff
@@ -25,13 +25,10 @@ import (
...
-}
-
 func obcClassicHelper(id uint64, config *viper.Viper, stack consensus.Stack) pbftConsumer {
 	// It's not entirely obvious why the compiler likes the parent function, but not newObcClassic directly
-	return newObcClassic(id, config, stack)
+	config.Set("general.batchsize", 1)
+	return newObcBatch(id, config, stack)
 }
 
 func TestClassicNetwork(t *testing.T) {
```
kchristidis: Is this test really needed? And what do you say about moving the other two tests to the batch-test file? (With the proper renaming obviously.)

christo4ferris:
+1 to removing this 
LGTM

kchristidis:
As I've noted in the comments, moving over the last two tests to the obc-batch_test.go is probably the way to go.

consensus/obcpbft/obc-classic_test.go
jyellick: I can do this here, if we'd prefer.  I was thinking to wait to re-arrange the tests for another PR so as to minimize the diff, but I'm happy either way.

kchristidis: I'd argue that doing so here falls neatly the PR's scope and the changeset won't really explode. But it's a minor thing at any rate.

jyellick:
@kchristidis As requested, those classic tests have been consolidated under batch, in 7eeb2bb above.

kchristidis:
This is good to go.

jyellick:
Sorry for the churn, was cleaning up the dead and merged branches through the github UI and misclicked.

PR #1903, opened 2016-06-17
REST API: Fix JSON encoding

### LINKED ISSUE #1475: REST API can be tricked to respond with malformed JSON
## Description

In some places the REST API responses with inlined strings inside a JSON string, instead of encoding a proper JSON. This allows the client to cause the server to respond with malformed JSON. I can't yet find any vulnerability here, but it seems not too pretty (and [OWASP recommends against this practice](https://www.owasp.org/index.php/AJAX_Security_Cheat_Sheet#Avoid_building_XML_or_JSON_by_hand.2C_use_the_framework)).

As an example, the `GET /transactions/{UUID}` API endpoint will "paste" the given UUID as-is in an error response in case the UUID is not found. Since the given UUID can contain chars like " and ' and } it can cause the server to actually return malformed JSON. 
## Describe How to Reproduce

Make sure the peer node is running. Then from inside vagrant run this:

```
curl -v 'http://127.0.0.1:5000/transactions/garbage"\}%20%20%20%20\{"OK":"It%20does%20not%20matter%20that%20it'
```

Here is how it acts on my machine:

```
*   Trying 127.0.0.1...
> GET /transactions/garbage"}%20%20%20%20{"OK":"It%20does%20not%20matter%20that%20it HTTP/1.1
> User-Agent: curl/7.35.0
> Host: 127.0.0.1:5000
> Accept: */*
> 
< HTTP/1.1 404 Not Found
< Date: Fri, 13 May 2016 18:29:50 GMT
< Content-Length: 85
< 
{"Error": "Transaction garbage"}    {"OK":"It does not matter that it is not found."}
```

dubek:
Use proper JSON encoder when building responses of the REST API. This PR fixes all the endpoints **except**:
-  `/registrar/*` endpoints - still in progress (I'd appreciate feedback before I keep working on those)
- `/devops/*` endpoints - these are deprecated, so I skipped them.
## Description

See #1475 for the original problem of hand-crafting JSON responses. This PR adds unit-tests of the REST API, and fixes the JSON encoding to use a proper JSON encoder even for simple `{"Error": "..."}` responses.

Note that in order to add unit-tests for the REST API (which were non-existing) I did some small refactoring in `rest_api.go`.

The changes are split into several commits, each commit fixing (and testing) only one endpoint. It will therefore makes sense to read this review commit-by-commit. (Let me know if this should be broken to separate PRs.)
## Motivation and Context

Fixes #1475 
## How Has This Been Tested?

Added unit tests to the REST API, except the deprecated `/devops/*` endpoints (and currently also misses the `/registrar/*` - still in progress).

Also ran a peer and tried the problematic curl request from #1475 and verified that the response is indeed valid JSON (the error string is properly escaped).
## Checklist:

Signed-off-by: Dov Murik <dmurik@us.ibm.com>

### DIFF
--- a/core/rest/rest_api.go
+++ b/core/rest/rest_api.go
@@ -292,14 +292,16 @@ func (s *ServerOpenchainREST) GetEnrollmentID(rw web.ResponseWriter, req *web.Re
 	// Returns /var/hyperledger/production/client/
 	localStore := getRESTFilePath()
 
+	encoder := json.NewEncoder(rw)
+
 	// If the user is already logged in, return OK. Otherwise return error.
 	if _, err := os.Stat(localStore + "loginToken_" + enrollmentID); err == nil {
 		rw.WriteHeader(http.StatusOK)
-		fmt.Fprintf(rw, "{\"OK\": \"User %s is already logged in.\"}", enrollmentID)
+		encoder.Encode(restResult{OK: fmt.Sprintf("User %s is already logged in.", enrollmentID)})
 		restLogger.Infof("User '%s' is already logged in.\n", enrollmentID)
 	} else {
 		rw.WriteHeader(http.StatusUnauthorized)
-		fmt.Fprintf(rw, "{\"Error\": \"User %s must log in.\"}", enrollmentID)
+		encoder.Encode(restResult{Error: fmt.Sprintf("User %s must log in.", enrollmentID)})
 		restLogger.Infof("User '%s' must log in.\n", enrollmentID)
 	}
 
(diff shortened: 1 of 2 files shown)


dubek:
As I already said,

christo4ferris:
@dubek when you sign-off the signature needs the email address enclosed in angle brackets eg. &lt;dmurik@us.ibm.com>

srderson:
LGTM. Great to see new test cases in this area. Thanks!

dubek:
@srderson Thanks for the feedback. I'll complete fixing the remaining (non-deprectaed) REST API endpoints in a separate PR.

I also consider moving the code of the deprecated API endpoints as-is to `rest_api_deprecated.go`. Is that a good idea?

PR #1904, opened 2016-06-17
Fix viper flags binding so that command line options work Issue-578

### LINKED ISSUE #578: obc-peer command-line flags do not (can not) work

lehors:
## Description

This fixes the viper flag bindings so that settings via yaml configuration can be overridden using command line options (aka flags).
## Motivation and Context

Fixes #578
## How Has This Been Tested?

Before:

```
$ peer node start --peer-chaincodedev --peer-tls-enabled=true --peer-tls-cert-file=./bddtests/tlsca.cert --peer-tls-key-file=./bddtests/tlsca.priv
18:17:55.995 [crypto] main -> INFO 001 Log level recognized 'info', set to INFO
18:17:55.998 [main] serve -> INFO 002 Running in chaincode development mode
...
18:17:56.203 [rest] StartOpenchainRESTServer -> INFO 017 Initializing the REST service on 0.0.0.0:5000, TLS is disabled.

```

After:

```
$ peer node start --peer-chaincodedev --peer-tls-enabled=true --peer-tls-cert-file=./bddtests/tlsca.cert --peer-tls-key-file=./bddtests/tlsca.priv
18:20:54.909 [crypto] main -> INFO 001 Log level recognized 'info', set to INFO
18:20:54.914 [main] serve -> INFO 002 Running in chaincode development mode
...
18:20:55.073 [rest] StartOpenchainRESTServer -> INFO 018 Initializing the REST service on 0.0.0.0:5000, TLS is enabled.

```

Note: Several people have argued that the broken flags should simply be removed rather than fixed but not all flags can be removed because some settings can only be set using flags so I've concluded that until we do a complete overall of the configuration framework it is better to minimize the changes by simply fixing what's broken and leaving it at that for now.

Credits go to @kchristidis for finding the bug.
## Checklist:

### DIFF
--- a/peer/main.go
+++ b/peer/main.go
@@ -247,11 +247,11 @@ func main() {
 
 	flags.BoolVarP(&chaincodeDevMode, "peer-chaincodedev", "", false, "Whether peer in chaincode development mode")
 
-	viper.BindPFlag("peer_tls_enabled", flags.Lookup("peer-tls-enabled"))
-	viper.BindPFlag("peer_tls_cert_file", flags.Lookup("peer-tls-cert-file"))
-	viper.BindPFlag("peer_tls_key_file", flags.Lookup("peer-tls-key-file"))
-	viper.BindPFlag("peer_gomaxprocs", flags.Lookup("peer-gomaxprocs"))
-	viper.BindPFlag("peer_discovery_enabled", flags.Lookup("peer-discovery-enabled"))
+	viper.BindPFlag("peer.tls.enabled", flags.Lookup("peer-tls-enabled"))
+	viper.BindPFlag("peer.tls.cert.file", flags.Lookup("peer-tls-cert-file"))
+	viper.BindPFlag("peer.tls.key.file", flags.Lookup("peer-tls-key-file"))
+	viper.BindPFlag("peer.gomaxprocs", flags.Lookup("peer-gomaxprocs"))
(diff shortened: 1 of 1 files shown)


kchristidis:
👍

christo4ferris:
LGTM

binhn:
Good eyes. It took me a few seconds to figure out the differences :-)

PR #1908, opened 2016-06-17
Fixing asset-mgmt-with-roles unit test

angrbrd:
This PR contains the required changes for the asset-management-with-roles unit tests to pass.
## Description

The PR includes fixes to the sdk/node/test/unit/asset-mgmt-with-roles.js UT for the Node.js SDK. It also updates the asset_management_with_roles.go chaincode and asset_management_with_roles_test.go Go unit test file. This was required as changes to the chaincode itself had to be made for the Node.js SDK work, which broke the Go unit test.
## Motivation and Context

These fixes were required to complete the last unit test for the Node.js SKD.
## How Has This Been Tested?

To verify this change, pull these changes into a clean environment and execute the commands below.

(1) To run the complete Node.js sdk unit test suite.

```
cd $GOPATH/src/github.com/hyperledger/fabric
make node-sdk-unit-tests
```

(2) To run the asset-management-with-roles.go Go unit test:

```
cd $GOPATH/src/github.com/hyperledgerfabric/examples/chaincode/go/asset_management_with_roles
go test
```
## Checklist:

Signed-off-by:
Anna D Derbakova adderbak@us.ibm.com
Diego Masini damasini@us.ibm.com


binhn:
@angrbrd Looping on consensus tests which is waiting for 1877

core/ledger/pvtdatastorage/snapshot_data_importer.go
```diff
@@ -0,0 +1,327 @@
...
+)
+
+type SnapshotDataImporter struct {
+	namespacesVisited   map[string]struct{}
+	eligibilityBTLCache *eligibilityBTLCache
+
+	db *leveldbhelper.DBHandle
+}
+
+func newSnapshotDataImporter(
```
cendhu, line 29: A question related to future PRs. I assume that the `SnapshotDataImporter` would be passed as a [pvtdataHashesConsumer](https://github.com/hyperledger/fabric/blob/ef6ff5f76288de034c12f89afe6bc954b87a96dd/core/ledger/kvledger/txmgmt/privacyenabledstate/snapshot.go#L228) while importing states to level/couchdb. As `newSnapshotDataImporter()` is unexported, I assume that kvledger would open the pvtdatastore which would return the `SnapshotDataImporter` and then it will be passed on to the [pvtdataHashesConsumer](https://github.com/hyperledger/fabric/blob/ef6ff5f76288de034c12f89afe6bc954b87a96dd/core/ledger/kvledger/txmgmt/privacyenabledstate/snapshot.go#L228). Correct me if it going to be done in some other way.

core/ledger/pvtdatastorage/snapshot_data_importer.go
```diff
@@ -0,0 +1,327 @@
...
+type nsColl struct {
+	ns, coll string
+}
+
+type eligibility struct {
+	configBlockNum uint64
+	isEligible     bool
+}
+
+type eligibilityBTLCache struct {
```
cendhu: nit. eligibility**And**BTLCache can avoid the confusion I had. It got resolved after looking that `loadDataFor`.

core/ledger/pvtdatastorage/snapshot_data_importer.go
```diff
@@ -0,0 +1,327 @@
...
+			}
+			i.eligibilityHistory[key] = append(i.eligibilityHistory[key],
+				&eligibility{
+					configBlockNum: committingBlkNum,
+					isEligible:     eligible,
+				},
+			)
+			if staticCollection.BlockToLive > 0 {
+				i.btl[key] = staticCollection.BlockToLive
+			}
```
cendhu, line 174: Isn't it enough to store the following in the cache:
1. the most recent entry.
2. if the org is not eligible as per the most recent entry, then we need to store the next most recent entry where the org is eligible. If no such entry is available, we would store only (1). At most, we need two entries.

Am I missing something?

Further, the block to live is not going to change for a given `<ns, coll>`.

core/ledger/pvtdatastorage/snapshot_data_importer.go
```diff
@@ -0,0 +1,327 @@
...
+	}
+	return expiringBlk < math.MaxUint64, expiringBlk
+}
+
+type dbUpdater struct {
+	db    *leveldbhelper.DBHandle
+	batch *leveldbhelper.UpdateBatch
+}
+
+func (u *dbUpdater) upsertMissingDataEntry(key *missingDataKey, committingTxNum uint64, isEligible bool) error {
```
cendhu, line 237: Nit. To avoid the control flag `isEligible`, we can pass the encodedKey directly instead of the key.

core/ledger/pvtdatastorage/snapshot_data_importer.go
```diff
@@ -0,0 +1,327 @@
...
+	if err != nil {
+		return errors.WithMessage(err, "error while getting missing data bitmap from the store")
+	}
+
+	missingData := &bitset.BitSet{}
+	if encMissingData != nil {
+		if missingData, err = decodeMissingDataValue(encMissingData); err != nil {
+			return err
+		}
+	}
```
cendhu: Nit. For code symmetry, here we can repeat the pattern used from line 273 to 280 & 306 to 313

core/ledger/pvtdatastorage/persistent_msgs_helper.go
```diff
@@ -7,44 +7,57 @@ SPDX-License-Identifier: Apache-2.0
...
 		// Hence, we need to reinitialize the map.
-		if collections.Map == nil {
-			collections.Map = make(map[string]*TxNums)
+		if nsExpiryData.PresentData == nil {
+			nsExpiryData.PresentData = make(map[string]*TxNums)
 		}
-		if collections.MissingDataMap == nil {
-			collections.MissingDataMap = make(map[string]bool)
+		if nsExpiryData.MissingData == nil {
+			nsExpiryData.MissingData = make(map[string]bool)
```
cendhu, line 30: Is the following correct? We don't need to repeat the same logic for `BootKVHashes` because when the pvtstore is bootstrapped using snapshots, `BootKVHashes` would never be empty (during the sub-sequent fetches). Once the bootstrapping is done, in the normal flow, we wouldn't never add anything to `BootKVHashes` (even if it is empty for new entry/collections).

manish-sethi: Yes, that's correct. Still, let me add this for the sake consistency.

core/ledger/pvtdatastorage/snapshot_data_importer.go
manish-sethi: You guessed it right :-)

core/ledger/pvtdatastorage/snapshot_data_importer.go
manish-sethi: sure, I can rename this.

core/ledger/pvtdatastorage/snapshot_data_importer.go
manish-sethi: I prefer to keep it as consistent with the other functions where they take the keys in the struct form - as this segregates the data access logic from the rest of the logic a little cleaner way. In fact, now this package has so many coding encoding, I was thinking in the lines of broadly to introduce a layer that deals with DB and exposes APIs in the terms of structs only.

core/ledger/pvtdatastorage/snapshot_data_importer.go
manish-sethi: sure

core/ledger/pvtdatastorage/store.go
```diff
@@ -577,6 +584,10 @@ func (s *Store) purgeExpiredData(minBlkNum, maxBlkNum uint64) error {
 			)
 		}
 
+		for _, bootKVHashesKey := range bootKVHashesKeys {
+			batch.Delete(encodeBootKVHashesKey(bootKVHashesKey))
+		}
```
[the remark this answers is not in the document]
manish-sethi: Yes, I intend to enhance the tests related to persisted values in the upcoming PRs. In this PR I focused mainly on the data loading and still this PR went much oversize.

core/ledger/pvtdatastorage/snapshot_data_importer.go
```diff
@@ -0,0 +1,327 @@
...
+			)
+	}
+
+	if dataBlockNum <= history[len(history)-1].configBlockNum {
+		return false,
+			errors.Errorf(
+				"unexpected error - no collection config found below block number [%d] for <namespace=%s, collection=%s>",
+				dataBlockNum, namespace, collection,
+			)
+	}
```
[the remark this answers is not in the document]
manish-sethi: No, as I mentioned that I was planning to retain only a single element and in that scheme of thing, I would have retained the first block number for the config.

core/ledger/pvtdatastorage/snapshot_data_importer.go
manish-sethi: No, you are not missing much except that in the case 2, we need to store most recent entry when the org became ineligible, instead of the the latest one. In other words, the change point is important in the case 2. I would not be keen in shrinking the history as it does not still make us get rid of array, it just shrinks the history and I don't think that memory saving is our motive here.

I first was rather thinking of making it a single element always (no arrary) by recording only the highest point till which the org was eligible. For always ineligible, the point could be used as maxuint64 but then with array for maintaining full list. My main motive here was to investigate if we could use the same logic here as well as in the pvtdata puller/sender by sharing this cache and `isEligibile` function on this to be sure of consistency of logic but before we discovered the gap in the pvtdata puller/sender code, I had coded this array based structure.

Finally, as we discussed offline that from the point of view of ease of readability, the current structure makes better sense so keeping this as is for now.

PR #1912, opened 2016-06-18
Chaincode tls

### LINKED ISSUE #1848: Chaincode container should communicate over TLS with the peer
## Description

The chaincode containers do not communicate over TLS. Thus, if TLS is turned on on the peer, the chaincode container will fail to connect to the peer.
## Describe How to Reproduce

1. Turn TLS on on the peer
2. Deploy chaincode
## Proposal

Ref: https://github.com/masterDev1985/fabric/commit/263dcc3dabf0c71821b3c2d4f505a6fb658b5b0f
In addition to above, we need to modify the chaincode Dockerfile in core.yaml to copy the certs into the chaincode container in the folder provided at peer.tls.cert.file & peer.tls.key.file.

muralisrini:
We typically run peer without TLS. But when we do, we need to run chaincode in TLS too.
## Description

Chaincode and peer use the same transport (currently at least). So TLS/no-TLS should apply to both. Added support for chaincode TLS using the peer.tls.\* properties and passing the creds and the files to the golang chaincode.
## Motivation and Context

Without this change peer cannot be used with TLS (as chaincodes communications will fail)

Fixes #1848
## How Has This Been Tested?

Enabled ALL chaincode tests except one (for non-TLS coverage) to use TLS. Created self-signed certs under chaincode/testdata valid for 3650 days for TLS.
## Checklist:

Signed-off-by: mrshah@us.ibm.com, muralisr@us.ibm.com


core/chaincode/chaincode_support.go
```diff
@@ -122,6 +122,13 @@ func NewChaincodeSupport(chainname ChainName, getPeerEndpoint func() (*pb.PeerEn
 		s.chaincodeInstallPath = chaincodeInstallPathDefault
 	}
 
+	s.peerTLS = viper.GetBool("peer.tls.enabled")
+	if s.peerTLS {
+		s.peerTLSCertFile = viper.GetString("peer.tls.cert.file")
+		s.peerTLSKeyFile = viper.GetString("peer.tls.key.file")
```
mastersingh24, line 128: I don't think there is any need to add the `peer.tls.key.file` since this is a client connection and we only need the cert so that it can be trusted in the TLS client connection.

core/chaincode/chaincode_support.go
```diff
@@ -251,6 +262,18 @@ func (chaincodeSupport *ChaincodeSupport) sendInitOrReady(context context.Contex
 func (chaincodeSupport *ChaincodeSupport) getArgsAndEnv(cID *pb.ChaincodeID) (args []string, envs []string, err error) {
 	envs = []string{"CORE_CHAINCODE_ID_NAME=" + cID.Name}
 
+	//if TLS is enabled, pass TLS material to chaincode
+	if chaincodeSupport.peerTLS {
+		envs = append(envs, "CORE_PEER_TLS_ENABLED=true")
+		envs = append(envs, "CORE_PEER_TLS_CERT_FILE="+chaincodeSupport.peerTLSCertFile)
+		envs = append(envs, "CORE_PEER_TLS_KEY_FILE="+chaincodeSupport.peerTLSKeyFile)
```
mastersingh24: see previous comment on not requiring the key

core/chaincode/platforms/golang/package.go
```diff
@@ -61,6 +61,15 @@ func writeChaincodePackage(spec *pb.ChaincodeSpec, tw *tar.Writer) error {
 	//let the executable's name be chaincode ID's name
 	newRunLine := fmt.Sprintf("RUN go install %s && cp src/github.com/hyperledger/fabric/peer/core.yaml $GOPATH/bin && mv $GOPATH/bin/%s $GOPATH/bin/%s", urlLocation, chaincodeGoName, spec.ChaincodeID.Name)
 
+	//NOTE-this could have been abstracted away so we could use it for all platforms in a common manner
+	//However, it would still be docker specific. Hence any such abstraction has to be done in a manner that
+	//is not just language dependent but also container depenedent. So lets make this change per platform for now
+	//in the interest of avoiding over-engineering without proper abstraction
+	if viper.GetBool("peer.tls.enabled") {
+		newRunLine = fmt.Sprintf("%s\nCOPY src/certs/cert.pem %s", newRunLine, viper.GetString("peer.tls.cert.file"))
+		newRunLine = fmt.Sprintf("%s\nCOPY src/certs/key.pem %s", newRunLine, viper.GetString("peer.tls.key.file"))
```
mastersingh24: see comments above on not including the key

muralisrini: @mastersingh24 thanks much. Will change

PR #1920, opened 2016-06-20
Add config to avoid chaincode logging fills up the disk space.

### LINKED ISSUE #709: chaincode logging fills up system drive
I ran an invoke loop over night, and the chaincode logging inside the docker container filled up my (host) filesystem; 9GB of JSON logs.  This may also explain why I observe high CPU utilization of the docker process.

genggjh:
## Description

We need set the default log settings for the docker container to avoid logging file fills up the disk space.
## Motivation and Context

Fixes #709 
## How Has This Been Tested?

Only the configuration file has been updated, no code related.
## Checklist:

- [] Either no new documentation is required by this change, OR I added new documentation

### DIFF
--- a/peer/core.yaml
+++ b/peer/core.yaml
@@ -257,10 +257,17 @@ vm:
         # NetworkMode Sets the networking mode for the container. Supported standard values are: `host`(default),`bridge`,`ipvlan`,`none`
         # dns A list of DNS servers for the container to use.
         # note: not support customize for `Privileged` 
+        # LogConfig sets the logging driver (Type) and related options (Config) for Docker
+        # you can refer https://docs.docker.com/engine/admin/logging/overview/ for more detail configruation.
         hostConfig:
             NetworkMode: host
             Dns:
                # - 192.168.0.1
+            LogConfig:
+                Type: json-file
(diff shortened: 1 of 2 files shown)


peer/core.yaml
```diff
@@ -257,10 +257,17 @@ vm:
...
+        # you can refer https://docs.docker.com/engine/admin/logging/overview/ for more detail configruation.
         hostConfig:
             NetworkMode: host
             Dns:
                # - 192.168.0.1
+            LogConfig:
+                Type: json-file
+                Config:
+                    max-size: "50m"
+                    max-file: "5"
```
muralisrini, line 270: @genggjh I couldn't find a test case in https://github.com/fsouza/go-dockerclient using LogConfig so we can compare the "max-size" and "max-file" uses. Were you able to confirm this works as expected ? Perhaps we should add a test for this ?

genggjh, line 270: I think we can test if the max-size and max-file can be loaded as expected from configure file and create the LogConfig obj. We don't need to cover the docker related function test, right?

srderson:
LGTM, thanks

PR #1924, opened 2016-06-20
improve pbft performance of reqstore

corecode:
## Description

Change complexity of batch outstanding requests datastructure from O(N^2) to O(1).
## Motivation and Context

Fixes a temporary performance regression.  Slack #performance-benchmark reports 10x performance increase.
## How Has This Been Tested?

new tests
## Checklist:

### DIFF
--- a/consensus/obcpbft/obc-batch_test.go
+++ b/consensus/obcpbft/obc-batch_test.go
@@ -94,7 +94,7 @@ func TestClearOustandingReqsOnStateRecovery(t *testing.T) {
 
 	b.manager.Queue() <- nil
 
-	if len(*(b.reqStore.outstandingRequests)) != 0 {
+	if b.reqStore.outstandingRequests.Len() != 0 {
 		t.Fatalf("Should not have any requests outstanding after completing state transfer")
 	}
 }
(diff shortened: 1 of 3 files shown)


jyellick:
@corecode Great work! The initial implementation was extremely naive from a performance perspective, and this looks to be a significant improvement, also the inclusion of new tests was something sorely needed.

It looks like ordering is respected based on insertion, rather than by timestamp now?

corecode:
Yes, correct

jyellick:
We'll see if this causes any problems with the behave tests, but I think that is reasonable behavior.  We should not encourage users to depend on transactions being executed in timestamp order, as this generally cannot be assured.

genggjh:
Cool, have any data or log to show us how much performance improvement after this changes?
I need write this achievement down :-)

corecode:
It was introduced recently, so I don't think we need to brag about fixing performance regressions we added a few days before.  It's from O(N^2) to O(1).

corecode:
good to pull

genggjh:
👍 LGTM

binhn:
@corecode the code LGTM, but need some description on the PR

srderson:
LGTM as it looks like a description has been added. Thanks!

PR #1928, opened 2016-06-20
Ensure message order of broadcasts

jyellick:
## Description

This changeset modifies the obcpbft `broadcast.go` to always send messages in the order in which they were received.

As @corecode was the original author of the `broadcast.go` his review especially is requested.
## Motivation and Context

The previous implementation of the broadcaster would spawn a goroutine per broadcast request, which could wake up in an arbitrary order, delivering messages to recipients in that arbitrary order.  This caused problems for sections of the code which expected and depended on messages arriving in order.

One manifestation of this was request duplication as a request might be prepared and committed before the broadcast of its original contents was received, causing the network to view the request as new and invoke it twice.

The other problem was that these spawned go routines would stack and stack until the underlying gRPC stack would timeout and cause these calls to unwind and be garbage collected.
## How Has This Been Tested?

The existing unit tests and CI, plus the busywork stress2b tool.
## Checklist:

### DIFF
--- a/consensus/obcpbft/broadcast.go
+++ b/consensus/obcpbft/broadcast.go
@@ -34,26 +34,32 @@ type broadcaster struct {
 	comm communicator
 
 	f        int
-	msgChans map[uint64]chan *pb.Message
+	msgChans map[uint64]chan *sendRequest
 	closed   sync.WaitGroup
 	closedCh chan struct{}
+}
+
+type sendRequest struct {
+	msg  *pb.Message
+	done chan struct{}
 }
 
 func newBroadcaster(self uint64, N int, f int, c communicator) *broadcaster {
 	queueSize := 10 // XXX increase after testing
 
-	chans := make(map[uint64]chan *pb.Message)
-	for i := 0; i < N; i++ {
-		if uint64(i) == self {
-			continue
-		}
-		chans[uint64(i)] = make(chan *pb.Message, queueSize)
-	}
+	chans := make(map[uint64]chan *sendRequest)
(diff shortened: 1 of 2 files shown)


consensus/obcpbft/broadcast.go
```diff
@@ -67,35 +73,68 @@ func (b *broadcaster) Wait() {
...
 	}
 }
 
+func (b *broadcaster) unicastOne(msg *pb.Message, dest uint64, wait chan struct{}) {
+	select {
+	case b.msgChans[dest] <- &sendRequest{
+		msg:  msg,
+		done: wait,
+	}:
+	default:
```
corecode, line 131: but doesn't that mean that unresponsive replicas appear like very fast ones, leading us to pace too quickly - creating in essence a positive feedback?

Maybe here is the place to sleep a second?

jyellick, line 131: Yes, I considered that briefly, I was thinking maybe instead we could only count 'successful' sends towards the total count, so instead of replying with that struct, switch it back to a bool.

Maybe the logic could then be "we wait for 2f+1 successes, or after n seconds, 2f+1 replies of success or failure". What do you think?

corecode, line 131: Yea, the timeout would handle that

jyellick, line 131: @srderson please see @corecode's comments in #1938 which affirms the timeout has been implemented.

srderson:
Thanks, LGTM

PR #1926, opened 2016-06-20
use a different db for system chaincode UT than chaincode UTs

muralisrini:
Make unit test for system chaincode use a different ledger DB than those used for chaincode UTs.
## Description

We have seen errors with following signature recently

```
Error opening DB IO error: lock /var/hyperledger/test/tmpdb/db/LOCK: Resource temporarily unavailable
--- FAIL: TestExecuteDeploySysChaincode (0.00s)
panic: Could not open openchain db error = [IO error: lock /var/hyperledger/test/tmpdb/db/LOCK: Resource temporarily unavailable] [recovered]
    panic: Could not open openchain db error = [IO error: lock /var/hyperledger/test/tmpdb/db/LOCK: Resource temporarily unavailable]
```

While we don't know what the underlying reason is - best we have is somehow tests are run in parallel causing DB contention - it is not a bad idea to run tests under different test dbs.

This PR does this for just the system chaincode UT to see if this takes us past the recent errors in this Unit Test case. We could do the same for other tests in the chaincode UT suite as well in a future PR.
## Motivation and Context

Fixes #1921
## Checklist:

- [X] I have added a [Signed-off-by](https://github.com/hyperledger/fabric/blob/master/CONTRIBUTING.md#legal-stuff)
- [X] Either no new documentation is required by this change, OR I added new documentation
- [X] Either no new tests are required by this change, OR I added new tests
- [X] I have run [goimports](https://godoc.org/golang.org/x/tools/cmd/goimports), [go vet](https://golang.org/cmd/vet/), and [golint](https://github.com/golang/lint). I have cleaned up all valid errors and warnings in code I have added or modified. These tools may generate false positives. Don't be worried about ignoring some errors or warnings. The goal is clean, consistent, and readable code.

Signed-off-by: muralisr@us.ibm.com

PR #1930, opened 2016-06-20
Turn off test parallelism - issue #1926

srderson:
## Description

Turns of test parallelism while running `make unit-test`
## Motivation and Context

Fixes #1926 
Failures are starting to pop-up due to Go running package tests in parallel by default. This has generally been caused by two processes accessing the database at the same time which is not supported
## How Has This Been Tested?

Ran `make checks`. No new tests are required as this does not alter functionality.
## Checklist:


binhn:
Thanks. This will fix many random problems we've been seeing when running unit tests

PR #1937, opened 2016-06-21
check if member services URL served/exists (#1893)

### LINKED ISSUE #1893: Client-sdk: if the wrong membership service IP is inserted the client-sdk stalls

th-ei:
Member services RPCs in Nodejs client SDK are protected with a URL sanity check
## Description

Upon first connection to the member services the client SDK will check if the URL refers to an existing host and a serviced TCP port. If this test is successful it will directly run gRPCs against this URL. If this fails the error is reported back to the caller of the client SDK. The next call using member services will again sanity check the URL.
## Motivation and Context

Fixes #1893
## How Has This Been Tested?

Run test/unit/chain-tests.js with an invalid member service URL
- localhost but unserviced port
- non-existing host
## Checklist:

- [n/a] I have run [golint](https://github.com/golang/lint) and have fixed valid warnings in code I have added or modified. This tool generates false positives so you may choose to ignore some warnings. The goal is clean, consistent, and readable code.

Signed-off-by: Thomas Eirich eir@zurich.ibm.com


srderson:
@smithbk or @angrbrd - please review

smithbk:
We can tell if the gRPC call fails, right?  If yes, then for performance, we should call checkUrl only if the gRPC call fails so that we can provide a better error message; otherwise, there is no need to incur the overhead of another connection.

Another minor style comment is that the standard node style is to provide a single callback that can optionally return an error.  It seems the checkUrl function has one error callback and a separate success callback.  This is not an issue but made me look twice to see what was going on.

srderson:
@th-ei please respond to @smithbk comments so we can merge this PR. Thanks.

th-ei:
The issue states that the gRPC does _not_ return. This fix uses the same strategy as checking for a working peer. Could you explain what you mean by tell if the gRPC fails in this context? The only other option is to cancel the gRPC if the response takes too long. What's too long? This also has another problem: the gRPC might have actually run/is running but get's cancelled anyway. All code sitting on top must be able to deal with this. The gRPC reports cancelled the effects have been or have not been carried out.

smithbk:
I get a 404 for https://github.com/hyperledger/fabric/issues/1893, so I couldn't see the issue description.
But looking more closely at checkUrl, I see that you set url to null and therefore will only make the extra TCP connection once.  Sorry, I missed that the 1st time.
So this LGTM.

srderson:
@th-ei There's a merge conflict and this needs to be rebased on master. Also, could you double check the issue #? Thanks.

christo4ferris:
@th-ei please rebase on master and resolve merge conflicts so that we can merge.

angrbrd:
@th-ei 

Has this been merged? Can the PR and the associated issue be closed? Thanks!

christo4ferris:
@th-ei we will be transferring the fabric and fabric-api repository and all the corresponding issues and pull requests to the hyperledger-archives org on GH and replacing this repository with a read-only mirror of our Gerrit source. As a result, any automatic GH redirects will be broken. If you haven't already migrated your PR to Gerrit, and intend to do so, you will need to change your git remote to reference the transferred repository (e.g. git@github.com:hyperledger-archives/fabric) on your local clone.

PR #1938, opened 2016-06-21
Ensure message order of broadcasts

jyellick:
## Description

This changeset modifies the obcpbft `broadcast.go` to always send messages in the order in which they were received.

This is a companion to PR#1928 and incorporates some feedback from @corecode.
## Motivation and Context

The previous implementation of the broadcaster would spawn a goroutine per broadcast request, which could wake up in an arbitrary order, delivering messages to recipients in that arbitrary order.  This caused problems for sections of the code which expected and depended on messages arriving in order.

One manifestation of this was request duplication as a request might be prepared and committed before the broadcast of its original contents was received, causing the network to view the request as new and invoke it twice.

The other problem was that these spawned go routines would stack and stack until the underlying gRPC stack would timeout and cause these calls to unwind and be garbage collected.
## How Has This Been Tested?

The existing unit tests and CI, plus the busywork stress2b tool.
## Checklist:

### DIFF
--- a/consensus/obcpbft/broadcast.go
+++ b/consensus/obcpbft/broadcast.go
@@ -73,41 +73,41 @@ func (b *broadcaster) Wait() {
 	b.closed.Wait()
 }
 
-func (b *broadcaster) drainerSend(dest uint64, send *sendRequest, printedValidatorNotFound bool) bool {
+func (b *broadcaster) drainerSend(dest uint64, send *sendRequest, successLastTime bool) bool {
+	// Note, successLastTime is purely used to avoid flooding the log with unnecessary warning messages when a network problem is encountered
 	defer func() {
 		b.closed.Done()
 	}()
 	h, err := getValidatorHandle(dest)
 	if err != nil {
-		if !printedValidatorNotFound {
+		if successLastTime {
 			logger.Warningf("could not get handle for replica %d", dest)
 		}
-		time.Sleep(time.Second)
 		send.done <- false
-		return true
-	}
-
-	if printedValidatorNotFound {
-		logger.Infof("Found handle for replica %d", dest)
-		printedValidatorNotFound = false
+		return false
 	}
 
 	err = b.comm.Unicast(send.msg, h)
 	if err != nil {
-		logger.Warningf("could not send to replica %d: %v", dest, err)
+		if successLastTime {
+			logger.Warningf("could not send to replica %d: %v", dest, err)
+		}
 		send.done <- false
-	} else {
-		send.done <- true
+		return false
 	}
 
-	return false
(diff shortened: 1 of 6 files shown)


consensus/obcpbft/broadcast.go
```diff
@@ -67,35 +73,71 @@ func (b *broadcaster) Wait() {
 	b.closed.Wait()
 }
 
-func (b *broadcaster) unicastOne(msg *pb.Message, dest uint64, wait chan bool) {
+func (b *broadcaster) drainerSend(dest uint64, send *sendRequest, printedValidatorNotFound bool) bool {
```
corecode: returning false on success seems weird

consensus/obcpbft/broadcast.go
```diff
@@ -67,35 +73,71 @@ func (b *broadcaster) Wait() {
...
 	defer func() {
 		b.closed.Done()
-		wait <- true
 	}()
+	h, err := getValidatorHandle(dest)
+	if err != nil {
+		if !printedValidatorNotFound {
+			logger.Warningf("could not get handle for replica %d", dest)
+		}
+		time.Sleep(time.Second)
```
corecode: given that we wait below, do we still need to sleep here?

consensus/obcpbft/broadcast.go
```diff
@@ -67,35 +73,71 @@ func (b *broadcaster) Wait() {
...
+
+	return false
+}
+
+func (b *broadcaster) drainer(dest uint64) {
+	printedValidatorNotFound := false
+
+	for {
 		select {
 		case <-b.closedCh:
```
corecode, line 111: the order of the case statements is a bit confusing.  I'd put the main work case on top.

consensus/obcpbft/broadcast.go
jyellick: Agreed, I considered flipping this, and will do so now.  The status is only used to avoid flooding our logs when some peer is not connected.

consensus/obcpbft/broadcast.go
jyellick: This is likely harmless but superfluous, I will remove.

consensus/obcpbft/broadcast.go
jyellick, line 111: Sure, I can re-order.

jyellick:
@corecode Commit 0051f52 should address your feedback.

srderson:
@corecode please leave a LGTM comment if @jyellick has resolved all of your feedback. Thanks.

corecode:
looks good

srderson:
Thanks @corecode

Issue #1931, opened 2016-06-20
All peers report identical blockchains that contain duplicated transactions


bcbrock:
## Description

~~This is a very rare failure mode that I have only seen 3 or 4 times now. This
is the reason that I added the "duplicate check" to the "checkAgreement"
script.~~

This error appears to be 100% repeatable on my system when run as described under **How to Reproduce**, always showing 11 to 15 duplicates.

We see from the busywork/counters/driver script that all peers appear to be
working perfectly : The counts are correct and the blockchains are
identical. However, it turns out in this case that 11 transactions have been
duplicated at the end of every blockchain. Note that the duplicates come from
all over the place. (I should say that based on the responses to the
chain/blocks API it would _appear_ that the transactions were duplicated, but
I have no other knowledge of the actual internal state of the system.)

I've attached the contents of the BUSYWORK_HOME directory. Unfortunately
this run did not have DEGUG-level logs.

[busywork.tar.gz](https://github.com/hyperledger/fabric/files/324446/busywork.tar.gz)

```
network: ../bin/userModeNetwork -batch 4
network: Creating a pristine environment
network:     Killing all peer and membersrvc processes
network:     Killing all running and stopped Docker containers
network:     Removing all chaincode containers
network:     Cleaning out BUSYWORK_HOME
network:     Cleaning out /var/hyperledger
network: Running with PBFT Batch consensus
network: Network Port and PID Map:
network: ------------------------------------------------------------
network:  PeerID |  gRPC  |  REST | Events |  CLI  | Profile |  PID 
network: ------------------------------------------------------------
network:     vp0 |  38333 | 33346 |  45344 | 44592 |  45596  | 38178
network:     vp1 |  42073 | 33389 |  41646 | 42326 |  41317  | 38182
network:     vp2 |  41149 | 38904 |  46806 | 37063 |  35758  | 38190
network:     vp3 |  45836 | 46428 |  35879 | 40957 |  42482  | 38198
network: ------------------------------------------------------------
network: Waiting for peers to start
network: Network is up
driver: ./driver -clients 1 -transactions 16384 -peerBurst 10 -netBurst 10 -interlock -keepLog -checkAgreement -dupCheck
driver: List of targeted peers with REST API ports
driver:     0.0.0.0:33346
driver:     0.0.0.0:33389
driver:     0.0.0.0:38904
driver:     0.0.0.0:46428
driver: 1 chaincode(s) will be deployed
[...]

PR #1946, opened 2016-06-21
Busywork: Changes to allow PBFT batch runs to pass busywork tests

bcbrock:
## Description

The definition of correctness for **busywork** tests has been relaxed due to
the realities of the current PBFT consensus algorithms. Because of state transfer,
we can no longer assume that each peer has actually executed each
transaction. Instead, we can only demand that the state of the peer is
consistent with having executed every transaction. We also now wait until the
peers reach a consensus on the blockchain height at the end of the run before
querying for final status, since queries now fail on peers that are in the
middle of state transfer. 

Once a couple of outstanding consensus PRs are
processed we will probably be able to state that the busywork/counters test
case _stress2b_ is working. The documentation has already been updated as if that were true.
## Motivation and Context

This does not fix any open issues, but it is the result of working through several issues with @jyellick related to what happens during state transfer, and what it means for blockchain invokes and queries to be "correct".
## How Has This Been Tested?

I have been testing this over a couple of weeks, for example in testing related to issues #1931 and #1857.
## Checklist:

### DIFF
--- a/tools/busywork/README.md
+++ b/tools/busywork/README.md
@@ -88,25 +88,30 @@ to build the `peer` and `membersrvc` images. Then switch to the
 to run a simple stress test. This stress test exercises a single peer
 running `NOOPS` consensus.
 
-Currently Sieve is the only PBFT consensus algorithm that is stable enough
-to survive stress testing. The target
+At present the PBFT *batch* algorithm is the only true consensus algorithm
+being suported by the development team. (The *classic* and *sieve* algorithms
+may work for some cases.) The target 
 
-    make stress2s
+    make stress2b
 	
-runs Sieve on a 4-peer network. There is also a pre-canned target
+runs PBFT *batch* on a 4-peer network. 
+
+There is also a pre-canned target
 
     make secure1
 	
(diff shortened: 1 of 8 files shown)


bcbrock:
In terms of review, we might impose on @jyellick or another user to do a functional review, that is to try the `busywork/counters/stress2b` make target a few times to make sure that it works in their environment.

jyellick:
@bcbrock I've run the stress2b test a few times in my environment with no regressions.  I have noted some failures, in particular:

```
driver:ERROR: Peer 0.0.0.0:50138: Final count mismatch for chaincode cc0, array a1
```

which I believe is because chaincode local state is not modified on state transfer.  I also think I have observed the latest incarnation of #1857.  Both of these issues are not new though, so should not count against this PR.

I can't speak to the bulk of the changes, but I do endorse the modification of the stress2b definition to include the null requests.

Generally, this looks good to me.

srderson:
Thanks @bcbrock and @jyellick LGTM

bcbrock:
Thank you @srderson @jyellick

PR #1957, opened 2016-06-22
Add chaincode ping2#issue1850

### LINKED ISSUE #1850: Connection times out between chaincode & peer when running behind a proxy
## Description

When running behind a proxy, there is a definite timeout if the connection is inactive. This timeout breaks the connection between the peer and the chaincode, the chaincode stops.
## Describe How to Reproduce

1. Run a peer behind a proxy
2. Deploy a chaincode
3. The chaincode connects to the peer via proxy
4. Do not invoke/query for a long period. The chaincode container exits due to grpc error, connection broke.
## Proposal

Add a keep-alive message between chaincode & peer to keep the communication active - https://github.com/masterDev1985/fabric/commit/1e8eabb94b48862f29c4d95db40c749f261fa18e

muralisrini:
Allow a keep alive messages between fabric and chaincode. Default is no-keepalive.
## Description

It is possible that an inactive gRPC connection between the chaincode and the peer can get broken by a intermediate proxy. Introduced a "keepalive" parameter (in seconds) which can be used to send ChaincodeMessage of type KEEPALIVE between the peer and chaincode. A chaincode just echoes the KEEPALIVE message received from the peer. 
## Motivation and Context

Keeps the connection between peer and chaincode going with application protocol level messages. This makes the connection liveness non dependent upon lower layer keepalive mechanisms.

Fixes #1850
## How Has This Been Tested?

Enabled all chaincode tests (except one for covering no keep-alive scenario) to use an aggressive keepalive timer of 1 second.
## Checklist:

### DIFF
--- a/peer/core.yaml
+++ b/peer/core.yaml
@@ -314,6 +314,13 @@ chaincode:
     # the image
     installpath: /opt/gopath/bin/
 
+    #keepalive in seconds. In situations where the communiction goes through a 
+    #proxy that does not support keep-alive, this parameter will maintain connection
+    #between peer and chaincode
+    keepalive: 30
+
+###############################################################################
+#
 ###############################################################################
 #
 #    Ledger section - ledger configuration encompases both the blockchain
--- a/peer/core.yaml
+++ b/peer/core.yaml
@@ -316,8 +316,9 @@ chaincode:
 
     #keepalive in seconds. In situations where the communiction goes through a 
     #proxy that does not support keep-alive, this parameter will maintain connection
-    #between peer and chaincode
-    keepalive: 30
+    #between peer and chaincode.
+    #A value <= 0 turns keepalive off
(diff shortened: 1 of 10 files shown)


peer/core.yaml
```diff
@@ -314,6 +314,14 @@ chaincode:
     # the image
     installpath: /opt/gopath/bin/
 
+    #keepalive in seconds. In situations where the communiction goes through a 
+    #proxy that does not support keep-alive, this parameter will maintain connection
+    #between peer and chaincode.
+    #A value <= 0 turns keepalive off
```
srderson: Could you add a space after the `#` characters so it is consistent with the other comments in core.yaml?

muralisrini: Sure, will do.

srderson:
Thanks, LGTM

PR #1906, opened 2016-06-17
Add make checks to PR template. Resolves #1905

srderson:
## Description

An updated pull request template that discusses `make checks`
## Motivation and Context

Fixes #1905 
This change was discussed with @ghaskins and @rameshthoomu. Please review this proposal. I have remove `go vet` and `goimports` from the checklist and instead discuss how `make checks` will be run in CI and how it can be run locally.
## How Has This Been Tested?

No new tests required and this is only a change to the PR template.
## Checklist:

- [x] I have added a [Signed-off-by](https://github.com/hyperledger/fabric/blob/master/CONTRIBUTING.md#legal-stuff)
- [x] Either no new documentation is required by this change, OR I added new documentation
- [x] Either no new tests are required by this change, OR I added new tests
- [x] I have run [goimports](https://godoc.org/golang.org/x/tools/cmd/goimports), [go vet](https://golang.org/cmd/vet/), and [golint](https://github.com/golang/lint). I have cleaned up all valid errors and warnings in code I have added or modified. These tools may generate false positives. Don't be worried about ignoring some errors or warnings. The goal is clean, consistent, and readable code.

Signed-off-by: Sheehan Anderson sheehan@us.ibm.com

PR #1958, opened 2016-06-23
Add chaincode ping2#issue1850

### LINKED ISSUE #1850: Connection times out between chaincode & peer when running behind a proxy

muralisrini:
Allow a keep alive messages between fabric and chaincode. Default is no-keepalive.
## Description

It is possible that an inactive gRPC connection between the chaincode and the peer can get broken by a intermediate proxy. Introduced a "keepalive" parameter (in seconds) which can be used to send ChaincodeMessage of type KEEPALIVE between the peer and chaincode. A chaincode just echoes the KEEPALIVE message received from the peer. 
## Motivation and Context

Keeps the connection between peer and chaincode going with application protocol level messages. This makes the connection liveness non dependent upon lower layer keepalive mechanisms.

Fixes #1850
## How Has This Been Tested?

Enabled all chaincode tests (except one for covering no keep-alive scenario) to use an aggressive keepalive timer of 1 second.
## Checklist:

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -17,8 +17,8 @@ Fixes #
 <!-- To uncheck box, add a space: [ ] -->
 <!-- If you're unsure about any of these, don't hesitate to ask. We're here to help! -->
 - [] I have added a [Signed-off-by](https://github.com/hyperledger/fabric/blob/master/CONTRIBUTING.md#legal-stuff).
-- [] I have added documentation to cover my changes or this change requires no new documentation.
-- [] I have added unit tests to cover my changes or this change requires no new tests.
+- [] I have either added documentation to cover my changes or this change requires no new documentation.
+- [] I have either added unit tests to cover my changes or this change requires no new tests.
 - [] I have run [golint](https://github.com/golang/lint) and have fixed valid warnings in code I have added or modified. This tool generates false positives so you may choose to ignore some warnings. The goal is clean, consistent, and readable code.
 
 The continuous integration build process will run [make checks](https://github.com/hyperledger/fabric/blob/master/Makefile#L22) to confirm that tests pass and that code quality meets minimum standards. You may optionally run this locally as PRs will not be accepted until they pass.
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,23 +1,26 @@
-<!--- Provide a general summary of your changes in the Title above -->
+<!-- Provide a general summary of your changes in the Title above -->
 
 ## Description
-<!--- Describe your changes in detail. -->
+<!-- Describe your changes in detail. -->
 
 ## Motivation and Context
-<!--- Why is this change required? What problem does it solve? -->
-<!--- If it fixes an open issue, please link to the issue here. -->
+<!-- Why is this change required? What problem does it solve? -->
+<!-- If it fixes an open issue, please link to the issue here. -->
 Fixes #
 
 ## How Has This Been Tested?
-<!--- Please describe in detail how you tested your changes. -->
-<!--- If this PR does not contain a new test case, explain why. -->
+<!-- If this PR does not contain a new test case, explain why. -->
+<!-- Describe in detail how you tested your changes. -->
 
 ## Checklist:
-<!--- Go over all the following points, and put an `x` in all the boxes that apply. -->
-<!--- If you're unsure about any of these, don't hesitate to ask. We're here to help! -->
-- [] I have added a [Signed-off-by](https://github.com/hyperledger/fabric/blob/master/CONTRIBUTING.md#legal-stuff)
-- [] Either no new documentation is required by this change, OR I added new documentation
-- [] Either no new tests are required by this change, OR I added new tests
-- [] I have run [goimports](https://godoc.org/golang.org/x/tools/cmd/goimports), [go vet](https://golang.org/cmd/vet/), and [golint](https://github.com/golang/lint). I have cleaned up all valid errors and warnings in code I have added or modified. These tools may generate false positives. Don't be worried about ignoring some errors or warnings. The goal is clean, consistent, and readable code.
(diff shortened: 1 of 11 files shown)


srderson:
@muralisrini It looks like this PR somehow contains additional commits that are not part of your change. Specifically, it looks like PR #1906 was included which is not part of the 0.5-developer-preview branch. Could you see if you can remove those?

muralisrini:
@srderson I'll delete this PR and create a fresh one

integration/chaincode/kvexecutor/chaincode.go
```diff
@@ -101,6 +106,66 @@ func (t *KVExcutor) readWriteKVs(stub shim.ChaincodeStubInterface, args []string
...
+		buffer.WriteString(", \"IsDelete\":")
+		buffer.WriteString("\"")
+		buffer.WriteString(strconv.FormatBool(response.IsDelete))
+		buffer.WriteString("\"")
+
+		buffer.WriteString("}")
+		bArrayMemberAlreadyWritten = true
+	}
+	buffer.WriteString("]")
+
```
manish-sethi: It will be much cleaner to prepare a map and then use `json.Marshal` function. I guess that marble chaincode also has this pattern and that can also be improved.

integration/chaincode/kvexecutor/chaincode.go
```diff
@@ -101,6 +113,45 @@ func (t *KVExcutor) readWriteKVs(stub shim.ChaincodeStubInterface, args []string
 	return shim.Success(resultBytes)
 }
 
+func (t *KVExcutor) getHistory(stub shim.ChaincodeStubInterface, args []string) pb.Response {
+	if len(args) != 1 {
+		return shim.Error("Incorrect number of arguments. Expecting 1 (key)")
+	}
+
+	key := args[0]
```
manish-sethi: Not used in the latest patch. leftover from previous patch?

integration/ledger/snapshot_test.go
```diff
@@ -0,0 +1,816 @@
...
+		// verify snapshot commands with different parameters")
+		blockNum := nwo.GetLedgerHeight(setup.network, org1peer0, testchannelID) - 1
+		verifySnapshotRequestCmds(setup.network, org1peer0, testchannelID, blockNum)
+
+		// test 1: generate snapshots on 2 peers for the same blockNum and verify they are same
+		_, snapshotDir2 := generateAndCompareSnapshots(setup.network, org1peer0, org2peer0, blockNum)
+
+		// test 2: bootstrap a peer in an existing org from snapshot and verify
+		By("starting new peer org2peer1 in existing org (couchdb)")
+		org2peer1, couchProc := startPeerInExistingOrgFromSnapshot(setup, "Org2", "peer1", testchannelID, snapshotDir2, blockNum, true)
```
manish-sethi: `startNewPeerInExistingOrg` (as snapshot has no use in starting the peer). Also, the parameters snapshotDir2 and blockNum are perhaps, leftover from previous patch.

integration/ledger/snapshot_test.go
```diff
@@ -0,0 +1,816 @@
...
+			MSPID:         "Org3MSP",
+			Name:          "Org3",
+			Domain:        "org3.example.com",
+			EnableNodeOUs: true,
+			Users:         2,
+			CA: &nwo.CA{
+				Hostname: "ca",
+			},
+		}
+		org3peer0, _ := startPeerInNewOrgFromSnapshot(setup, org3, "peer0", testchannelID, snapshotDir2, blockNum, false)
```
manish-sethi: While this step adds the org to the channel but it does so after the snapshot has been generated.
This seem to have worked but I am not sure whether the new peer in the org3 may show some errors in the logs immediately after bootstrapping (either now or with some future changes), as ideally it is not expected to be part of the channel as yet (till it fetches the latest block). My suggestion would be to update the channel config upfront before taking the snapshot - this is what we considered as an expected scenario in the rfc.

integration/ledger/snapshot_test.go
```diff
@@ -0,0 +1,716 @@
...
+
+	By("configuring peer to couchdb address " + couchDB.Address())
+	s.network.WritePeerConfig(peer, core)
+
+	return couchProc
+}
+
+// initAndStartMultiOrgsNetwork creates a network with multiple orgs.
+// Initially only start Org1.peer0 and Org2.peer0 are started and join the channel.
+func initAndStartMultiOrgsNetwork() *setup {
```
manish-sethi, line 295: Since this function does not take a parameter and starts the network exactly with four orgs, better to explicitly reveal that in the function name.

integration/ledger/snapshot_test.go
```diff
@@ -0,0 +1,716 @@
...
+	core.Ledger.State.CouchDBConfig.CouchDBAddress = couchDB.Address()
+
+	By("configuring peer to couchdb address " + couchDB.Address())
+	s.network.WritePeerConfig(peer, core)
+
+	return couchProc
+}
+
+// initAndStartMultiOrgsNetwork creates a network with multiple orgs.
+// Initially only start Org1.peer0 and Org2.peer0 are started and join the channel.
```
manish-sethi, line 294: ~~start~~

integration/ledger/snapshot_test.go
```diff
@@ -0,0 +1,716 @@
...
+func initAndStartMultiOrgsNetwork() *setup {
+	var err error
+	testDir, err := ioutil.TempDir("", "reset-rollback")
```
manish-sethi, line 297: `reset-rollback` -> snapshot

Issue #856, opened 2016-03-21
Performance measurements


binhn:
We need to understand the time it takes various modules to complete a transaction. Starting at the Ledger, we know we can do around 55k tps (1 write), but at the client API level, we see max of around 400 tps. We need to instrument the code (or tests) with time in log INFO using the technique here https://coderwall.com/p/cp5fya/measuring-execution-time-in-go

ibmmark:
Check for 15 nodes

genggjh:
I want to contribute something for this, but first we need a list of modules or functions we want to understand the time spend.

genggjh:
maybe we can use this to reduce our coding effort which likes the decorator grammar in python : https://gist.github.com/alehano/8785933

PR #1960, opened 2016-06-23
a few time measuring and some goimports fixes

yacovm:
Added some time measurements
## Description

Just added some debug information regarding how long it takes to invoke CommitTxBatch and ProcessTransactionMsg (to see how fast the other goroutine processes the queue)

Also fixed some goimports because make linter complained.

### DIFF
--- a/consensus/executor/executor.go
+++ b/consensus/executor/executor.go
@@ -17,11 +17,12 @@ limitations under the License.
 package executor
 
 import (
+	"time"
+
 	"github.com/hyperledger/fabric/consensus"
 	"github.com/hyperledger/fabric/consensus/obcpbft/events"
 	"github.com/hyperledger/fabric/core/peer/statetransfer"
 	pb "github.com/hyperledger/fabric/protos"
-
 	"github.com/op/go-logging"
 )
 
@@ -79,7 +80,7 @@ func (co *coordinatorImpl) ProcessEvent(event events.Event) events.Event {
 
 		co.consumer.Executed(et.tag)
 	case commitEvent:
-		logger.Debug("Executor is processing an commitEvent")
+		logger.Debug("Executor is processing a commitEvent")
 		if co.skipInProgress {
 			logger.Error("Likely FATAL programming error, attempted to commit a transaction batch during state transfer")
 			return nil
@@ -90,6 +91,7 @@ func (co *coordinatorImpl) ProcessEvent(event events.Event) events.Event {
 			return nil
 		}
 
+		start := time.Now()
 		_, err := co.rawExecutor.CommitTxBatch(co, et.metadata)
 		_ = err // TODO This should probably panic, see issue 752
 
@@ -97,11 +99,11 @@ func (co *coordinatorImpl) ProcessEvent(event events.Event) events.Event {
 
 		info := co.rawExecutor.GetBlockchainInfo()
 
-		logger.Debugf("Committed block %d with hash %x to chain", info.Height-1, info.CurrentBlockHash)
+		logger.Debugf("Committed block %d with hash %x to chain, took %v", info.Height-1, info.CurrentBlockHash, time.Since(start))
 
 		co.consumer.Committed(et.tag, info)
 	case rollbackEvent:
-		logger.Debug("Executor is processing an rollbackEvent")
+		logger.Debug("Executor is processing a rollbackEvent")
 		if co.skipInProgress {
 			logger.Error("Programming error, attempted to rollback a transaction batch during state transfer")
 			return nil
--- a/consensus/executor/executor_test.go
+++ b/consensus/executor/executor_test.go
@@ -22,10 +22,9 @@ import (
 	"testing"
 
 	"github.com/hyperledger/fabric/consensus/obcpbft/events"
+	"github.com/op/go-logging"
 
 	pb "github.com/hyperledger/fabric/protos"
-
-	"github.com/op/go-logging"
 )
 
 func init() {
--- a/consensus/obcpbft/obc-batch.go
+++ b/consensus/obcpbft/obc-batch.go
@@ -18,16 +18,15 @@ package obcpbft
 
 import (
 	"fmt"
+	"google/protobuf"
 	"time"
 
 	"github.com/hyperledger/fabric/consensus"
 	"github.com/hyperledger/fabric/consensus/obcpbft/events"
 	pb "github.com/hyperledger/fabric/protos"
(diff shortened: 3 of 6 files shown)


yacovm:
Signed-off-by: yacovm@il.ibm.com

genggjh:
It's better we can provide a util function to handle this performance measurement logs related. I believe we should have a lot of check points regarding how long it taks, the issue #856 is using to trace this.

yacovm:
But, if we do something generic it'll have to be with interface{} and then we lose the compile time type checking, won't we?

GrapeBaBa:
Does it exist a metrics library in golang world which like dropwizzard metrics in java?

christo4ferris:
I've made this point previously. There are plenty of metrics/benchmarking packages for go that can be used for performance measurement. Some of the tests have started to use Ginkgo which includes [benchmarking capabilities](http://onsi.github.io/ginkgo/#benchmark-tests), and I think it would be preferable to instrument the tests than littering the code with more debug log output.

GrapeBaBa:
If we have this metrics in code, we can collect the metrics from a dedicated monitoring system real time. i saw many large scale system have this feature.

christo4ferris:
we aren't going to run production with DEBUG level logging enabled. Further, you can track things via events that we're already publishing. Further, it would make more sense to have metrics collection that can be periodically published rather than constantly spewing log messages that need to be correlated to yield value. As noted, Google is our friend. There are an abundance of metrics packages we might employ vs rolling and maintaining our own.

GrapeBaBa:
@christo4ferris Sure. I totally agree you, my point is not related this PR, I just wonder if we already have a plan to do some instrument for metrics statistic, request tracing,  maybe it is better to create a issue to track.

christo4ferris:
@yacovm we will be transferring the fabric and fabric-api repository and all the corresponding issues and pull requests to the hyperledger-archives org on GH and replacing this repository with a read-only mirror of our Gerrit source. As a result, any automatic GH redirects will be broken. If you haven't already migrated your PR to Gerrit, and intend to do so, you will need to change your git remote to reference the transferred repository (e.g. git@github.com:hyperledger-archives/fabric) on your local clone.

docs/source/private_data_tutorial.rst
```diff
@@ -152,8 +155,19 @@ collection allows only members of Org2 to have the private data in their private
     }
  ]
 
-The data to be secured by these policies is mapped in chaincode and will be
-shown later in the tutorial.
+
+The ``policy`` property in the ``assetCollection`` definition specifies that both
+Org1 and Org2 can store the collection on their peers. The ``memberOnlyRead``
+and ``memberOnlyWrite`` parameters are used to specify that only Org1 and
+Org2 can read and write to this collection.
```
denyeart: People are always confused about whether memberOnlyWrite applies to both client and endorsing peer... let's clarify that we're talking about client calls for these properties...

```suggestion
Org2 clients can read and write to this collection.
```

docs/source/private_data_tutorial.rst
```diff
@@ -152,8 +155,19 @@ collection allows only members of Org2 to have the private data in their private
...
+
+The ``Org1MSPPrivateCollection`` collection allows only members of Org1 to have
```
denyeart:
```suggestion
The ``Org1MSPPrivateCollection`` collection allows only peers of Org1 to have
```

docs/source/private_data_tutorial.rst
```diff
@@ -152,8 +155,19 @@ collection allows only members of Org2 to have the private data in their private
...
+the private data in their private database, while the ``Org2MSPPrivateCollection``
+collection can only be stored by members of Org2. The ``endorsementPolicy`` parameter
```
denyeart:
```suggestion
collection can only be stored by peers of Org2. The ``endorsementPolicy`` parameter
```

docs/source/private_data_tutorial.rst
```diff
@@ -731,16 +746,30 @@ When successful, should see something similar to the following result:
 Query private data Org2 is not authorized to
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Peers in Org2 do not have the Org1 asset1 ``appraisedValue`` private data in their side database.
-When they try to query for this data, they get back a hash of the key matching
-the public state but will not have the private state.
+Because the asset was created by Org1, peers in Org2 do not store the
```
denyeart:
```suggestion
Because the asset was created by Org1 and stored in the ``Org1MSPPrivateCollection`` collection, peers in Org2 do not store the
```

docs/source/private_data_tutorial.rst
```diff
@@ -731,16 +746,30 @@ When successful, should see something similar to the following result:
...
 :guilabel:`Try it yourself`
 
+.. code:: bash
+
+    peer chaincode query -o localhost:7050 --ordererTLSHostnameOverride orderer.example.com --tls --cafile ${PWD}/organizations/ordererOrganizations/example.com/orderers/orderer.example.com/msp/tlscacerts/tlsca.example.com-cert.pem -C mychannel -n private -c '{"function":"ReadAssetPrivateDetails","Args":["Org2MSPPrivateCollection","asset1"]}'
+
+The empty response shows that the asset1 private details do not exist in buyer
+(Org2) private collection.
+
+Nor can a member of Org2 read the Org1 private data collection:
```
denyeart:
```suggestion
Nor can a client of Org2 read the Org1 private data collection:
```

docs/source/private_data_tutorial.rst
```diff
@@ -731,16 +746,30 @@ When successful, should see something similar to the following result:
...
+
 .. code:: bash
 
     peer chaincode query -C mychannel -n private -c '{"function":"ReadAssetPrivateDetails","Args":["Org1MSPPrivateCollection","asset1"]}'
 
+By setting ``"memberOnlyRead": true`` in the collection configuration file, we
+specify that only members of Org1 can read data from the collection. An Org2 member
```
denyeart:
```suggestion
specify that only clients of Org1 can read data from the collection. An Org2 client
```

docs/source/private_data_tutorial.rst
```diff
@@ -731,16 +746,30 @@ When successful, should see something similar to the following result:
...
+who tries to read the collection would only get the following response:
```
denyeart, line 771: Remove the next line.

PR #1969, opened 2016-06-23
Fix busywork counters non-determinism

jyellick:
## Description

This changeset fixes the busywork counters chaincode to use the ledger retrieved state, rather than the chaincode local variable for updating the ledger state.
## Motivation and Context

This is invalid chaincode and has produced spurious test failures.

Fixes #1857, though #1857 is also being used to track a second checkpoint garbage collection issue that this non-deterministic testing surfaced.

As @bcbrock is the primary author of the busywork tests, his signoff should be required.
## How Has This Been Tested?

This is part of the busywork tests, and so this has been tested by running the busywork stress2b test.
## Checklist:

- [X] I have added a [Signed-off-by](https://github.com/hyperledger/fabric/blob/master/CONTRIBUTING.md#legal-stuff).
- [X] I have either added documentation to cover my changes or this change requires no new documentation.
- [X] I have either added unit tests to cover my changes or this change requires no new tests.
- [X] I have run [golint](https://github.com/golang/lint) and have fixed valid warnings in code I have added or modified. This tool generates false positives so you may choose to ignore some warnings. The goal is clean, consistent, and readable code.

The continuous integration build process will run [make checks](https://github.com/hyperledger/fabric/blob/master/Makefile#L22) to confirm that tests pass and that code quality meets minimum standards. You may optionally run this locally as PRs will not be accepted until they pass.

Signed-off-by: Jason Yellick jyellick@us.ibm.com

PR #1973, opened 2016-06-23
Invalidate state on checkpoint quorum peer disagrees with

jyellick:
## Description

This changeset verifies that the checkpoint quorum id agrees with its own checkpoint id.  If there is a quorum which disagrees with the local replica, the replica will invalidate its state, and perform state transfer.

This is a companion PR to #1971.  SMEs include @kchristidis @corecode or @tuand27613.
## Motivation and Context

The code previously naively assumed that its state must match the any checkpoint quorum, assuming it had reached that sequence number.  This caused the replica not to detect non-determinism if the replica generated its own checkpoint before the last matching quorum checkpoint was received.

This should eliminate the secondary behavior seen in #1857, where a replica would believe itself to be up to date with a diverged state.  The primary cause of #1857 (non-deterministic chaincode) has been fixed in pr #1969 for master.
## How Has This Been Tested?

This PR contains a new testcase which verifies the new behavior.
## Checklist:

### DIFF
--- a/consensus/obcpbft/pbft-core.go
+++ b/consensus/obcpbft/pbft-core.go
@@ -1222,7 +1222,8 @@ func (instance *pbftCore) recvCheckpoint(chkpt *Checkpoint) events.Event {
 	// we have reached this checkpoint
 	// Note, this is not divergent from the paper, as the paper requires that
 	// the quorum certificate must contain 2f+1 messages, including its own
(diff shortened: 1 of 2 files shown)


corecode:
must go in

srderson:
@corecode Thanks, LGTM. Also review in the 0.5 branch.

PR #1977, opened 2016-06-23
Remove duplicated storeRequest call per @kchristidis

jyellick:
## Description

This removes a duplicated call to store an outstanding request.
## Motivation and Context

This was [noted](https://github.com/hyperledger/fabric/commit/895dd2da3922a5428f81096134c2aff1d582a2c6#commitcomment-17982896) by @kchristidis but the PR was including that commit was merged before @kchristidis made the remark.  It is a simple case of unnecessarily storing the same request in the outstanding requests list twice.

This is inefficient, and should be removed.
## How Has This Been Tested?

This is equivalent to writing the same value into a map key twice, so there is no real value in writing a test case.  The existing CI should provide sufficient regression testing.
## Checklist:


srderson:
Review as part of PR #1971 in the 0.5 branch. LGTM

PR #1987, opened 2016-06-24
Check for state transfer on view change before ordering

jyellick:
## Description

This changeset adds a check to the batch ordering, that when a replica becomes a new primary, it first checks to see if it is in state transfer before attempting to order any outstanding requests. 

It also prevents null requests from permanently disabling the outstanding request timer, and adds outstanding requests to the primary before ordering.
## Motivation and Context

The outstanding requests are cleared at the conclusion of state transfer, which is sufficient in most cases, as a replica in state transfer is usually not the primary.  However, a replica may become the new primary in a view change which also forces a state transfer.  In this case, the replica is the primary, and can order requests until it exhausts half its log size (4 of them by default).  This can lead to duplication of up to 4 transactions on certain view changes.

Typical SMEs include @kchristidis @corecode and @tuand27613 .

The corresponding issue which this fixes, #1942 , has been tagged for inclusion in v0.5 (@srderson)
## How Has This Been Tested?

This PR contains a behave test which exhibits the failure (thanks @lhaskins!)
## Checklist:

- [X] I have added a [Signed-off-by](https://github.com/hyperledger/fabric/blob/master/CONTRIBUTING.md#legal-stuff).
- [X] I have either added documentation to cover my changes or this change requires no new documentation.
- [X] I have either added unit tests to cover my changes or this change requires no new tests.
- [X] I have run [golint](https://github.com/golang/lint) and have fixed valid warnings in code I have added or modified. This tool generates false positives so you may choose to ignore some warnings. The goal is clean, consistent, and readable code.

The continuous integration build process will run [make checks](https://github.com/hyperledger/fabric/blob/master/Makefile#L22) to confirm that tests pass and that code quality meets minimum standards. You may optionally run this locally as PRs will not be accepted until they pass.

Signed-off-by: Jason Yellick jyellick@us.ibm.com

PR #1988, opened 2016-06-24
 Check for state transfer on view change before ordering

jyellick:
## Description

This changeset adds a check to the batch ordering, that when a replica becomes a new primary, it first checks to see if it is in state transfer before attempting to order any outstanding requests. 

It also prevents null requests from permanently disabling the outstanding request timer, and adds outstanding requests to the primary before ordering.
## Motivation and Context

The outstanding requests are cleared at the conclusion of state transfer, which is sufficient in most cases, as a replica in state transfer is usually not the primary.  However, a replica may become the new primary in a view change which also forces a state transfer.  In this case, the replica is the primary, and can order requests until it exhausts half its log size (4 of them by default).  This can lead to duplication of up to 4 transactions on certain view changes.

Typical SMEs include @kchristidis @corecode and @tuand27613 .

This is a companion PR to #1987 
## How Has This Been Tested?

This PR contains a behave test which exhibits the failure (thanks @lhaskins!)
## Checklist:

### DIFF
--- a/bddtests/peer_basic.feature
+++ b/bddtests/peer_basic.feature
@@ -1116,3 +1116,67 @@ Feature: lanching 3 peers
     Examples: Composition options
         |          ComposeFile     |
         |   docker-compose-2.yml   |
+
+
+@issue_1942
+#@doNotDecompose
+Scenario: chaincode example02 with 4 peers, stop and start alternates, reverse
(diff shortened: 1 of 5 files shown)


corecode:
great

srderson:
LGTM. Also reviewed in PR #1987

Issue #1338, opened 2016-05-03
PBFT batch should be defined as default


binhn:
Currently batch is a configurable mode and the default is `batchsize: 2`, which is not reasonable default. We should set the default to be batch and size around 500; no magic about this number, but it is supposed to be batched for better performance :-)

tuand27613:
A few of us are thinking about combining batch and classic into one codebase with classic becoming the special case where batchsize=1   

@corecode  @jyellick 

Also,  now that we've merged PR #1325,  trying to find some time to do performance measurements at different batch sizes

corecode:
I don't know whether the number of transactions is a good metric to limit a batch.  If the reason is to limit the size of the resulting block, we should limit by size, not by count.

corecode:
In any case, this is not a bug.

binhn:
Is it better performance with batch = 2 or batch > 2 ?

ibmmark:
I agree with @corecode this should be removed from the bug list and tagged as consensus.

PR #2003, opened 2016-06-24
increase consensus batchsize to 1000

dongmingh:
## Description

To increase consensus batchsize to 1000 in config.yaml
## Motivation and Context

This is to increase the performance with pbft consensus
## How Has This Been Tested?

Compare the performance results under stress test with various batchsizes: 2 (default), 100, 500, and 1000.
## Checklist:

- [ ] I have run [golint](https://github.com/golang/lint) and have fixed valid warnings in code I have added or modified. This tool generates false positives so you may choose to ignore some warnings. The goal is clean, consistent, and readable code.

Signed-off-by: dming@us.ibm.com

### DIFF
--- a/consensus/pbft/batch_test.go
+++ b/consensus/pbft/batch_test.go
@@ -139,6 +139,8 @@ func TestOutstandingReqsIngestion(t *testing.T) {
 
 func TestOutstandingReqsResubmission(t *testing.T) {
 	omni := &omniProto{}
+	config := loadConfig()
+	config.Set("general.batchsize", 2)
 	b := newObcBatch(0, loadConfig(), omni)
 	defer b.Close() // The broadcasting threads only cause problems here... but this test stalls without them
 
--- a/consensus/pbft/batch_test.go
+++ b/consensus/pbft/batch_test.go
@@ -187,7 +187,9 @@ func TestOutstandingReqsResubmission(t *testing.T) {
 	execute()
 
 	if b.reqStore.outstandingRequests.Len() != 0 {
-		t.Fatalf("All requests should have been executed and deleted after exec")
+		config := loadConfig()
+		config.Set("general.batchsize", 2)
+		newObcBatch(0, config, omni)
 	}
 
 	// Simulate changing views, with a request in the qSet, and one outstanding which is not
--- a/consensus/pbft/batch_test.go
+++ b/consensus/pbft/batch_test.go
@@ -141,7 +141,7 @@ func TestOutstandingReqsResubmission(t *testing.T) {
 	omni := &omniProto{}
 	config := loadConfig()
 	config.Set("general.batchsize", 2)
-	b := newObcBatch(0, loadConfig(), omni)
+	b := newObcBatch(0, config, omni)
 	defer b.Close() // The broadcasting threads only cause problems here... but this test stalls without them
 
 	transactionsBroadcast := 0
@@ -189,9 +189,7 @@ func TestOutstandingReqsResubmission(t *testing.T) {
 	execute()
 
 	if b.reqStore.outstandingRequests.Len() != 0 {
-		config := loadConfig()
-		config.Set("general.batchsize", 2)
-		newObcBatch(0, config, omni)
+		t.Fatalf("All requests should have been executed and deleted after exec")
 	}
 
 	// Simulate changing views, with a request in the qSet, and one outstanding which is not
--- a/consensus/pbft/config.yaml
+++ b/consensus/pbft/config.yaml
@@ -32,7 +32,7 @@ general:
     logmultiplier: 4
 
     # How many requests should the primary send per pre-prepare when in "batch" mode
-    batchsize: 1000
+    batchsize: 500
 
     # Whether the replica should act as a byzantine one; useful for debugging on testnets
     byzantine: false
--- a/consensus/pbft/config.yaml
+++ b/consensus/pbft/config.yaml
@@ -32,7 +32,7 @@ general:
     logmultiplier: 4
 
     # How many requests should the primary send per pre-prepare when in "batch" mode
-    batchsize: 2
+    batchsize: 1000
 
     # Whether the replica should act as a byzantine one; useful for debugging on testnets
     byzantine: false


srderson:
Where are the performance results? These would likely be very useful for the community if you're able to share.

JonathanLevi:
Yes, I jumped on this too, out of interest/curious to see some perf numbers! 

Specifically for this PR, it will be good to know by how much does the proposed change improve things (even relatively... if that's easier to report on a single system).

dongmingh:
With commit: 9777406bffeb68e35f767f89d0e0c0b72626329b dated 6/23/2016
4 peer network with pbft-batch, security and privacy enabled
chaincode: example02
hardware: Lenovo W540
run duration: 10 min and 60 min

TPS measured at SoapUI for invoke for 10 min/60 min runs:
batchsize = 2, 31/13 
batchsize = 100, 36/15
batchsize = 500, 47/21
batchsize=1000, 48/21

TPS measured at SoapUI for query for 10 min/60 min runs:
batchsize = 2, 254/250
batchsize = 100, 255/252
batchsize = 500, 252/247
batchsize=1000, 252/249

For commit level cdfbdcd66de75cb832114d8815999189c7f6da9b dated 6/24/2016 under the same test as above,
TPS measured at SoapUI for invoke for 10 min/60 min runs:
batchsize=1000, 72/66

TPS measured at SoapUI for query for 10 min/60 min runs:
batchsize=1000, 252/248

JonathanLevi:
Interesting, thank you very much!

Noting that the biggest improvement is in the SoapUI for invoke, 

> batchsize = 2, 31/13
> batchsize = 100, 36/15
> batchsize = 500, 47/21
> batchsize=1000, 48/21

and that this is a considerable improvement (from the existing batchsize value of 2!).

So it seems as the biggest 'bang for the batch(size)' ;-), comes from the increase from 2, 100 and 500. From 500 to 1000 does not give us that much, right? (I see 1 extra 'unit')

JonathanLevi:
If you don't mind, I have 3 questions please (but if I'm a pain ;-) and/or these questions do not make sense, then we can leave this/chat on Slack:
1. The batchsize in Consensus is the number of pre-prepare sends... right? **What's the main trade-off/risk using too high a batchsize** (for Concensus) is used (is it memory/CPU/network bandwidth/etc...) ? I bet that something 'has to give'? I honestly don't know enough about this, so happy to learn.
2. What happens **if we go above 1000**? Were 1200, 1500 or 2000 feasible to try at all? Or are these simply too high and something goes wrong, or is the performance improvement too negligible, or maybe even worse?
3. Probably related to the answers (1 and 2) above, but I just want to frame my questions:
   Given these results I am wondering **whether we should set it to 500**, as the improvement from 500 to 1000 seems (to me) like an extra 1 TPS (assuming that a high batch size comes with a price) ?

BTW: I'm interested in this also because we were looking at various TCerts batchsize/levels, and it wasn't always clear what's the right thing to do, tbh.

tuand27613:
@JonathanLevi ,  your question 1.  batchsize is how many requests are batched together before passing on to consensus. Basically, we're saving on the PBFT protocol messages.

kchristidis:
I have another question: for the invoke case, what do you count as a transaction in those metrics?

Do you make sure that the transaction is added to the blockchain? (Of any node. As long as 1 node has it, at least 2f+1 have participated in the ordering of it.)

JonathanLevi:
@tuand27613 Thank you for the prompt reply. This feels like an important change!

Yes, so while I am purposely over-simplifying this, but broadly speaking, does higher batchsize mean?
    "**Saving on the PBFT protocol messages**"
    vs.
    "**Increasing the workload/stress at the chaincode execution point**" [UPDATED:] "**... and probably at other points in the system as well**"

JonathanLevi:
My 2 cents are that given the limited (at this stage) performance information/data we have at the moment for each component in the system, let alone the full end-to-end integration, and in view of the above performance numbers, we should probably try to change to 500 (and not to 1000 right away).

To me it feels as if 500 is already a big improvement, and I suspect that other components of the **system (!)** might need to be able to keep up/adapt to this new batchsize, more issues will surface that will need addressing, etc. FWIW.

Maybe I'm taking a too simplistic/prudent approach and I'd love to hear/learn/discuss more, especially in case, there are other benefits to having the batchsize of 1000 ?

christo4ferris:
I would agree that we need more experience with something like this and not just under performance stress but various other failure modes tested as well.

tuand27613:
@JonathanLevi  I'd say increased batchsize ==> increased workload at cc execution point and probably at other points in the system as well.  

linking to #1338

JonathanLevi:
Thanks again @tuand27613. Yes, 'and probably at other points in the system as well'... I hear you. Updating the above.

For the record, I agreed with @binhn 's 'no magic' number before the link to #1338 ;-)

dongmingh:
I did a test with batchsize=500 on the commit cdfbdcd66de75cb832114d8815999189c7f6da9b, and the invoke tps is 64/66 for 10 min/60 min runs compare to 72/66 for batchsize=1000.

tuand27613:
@dongmingh, suggest you create a wiki page where you record your results in tabular format.  It would be easier to read and compare  rather than checking comments in a PR.

Also,  is this something to automate and make part of CI ?  Maybe pick a build a day and build up some historical data.

dongmingh:
Yes, from the performance perspective, batchsize 1000 performs better then batchsize 500 only in the beginning (first 10 mins) probably because the queue is bigger.
However, I agree that both batchsize of 500 and 1000 achieve the exact same level of performance (at the end of the 60 min runs).

dongmingh:
@tuand27613 I will add these data to performance wiki page.

kchristidis:
(Not sure if the answer is obvious to everybody else, but I'll repeat the question: https://github.com/hyperledger/fabric/pull/2003#issuecomment-228436159)

dongmingh:
@kchristidis I use SoapUI to generate REST API.  SoapUI calculates numbers of transaction and tps.  The tps in the metrics are from SoapUI.

At the end of test, I calculate the values of a and b based on the numbers of invoke transactions sent from SoapUI.  Then I query a and b to see if they match with the calculated values.  Then I send an invoke followed by a query to see if the transactions get executed correctly.

corecode:
The tradeoff for larger batches is that sending/receiving them takes more time, which delays other protocol packets.  Imagine a 1GB block into a 4-node network: the primary would try to send out 3GB.  Any commit message (for a previous block) would be delayed until this 1GB block has been sent.

Maybe we should look into specifying the batch size in bytes, not transactions.  More difficult to count in the code, easier to reason about everywhere else.

binhn:
@corecode I agreed size would be a better metric to balance network io. The question is what size is reasonable.  MTU is about 1500 bytes. I think (not sure) databases default to 4k. We would need some work to figure out the optimal balance.

I would suggest for now, if folks are concerned about 1000 tx batch vs 500, then change to 500 and merge that until we can figure out the size in bytes.

vukolic:
batch size of 1000 on <100 tps throughput means >10s latency. This seems suboptimal as a default value.

JonathanLevi:
Hi @vukolic, I agree. There is also an agreement that we can clearly 'do better than 2'.

Shall we proceed with 500 (for now, until we re-visit the metric, resolve other issues, etc.) ?

JonathanLevi:
Hi @vukolic, I agree. There is also an agreement over the 'we can clearly do better than 2' statement. Feels more like something in between.

Shall we proceed with the 500 value (for now, until we re-visit the metric, test & resolve other issues, etc.) ?

christo4ferris:
I restarted Travis again, but it appears that this build terminates after overflowing the log max size. We really need to look at better approaches for running builds that don't require DEBUG level logging throughout. Possibly we could reserve that for failed builds?

srderson:
@vukolic any updates regarding @JonathanLevi's question? Still unsure if this should be merged or not

vukolic:
@srdeson @JonathanLevi Sorry, I did not perceive my opinion is blocking. I think 500 is probably ok.

srderson:
This PR is failing during unit tests

```
--- FAIL: TestOutstandingReqsResubmission (0.00s)
    obc-batch_test.go:190: All requests should have been executed and deleted after exec
```

@corecode @jyellick any ideas?

jyellick:
The batch tests generally assume a batch size of 2.  You can see [this line](https://github.com/hyperledger/fabric/blob/master/consensus/obcpbft/obc-batch_test.go#L142) is loading the default config.

My guess is you should be able to fix this by changing that line to the following:

```
config := loadConfig()
config.Set("general.batchsize", 2)
newObcBatch(0, config, omni)
```

srderson:
Thanks @jyellick. @dongmingh would you mind changing that line pushing the commit to see if this passes Travis? This also needs to be rebased on master as there's a merge conflict currently. Thanks.

srderson:
@jyellick @dongmingh Seems this is still failing with the suggested fix?

srderson:
Thanks @jyellick and @dongmingh, looks to be passing now. LGTM

PR #2006, opened 2016-06-24
core/rest: DELETE /registrar// deletes entire local crypto/client directory

dubek:
core/rest: Add format validation on URL parameters to prevent remote deletion of local dirs.
## Description

Added simple validation via regexp on URL path components to prevent these parameters being empty or with unexpected chars.

This commit adds simple format validation:
- ID for `/registrar/:id` endpoints must be at least 1 char long and all chars must be `\w` = `[a-zA-Z0-9_]`
- Block number for `/chain/blocks/:id` must be at least 1 digit long and all chars must be digits
- UUID for `/transactions/:uuid` endpoint must be `[\w-]+` (we can be more      strict here)

Note that the first limit will prevent user-IDs from having minus chars, dots, equals, at-signs, etc. This might be too strict and break existing users. Alternatively, we can replace it to something more relaxed like `[^/]+` (one or more chars that are not slash - we definitely don't support slashes in usernames; I suggest avoiding backslashes too if we're supposed to work on Windows).
## Motivation and Context

Passing an empty :id parameter to DELETE /registrar/:id endpoint causes     deletion of the entire `/var/hyperledger/production/crypto/client`     directory. For example, run this in vagrant:

```
$ mkdir -p /var/hyperledger/production/crypto/client
$ echo "important_data" > /var/hyperledger/production/crypto/client/myfile
$ curl -v -X DELETE http://127.0.0.1:5000/registrar//
*   Trying 127.0.0.1...
> DELETE /registrar// HTTP/1.1
> User-Agent: curl/7.35.0
> Host: 127.0.0.1:5000
> Accept: */*
> 
< HTTP/1.1 200 OK
< Date: Fri, 24 Jun 2016 19:03:23 GMT
< Content-Length: 54
< 
{"OK": "Deleted login token and directory for user ."}

$ ls -l /var/hyperledger/production/crypto/client
ls: cannot access /var/hyperledger/production/crypto/client: No such file or directory

# Whoops important data is gone!
```
## How Has This Been Tested?

core/rest unit tests still pass.

Verified the above scenario. Now the curl call response is:

```
< HTTP/1.1 404 Not Found
...
{"Error":"Openchain endpoint not found."}
```

because the route is no longer matched (due to the stricter URL parameter validation).
## Checklist:

### DIFF
--- a/core/rest/rest_api.go
+++ b/core/rest/rest_api.go
@@ -26,6 +26,7 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"regexp"
 	"strconv"
 	"strings"
 
@@ -182,6 +183,37 @@ func getRESTFilePath() string {
 	return localStore
 }
 
+// isEnrollmentIDValid returns true if the given enrollmentID matches the valid
+// pattern defined in the configuration.
+func isEnrollmentIDValid(enrollmentID string) (bool, error) {
+	pattern := viper.GetString("rest.validPatterns.enrollmentID")
+	if pattern == "" {
+		return false, errors.New("Missing configuration key rest.validPatterns.enrollmentID")
+	}
+	return regexp.MatchString(pattern, enrollmentID)
+}
+
+// validateEnrollmentIDParameter checks whether the given enrollmentID is
+// valid: if valid, returns true and does nothing; if not, writes the HTTP
+// error response and returns false.
+func validateEnrollmentIDParameter(rw web.ResponseWriter, enrollmentID string) bool {
+	validID, err := isEnrollmentIDValid(enrollmentID)
+	if err != nil {
+		rw.WriteHeader(http.StatusInternalServerError)
+		json.NewEncoder(rw).Encode(restResult{Error: err.Error()})
+		restLogger.Errorf("Error when validating enrollment ID: %s", err)
+		return false
+	}
+	if !validID {
+		rw.WriteHeader(http.StatusBadRequest)
+		json.NewEncoder(rw).Encode(restResult{Error: "Invalid enrollment ID parameter"})
+		restLogger.Errorf("Invalid enrollment ID parameter '%s'.\n", enrollmentID)
+		return false
+	}
+
+	return true
+}
+
 // Register confirms the enrollmentID and secret password of the client with the
 // CA and stores the enrollment certificate and key in the Devops server.
 func (s *ServerOpenchainREST) Register(rw web.ResponseWriter, req *web.Request) {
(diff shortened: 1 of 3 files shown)


kchristidis:
This feels relevant: https://xkcd.com/327/

srderson:
Thanks for submitting this.

Should these filters be added to the functions `GetEnrollmentID` `DeleteEnrollmentID` `GetEnrollmentCert`, etc. instead of adding them to the routes? Seems like this may offer additional security if anyone else tries calls these functions in the future.

For a separate PR
1. REST RPC should likely be disabled by default
2. There should be an `allowip` setting that limits the addresses which can access the RPC API.

srderson:
In terms of the allowed characters, I'd rather be too restrictive at the start and allow additional characters as required. It's much more difficult to go in the other direction.

dubek:
@srderson Thanks for the feedback. Initially I looked at an option to add a `ValidateIDParameter` middleware function to the router (to wrap those specific functions), then found about the validation via regexp in the route itself. We can perform the validation in the functions themselves (and then return HTTP 400 instead of 404 with the current patch). 

Would you like me to re-write this PR with explicit calling to a validation function at the beginning of GetEnrollmentID DeleteEnrollmentID GetEnrollmentCert, etc ?

srderson:
That would be my preference. Then it would be easy to be very restrictive by default and have the validation function read a regex from core.yaml. If anyone wants to be less restrictive, it's possible, but they are doing so at their own risk.

dubek:
So: fetch the regexp from the config with fallback to the strict patterns (`\w+` for enrollment ID).

It might need to wait for Monday though...

srderson:
Yes, or just have the strict patterns as the default values we provide in the initial `core.yaml`.

I think waiting until early next week is ok as Fabric is still in developer preview. If you want though, we can merge this as a stop gap and open up an issue. I'm slightly concerned about breaking existing users as you suggested due to the new ID restrictions with no ability to relax the restrictions if needed.

dubek:
I'm OK with waiting for the proper solution; if someone needs to patch their REST endpoint urgently, they can apply this patch.

dubek:
Another try - 38e28c825c04b8426993b44ad631427ba900c54e

The valid pattern is set in `core.yaml` key `rest.validPatterns.enrollmentID` . The default pattern we ship with is `^\w+$` .

Do we need to copy this new configuration setting to other *yaml files in the project? For tests etc?

Note that with the updated fix the response for the original bad curl is 400 (and not 404):

```
$ curl -v -X DELETE http://127.0.0.1:5000/registrar//
*   Trying 127.0.0.1...
> DELETE /registrar// HTTP/1.1
> User-Agent: curl/7.35.0
> Host: 127.0.0.1:5000
> Accept: */*
> 
< HTTP/1.1 400 Bad Request
< Date: Mon, 27 Jun 2016 16:45:41 GMT
< Content-Length: 44
< 
{"Error":"Invalid enrollment ID parameter"}
```

srderson:
"Do we need to copy this new configuration setting to other *yaml files in the project?"
I wish we were not copying the core.yaml, but that's another issue. Given that Travis passed and it looks like your changes would result in a failure if someone tried to use REST without the config set, I think it's good.

LGTM. Thanks for the fix!

srderson:
@dubek I think this needs to go into the 0.5-developer-preview branch. Would you like to submit a PR against that branch or do you want me to open one?

dubek:
@srderson You got it - #2094 . It doesn't not include the unit-tests and not all the other JSON-related fixes in core/rest.

Issue #1874, opened 2016-06-16
Peer PBFT Batch does not work on queue after two VPs bounce


scottz64:
## Description

Some basic behavior with Consensus seems to be no longer working, in a specific scenario. This test fails on test images from 6/15 (f8ffcd1) and 6/16 (04f076b). The test passed on 6/9 image (a94f5a7). 

In the following scenario, after recovery, the values for A and B seem to remain stuck with the same values prior to the failures. The Invokes that were requested while there were only 2 peers active, AND the Invokes that were requested after all 4 nodes recovered, are NOT processed and not reflected in the query results. Caveat: the failure occurs ONLY if we send Invoke requests WHILE only 2 peers active; if we stop and start two nodes and THEN Invoke and Query, then all seems to work fine. Another note: it does not make any difference whether we request Invokes while 4 peer nodes are up PRIOR to stopping two of them.
## Describe How to Reproduce
1.     Using Vagrant and docker containers, running go script using our SDK toolkit, we started 4 peer nodes with consensus and security enabled, running PBFT Batch and batchsize=2.
2.     Deploy chaincode_example02 values for A (200) and B (100)
3.     stop 2 peers
4.     invokes to transfer from A to B
5.     start the 2 peers again
6.     invokes to transfer from A to B
7.     query  <----- FAILURE : This query returns the original values of A (200) and B (100), essentially ignoring all the invokes.

Note: Pull Request #1867  represents a similar scenario with a Behave test - which PASSES. Our team also confirmed the 1867 passing result by using a similar Behave test with PBFT Batch (although we could reproduce the failure with the Behave test case using PBFT Classic). Not sure why the Behave test passes and our GO test fails. Maybe the config parameters being used are different. Maybe the logs will help. Attachments include the GO test script, the local_fabric.sh script used to start the docker peers and pass config parameters, and the logfiles. PEER3 may be of interest.

[InvokesWhenNoConsensusCauseFreeze.tar.gz](https://github.com/hyperledger/fabric/files/318934/InvokesWhenNoConsensusCauseFreeze.tar.gz)

corecode:
You need to keep submitting invokes, or switch on periodic null requests.  If there is no activity in the network, the peers will not catch up.

tuand27613:
@corecode,  i tried running the behave test from PR #1867 . Looks like we're not receiving any more requests after the initial deploy and invoke ? After restarting vp3 , I don't see vp0 saying it "received request"

[...]

PR #2008, opened 2016-06-24
Clear batch store on view change

jyellick:
## Description

This changeset clears the batch store on view change, does not start the request timer when not in an active view, and decreases the batch timeout to be less than the request timeout.

Usual SMEs are @kchristidis @corecode and @tuand27613, though you may wish to check PR #2007 (the companion to this PR against 0.5) for signoffs.
## Motivation and Context

This is intended to address symptoms of #1874 as well as a duplicated deploy transaction seen in busywork.

If a replica was a primary and had a transaction in its batch store when the view change timer expired, it would include that request in the next batch when it became primary again, leading to transaction duplication.

If only a single request came in, then the view change timer would expire before the batch was created, and the network would view change over and over again until another request came in, never executing the first request.

Finally, while processing a new view, it was possible for the request timer to start, negating the doubling of the new view timer.
## How Has This Been Tested?

This has been tested against the busywork stress2b test, and a new unit test checking for the batchStore emptying on view change has been added.
## Checklist:

### DIFF
--- a/consensus/obcpbft/config.yaml
+++ b/consensus/obcpbft/config.yaml
@@ -47,7 +47,7 @@ general:
         # and this much time has elapsed since the current batch was formed
         batch: 1s
 
-        # How long may a request take between reception and execution
+        # How long may a request take between reception and execution, must be greater than the batch timeout
         request: 2s
 
         # How long may a view change take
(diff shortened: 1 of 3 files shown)


srderson:
Reviewed as part of PR #2007 in the 0.5 branch. LGTM

PR #2013, opened 2016-06-25
Fixed GetRows() by moving defer into dependent goroutine

bren3582:
## Description

Within the `GetRows` function in `core/chaincode/shim/chaincode.go`, I have moved `defer iter.Close()` inside the goroutine which depends on `iter` for the reason explained below.
## Motivation and Context

The `GetRows` function uses a goroutine to get rows from a `StateRangeQueryIterator`. This iterator is used within the goroutine. Defer is used to wait until the function completes to clean up the iterator. However, the function can finish before the goroutine (in my case when there are a large number of rows requested), causing the iterator to be cleaned up before the goroutine is finished using it. This results in a chaincode error:

```
00:00:00 [shim] DEBU : [xxx]Sending RANGE_QUERY_STATE_NEXT
00:00:00 [shim] DEBU : [xxx]Received message ERROR from shim
00:00:00 [shim] DEBU : [xxx]Handling ChaincodeMessage of type: ERROR(state:ready)
Error starting chaincode: Error handling message: [xxx]Chaincode handler FSM cannot handle message (ERROR) with payload size (30) while in state: ready
```
## How Has This Been Tested?

I tested this change against a chaincode project which requests a large number of rows from a table using the `GetRows` function. It solves the error shown above in a logical way. I created a separate [test to show the issue simply](https://play.golang.org/p/TNEh2gaB_m).
## Checklist:

Signed-off-by: Brian Ensor briancensor@gmail.com

### DIFF
--- a/core/chaincode/shim/chaincode.go
+++ b/core/chaincode/shim/chaincode.go
@@ -572,11 +572,11 @@ func (stub *ChaincodeStub) GetRows(tableName string, key []Column) (<-chan Row, 
 	if err != nil {
 		return nil, fmt.Errorf("Error fetching rows: %s", err)
 	}
-	defer iter.Close()
 
 	rows := make(chan Row)
 
 	go func() {
+		defer iter.Close()
 		for iter.HasNext() {
 			_, rowBytes, err := iter.Next()
 			if err != nil {


ghaskins:
This patch LGTM.  Normally I do not like to decouple a defer from its underlying acquisition since it invites leaks.  However, in this case, I do not think there is a better way to do it.

ghaskins:
@srderson @binhn @muralisrini @jeffgarratt any comments?

core/chaincode/shim/chaincode.go
```diff
@@ -572,11 +572,11 @@ func (stub *ChaincodeStub) GetRows(tableName string, key []Column) (<-chan Row,
...
 	}
-	defer iter.Close()
 
 	rows := make(chan Row)
 
 	go func() {
+		defer iter.Close()
 		for iter.HasNext() {
 			_, rowBytes, err := iter.Next()
 			if err != nil {
```
muralisrini, line 582: That change LGTM.

However, should we also exit the loop when err != nil instead of calling close(..) and continuing ? If the channel is closed any point in adding other rows to it ? Also, if there _is_ an error, don't we want to provide that to the user as well ? 

But defer to @srderson

muralisrini:
@ghaskins @srderson added an inline comment.

ghaskins:
I suspect this needs to be fixed in the dp branch too.

bren3582:
If we were to replace the first to instances within the goroutine of `close(rows)` with `break` that would exit the loop as @muralisrini suggested and the `close(rows)` line immediately after the loop would still be hit. However, that still doesn't pass an error back to the user. One way to handle this would be something like:

bren3582:
If we were to replace the first two instances within the goroutine of `close(rows)` with `break` that would exit the loop as @muralisrini suggested and the `close(rows)` line immediately after the loop would still be hit. However, that still doesn't pass an error within the goroutine back to the user.

One way to accomplish this would be changing the `rows` channel to a struct channel in which the struct contains a `row` and an `error`. Another option is creating a new `error` channel, necessitating access using `select`. I've thought quite a bit about how either of these changes could be made in a backward-compatible way, without much luck. If we went with the error channel option, it is possible to achieve backward-compatibility as shown below but it's not very clean.

``` go
func (stub *ChaincodeStub) GetRows(tableName string, key []Column) (<-chan Row, error) {
    rows, errorChan, err := stub.GetRowsFixed(tableName, key)
    return rows, err
}

func (stub *ChaincodeStub) GetRowsFixed(tableName string, key []Column) (<-chan Row, <-chan error, error) {
    ...
}
```

As for @ghaskins's comment about decoupling the defer from its underlying acquisition, that can be solved by rearranging a bit more within `GetRows`, but this introduces another possible error within the goroutine that would need to be made accessible somehow.

``` go
func (stub *ChaincodeStub) GetRows(tableName string, key []Column) (<-chan Row, error) {
    ...
    go func() {
        // next 4 lines moved into goroutine from outside
        iter, err := stub.RangeQueryState(keyString+"1", keyString+":")
        if err != nil {
            //return nil, fmt.Errorf("Error fetching rows: %s", err)
        }

        defer iter.Close()
        ...
    }
    ...
}
```

Comments?

core/chaincode/shim/chaincode.go
sachikoy, line 582: I just want to comment that this PR did not work for us. The problem is that we query two tables consequtively, and the second query fails because (IMU) the access to the first table has not finished.

ghaskins, line 582: That doesn't sound like this PR is the root cause, but rather the fact that the iterators are asynchronous and/or concurrently created, to begin with.  If your program was working prior to this patch, it sounds like it was relying on the race with the (clearly incorrect) premature close of the first iterator.  My question is: why does the second iterator seemingly rely on the first being closed before it will function properly?  This sounds like the real bug.

bren3582:
@sachikoy Do you have any example code that exposes the issue with querying two tables consecutively? I'd like to figure out why it fails. This PR fixes what seems like an obvious bug to me, so perhaps it is revealing a new bug which is causing issues with your code?

@muralisrini I'd like to separate the error handling design issue from this defer placement bug fix if that's alright with you.

srderson:
@bren3582 Agreed that it makes sense to separate the error handling from this bug fix. We may be able to change API in the Fabric Next Architecture so we can make the change without worrying about backwards compatibility.

core/chaincode/shim/chaincode.go
srderson, line 582: Agree with @ghaskins, not sure it's related. The fix LGTM, but I'll run some local tests to see if I encounter anything.

muralisrini:
+1 @bren3582 @srderson for separation of issues. Thanks!

core/chaincode/shim/chaincode.go
srderson, line 582: @sachikoy Could you provide a test case? I'm not seeing the same error in my testing.

sachikoy, line 582: @ghaskins Good point. In the chaincode log, an error message which says "Another state request pending for this Uuid. Cannot process." is recorded. It looks the code is checking and make sure that there is only one channel per uuid. I don't know why.

sachikoy, line 582: @srederson The code is part of a bigger prototype which we cannot share in public. We'll make a sample code which reproduce the problem.

sachikoy:
@bren3582 @srderson The code is part of a bigger prototype which we cannot share in public. We'll make a sample code which reproduce the problem.
A simple description of what the code does is as follows:
- Read many rows from a table using GetRows() and store it into an array,
- then while iterating over the array, access another table using GetRow().

srderson:
@sachikoy Do you only see the issue with this patch or were you seeing this issue previously?

bren3582:
@srderson @sachikoy I'm investigating what seems to be a new bug as a result of this patch that I haven't yet reproduced consistently. Perhaps it's related to the issue you mentioned.

bren3582:
~~It seems that if QUERY_COMPLETED occurs before RANGE_QUERY_STATE_CLOSE which can happen sometimes, why I'm not sure, that causes a crash.~~ I've now seen the crash with RANGE_QUERY_STATE_CLOSE after QUERY_COMPLETED, so that may not be of much help. I'm hoping that someone more knowledgeable about the architecture of chaincode.go and handler.go may have some insight. I am curious why the StateRangeQueryIterator struct stores its own handler reference instead of using the one already existing in chaincode.go, but that may be completely unrelated.

In order to reproduce the crash, make the change suggested by this pull request and also within chaincode.go import time and put `time.Sleep(5 * time.Second)` on line 380 right after `func (iter *StateRangeQueryIterator) Close() error {`

bren3582:
Here is the output from the peer panic:

```
panic: runtime error: invalid memory address or nil pointer dereference
    panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xb code=0x1 addr=0x0 pc=0x638e21]

goroutine 69 [running]:
panic(0xd7ff00, 0xc82000e0f0)
    /opt/go/src/runtime/panic.go:464 +0x3e6
github.com/hyperledger/fabric/core/chaincode.(*Handler).handleRangeQueryStateClose.func1.1(0xc82009b500, 0xc820343db0, 0xc82038bea0)
    /opt/gopath/src/github.com/hyperledger/fabric/core/chaincode/handler.go:911 +0x261
panic(0xd7ff00, 0xc82000e0f0)
    /opt/go/src/runtime/panic.go:426 +0x4e9
github.com/hyperledger/fabric/core/chaincode.(*Handler).getRangeQueryIterator(0xc82009b500, 0x0, 0xc820486f30, 0x24, 0x0, 0x0)
    /opt/gopath/src/github.com/hyperledger/fabric/core/chaincode/handler.go:155 +0x89
github.com/hyperledger/fabric/core/chaincode.(*Handler).handleRangeQueryStateClose.func1(0xc82009b500, 0xc820343db0)
    /opt/gopath/src/github.com/hyperledger/fabric/core/chaincode/handler.go:925 +0x52f
created by github.com/hyperledger/fabric/core/chaincode.(*Handler).handleRangeQueryStateClose
    /opt/gopath/src/github.com/hyperledger/fabric/core/chaincode/handler.go:945 +0x3f
```

sachikoy:
@srderson @bren3582 After applying this PR, the problem occurs more consistently, i.e., the second request almost always fails due to the error "Another state request pending for this Uuid. Cannot process." However, before the patch, we observed that the validating peer crashes, when the number of the rows in the first table becomes large.

And I think the reason we get "Another state request pending for this Uuid. Cannot process."  is a kind of obvious, because the current implementation of handler.go allows only one channel to be used per UUID.
That is, there are multiple threads,
- thread 1 : chaincode invocation 
- thread 2 : new thread created inside GetRows using the go-routine, which opens a channel to query the state

What happens is as follows:
1) first, chaincode is invoked in thread 1
2) then when the chancode calls GetRows, thread 2 is created, opens the channel
3) then in the thread 1, chaincode calls the second GetRow(), while the channel in thread 2 is still open.

So I was thinking about two approaches to fix the issue. 
- Since chaincode can be executed only synchronously anyway, maybe we don't need to create thread 2 in GetRows, and let the thread 1 do all the job
- Or, why do we need to limit the number of channel per UUID (i.e., per chaincode invocation)? Perhaps we could allow multiple channels per each chaincode invocation? 

Any thoughts?

sachikoy:
Attached is a sample chaincode to reproduce the problem that occurs with this PR.
(Thanks @Ohsaworks  !)
[race_condition.zip](https://github.com/hyperledger/fabric/files/342895/race_condition.zip)

sachikoy:
I analysed detailed behaviour of handler to see why the channel for the fist GetRows() is not closed when the second GetRow() is called.
The reason is when RANGE_QUERY_STATE_CLOSE is sent from shim, it  does not get RESPONSE immediately. The RESPONSE is delayed, and so that the second GetRow() request is issued before deleteChannel is called.
In other case such as GET_STATE or RANGE_QUERY_STATE, the RESPONSE is returned almost immediately.

08:47:25.853 [shim] DEBU : [45b4fc60]Received message QUERY from shim
08:47:25.853 [shim] DEBU : [45b4fc60]Handling ChaincodeMessage of type: QUERY(state:ready)
08:47:25.854 [HDLS] INFO : Query called, function=shortTable
08:47:25.854 [HDLS] INFO : calling GetRows(), model=ShortTable
08:47:25.854 [shim] INFO : GetRows() table=ShortTable, key[0]=[]
**08:47:25.855 [shim] DEBU : [45b4fc60] handleGetState is called, key=10ShortTable**
08:47:25.855 [shim] INFO : [45b4fc60] createChannel called
_08:47:25.855 [shim] DEBU : [45b4fc60]Sending GET_STATE_
_08:47:25.856 [shim] DEBU : [45b4fc60]Received message RESPONSE from shim_
08:47:25.856 [shim] DEBU : [45b4fc60]Handling ChaincodeMessage of type: RESPONSE(state:ready)
08:47:25.857 [shim] DEBU : [45b4fc60]before send
08:47:25.857 [shim] DEBU : [45b4fc60]after send
_08:47:25.857 [shim] DEBU : [45b4fc60]Received RESPONSE, communicated (state:ready)_
_08:47:25.857 [shim] DEBU : [45b4fc60]GetState received payload RESPONSE_
08:47:25.858 [shim] INFO : [45b4fc60] deleteChannel called
**08:47:25.858 [shim] DEBU : [45b4fc60] handleRangeQueryState()**
08:47:25.858 [shim] INFO : [45b4fc60] createChannel called
_08:47:25.858 [shim] DEBU : [45b4fc60]Sending RANGE_QUERY_STATE_
_08:47:25.865 [shim] DEBU : [45b4fc60]Received message RESPONSE from shim_
08:47:25.865 [shim] DEBU : [45b4fc60]Handling ChaincodeMessage of type: RESPONSE(state:ready)
08:47:25.866 [shim] DEBU : [45b4fc60]before send
08:47:25.866 [shim] DEBU : [45b4fc60]after send
_08:47:25.866 [shim] DEBU : [45b4fc60]Received RESPONSE, communicated (state:ready)_
_08:47:25.866 [shim] DEBU : [45b4fc60]Received RESPONSE. Successfully got range_
08:47:25.867 [shim] INFO : [45b4fc60] deleteChannel called
08:47:25.867 [HDLS] INFO : fetch all the rows, model=ShortTable
08:47:25.868 [HDLS] INFO : fetching row, count=1
...
08:47:25.872 [HDLS] INFO : fetching row, count=10
**08:47:25.873 [shim] DEBU : [45b4fc60] handleRangeQueryStateClose()**
08:47:25.873 [shim] INFO : [45b4fc60] createChannel called
_08:47:25.873 [shim] DEBU : [45b4fc60]Sending RANGE_QUERY_STATE_CLOSE_
08:47:25.874 [HDLS] INFO : Querying the table using GetRow()
**08:47:25.875 [shim] DEBU : [45b4fc60] handleGetState is called, key=10ShortTable12 <-- this corresponds to the second GetRow()**
08:47:25.875 [shim] INFO : [45b4fc60] createChannel called
08:47:25.875 [shim] INFO : [45b4fc60] Channel exists
_08:47:25.876 [shim] DEBU : [45b4fc60]Received message RESPONSE from shim *\* <-- this RESPONSE corresponds to the RANGE_QUERY_STATE_CLOSE for the first GetRows() *\* _
_08:47:25.876 [shim] DEBU : [45b4fc60]Handling ChaincodeMessage of type: RESPONSE(state:ready)_
08:47:25.881 [shim] DEBU : handleGetState : Another state request pending for this Uuid. Cannot process.
08:47:25.881 [HDLS] INFO : GetRow() returned error : Error fetching row from DB: [45b4fc60]Channel exists
08:47:25.881 [shim] DEBU : [45b4fc60]Query execution failed. Sending QUERY_ERROR
08:47:25.881 [shim] DEBU : [45b4fc60]before send
08:47:25.881 [shim] DEBU : [45b4fc60]after send
_08:47:25.881 [shim] DEBU : [45b4fc60]Received RESPONSE, communicated (state:ready)_
_08:47:25.881 [shim] DEBU : [45b4fc60]Received RESPONSE. Successfully got range_
08:47:25.881 [shim] INFO : [45b4fc60] deleteChannel called

muralisrini:
thanks @sachikoy for the detailed input, sample and analysis!.... will check today.

muralisrini:
@bren3582 could you take a look at PR #2095 please ? It attempts to fix all holes stemming from this investigation and #2088.

bren3582:
Complete fix in PR #2095. Thanks for your help everyone.

PR #2016, opened 2016-06-26
Discovery: maintain and persist list of recent connections

kchristidis:
## Description

With this set of changes, every node should now maintain a list of recent connections via the `discovery` service. When the appropriate flag (`peer.discovery.persist`) in the config file is set to `true` (default), the node also persists the discovery list to the DB, and repeats this task whenever the list is updated. If the node crashes, it will load the list upon restarting, and attempt to connect to the addresses it carries, via the touch service.

This PR extends the existing `discovery` service by including non-anchor nodes in its list, and by adding persistence support. It moves the `discovery` package in the `core` directory as this seemed a much more logical fit.

This set of changes also simplifies the peer connection logic in the `peer` package, and adds support for the `discovery.maxTouchNodes` configuration option originally added by @jeffgarratt; now the touch service will attempt to connect to a maximum of `maxTouchNodes` nodes. (This new list is generated randomly from the discovery list.)
## Motivation and Context

Consider this scenario:
1. All the nodes in a network use the same anchor node.
2. The anchor node crashes.
3. A node crashes and comes back online while the anchor node is still down.

This node will not be able to connect to its peers, i.e. we will have a network partition. Do note that the practice listed in step 1 is how we actually launch all of our test networks today. This problem was discovered in #1874.

With this PR, such issues are fixed.
## How Has This Been Tested?
- New unit tests for the new functions added in the `discovery` and `util` package.
-  2 new behave tests have been added, provided by @lhaskins and @scottz64.
- Passes all tests locally.
## Checklist:

Signed-off-by: Kostas Christidis kchristidis@us.ibm.com.


kchristidis:
@jeffgarratt @jyellick @srderson: you guys are good candidates for reviewing this. Let me know of any issues you see with the code and I'll gladly fix them.

PS: @srderson as I noted on Slack, I expect the CI test to fail because I have moved the `discovery` directory, which seems to be hardcoded in in the CI's check scripts. With the modified scripts, this PR passes all tests (at least on my end).

binhn:
LGTM - Thanks.

core/discovery/discovery.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+// RemoveNode removes an address from the discovery list
+func (di *DiscoveryImpl) RemoveNode(address string) bool {
+	di.Lock()
+	defer di.Unlock()
+	if _, ok := di.nodes[address]; ok {
+		di.nodes[address] = false
+	}
+	return !di.nodes[address]
```
jyellick: Is there a scenario where this does not return true?

core/discovery/discovery.go
```diff
@@ -0,0 +1,109 @@
...
+func (di *DiscoveryImpl) GetRandomNodes(n int) []string {
+	var pick string
+	randomNodes := make([]string, n)
+	di.Lock()
+	defer di.Unlock()
+	for i := 0; i < n; i++ {
+		for {
+			pick = di.seq[di.random.Intn(len(di.nodes))]
+			if di.nodes[pick] {
+				break
```
jyellick, line 96: It looks like the same node could be picked multiple times, is that okay here?

core/peer/peer.go
```diff
@@ -322,6 +322,14 @@ func (p *PeerImpl) GetPeers() (*pb.PeersMessage, error) {
 	return peersMessage, nil
 }
 
+func getPeerAddresses(peersMsg *pb.PeersMessage) []string {
+	var addresses []string
+	for _, v := range peersMsg.GetPeers() {
+		addresses = append(addresses, v.Address)
```
jyellick: Not certain how frequently this is called, but the `addresses` length looks like it could be computed to reduce allocations rather than appended up to size.

core/discovery/discovery.go
kchristidis: Fixed in https://github.com/hyperledger/fabric/pull/2016/commits/9aae07b9d6fd76f2ee45d7c2635846772474ad92

core/peer/peer.go
```diff
@@ -808,3 +794,109 @@ func (p *PeerImpl) GetTransactionResultByUUID(txUuid string) (*pb.TransactionRes
 	defer p.ledgerWrapper.RUnlock()
 	return p.ledgerWrapper.ledger.GetTransactionResultByUUID(txUuid)
 }
+
+// initDiscovery load the addresses from the discovery list previously saved to disk and adds them to the current discovery list
+func (p *PeerImpl) initDiscovery() []string {
+	p.discHelper = discovery.NewDiscoveryImpl()
+	p.discPersist = viper.GetBool("peer.discovery.persist")
+	if !p.discPersist {
+		peerLogger.Warning("Discovery list will not be persisted to disk")
```
jyellick, line 803: Should this also abort loading the list from disk?

core/peer/peer.go
kchristidis: Fixed in https://github.com/hyperledger/fabric/pull/2016/commits/91aa90b9f4fbe3452510398f1745b5127367bebb

core/peer/peer.go
kchristidis, line 803: I claim that it should not because you may run the peer with persistence and then decide for whatever reason to switch it off at a later point. Let me know if you disagree. Also note that at some point we'll need to define when exactly we want nodes removed from the discovery list. (The method is already there, but the process that will result in this invocation is undefined.)

jyellick, line 803: Works for me, just wanted to make sure it had been considered.

core/discovery/discovery.go
kchristidis, line 96: Good catch, thank you. Fixed in https://github.com/hyperledger/fabric/pull/2016/commits/fc0d06e79d20f3d37a4d7d5a52b68354d9dc06d3

jyellick:
Looks good to me, thanks for the updates @kchristidis

Issue #769, opened 2016-03-02
Confusion about or new specification of module-level logging setup [re: crypto]?


bcbrock:
Recently a `logging.crypto` key was added to openchain.yaml, which is interpreted in the file

```
obc-peer/openchain/crypto/crypto_settings.go
```

"crypto" here is what the logging package refers to as logging _module_. The other `logging.*` keys in openchain.yaml refer to _subcommands_ of the `obc-peer` command. Module-level logging control in obc-peer subcommands is already implemented and documented - it is not necessary to add new configuration keys to control module-level logging.

Currently a module-level override specification is potentially required for each peer subcommand. However if the desire is to add a way to specify module-level overrides that apply to all obc-peer subcommands that would also be easy to do in a way that would not require defining a unique configuration key for each logging module.

kchristidis:
I'm not sure I follow @bcbrock's phrasing here as closely as I'd like to (sorry Bishop...), and apologies in advance if I'm derailing this thread, but the issue that I'm seeing is this:

If you launch the `membersrvc` binary, the config file that `viper` reads is the `membersrvc/membersrvc.yaml` one. Then when we call into the `crypto` package, then `Init()` function attempts to read the key `logging.crypto` which is **NOT** present in `membersrvc.yaml`. (It is however present in `core.yaml`, but that config file is not loaded when running the `membersrvc` binary.)

As a result, we get a "Log level not recognized" error.

Would it be easy to have the `membersrvc` binary run on the same logic as the `peer` one when it comes to control its logging modules? (I _believe_ that this is what Bishop is referring to.)

If not, can we at least add a `logging.crypto` key to `membersrvc.yaml` to make the error message go away?

Copying @adecaro who may be able to help on this.

kchristidis:
I think I see the other issue that Bishop noted.

When running the `peer` binary, the `crypto` package does not respect the value set via the `logging-level` flag, and instead logs at the level set by the `logging.crypto` key.

This behavior is inconsistent with the other modules. Any way we could fix this?

bcbrock:
@kchristidis Yes, your second point was my point. The YAML entries were supposed to provide default logging levels for 'peer' subcommands, and the syntax of the entry then allowed individual module level control. After this mechanism was introduced the "crypto" system was subsequently modified to take its logging level directly from the YAML file.

juliancarrivick-ibm:
[...]

PR #2024, opened 2016-06-27
add log-level support for chaincode

### LINKED ISSUE #2023: Add log-level support for chaincode
## Description

Docker daemon has a very high cpu usage when run performance test.
Profile shows that the high cpu usage is caused by verbose logging in chaincode container.
When docker daemon started with option --log-driver="none", the problem gone.
## Describe How to Reproduce

1. Start 4 validator peer nodes on single node.
2. Deploy chaincode_example02, start 20 clients to invoke chaincode_example02
3. Use top to see docker daemon's cpu usage

zuowang:
## Description
1. read in config file from current path
2. config log-level for chaincode 
## Motivation and Context

Fixes #2023
## How Has This Been Tested?

deploy a chaincode, check log-level with command "docker logs chaincode-container-id"
## Checklist:

Signed-off-by: Zuo Wang wangzuo3@wanda.cn

### DIFF
--- a/core/chaincode/shim/chaincode.go
+++ b/core/chaincode/shim/chaincode.go
@@ -97,15 +97,15 @@ func Start(cc Chaincode) error {
 		panic(fmt.Errorf("Fatal error config file: %s \n", err))
 	}
 
-	level, err := LogLevel(viper.GetString("shim.logging.shim"))
+	level, err := LogLevel(viper.GetString("chaincode.logging.shim"))
 	if err == nil {
 		// No error, use the setting
 		SetLoggingLevel(level)
-		chaincodeLogger.Infof("Log level recognized '%s', set to %s", viper.GetString("shim.logging.shim"),
+		chaincodeLogger.Infof("Log level recognized '%s', set to %s", viper.GetString("chaincode.logging.shim"),
 			logging.GetLevel("shim"))
 	} else {
-		SetLoggingLevel(LogInfo)
-		chaincodeLogger.Warningf("Log level not recognized '%s', defaulting to %s: %s", viper.GetString("shim.logging.shim"),
+		SetLoggingLevel(shimLoggingLevel)
+		chaincodeLogger.Warningf("Log level not recognized '%s', defaulting to %s: %s", viper.GetString("chaincode.logging.shim"),
 			logging.GetLevel("shim"), err)
 	}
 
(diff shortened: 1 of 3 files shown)


srderson:
@muralisrini @bcbrock As you have both been active in this area do you have any comments?

bcbrock:
@srderson -  If merged as-is this would be the second time that people have misunderstood or ignored the documentation in the core.yaml file surrounding logging control. The logging.<subcommand> keys are clearly documented as applying to subcommands of the peer command. The first mistake is the presence of the "crypto" key here - see #769. 

Note that the subcommand controls include a syntax for fine-grained control of logging "modules" within the peer program, which is very different from what is being proposed here - namely a simple level control over a single logging object in the shim.

I think if we want to control this from a YAML file then it should at least be done as a completely separate set of keys from logging.<subcommand>, to avoid the types of confusion listed above. Also note that the control suggested here only applies to user-mode chaincodes but not to system chaincodes. Is that an oversight?

As long as we're doing the shim we should also go ahead and have a control that would apply a default to the objects created by the shim.NewLogger() function. Of course the PR also needs to include updates to documentation before being accepted.

zuowang:
@bcbrock could you help to review the code?

bcbrock:
Hi @zuowang you may be sorry you asked - I am hard to please.

This utility only currently only applies to go-language chaincodes. That needs to be made clear in the docs. I think. We also need more than a single incomplete sentence to describe the operation of these parameters.

Also, your key 'shim.logging.chaincode' only applies to system chaincodes. I was suggesting that there should be a way to apply this setting as a default to the logging objects created by shim.NewLogger(). So we probably want 3 settings: shim, system chaincode, user chaincode.

Rather than creating a new top-level hierarchy called 'shim', maybe these keys should be placed under the pre-existing 'chaincode' hierarchy?

Also, if I am not mistaken it is not necessary to reload the Viper file in startInProc() since by definition you are already running in the context of a running peer.

zuowang:
@bcbrock 
It's already in the 'Go chaincodes' section. I made it more clear by adding to example.

I added 3 key "chaincode.logging.shim" "chaincode.logging.scc" "chaincode.logging.ucc" for setting default log level of shim, system chaincode, user chaincode.

no need to reload the Viper file.

bcbrock:
@srderson I believe this PR should be closed. I apologize to @zuowang for not coming to this conclusion  earlier, but some of the issues he had to solve in his latest code leads me to this conclusion.

First, the logging level of the shim for system chaincodes can already be controlled in no less than 4 different ways. If anything is missing, it is only an explicit line of documentation describing how to do this.

Second, I have become a big anti-fan of Viper and YAML files for configuration. The YAML files are part of the source code of the system. We need to allow users change the configuration without changing the source code of the system. Also consider that logging levels are user preferences that change day by day. Why should ephemeral chaincode logging settings be stored in the same configuration file that the system administrator uses to configure critical functions of the server?

At the moment I will argue that it is cleaner for the chaincode writer to explicitly control the chaincode operations through private configurations. For example in `tools/busywork/counters/counters.go` the chaincode accepts "command line parameters" (iinitialization arguments) that include logging levels for the shim and the chaincode. These level default according to a personal preference but can can be changed by anyone without modifying any system source code.

Finally, I believe I've heard @muralisrini lament in the past that the user-chaincode shim should be a lightweight interface, but instead currently brings in the kitchen sink. This shows up in the fact that this PR has to (redundantly) reload the Viper file every time a chaincode allocates a new logging object - the user-chaincode interfaces are not cleanly separated from the system chaincode support.

I think the chaincode interafaces should be refactored, perhaps with a well-thought-out, comprehensive approach to configuring chaincodes. Since there are already other way to accomplish what this PR does, I don't see a need for it at this time.

zuowang:
@bcbrock  The log level of shim is hard coded to be LogDebug. And it caused a very high CPU usage in my performance test. Could you help to point out if there is any way to config it?

```
var shimLoggingLevel = LogDebug // Necessary for correct initialization; See Start()
```

I admit the chaincode interafaces should be refactored. But it's a quick fix to config the log level of the user-chaincode shim.

zuowang:
this part of code requires non-trivial refactor work. close the PR.

bcbrock:
@zuowang The shim uses a logging object that logs as the "shim" _module_. Module-level logging control is documented in the core.yaml file and in docs/dev-setup/logging-control.md. These settings will only apply to the shim used by system chaincodes, however. For example if you want the peer to log at INFO and the system-chaincode shim to log at WARNING:

```
export CORE_LOGGING_LEVEL=info:shim=warning
```

If your own user-chaincode is causing the problem then you can use the shim.SetLoggingLevel API to control it. If it's somebody else's user-chaincode then you could either try to convince them to add an explicit control, copy/modify their code and do it yourself, or modify the line of source code above in your installation. I realize that these are not perfect options. Thank you for your agreement to hold off this PR in lieu of a refactoring if the chainode interface.

Issue #1610, opened 2016-05-26
Modify ledger.GetBlockByNumber to additionally return OK


jyellick:
## Description

The `ledger.GetBlockByNumber` returns `nil,nil` when it is asked to retrieve a block which is not in the DB.  This is not a bug, but seems to cause considerable confusion to consumers.

Most code seems to follow the pattern:

```
block, err := ledger.GetBlockByNumber(n)
if err != nil {
    return err
}
block.doStuff()
```

This causes a nil pointer dereference, and was seen in state transfer which was fixed, but `api.go` currently follows this pattern as well and would conceivably cause a crash. 

After discussions with @srderson he convincingly argued that this is not an error condition, it is an expected condition, so setting error is inappropriate.

Instead, to prevent this misuse, the `GetBlockByNumber` call should return `*pb.Block, bool, error` to indicate whether or not the block was actually retrieved and the callers should wrap and propagate this pattern.

srderson:
This would be a great first issue if someone wants to try it out. Nothing too complex and will give you a chance to explore the code.

corecode:
I don't understand why we need to have 3 values returned?  What are the possible combinations?  Are there really 8 different possible outcomes?

jyellick:
@corecode I think the argument would be that, this is essentially a read from an IO based map, so following the golang idioms, a map returns `(value <type>, exists bool)` while IO returns `(value <type>, err error)`, so, smashing these together we have `(value <type>, exists bool, err error)`.

I agree that there aren't that many possible outcomes, I can only think of three.
1. `nil, false, err`
2. `block, true, nil`
3. `nil, false, nil`

Since the block being nil implies `ok = false`, you can argue that the argument is superfluous, but, based on the usage patterns we've seen, people interpret the current return parameters in a golang idiomatic way, and do not check for a nil result when err is unset.  It's caused nil pointer dereference bugs in two different people's codes, and worse, because the `nil,nil` case is so rare, it is not generally discovered by routine testing.

corecode:
Or we just return an error, because there is no such block (yet)?

jyellick:
Maybe @srderson can be more articulate, as he convinced me, but setting errors should be reserved for conditions which are abnormal.  Simply because a query is made for a block that isn't there doesn't mean there is anything 'wrong' with the system, in fact, it can be a perfectly expected event.

corecode:
Then maybe we should return `ok` only?

jyellick:
[...]

PR #2027, opened 2016-06-27
core/rest: Fix panic when block not found in DB

### LINKED ISSUE #2012: Panic: "invalid memory address or nil pointer dereference" in response to /chain/blocks/<N> API
## Description

This is a panic that appears to be coming from the /chain/blocks/<N> API.  The panic is not fatal, but is being caught by the gocraft/web stuff, which carries on. However it does cause the test to fail (or, I assume this is what's causing it to fail.)

Here is the log file of the failing peer. The panic is buried in the middle of it.

[stderr.txt](https://github.com/hyperledger/fabric/files/332807/stderr.txt)
## Describe How to Reproduce

I am not yet sure if this is reproducible. Here is the context:

```
commit 2aeddd6929d7108bda6a498d1d64e771177c5270
Merge: ef5c8c7 99bf9b3
Author: Sheehan Anderson <sranderson@gmail.com>
Date:   Fri Jun 24 17:41:14 2016 -0400

    Merge pull request #1915 from frankyclu/issue-1844-workaround

    Issue 1844 (multi-account with TCert attributes) workaround
```

This is the soon-to-be "official" form of the busywork `stress2b` test I was trying to run:

```
STRESS2 = ./driver \
        -transactions 10000 \
        -peerBurst 10000 \
        -interlock \
        -keepLog \
        -checkAgreement -dupCheck

STRESS2_CONFIG = \
    @CORE_PBFT_GENERAL_TIMEOUT_NULLREQUEST=3s \
    CORE_PEER_VALIDATOR_CONSENSUS_BUFFERSIZE=100000 

stress2b: 
    @$(STRESS2_CONFIG) \
    $(NETWORK) -batch 4
    @echo "makefile: Sleeping 10 seconds to allow peers to interlock"; sleep 10
    @$(STRESS2)

```

dubek:
## Description

When querying REST API using `/chain/blocks/N`, if N refers to a block which is _lower_ than the current chain height but the block doesn't exist in the DB, the REST API panics.  This PR fixes this.
## Motivation and Context

This deals with a `nil,nil` response from `ledger.GetBlockByNumber`; see #1610 for discussion whether this response should be valid at all.

Fixes #2012
## How Has This Been Tested?

Current REST API unit tests past. Unfortunately, I could find a way to mimic this scenario in unit tests - I tried reducing `deltaHistorySize` to 3 but it only deletes records from the state delta columnfamily and not from the blocks columnfamily. If anyone has an idea how to cause this scenario in the unit-tests, please let me know and I'll add it.
## Checklist:

### DIFF
--- a/core/rest/rest_api.go
+++ b/core/rest/rest_api.go
@@ -670,26 +670,27 @@ func (s *ServerOpenchainREST) GetBlockByNumber(rw web.ResponseWriter, req *web.R
 		// Failure
 		rw.WriteHeader(http.StatusBadRequest)
 		encoder.Encode(restResult{Error: "Block id must be an integer (uint64)."})
-	} else {
-		// Retrieve Block from blockchain
-		block, err := s.server.GetBlockByNumber(context.Background(), &pb.BlockNumber{Number: blockNumber})
-
-		// Check for error
-		if err != nil || block == nil {
-			// Failure
-			switch {
-			case err == ErrNotFound || block == nil:
-				rw.WriteHeader(http.StatusNotFound)
-			default:
-				rw.WriteHeader(http.StatusInternalServerError)
-			}
-			encoder.Encode(restResult{Error: err.Error()})
-		} else {
-			// Success
-			rw.WriteHeader(http.StatusOK)
-			encoder.Encode(block)
-		}
+		return
 	}
+
+	// Retrieve Block from blockchain
+	block, err := s.server.GetBlockByNumber(context.Background(), &pb.BlockNumber{Number: blockNumber})
+
+	if (err == ErrNotFound) || (err == nil && block == nil) {
+		rw.WriteHeader(http.StatusNotFound)
+		encoder.Encode(restResult{Error: ErrNotFound.Error()})
+		return
+	}
+
+	if err != nil {
+		rw.WriteHeader(http.StatusInternalServerError)
+		encoder.Encode(restResult{Error: err.Error()})
+		return
+	}
+
+	// Success
+	rw.WriteHeader(http.StatusOK)
+	encoder.Encode(block)
 }
 
 // GetTransactionByUUID returns a transaction matching the specified UUID
--- a/core/rest/rest_api_test.go
+++ b/core/rest/rest_api_test.go
@@ -239,6 +239,14 @@ func TestServerOpenchainREST_API_GetBlockByNumber(t *testing.T) {
 	if res.Error == "" {
 		t.Errorf("Expected an error when URL doesn't have a number, but got none")
 	}
+
+	// Add a fake block number 9 and try to fetch non-existing block 6
+	ledger.PutRawBlock(&block0, 9)
+	body = performHTTPGet(t, httpServer.URL+"/chain/blocks/6")
+	res = parseRESTResult(t, body)
+	if res.Error == "" {
+		t.Errorf("Expected an error when block doesn't exist, but got none")
+	}
 }
 
 func TestServerOpenchainREST_API_GetTransactionByUUID(t *testing.T) {


srderson:
@dubek For the test, could you call ledger.PutRawBlock directly and add some fake block larger than the current chain height, and then GET `/chain/blocks/N` where N is a value between the actual chain height and the fake block that you added?

dubek:
@srderson Thanks for the advice - it worked (indeed triggers the `nil,nil` response). I added the test case and rebased on top of master.

srderson:
Thanks, LGTM

common/ledger/blockledger/fileledger/factory.go
```diff
@@ -54,8 +54,8 @@ func (f *fileLedgerFactory) GetOrCreate(channelID string) (blockledger.ReadWrite
 }
 
 // Remove removes an existing ledger and its indexes. This operation
-// is blocking.
-func (f *fileLedgerFactory) Remove(channelID string) error {
+// performs an async deletion.
+func (f *fileLedgerFactory) Remove(channelID string, finishRemove func(string, bool)) error {
```
tock-ibm: I think a better approach would be to implement the asynchrony in the Registrar, launching the go routine over there.

common/ledger/blockledger/fileledger/factory.go
```diff
@@ -70,10 +70,15 @@ func (f *fileLedgerFactory) Remove(channelID string) error {
...
-	}
+	go func() {
+		err := f.blkstorageProvider.Drop(channelID)
+		if err != nil {
+			logger.Info(err.Error())
+			finishRemove(channelID, false)
+		} else {
+			finishRemove(channelID, true)
+		}
+	}()
```
tock-ibm: The problem here is that the Drop() will be executed with one routine, and the rest will execute with another.... the channelID will be removed from the ledger, the .remove file removed from the file repo... all this while for a long Drop() the other goroutine is still working. This will break the crash fault tolerance.

common/ledger/blockledger/fileledger/factory.go
```diff
@@ -70,10 +70,15 @@ func (f *fileLedgerFactory) Remove(channelID string) error {
 		return err
 	}
```
tock-ibm: I think the `removeFileRepo.Save()` should execute before the `blockStore.Shutdown()`, no?
I seems to me `removeFileRepo.Save()`  should be the first think we do.

orderer/common/multichannel/registrar.go
```diff
@@ -60,7 +60,8 @@ type Registrar struct {
 	bccsp              bccsp.BCCSP
 	clusterDialer      *cluster.PredicateDialer
 
-	joinBlockFileRepo *filerepo.Repo
+	joinBlockFileRepo, removeFileRepo *filerepo.Repo
```
tock-ibm: I don't think we should expose the `removeFileRepo` here, but rather keep internal to the `fileledger` if we can

orderer/common/multichannel/registrar.go
```diff
@@ -102,41 +103,53 @@ func NewRegistrar(
...
-	r.init(consenters)
+	r.init(consenters, "")
 
 	r.lock.Lock()
 	defer r.lock.Unlock()
 	r.startChannels()
 }
 
-func (r *Registrar) init(consenters map[string]consensus.Consenter) {
+func (r *Registrar) init(consenters map[string]consensus.Consenter, systemChannelID string) {
```
tock-ibm: we don't need that `systemChannelID string`

orderer/common/multichannel/registrar.go
```diff
@@ -188,19 +204,22 @@ func (r *Registrar) discoverLedgers(channelsWithJoinBlock map[string]*cb.Block)
 		if rl.Height() == 0 {
 			if _, ok := channelsWithJoinBlock[channelID]; !ok {
 				logger.Warnf("Channel '%s' has an empty ledger without a join-block, removing it", channelID)
-				if err := r.ledgerFactory.Remove(channelID); err != nil {
+				if err := r.ledgerFactory.Remove(channelID, r.removeChannelFromPendingRemovals); err != nil {
```
tock-ibm: This for example should be done synchronously, as it is a small ledger, and we don't want to start before it is deleted

common/ledger/blockledger/fileledger/factory.go
```diff
@@ -115,24 +120,10 @@ func New(directory string, metricsProvider metrics.Provider) (blockledger.Factor
...
-		for _, fileName := range files {
-			channelID := factory.removeFileRepo.FileToBaseName(fileName)
-			err = factory.Remove(channelID)
-			if err != nil {
-				logger.Errorf("Failed to remove channel %s: %s", channelID, err.Error())
-				return nil, err
-			}
-			logger.Infof("Removed channel: %s", channelID)
-		}
-	}
```
tock-ibm: why was this removes / moved? 
In addition, because the filerepo .remove file is deleted before the Drop() completed, we could be missing on the retry of removals that crashed in the middle of a Drop...

orderer/common/multichannel/registrar.go
```diff
@@ -61,6 +61,7 @@ type Registrar struct {
 	clusterDialer      *cluster.PredicateDialer
 
 	joinBlockFileRepo *filerepo.Repo
+	pendingRemoval    map[string]bool
```
tock-ibm: A nit: I would place the `pendingRemoval` in the block of fields below the `lock` right below `follower`, to send the message that it is a fields that  should be accessed with the lock. (In contrast to all the rest that are initialized upon creation and stay static throughout).

orderer/common/multichannel/registrar.go
```diff
@@ -102,6 +103,7 @@ func NewRegistrar(
 		callbacks:          callbacks,
 		bccsp:              bccsp,
 		clusterDialer:      clusterDialer,
+		pendingRemoval:     make(map[string]bool),
```
tock-ibm: As above, place this field below `followers`.

orderer/common/multichannel/registrar.go
```diff
@@ -696,6 +698,11 @@ func (r *Registrar) JoinChannel(channelID string, configBlock *cb.Block, isAppCh
 	r.lock.RLock()
 	defer r.lock.RUnlock()
 
+	removalStatus, ok := r.pendingRemoval[channelID]
+	if removalStatus && ok {
+		return types.ChannelInfo{}, types.ErrChannelNotExist
+	}
+
```
tock-ibm: Two things:
1. I think this is the wrong error to return. As long as the channel is pending removal, it "exists". So maybe define a new type:
```
var ErrChannelPendingRemoval = errors.New("channel pending removal")
```
If you do this, check how this new error type affects the REST API handler behavior. 
The minimum I would do is to use `ErrChannelExists`.

2. `pendingRemoval` is `map[string]bool`, and the cool thing about it is that when the value is missing, false is returned, so you can do this:
```
if pendingRemoval[channelID] {
	return types.ChannelInfo{}, types.ErrChannelPendingRemoval 
}
```

orderer/common/multichannel/registrar.go
tock-ibm: I think that the "List" and "ListAll" should return channels that are pending removal, with:
ClusterRelation = "none"
Status = "pending-remove" (or maybe just reuse "inactive"?)
This way an operator would be able to monitor when a channel was finally removed.

orderer/common/multichannel/registrar.go
```diff
@@ -1044,3 +1041,20 @@ func (r *Registrar) removeSystemChannel() error {
 
 	return nil
 }
+
+func (r *Registrar) removeLedgerAsync(channelID string) {
+	removalStatus, ok := r.pendingRemoval[channelID]
+	if removalStatus && ok {
+		return
+	}
```
tock-ibm: I think this is redundant, because we only get here if the channel is in either in the followers or channels map, but never in the pendingRemoval map... If anything, this signals an error condition, no?

orderer/common/multichannel/registrar.go
```diff
@@ -1044,3 +1041,20 @@ func (r *Registrar) removeSystemChannel() error {
...
+
+	r.pendingRemoval[channelID] = true
+	go func() {
+		if err := r.ledgerFactory.Remove(channelID); err != nil {
+			r.pendingRemoval[channelID] = false
+			logger.Errorf("Ledger factory failed to remove empty ledger '%s', error: %s", channelID, err)
+			return
+		}
+		delete(r.pendingRemoval, channelID)
```
tock-ibm, line 1098: Must take the lock to delete from the r.pendingRemoval map.

orderer/common/multichannel/registrar_test.go
```diff
@@ -1543,6 +1565,60 @@ func TestRegistrar_RemoveChannel(t *testing.T) {
...
+		require.NotContains(t, ledgerFactory.ChannelIDs(), "my-raft-channel")
+		info, err := registrar.JoinChannel("my-raft-channel", genesisBlockAppRaft, true)
+		require.NoError(t, err)
+		require.Equal(t, types.ChannelInfo{Name: "my-raft-channel", URL: "", ClusterRelation: "member", Status: "active", Height: 1}, info)
+		require.NotNil(t, registrar.GetChain("my-raft-channel"))
+		require.Contains(t, ledgerFactory.ChannelIDs(), "my-raft-channel")
+
+		registrar.pendingRemoval["my-raft-channel"] = true
+		err = registrar.RemoveChannel("my-raft-channel")
+		require.NoError(t, err)
```
tock-ibm: I would actually have this return `ErrChannelPendingRemove`

orderer/common/multichannel/registrar_test.go
```diff
@@ -1543,6 +1565,60 @@ func TestRegistrar_RemoveChannel(t *testing.T) {
...
+
+		// After removing the channel, it no longer exists in the registrar or the ledger
+		require.Nil(t, registrar.GetChain("my-raft-channel"))
+		require.Equal(t, len(ledgerFactory.ChannelIDs()), 1)
+		require.Contains(t, ledgerFactory.ChannelIDs(), "my-raft-channel")
```
tock-ibm: You say in the comment it is not in the ledger, but verify that it is?

orderer/common/multichannel/registrar.go
```diff
@@ -908,17 +916,12 @@ func (r *Registrar) RemoveChannel(channelID string) error {
 	return types.ErrChannelNotExist
 }
```
tock-ibm, line 948: I think we need to check if it is pendingRemoval, and if so, return another error: 
`var ErrChannelPendingRemoval = errors.New("channel pending removal")`

orderer/common/multichannel/registrar.go
```diff
@@ -48,6 +48,7 @@ type Registrar struct {
 	lock            sync.RWMutex
 	chains          map[string]*ChainSupport
 	followers       map[string]*follower.Chain
+	pendingRemoval  sync.Map
```
tock-ibm: The use of sync.Map is not needed here. The map is mostly used within the `lock`.
All you have to do is when the goroutine you assign to remove finishes, delete the key from the map under a lock. Referring to your earlier commit:

```
func (r *Registrar) removeLedgerAsync(channelID string) {
	removalStatus, ok := r.pendingRemoval[channelID]
	if removalStatus && ok {
		return
	}

	r.pendingRemoval[channelID] = true
	go func() {
		if err := r.ledgerFactory.Remove(channelID); err != nil {
			r.pendingRemoval[channelID] = false
			logger.Errorf("Ledger factory failed to remove empty ledger '%s', error: %s", channelID, err)
			return
		}
                r.lock.Lock() \\<<<
		delete(r.pendingRemoval, channelID)
                r.lock.Unlock() \\<<<
	}()
}
```
So you can use map[string]bool as you did before.

See [here](https://golang.org/pkg/sync/#Map) and [here](https://medium.com/@deckarep/the-new-kid-in-town-gos-sync-map-de24a6bf7c2c), that explain that the `sync.Map` is better avoided unless you hit a very specific use-case.

orderer/common/multichannel/registrar.go
```diff
@@ -696,6 +698,11 @@ func (r *Registrar) JoinChannel(channelID string, configBlock *cb.Block, isAppCh
 	r.lock.RLock()
 	defer r.lock.RUnlock()
 
+	removalStatus, ok := r.pendingRemoval.Load(channelID)
```
tock-ibm: map[string]bool will do fine, as explained above. Here it is protected by `r.lock`.

orderer/common/multichannel/registrar.go
```diff
@@ -886,6 +893,12 @@ func (r *Registrar) joinSystemChannel(
 func (r *Registrar) RemoveChannel(channelID string) error {
 	r.lock.Lock()
 	defer r.lock.Unlock()
+
+	removalStatus, ok := r.pendingRemoval.Load(channelID)
+	if ok && removalStatus.(bool) {
```
tock-ibm: map[string]bool will do fine, as explained above. Here it is protected by `r.lock`.

orderer/common/multichannel/registrar.go
```diff
@@ -1002,10 +1008,26 @@ func (r *Registrar) removeSystemChannel() error {
 	r.systemChannel.Halt()
 	delete(r.chains, systemChannelID)
 
+	pendingRemovals := []string{}
+
 	// halt all application channels
 	for channel, cs := range r.chains {
 		cs.Halt()
+		pendingRemovals = append(pendingRemovals, channel)
+	}
```
tock-ibm: This is just a list of all the channels.... I think it is simpler to append to this list only if the channel does not include the orderer in the consenters set... that is, move the part that computes the `isChannelMember` in the second loop here, and then the second loop will be simpler, as it would iterate on channels that actually need to be removed.

orderer/common/channelparticipation/restapi.go
```diff
@@ -289,6 +290,8 @@ func (h *HTTPHandler) serveRemove(resp http.ResponseWriter, req *http.Request) {
 		h.sendResponseNotAllowed(resp, errors.WithMessage(err, "cannot remove"), http.MethodGet)
 	case types.ErrChannelNotExist:
 		h.sendResponseJsonError(resp, http.StatusNotFound, errors.WithMessage(err, "cannot remove"))
+	case types.ErrChannelPendingRemoval, types.ErrChannelRemoval:
+		h.sendResponseJsonError(resp, http.StatusConflict, errors.WithMessage(err, "cannot remove"))
```
tock-ibm: `types.ErrChannelRemoval` - I suggest `types.ErrChannelRemovalFailure`, to be more explicit.
In addition, I think it should return the error code "InternalServerError" 500. What do you think?

orderer/common/multichannel/registrar.go
```diff
@@ -1014,31 +1057,44 @@ func (r *Registrar) removeSystemChannel() error {
...
+			delete(r.chains, channel)
 		}
 	}
 
-	// remove system channel resources
-	err := r.ledgerFactory.Remove(systemChannelID)
-	if err != nil {
-		return errors.WithMessagef(err, "error removing ledger for system channel %s", r.systemChannelID)
+	if len(failedRemovals) > 0 {
+		return fmt.Errorf("failed removing %s ledger for channel(s)", strings.Join(failedRemovals, ", "))
```
tock-ibm: I suggest:
`fmt.Errorf("failed removing ledger for channel(s): %s", strings.Join(failedRemovals, ", "))

orderer/common/multichannel/registrar.go
```diff
@@ -1014,31 +1057,44 @@ func (r *Registrar) removeSystemChannel() error {
...
 	return nil
 }
+
+func (r *Registrar) removeLedgerAsync(channelID string) {
+	r.pendingRemoval[channelID] = true
+
+	go func() {
+		if err := r.ledgerFactory.Remove(channelID); err != nil {
+			r.lock.Lock()
+			r.pendingRemoval[channelID] = false
```
tock-ibm, line 1092: I like this idea. Please put a comment on the map that clarifies what the boolean means, i.e. "True=pending, False=removal-failed", etc.

orderer/common/channelparticipation/restapi.go
```diff
@@ -254,8 +254,9 @@ func (h *HTTPHandler) sendJoinError(err error, resp http.ResponseWriter) {
 		// The client is trying to join an app-channel that exists, but the system channel does not;
 		// The client is trying to join the system-channel, and it exists. GET & DELETE are allowed on the channel.
 		h.sendResponseNotAllowed(resp, errors.WithMessage(err, "cannot join"), http.MethodGet, http.MethodDelete)
-	case types.ErrAppChannelsAlreadyExists:
+	case types.ErrAppChannelsAlreadyExists, types.ErrChannelPendingRemoval:
```
tock-ibm: I think there should be a symmetry between the "Join" & "Remove" w.r.t. `types.ErrChannelPendingRemoval` - i.e. both should respond with "Conflict" to reflect the transitory nature of the error.

In addition, "Join" does not handle `types.ErrChanneRemovalFailure` as the "Remove" below, and it would fall through to the "Bad Request" which is not what we want. I think this should return "500 Internal Server Error", since this  is the server's fault, not a state or a bad request, something that the user can do anything about. What do you think?

orderer/common/multichannel/registrar_test.go
```diff
@@ -1543,6 +1571,119 @@ func TestRegistrar_RemoveChannel(t *testing.T) {
...
+
+		require.Nil(t, registrar.GetChain("my-raft-channel"))
+		require.NotContains(t, ledgerFactory.ChannelIDs(), "my-raft-channel")
+		info, err := registrar.JoinChannel("my-raft-channel", genesisBlockAppRaft, true)
+		require.NoError(t, err)
+		require.Equal(t, types.ChannelInfo{Name: "my-raft-channel", URL: "", ClusterRelation: "member", Status: "active", Height: 1}, info)
+		require.NotNil(t, registrar.GetChain("my-raft-channel"))
+		require.Contains(t, ledgerFactory.ChannelIDs(), "my-raft-channel")
+
+		// This will force the failure of the underlying factory.Remove() because the pendingops/remove will not exist
```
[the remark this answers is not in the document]
tock-ibm: I like this idea. This goes towards what we want to achieve in this issue [FAB-18106](https://jira.hyperledger.org/browse/FAB-18106). The intention in that task is that a channel that fails catastrophically will stop and report a new status `types.StatusFailure` instead of panicking and crashing the server. 

If we had that error code already, I would be inclined to suggest that the map would include a struct with
```
{
  cRel     types.ClusterRelation
  status  types.Status
}
```
you could then insert into the map {"member", "inactive"} or {"follower", "inactive"} when you start the removal, delete that when removal ends successfully, and put  {"member", "failure"} or {"follower", "failure"} if something goes wrong.
What do you think? This can be done  in this task or open another Jira for it - up to you.

orderer/common/types/errors.go
```diff
@@ -21,3 +21,9 @@ var ErrAppChannelsAlreadyExists = errors.New("application channels already exist
 
 // ErrChannelNotExist is returned when trying to remove or list a channel that does not exist
 var ErrChannelNotExist = errors.New("channel does not exist")
+
+// ErrChannelPendingRemoval is returned when trying to remove or list a channel that is being removed.
+var ErrChannelPendingRemoval = errors.New("channel pending removal")
+
+// ErrChannelRemoval is returned when a removal attempt failure has been recorded.
+var ErrChannelRemoval = errors.New("channel removal failure")
```
tock-ibm: I think 
`var ErrChannelRemovalFailure = errors.New("channel removal failure")`
is clearer

PR #2029, opened 2016-06-27
Discovery: maintain and persist list of recent connections

kchristidis:
## Description

With this set of changes, every node should now maintain a list of recent connections via the `discovery` service. When the appropriate flag (`peer.discovery.persist`) in the config file is set to `true` (default), the node also persists the discovery list to the DB, and repeats this task whenever the list is updated. If the node crashes, it will load the list upon restarting, and attempt to connect to the addresses it carries, via the touch service.

This PR extends the existing `discovery` service by including non-anchor nodes in its list, and by adding persistence support. It moves the `discovery` package in the `core` directory as this seemed a much more logical fit.

This set of changes also simplifies the peer connection logic in the `peer` package, and adds support for the `discovery.maxTouchNodes` configuration option originally added by @jeffgarratt; now the touch service will attempt to connect to a maximum of `maxTouchNodes` nodes. (This new list is generated randomly from the discovery list.)
## Motivation and Context

Consider this scenario:
1. All the nodes in a network use the same anchor node.
2. The anchor node crashes.
3. A node crashes and comes back online while the anchor node is still down.

This node will not be able to connect to its peers, i.e. we will have a network partition. Do note that the practice listed in step 1 is how we actually launch all of our test networks today. This problem was discovered in #1874.

With this PR, such issues are fixed.
## How Has This Been Tested?
- New unit tests for the new functions added in the `discovery` and `util` package.
-  2 new behave tests have been added, provided by @lhaskins and @scottz64.
- Passes all tests locally.
## Checklist:

Signed-off-by: Kostas Christidis kchristidis@us.ibm.com.


jyellick:
This looks like a rebase of #2016 that already went into 0.5, and still looks good to me.

srderson:
Thanks, LGTM

PR #2054, opened 2016-06-28
Remove all command-line flags that shadow keys in core.yaml

### LINKED ISSUE #1962: Remove all redundant command-line flags from the Hyperledger Fabric Peer
## Description

All command-line flags that overlap variables in the YAML files should be removed from the peer. Why? Because both the command line flags and the YAML files define defaults for the options. Unless these options are identical, one or the other will be selected, and in fact, the YAML file does not override the command-line defaults in main.go. This is a maintenance nightmare and a severe usability bug. (I have just lived through this nightmare with the fact that GOMAXPROCS is now silently being set to 2 based on the command-line default, which now overrides the -1 default in the YAML file.)

All of the redundant command line flags should be removed until we implement a better way of handling configuration options, e.g. as discussed in #1309

bcbrock:
## Description

Removed all command-line flags from main.go that shadow keys in core.yaml. I also fixed an incorrect usage of StringVarP. I've also grep-ed the docs and I can't find these used anywhere.
## Motivation and Context

Fixes #1962
## How Has This Been Tested?

Passes tests in Vagrant environment
## Checklist:

### DIFF
--- a/peer/main.go
+++ b/peer/main.go
@@ -242,18 +242,7 @@ func main() {
 
 	// Set the flags on the node start command.
 	flags := nodeStartCmd.Flags()
-	flags.Bool("peer-tls-enabled", false, "Connection uses TLS if true, else plain TCP")
-	flags.String("peer-tls-cert-file", "testdata/server1.pem", "TLS cert file")
-	flags.String("peer-tls-key-file", "testdata/server1.key", "TLS key file")
-	flags.Int("peer-gomaxprocs", 2, "The maximum number threads excuting peer code")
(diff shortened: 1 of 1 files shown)


srderson:
Did a quick doc search also and saw nothing to update. LGTM. Thanks

lehors:
It seems that the title of this PR is misleading in that there is at least still one command line option left -logging-level which shadows the configuration file.

lehors:
Note that I'm not arguing for the removal of -logging-level, I'm only arguing for a more careful description of the change.

PR #2062, opened 2016-06-28
Fixed issue #1824

### LINKED ISSUE #1824: When peer fails to get the certificate, it can never start or restart
## Description

The function at Line https://github.com/hyperledger/fabric/blob/fcc6a7dc599f20d632fdb87beb785f73fa8796bd/core/crypto/node_crypto.go#L49 writes a file at https://github.com/hyperledger/fabric/blob/fcc6a7dc599f20d632fdb87beb785f73fa8796bd/core/crypto/node_eca.go#L90

The line https://github.com/hyperledger/fabric/blob/fcc6a7dc599f20d632fdb87beb785f73fa8796bd/core/crypto/node_impl.go#L95 checks this file and considers the node is registered and assumes the peer has the TLS certificates.

If the function at line https://github.com/hyperledger/fabric/blob/fcc6a7dc599f20d632fdb87beb785f73fa8796bd/core/crypto/node_crypto.go#L55 fails, restarting the peer will always fail as the peer doesn't have the TLS certificate but it thinks it is registered because it has the enrollment file.

The peer will never start (or restart) successfully.
## Describe How to Reproduce

How to reproduce it:
1. Start a CA (membership services) and a peer.
2. Delete the file `/var/hyperledger/production/crypto/validator/peer1/ks/raw/tls.cert` on the peer.
3. Restart the peer.

jeroiraz:
Registration flow wasn't taking into account failures while retrieving certificates from CAs. As a result, whenever a peer failed to get the TLS certificate it ended up in a bad state (not completely registered and not able to be initialized).
## Description

As part of this PR, a refactoring of the hierarchy of node, peer, client and validator nodes was performed, in order to have validation logic at the highest level of abstraction (node).
To prevent nodes from ending up in a bad state due to a failure, all the functions in charge of retrieving certificates were made idempotent by taking into account whether the corresponding certificate was already retrieved and stored.
## Motivation and Context

Fixes #1824
## How Has This Been Tested?

This PR does not require additional test cases as no new functionality was introduced.
- Unit tests have been run without errors.
- Behave tests have been run without errors.
- Node.js tests have been run without errors (compared with master branch)
## Checklist:

Signed-off-by: Jeronimo Irazabal jeronimo.irazabal@gmail.com / iraza@ar.ibm.com


core/crypto/validator_impl.go
```diff
@@ -176,10 +163,10 @@ func (validator *validatorImpl) initCryptoEngine() (err error) {
 		nil, validator.enrollChainKey.(*ecdsa.PrivateKey),
 	)
 	if err != nil {
-		return
+		return err
```
adecaro: It is not necessary to specify err here given that the return variable has been named.

core/crypto/validator_impl.go
```diff
@@ -176,10 +163,10 @@ func (validator *validatorImpl) initCryptoEngine() (err error) {
...
 	}
 
-	return
+	return nil
```
adecaro: There is no need to return nil explicitly here.

adecaro:
@jeronimoirazabal May you add tests in order to verify that #1824 is gone?

BTW, wasn't it simpler to add a cleanup of the key store upon error instead of delegating to the node the initialization of the subclasses? What if there are multiple levels of initialization?

christo4ferris:
+1 needs tests to confirm that the issue is resolved.

jeroiraz:
@adecaro Thanks for your comments. I've implemented a template pattern at the highest level (node) where the mandatory registration and initialization tasks and validations are performed. Thus only specific actions are implemented at lower levels of the hierarchy (peer, validator, etc).

@adecaro @christo4ferris Regarding the addition of test cases, I'm totally agree with that. In order to verify that referenced issue was solved I manually followed the steps to reproduce it. The reason I've not added test cases was to avoid white-box testing as I haven't found mocked CAs. Shall I add a test case that performs the removal of the files from the filesystem? Is there some mocked CA or do you think it would work doing that (maybe in the scope of another PR)?

jeroiraz:
@adecaro @christo4ferris Could you please take a look at last changes? I've added test cases to verify referenced issue was solved

srderson:
@adecaro @christo4ferris Please check the update when you have a chance. Thanks

core/crypto/crypto_test.go
```diff
@@ -264,6 +264,55 @@ func TestRegistrationSameEnrollIDDifferentRole(t *testing.T) {
 	}
 }
 
+func TestTLSCertificateDeletion(t *testing.T) {
+	conf := utils.NodeConfiguration{Type: "peer", Name: "peer"}
+
+	peer, err := registerAndReturnPeer(conf.Name, nil, conf.GetEnrollmentID(), conf.GetEnrollmentPWD())
```
christo4ferris, line 270: shouldn't we assert that peer is not nil?

core/crypto/crypto_test.go
```diff
@@ -264,6 +264,55 @@ func TestRegistrationSameEnrollIDDifferentRole(t *testing.T) {
...
+
+	if !peer.ks.certMissing(peer.conf.getTLSCertFilename()) {
+		t.Fatal("TLS certificate should be missing after deletion")
+	}
+}
+
+func TestRegistrationAfterDeletingTLSCertificate(t *testing.T) {
+	conf := utils.NodeConfiguration{Type: "peer", Name: "peer"}
+
+	peer, err := registerAndReturnPeer(conf.Name, nil, conf.GetEnrollmentID(), conf.GetEnrollmentPWD())
```
christo4ferris, line 291: same comment

christo4ferris:
Just left a couple of comments.

core/crypto/crypto_test.go
jeroiraz, line 270: @christo4ferris 
As registerAndReturnPeer is a helper function being written inside the test file and not part of the functionality I think it wouldn't be needed. Actually this helper could be removed if register (peer/client/validator) function just returned the registered peer, that change could be accomplished in a new PR as some refactoring would be needed. Does it make sense?

jeroiraz:
@christo4ferris @srderson 
Please let me know if there is something else pending from my side

core/crypto/node_ks.go
```diff
@@ -149,6 +149,10 @@ func (ks *keyStore) storePrivateKeyInClear(alias string, privateKey interface{})
 	return nil
 }
 
+func (ks *keyStore) deletePrivateKeyInClear(alias string) error {
+	return os.Remove(ks.node.conf.getPathForAlias(alias))
```
srderson, line 153: Is the `alias` in any way user input? Is there any need to be concerned about directory traversal attacks?

srderson:
Added a comment with a question as I'm not that familiar with this code. Otherwise LGTM. Would prefer to have a LGTM comment from @adecaro or @JonathanLevi also if possible.

core/crypto/node_ks.go
jeroiraz, line 153: thanks @srderson
The user is not able to alter those values as their inputs are statically specified in yaml files (e.g. TLSCertFilename and TLSKeyFilename). Besides that, I think your point is worth to be addressed in a PR to cover the entire project.

JonathanLevi:
I agree with both of you (@srderson & @jeronimoirazabal): even if don't end up with the full AccessControlContext, AccessControlException kind of setups... a separate, project/fabric-level PR to prevent some of such basic attacks, is needed.

@srderson: LGTM, thank you.

srderson:
Thanks @jeronimoirazabal and @JonathanLevi

PR #2069, opened 2016-06-29
core/rest: Proper JSON encoding in responses to /registrar/* endpoints

### LINKED ISSUE #1475: REST API can be tricked to respond with malformed JSON

dubek:
## Description

Use proper JSON encoding for crafting responses of the following REST API endpoints:
- `POST /registrar`
- `DELETE /registrar/:id`
- `GET /registrar/:id/ecert`
- `GET /registrar/:id/tcert`

This finishes fixing all the JSON responses of the REST API except the deprecated endpoints. 

Future notes/questions:
1. The tcert and ecert APIs return the PEM file in URL-encoding (so the newlines are replaced by `%0A`). Now that we have proper JSON encoding this extra encoding is no longer needed. However, removing it will break existing REST clients which try to URL-decode the cert strings in the response. Do we have a policy regarding such changes?
2. I suggest moving the deprecated endpoints code to a separate go file (in the same package) - maybe in another PR?
## Motivation and Context

Using only proper JSON-encoding functions for generating JSON responses will reduce the risk of returning a malformed JSON when an error message (or OK reponse) contains special JSON chars such as quotes or newlines.

Fixes #1475
## How Has This Been Tested?

Added some unit tests for the given endpoints.

I haven't found an easy way to test the `/registrar/:id/{ecert,tcert}` endpoints, because the handlers of those endpoints try to contact a members service process (which is not up during unit-tests). Unlike the other parts of the Devops interface which I was able to mock, these I believe are more hard-coded.
## Checklist:

### DIFF
--- a/core/rest/api_test.go
+++ b/core/rest/api_test.go
@@ -19,10 +19,9 @@ package rest
 import (
 	"bytes"
 	"fmt"
+	"google/protobuf"
 	"os"
 	"testing"
-
-	"google/protobuf"
 
 	"github.com/hyperledger/fabric/core/ledger"
 	"github.com/hyperledger/fabric/core/util"
@@ -37,12 +36,10 @@ func TestMain(m *testing.M) {
 }
 
 func setupTestConfig() {
-	viper.SetConfigName("core")         // name of config file (without extension)
-	viper.AddConfigPath("./")           // path to look for the config file in
-	viper.AddConfigPath("./..")         // path to look for the config file in
-	viper.AddConfigPath("./../../peer") // path to look for the config file in
-	err := viper.ReadInConfig()         // Find and read the config file
-	if err != nil {                     // Handle errors reading the config file
+	viper.SetConfigName("rest_test") // name of config file (without extension)
+	viper.AddConfigPath(".")         // path to look for the config file in
+	err := viper.ReadInConfig()      // Find and read the config file
+	if err != nil {                  // Handle errors reading the config file
 		panic(fmt.Errorf("Fatal error config file: %s \n", err))
 	}
 }
--- a/core/rest/rest_api.go
+++ b/core/rest/rest_api.go
@@ -68,6 +68,12 @@ type restResult struct {
 	Error string `json:",omitempty"`
 }
 
+// tcertsResult defines the response payload for the GetTransactionCert REST
+// interface request.
+type tcertsResult struct {
+	OK []string
+}
+
 // rpcRequest defines the JSON RPC 2.0 request payload for the /chaincode endpoint.
 type rpcRequest struct {
 	Jsonrpc *string           `json:"jsonrpc,omitempty"`
@@ -505,6 +511,8 @@ func (s *ServerOpenchainREST) GetTransactionCert(rw web.ResponseWriter, req *web
 
 	restLogger.Debugf("REST received transaction certificate retrieval request for registrationID '%s'", enrollmentID)
 
+	encoder := json.NewEncoder(rw)
+
 	// Parse out the count query parameter
 	req.ParseForm()
 	queryParams := req.Form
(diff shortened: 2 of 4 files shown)


srderson:
1. We do not currently. In https://github.com/hyperledger/fabric/wiki/Fabric-Next there is a comment that says "Depreciate REST from peer" in section 2.6, SDK. Here's one possible course of action.
   a. REST will be removed from the peer and peer will support gRPC only.
   b. REST will be moved into the SDK or a standalone process as we have a lot of existing Go REST code. All REST communication will be via gRPC to the server. This is already the case today for some REST APIs.
   c. REST API standalone process can be upgraded and is only impacted if gRPC API change on peer.
   I've also heard people suggest that it should remain in the peer. @dubek would like to get your thoughts.
2. +1 on moving deprecated endpoints to a separate Go file.

This change LGTM

dubek:
Re: "Depreciate REST from peer":

Even if we decide to remove the REST API from the peer main executable process, I still suggest keeping a small HTTP read-only API for Ops needs. So maybe a /stats handler that returns some stats like: chain height, last block addition timestamp, number of invokes processed, number of queries processed, etc. Consider something similar to [Redis INFO command](http://redis.io/commands/INFO) or [MySQL SHOW STATUS command](http://dev.mysql.com/doc/refman/5.7/en/show-status.html). The ability to get all this info with a simple `curl` is heaven for Ops teams (though need to consider if this should be behind authentication). Also see [Go's expvar package](https://golang.org/pkg/expvar/) for similar usage.

From the side right now it looks as though new endpoints/methods are added to the gRPC interface but most of them don't make it to the HTTP REST API. It seems as though most uses are via gRPC (otherwise no one could have used those new methods). If it's a decision that the REST API is not needed (for the core functions like deploy/invoke/query), then just remove it. Adding it again in Node or as a separate Go process won't solve the problem that it is "lagging".

I'm not sure there's an admin/dashboard app on your roadmap, but you'll probably need that at some point. I have some experience with RabbitMQ, which by default talks only AMQP. It is shipped with a [management plugin](https://www.rabbitmq.com/management.html) which is turned off by default (not sure why - maybe security or performance reasons). Once activated, it exposes an HTTP interface, both via REST-JSON API and via a web-based console (which is a single-page app that AJAXes to that REST-JSON API). For high-performance queueing operations you want to use the AMQP interface (you'll need a proper AMQP client/library); but for management, inspection, etc - you can use the web or curl.

An equivalent approach for Fabric can be something like (similar to your idea above):
1. peer exposes only gRPC interface
2. Fabric ships with a Node.js-based mgmt app (or Go-based) which talks gRPC with peer. (Should run on the same machine?)
3. The mgmt app exposes whatever API that is need for management operations and the web console. It doesn't have to mirror the gRPC interface functionality or semantics.

Then the team developing the mgmt app can develop independently and don't have to mirror the peer's interfaces.

As for the `/stats` endpoint I mentioned at the beginning, maybe expose it in gRPC and then also in the mgmt app / NodeSDK / CLI - for easier handling by Ops.

---

> +1 on moving deprecated endpoints to a separate Go file.

I'd rather not shake these files if there's a chance you remove them all in a few weeks. Let me know.

PR #2076, opened 2016-06-29
Fixes Issue 2073: Logging controls in core.yaml referring to non-existent peer commands

### LINKED ISSUE #2073: Logging control in core.yaml refers to no-longer-existent peer subcommands
## Description

I thought this had been taken care of when the peer subcommands were renamed; I guess not.
I will prepare a fix for this.

bcbrock:
## Description

There are now only 3 peer commands: node, network and chaincode. The logging.<command> keys in core.yaml now reflect this correctly. I've also added notes that describe how to override them. If there is complexity to this, my argument is that the complexity comes from the fact that a single executable functions both as a server and a client, and servers and clients naturally have different default logging requirements.

After some discussions with @lehors it was thought to be clearer not to document the developer-level logging control in the core.yaml file, so this is now only discussed in docs/dev-setup/logging-control.md.

I've set the default for the `node` command to 'info', which I believe is appropriate for a server. If developers want to run with debug logging all the time they can easily set CORE_LOGGING_LEVEL=debug in their .profile or .bashrc files.
## Motivation and Context

Fixes #2073
## How Has This Been Tested?

I've tested the changes and override ordering interactively at the command line. No Go code is changed by this fix. The unit and behave tests also work.
## Checklist:

### DIFF
--- a/docs/dev-setup/logging-control.md
+++ b/docs/dev-setup/logging-control.md
@@ -16,7 +16,13 @@ In pretty-printed logs the logging level is indicated both by color and by a 4-c
     16:47:09.635 [rest] StartOpenchainRESTServer -> INFO 035 Initializing the REST service...
     16:47:09.635 [main] serve -> INFO 036 Starting peer with id=name:"vp1" , network id=dev, address=9.3.158.178:30303, discovery.rootnode=, validator=true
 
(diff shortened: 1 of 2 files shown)


srderson:
@bcbrock LGTM. Thanks

Issue #2077, opened 2016-06-30
PBFT Batch: Infinite loop? Ignoring SyncStateSnapshot message....


bcbrock:
## Description

This is another "loop" that occurs in stressful busywork tests. I do not see this unless I run with multiple chaincodes and multiple data arrays in the chaincodes. 3 of the 4 peers have finished, but one peer is running non-stop issuing message streams like the below. I've sampled this stream at a point where it seems to give up on one thing it is trying and moving on to something else. It's been spinning like this for 20 minutes, generating gigabytes of INFO-level logs. The other peers are silent in their logs, even though they are also running about 30% of an X86 thread each.

```
[33m18:52:40.043 [peer] beforeSyncStateSnapshot -> WARN c56d5c[0m Ignoring SyncStateSnapshot message with correlationId = 9016, sequence = 1014, as current correlationId = 9017
[33m18:52:40.043 [peer] beforeSyncStateSnapshot -> WARN c56d5d[0m Ignoring SyncStateSnapshot message with correlationId = 9016, sequence = 1015, as current correlationId = 9017
[33m18:52:40.043 [peer] beforeSyncStateSnapshot -> WARN c56d5e[0m Ignoring SyncStateSnapshot message with correlationId = 9016, sequence = 1016, as current correlationId = 9017
[33m18:52:40.043 [peer] beforeSyncStateSnapshot -> WARN c56d5f[0m Ignoring SyncStateSnapshot message with correlationId = 9016, sequence = 1017, as current correlationId = 9017
[33m18:52:40.043 [peer] beforeSyncStateSnapshot -> WARN c56d60[0m Ignoring SyncStateSnapshot message with correlationId = 9016, sequence = 1018, as current correlationId = 9017
[33m18:52:40.043 [peer] beforeSyncStateSnapshot -> WARN c56d61[0m Ignoring SyncStateSnapshot message with correlationId = 9016, sequence = 1019, as current correlationId = 9017
[33m18:52:40.043 [peer] beforeSyncStateSnapshot -> WARN c56d62[0m Ignoring SyncStateSnapshot message with correlationId = 9016, sequence = 1020, as current correlationId = 9017
[33m18:52:40.043 [peer] beforeSyncStateSnapshot -> WARN c56d63[0m Ignoring SyncStateSnapshot message with correlationId = 9016, sequence = 1021, as current correlationId = 9017
[...]

PR #2078, opened 2016-06-30
busywork: Remove all non-determinism + New Stress targets

### LINKED ISSUE #1857: State Transfer Errors Running a **busywork** target with multiple clients

bcbrock:
## Description

This change set removes all\* vesitiges of "non-determinism" from the
`busywork/counters` chaincode. When @jyellick fixed a non-determinism bug here
a few days ago, I knew there were still other instances - but thought they
would be OK. I now realize that there are cases where peers never execute even
a single transaction, even when they are not under stress. So I'm throwing in
the towel, and I have effectively rewritten this chaincode, updated the
documentation and added a test program for the new verison. To be completely
honest, the data model manipulated by this chaincode is so simple that my
"non-determinism" for error checking never caught a single bug.

The PBFT Batch protocol is working well-enough now that the **busywork**
`stress2b` test can run in its originally intended form. I've restructured the
Makefile to show this, and I've also added a new `sweep1b` test that also
works. The bad news is that the more complex `sweep2b` test fails (see #2077).

I've also included a couple of minor bug fixes and enhancements to the
`busywork/counters/driver` that give the PBFT Batch runs every opportunity to
succeed.

*Depends on what's considered non-deterministic, but the chaincode no longer maintains an independent idea of what the state should be.
## Motivation and Context

Completely Fixes #1857
## How Has This Been Tested?

I added a new test for the new counters chaincode. I've also run the new make targets interactively and verified that they work.
## Checklist:

### DIFF
--- a/tools/busywork/README.md
+++ b/tools/busywork/README.md
@@ -89,11 +89,14 @@ to run a simple stress test. This stress test exercises a single peer
 running `NOOPS` consensus.
 
 At present the PBFT *batch* algorithm is the only true consensus algorithm
-being suported by the development team. The target 
+supported by the development team. The targets
 
     make stress2b
+	make sweep1b
 	
-runs PBFT *batch* on a 4-peer network. 
+both test 4-peer PBFT *batch* networks. The `stress2b` test is a single run
+with a single client. The `sweep1b` test sweeps a test setup over a range of
+from 1 to 64 clients.
 
 There is also a pre-canned target
 
@@ -109,10 +112,7 @@ counter (8 bytes) to 1000 counters (8000 bytes):
 
     .PHONY: stress2b1k
 	stress2b1k:
-            @CORE_PBFT_GENERAL_TIMEOUT_NULLREQUEST=3s \
-            $(NETWORK) -batch 4
-            @echo "makefile: Sleeping 10 seconds to allow peers to interlock"
-			sleep 10
+            @(STRESS2_CONFIG) $(NETWORK) -batch 4
 	        @$(STRESS2) -size 1000
 	
 ## Running Peers and Clients on Separate Machines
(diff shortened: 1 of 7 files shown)


jyellick:
I can confirm that stress2b completes on my laptop in Vagrant.  The sweep1b and sweep2b tests seem to have problems, I believe one of these may be #2077, and have no reason to believe it's a problem in this changeset.

Looks good to me.

bcbrock:
Please do not merge yet. It looks like I may need to tweak some parameters to allow some of the tests to work on a laptop under Vagrant.

bcbrock:
OK, I would like to reinstate the request that these changes be merged. For now we have to accept the fact that some of the busywork tests don't work on laptops. Perhaps the limitation of having only 1 or 2 cores to run 4 peers + 4 chaincode containers + N client processes stresses the system to the point that dropping transactions is currently unavoidable (as Jason explained to me privately).

srderson:
LGTM, thanks

PR #2082, opened 2016-06-30
Move wiki sections 'Proposals' and 'Release Notes' to docs directory of code repository

tamasblummer:
Move wiki sections 'Proposals' and 'Release Notes' to docs directory of code repository 
## Description

Move wiki sections 'Proposals' and 'Release Notes' to docs directory of code repository 
## Motivation and Context
1. Documentation should be versioned together with code
2. Enable PRs for proposals.

Fixes #
## How Has This Been Tested?

opened in browser
## Checklist:

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -115,3 +115,18 @@ Overview docs; Reference docs; API and chaincode developer docs; Network operati
      - Deploy Transaction
      - Execute Transaction
      - Chaincode Execution
+
+### Release Notes
+* [Fabric Releases](Fabric-Releases.md)
+* [Current API Changes](Sprint-Release-Notes.md)
+* [Release Process](Release-Process.md)
+
+### Proposals
+* [Fabric Next: Big Picture](Fabric-Next.md)
+* [Next Consensus Architecture Proposal](Next-Consensus-Architecture-Proposal.md)
+* [Next Ledger Architecture Proposal](Next-Ledger-Architecture-Proposal.md)
+* [Client SDK Specification](Client-SDK-Specification.md)
+* [Access Control Specification](Access-Control-Specification.md)
+* [Custom Events](Custom-Events-High-level-specification.md)
+* [System Chaincode](System-Chaincode-Specification.md)
+
--- a/docs/README.md
+++ b/docs/README.md
@@ -117,16 +117,16 @@ Overview docs; Reference docs; API and chaincode developer docs; Network operati
      - Chaincode Execution
 
 ### Release Notes
-* [Fabric Releases](Fabric-Releases.md)
-* [Current API Changes](Sprint-Release-Notes.md)
-* [Release Process](Release-Process.md)
+* [Fabric Releases](releasenotes/Fabric-Releases.md)
+* [Current API Changes](releasenotes/Sprint-Release-Notes.md)
+* [Release Process](releasenotes/Release-Process.md)
 
 ### Proposals
-* [Fabric Next: Big Picture](Fabric-Next.md)
-* [Next Consensus Architecture Proposal](Next-Consensus-Architecture-Proposal.md)
-* [Next Ledger Architecture Proposal](Next-Ledger-Architecture-Proposal.md)
-* [Client SDK Specification](Client-SDK-Specification.md)
-* [Access Control Specification](Access-Control-Specification.md)
-* [Custom Events](Custom-Events-High-level-specification.md)
-* [System Chaincode](System-Chaincode-Specification.md)
+* [Fabric Next: Big Picture](proposals/Fabric-Next.md)
+* [Next Consensus Architecture Proposal](proposals/Next-Consensus-Architecture-Proposal.md)
+* [Next Ledger Architecture Proposal](proposals/Next-Ledger-Architecture-Proposal.md)
+* [Client SDK Specification](proposals/Client-SDK-Specification.md)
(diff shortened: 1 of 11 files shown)


christo4ferris:
@hyperledger/fabric-maintainers I think it would make more sense to deal with each proposal independently in its own PR... that way, we organize discussion around each proposal in the PR comments. 

I think that this PR should be scaled back to be a top-level page that describes the purpose of this directory tree in the repo, and then the leaves would be each proposal. The process would be that a new page would be proposed via a PR, and that PR will be merged only when there is consensus amongst the maintainers to do so. Right now, I would suggest that we use a threshold of three (3) maintainers to merge a proposal.

Once merged, a proposal could be amended or augmented via subsequent PR, with the same threshold for merging as the original.

As for release notes page and release process pages, those could be merged at this point. I am not clear that we should preserve the sprint release page. I think that was an artefact of the IBM process that predated the contribution.

srderson:
I agree with @christo4ferris. Based on the mailing list discussion, it seems there is a phase when a proposal is not yet mature and could be worked on within the wiki. More mature proposals can be moved to the repository. I do think the Next-Consensus-Architecture is likely mature at this point and should be moved in a separate PR for consideration.

ijmitch:
Forgive me if I'm reading this wrong, but do we need to limit ourselves to working with a single PR at a time? The Next-Consensus-Architecture issue discussion is already crying out for some threading, which would be no better in a single PR.

I would see the idea of treating proposals as full-fledged documents which we can fork, raise issues against and submit (multiple, specific) PRs.

Perhaps we would need to formalise a lifecycle for proposals so it's clear what the status is (most importantly that it's a work-in-progress _proposal_ and not a reference doc for what's agreed and implemented). Maybe all this needs is a 'proposals' directory.

christo4ferris:
@ijmitch no, I think that multiple proposals can be active at a time... it is a function of how many people we have contributing to the development that will limit our concurrency. Note that this is also not advocating that a proposal be implemented in a single PR. The proposal document itself is merged when the maintainers/community have reached consensus that it is ready to be implemented. Then, we'd track with an issue(s) and fulfil theincremental development through multiple PRs as usual.

docs/README.md
```diff
@@ -115,3 +115,18 @@ Overview docs; Reference docs; API and chaincode developer docs; Network operati
      - Deploy Transaction
      - Execute Transaction
      - Chaincode Execution
+
+### Release Notes
+* [Fabric Releases](releasenotes/Fabric-Releases.md)
+* [Current API Changes](releasenotes/Sprint-Release-Notes.md)
+* [Release Process](releasenotes/Release-Process.md)
+
+### Proposals
```
christo4ferris, line 124: @tamasblummer I also think that given that we are publishing a pretty-print version of the docs from this tree that this isn't the appropriate location for the proposals. I would recommend a separate tree eg fabric/proposals. I would also propose that we have a top-level README that describes the process. I'd be happy to take this to email with a proposed readme describing the process.

joshhus:
Right, @tamasblummer as you say everything in /docs is being published as "documentation" of the Hyperledger Fabric, both in GitHub and as a "pretty" display version also. So anything that isn't documenting the Fabric, such as proposals, etc. should arguably be somewhere else, so it's not confused as something that is available with the fabric. Release notes does make sense to me as a subdirectory of /docs.

christo4ferris:
@tamasblummer have you plans to update this with the feedback above?

tamasblummer:
@joshhus it makes sense to have proposals in the code tree as soon as there is more than one contributor to a proposal, but at latest as soon as there is a PR implementing it, so it can amend it to match code. The goal is to keep code and its documentation in lockstep.

joshhus:
It's arguable whether to publish "proposals" to the world, or not to. My only point is that if it's in /fabric/docs, it will be published automatically at: http://hyperledger-fabric.readthedocs.io/en/latest/ ... so decide accordingly, thanks.

christo4ferris:
@tamasblummer I don't see proposals as documentation, excepting as archived decision-making. I completely agree we want to keep the code in lock-step with the documentation. That is critical. However, I would like to see the proposals under a /proposals folder in the fabric tree, maybe with a README that links to them all. We iterate on a proposal in a PR and when we reach consensus on the approach, we merge it and begin work. I don't know that we necessarily need to keep updating the proposal unless we want to head in a very different direction, then I could see a PR against a merged proposal that we used to affect that pivot.

christo4ferris:
@hyperledger/fabric-maintainers let's close on this one, please - in the end, it is your decision - but we need to make it now rather than have this drag on as it has (vacations aside... I realize that you have been rather sparse the past few weeks).

christo4ferris:
@tamasblummer I didn't mean for you to **close** the PR I meant close on the decision amongst the maintainers as to how to proceed.

srderson:
I vote to do this one proposal at a time. Trying to move all at once is too much. We need to decide on these one by one.

If a proposal is in flux, it can stay in the wiki. If it's finalized or near finalized, it can be moved to the /docs/proposals directory.

christo4ferris:
I think that the proposals should not be in docs... I don't consider them docs in the traditional sense. docs should be primarily for end-users, developers and operators. I think /proposals should be under fabric. my $0.02

christo4ferris:
@tamasblummer I am going to close this, as I think we have a process now. See email thread. Thanks

PR #2051, opened 2016-06-28
membershipsrv -  add membersrvc/ca/utils.go tests #1842 

gromeroar:
## Description

This PR solves the issue #1842 adding more tests and code fixes to membership services.
## Motivation and Context

Grow test coverage into membership services. Some function in util.go file were replaced by the equivalent existing functions in package primitives. The table below shows the function has been replaced in the first column and the function by what has been replaced in the second column. The function invocations were tested considering function compatibility making all the encrypt/decrypt/padding/unpadding combinations between both files. 
membersrvc/ca/util.go core/crypto/primitives/aes.go
**PKCS5Pad          ->           PKCS7Padding**
**PKCS5Unpad     ->            PKCS7UnPadding**
**CBCEncrypt        ->             CBCPKCS7Encrypt**
**CBCDecrypt       ->             CBCPKCS7Decrypt**

This task has been made in order to reach a more reusable code avoiding duplicate functions. 
## How Has This Been Tested?

- Compatibility tests: All valid combinations of encrypt/decrypt/padding/unpadding have been tested
- Two tests have been added into util_test.go file. 
- Unit tests have been run without errors. 
- Behave tests have been run without errors.
- Node.js tests have been run without errors.
## Checklist:

- [X] I have added a [Signed-off-by](https://github.com/hyperledger/fabric/blob/master/CONTRIBUTING.md#legal-stuff).
- [X] this change requires no new documentation.
- [X] I have either added unit tests to cover my changes.
- [X] I have run [golint](https://github.com/golang/lint) and have fixed valid warnings in code I have added or modified. This tool generates false positives so you may choose to ignore some warnings. The goal is clean, consistent, and readable code.

The continuous integration build process will run [make checks](https://github.com/hyperledger/fabric/blob/master/Makefile#L22) to confirm that tests pass and that code quality meets minimum standards. You may optionally run this locally as PRs will not be accepted until they pass.

Signed-off-by: 
Guillermo Romero : gromeroar guillermo.romero@gmail.com / rologui@ar.ibm.com

Issue #1780, opened 2016-06-10
Consolidate duplicate/bespoke CBCEncrypt/CBDecrypt implementations


JonathanLevi:
## Description

Just spotted some AES code duplication in [fabric/membersrvc/ca/util.go](https://github.com/hyperledger/fabric/blob/5dfbf946e158791b9934f3880907f21f8dd7f796/membersrvc/ca/util.go) specifically [CBCEncrypt](https://github.com/JonathanLevi/fabric/blob/ca8135b1a0f95127c96478518628a65a377ec123/core/crypto/primitives/aes.go#L68-L98) and [CBCDecrypt](https://github.com/JonathanLevi/fabric/blob/ca8135b1a0f95127c96478518628a65a377ec123/core/crypto/primitives/aes.go#L100-L134). 

At the same time, we are also sharing the AES functionality and wrappers here [core/crypto/primitives/aes.go](https://github.com/JonathanLevi/fabric/blob/ca8135b1a0f95127c96478518628a65a377ec123/core/crypto/primitives/aes.go#L68-L98), specifically [CBCEncrypt](https://github.com/JonathanLevi/fabric/blob/ca8135b1a0f95127c96478518628a65a377ec123/core/crypto/primitives/aes.go#L68-L98) and [CBCDecrypt](https://github.com/JonathanLevi/fabric/blob/ca8135b1a0f95127c96478518628a65a377ec123/core/crypto/primitives/aes.go#L100-L134).
## Describe How to Resolve

We should use only the shared functions in [core/crypto/primitives/aes.go](https://github.com/JonathanLevi/fabric/blob/ca8135b1a0f95127c96478518628a65a377ec123/core/crypto/primitives/aes.go#L68-L98), have one implementation, update existing calls to use the shared one, write some tests to verify this [especially as these 2 implementations are not identical] and remove the duplicate from the good ol' `fabric/membersrvc/ca` code.
## Good First Bug

I consider this a **Good First Bug** especially for those in the community who would like to pick up a coding task and get involved in fabric, build it, test it, and learn how to contribute to it. Need this done quickly and will be happy to spend some time guiding (pointing to the right Wiki's mostly), and verifying this all the way to submitting the PR [ usually at the modest tuition charge of 1-2 `MoreTests` digital currency ;-) ]

Interested? Hit me/us here, or on [HL's Slack](http://hyperledgerproject.slack.com/jonathanlevi)

JonathanLevi:
[...]

PR #2092, opened 2016-07-01
Updated AES CBC encryption/decrpytion and padding

itp4IBM:
## Description

In the original code, the PKCS7Padding function always adds an extra block of padding for messages that are multiple of the AES blockSize. An "if statement" has been added to take care of that. 

Also, in the PKCS7UnPadding function, an "if statement" has been added to take care of the condition when a message is empty, e.g message := "". In the current code, an empty message will cause a panic.

Finally, in AESEncryptCBC and AESDecryptCBC functions, changes are made to variable declaration.
## Motivation and Context

Added code in PKCS7Padding function removes the need for padding for plaintext messages with length of multiple of the AES blocksize.

Added code in PKCS7UnPadding function to prevent users from encrypting an empty message which will trigger a panic.

To satisfy compliance requirements.

## How Has This Been Tested?

This only changes implementation of existing functions and existing unit test cases (primitives_test.go) cover these changes.
## Checklist:

Signed-off-by: intelmind787@gmail.com

### DIFF
--- a/core/crypto/primitives/aes.go
+++ b/core/crypto/primitives/aes.go
@@ -22,7 +22,6 @@ import (
 	"crypto/cipher"
 	"crypto/rand"
 	"errors"
-	"fmt"
 	"io"
 )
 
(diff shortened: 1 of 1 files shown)


JonathanLevi:
Hi @itp4IBM, we have been talking/working exactly on this with @starsheriff as part of #1780 and following PR #2051.

We might have a few bunch of tests that can we useful to test and support this change/PR...

JonathanLevi:
Moved my comment to #2093.

JonathanLevi:
@itp4IBM Do you want to discuss this and #2093 over Slack, or shall I review these here? These need some work... but along with some other work that's in the making, we can do these together.

itp4IBM:
Hello @JonathanLevi . Let us discuss #2092 and #2093 over slack.

Thank you.

core/crypto/primitives/aes.go
```diff
@@ -39,108 +38,101 @@ func GenAESKey() ([]byte, error) {
...
 }
 
-// PKCS7Padding pads as prescribed by the PKCS7 standard
-func PKCS7Padding(src []byte) []byte {
-	padding := aes.BlockSize - len(src)%aes.BlockSize
-	padtext := bytes.Repeat([]byte{byte(padding)}, padding)
-	return append(src, padtext...)
+// PKCS7Padding adds padding for any block smaller than AES block size
+func PKCS7Padding(message []byte) []byte {
+	if len(message)%aes.BlockSize == 0 {
```
JonathanLevi: OK, that's the fundamental change here. Nice! Let's add a few tests, padding source/messages of various sizes.

core/crypto/primitives/aes.go
```diff
@@ -39,108 +38,101 @@ func GenAESKey() ([]byte, error) {
...
+		return message
+	}
+
+	padSize := aes.BlockSize - len(message)%aes.BlockSize
```
JonathanLevi, line 47: No need to rename these variables IMO[padtext/padBytes == potatoes/poTatoes ;-)], to make it easier to focus on the proposed change.

If you feel strongly about the names, we can do so/discuss in a separate issue/PR.

core/crypto/primitives/aes.go
```diff
@@ -39,108 +38,101 @@ func GenAESKey() ([]byte, error) {
...
+
+	return append(message, addByte...)
 }
 
-// PKCS7UnPadding unpads as prescribed by the PKCS7 standard
-func PKCS7UnPadding(src []byte) ([]byte, error) {
-	length := len(src)
-	unpadding := int(src[length-1])
+//PKCS7UnPadding removes any padding from plain text
+func PKCS7UnPadding(plainText []byte) ([]byte, error) {
```
JonathanLevi, line 54: `plaintext`, one word

core/crypto/primitives/aes.go
```diff
@@ -39,108 +38,101 @@ func GenAESKey() ([]byte, error) {
...
 // CBCPKCS7Encrypt combines CBC encryption and PKCS7 padding
 func CBCPKCS7Encrypt(key, src []byte) ([]byte, error) {
-	return CBCEncrypt(key, PKCS7Padding(src))
+	return AESEncryptCBC(key, PKCS7Padding(src))
 }
 
 // CBCPKCS7Decrypt combines CBC decryption and PKCS7 unpadding
 func CBCPKCS7Decrypt(key, src []byte) ([]byte, error) {
-	pt, err := CBCDecrypt(key, src)
+	pt, err := AESDecryptCBC(key, src)
```
JonathanLevi, line 135: +1 for adding the AES (despite working on `aes.go`. Golang does not really care much about filenames, eh?).

Since we use `AES` in `CBC` mode (and then have Encrypt/Decrypt), how about we go for: `AESCBCEncrypt` ?

core/crypto/primitives/aes.go
```diff
@@ -39,108 +38,101 @@ func GenAESKey() ([]byte, error) {
...
 	if err != nil {
 		return nil, err
 	}
 
-	// The IV needs to be unique, but not secure. Therefore it's common to
-	// include it at the beginning of the ciphertext.
-	if len(src) < aes.BlockSize {
-		return nil, errors.New("ciphertext too short")
+	if len(cipherText) < aes.BlockSize {
+		return nil, errors.New("ciphertext needs to be multiple of 16")
```
JonathanLevi, line 112:
``` go
if len(cipherText) < aes.BlockSize {
        return nil, errors.New("ciphertext needs to be multiple of 16")
```

Maybe a different error message?

core/crypto/primitives/aes.go
```diff
@@ -39,108 +38,101 @@ func GenAESKey() ([]byte, error) {
...
-	// be secure.
-	return ciphertext, nil
+	return tmpCiphertext, nil
 }
 
-// CBCDecrypt decrypts using CBC mode
-func CBCDecrypt(key, src []byte) ([]byte, error) {
-	block, err := aes.NewCipher(key)
+// AESDecryptCBC decrypts a AES block with CBC mode
+func AESDecryptCBC(decKey, cipherText []byte) ([]byte, error) {
```
JonathanLevi, line 104: Similarly, `ciphertext` etc.

core/crypto/primitives/aes.go
```diff
@@ -39,108 +38,101 @@ func GenAESKey() ([]byte, error) {
...
+	plaintextLen := len(plainText)
+	if plaintextLen == 0 {
+		return nil, errors.New("length of plaintext is 0")
+	}
 
-	if unpadding > aes.BlockSize || unpadding == 0 {
-		return nil, fmt.Errorf("invalid padding")
+	removepadSize := int(plainText[plaintextLen-1])
+	if removepadSize == 0 || removepadSize > aes.BlockSize {
+		return plainText, nil
```
JonathanLevi, line 62: So we are aware that this is the main change in functionality, right? Is this the intended behavior? Can we please test before/after this change? Again, happy to assist.

starsheriff:
Hi, 
if I may add some comments. The changes will break the tests we have prepared for #1780, so we would need to adjust them. If we include these changes before we accept this PR we could easily document the behavior before and after the changes.

The issue with the padding adding an extra block if the message is a multiple of `aes.BlockSize` came up there as well and I think changing that is a great idea. I would be ready to start a PR for #1780 right away.

itp4IBM:
hello @JonathanLevi:
Thank you for your comments.  I will update the code with your suggestions.

mastersingh24:
@itp4IBM  - I am not exactly sure what you were trying to fix here?  Maybe another way to say this is what led you to think this is a problem?

itp4IBM:
Hello Mr. Singh, @mastersingh24:
The changes made to CBC Encryption/Decryption was an attempt to remediate the compliance issue.
The changes made to PKCS7Padding/UnPadding aims to improve the efficiency of the decryption/encryption step. 
Thank you.

christo4ferris:
@itp4IBM what compliance issue? there's nothing but a vague assertion mentioned. compliance with what?

JonathanLevi:
@itp4IBM I think that a test/test-case (or a few) will make it easier for you to present the change/proposal, and clearer us for us to evaluate/review...

JonathanLevi:
@itp4IBM Can you please also elaborate on the second point too? That is, how do _The changes made to PKCS7Padding/UnPadding aims to improve the efficiency of the decryption/encryption step_?

Improving efficiency is what way? Is this a proposal for a better/faster/more-compact padding /encryption scheme?

smithbk:
@christo4ferris It is a legal issue because the method that was changed had a non-apache license.  This was just discovered.

itp4IBM:
@JonathanLevi . The purpose of the padding is to make the plaintext a multiple of 16 bytes. But the original code would add 16-byte padding to plaintexts that are already a multiple of the AES blocksize. The padding and unpadding does consume some computation time. For a large number of transactions where encryption/decryption that needs to be done, this could add extra time. 

Thank you.

JonathanLevi:
(Good morning) Yes, I wasn't sure at first whether this was the intended behavior or an oversight. Thank you for confirming @itp4IBM.

OK, so a few things:

1. The computation load/time it takes to add an extra (AES blocksize=) 16 bytes to the 'payload' is negligible compared to the computation required for the encryption/decryption.

2. I think these are your first PRs to fabric. I do not want to come across as too 'harsh', quite the opposite. I wish we had an open issue against which these 2 PRs were submitted, as then I would have been able to discuss this earlier.

3. While I have been told that this code/PR has been reviewed by others, please appreciate that I will be doing many people (/several Hyperledger projects?) a dis-service if I give my thumbs-up to this: These two naive looking changes make us not comply with the standard AES/PCKS#7 standard.

Even if people believe/told you that it is fine having this 'faster' version of AES/Padding (as fabric shares this crypto code both in the client and server...) we won't be able to interoperate with the rest of the world.

For example, think about Membership Services issuing x.509 certificates that only 'fabric' can produce/read or parse.

JonathanLevi:
(Good morning) Yes, I wasn't sure at first whether this was the intended behavior or an oversight. Thank you for confirming @itp4IBM.

OK, so a few things:
1. The computation load/time it takes to add an extra (AES blocksize=) 16 bytes to the 'payload' is negligible compared to the computation required for the encryption/decryption.
2. I think these are your first PRs to fabric. I do not want to come across as too 'harsh', quite the opposite. I wish we had an open issue against which these 2 PRs were submitted, as then I would have been able to discuss this earlier.
3. While I have been told that this code/PR has been reviewed by others, please appreciate that I will be doing many people (/several Hyperledger projects?) a dis-service if I give my thumbs-up to this: These two naive looking changes make us not comply with the standard AES/PKCS#7 standard [I'll add a reference below]

Even if people believe/told you that it is fine having this 'faster' version of AES/Padding (as fabric shares this crypto code both in the client and server...) we won't be able to interoperate with the rest of the world.

For example, think about Membership Services issuing x.509 certificates that only 'fabric' can produce/read or parse.

christo4ferris:
@JonathanLevi reference missing? what behavior does the standard require? Is this your concern:
" If some of your messages require padding, then you will have to add some sort of padding systematically, otherwise decryption will be ambiguous." http://security.stackexchange.com/questions/29993/aes-cbc-padding-when-the-message-length-is-a-multiple-of-the-block-size

JonathanLevi:
[@christo4ferris, sorry I was just typing this at the same time, yes!]

See [RFC2315, PKCS \#7: Cryptographic Message Syntax, v1.5, 1998](https://tools.ietf.org/html/rfc2315). 

Page 21:
Some content-encryption algorithms assume the input length is a multiple of k octets, where k > 1, and let the application define a method for handling inputs whose lengths are not a multiple of k octets. For such algorithms, the method shall be to pad the input at the trailing end with k - (l mod k) octets all having value k - (l mod k), where l is the length of the input.

In other words, the input is padded at the trailing end with one of the following strings:

```
                     01 -- if l mod k = k-1

                  02 02 -- if l mod k = k-2
                      .
                      .
                      .
            k k ... k k -- if l mod k = 0
```

The padding can be removed unambiguously since all input is padded and no padding string is a suffix of another. This padding method is well-defined if and only if k < 256; methods for larger k are an open issue for further study.

-------

So specifically, this line
```
k k ... k k -- if l mod k = 0
```
is how this padding works (rule of thumb/when in doubt in AES-CBC: 'we are always padding' => when it's a multiple (k=0) we append a full block.

The original code is/was doing the right thing, adding an extra block when k=0.

JonathanLevi:
[@christo4ferris, sorry I was just typing this at the same time, yes!]

See [RFC2315, PKCS #7: Cryptographic Message Syntax, v1.5, 1998](https://tools.ietf.org/html/rfc2315). 

Page 21:

> Some content-encryption algorithms assume the input length is a multiple of k octets, where k > 1, and let the application define a method for handling inputs whose lengths are not a multiple of k octets. For such algorithms, the method shall be to pad the input at the trailing end with k - (l mod k) octets all having value k - (l mod k), where l is the length of the input.
> 
> In other words, the input is padded at the trailing end with one of the following strings:

```
                     01 -- if l mod k = k-1

                  02 02 -- if l mod k = k-2
                      .
                      .
                      .
            k k ... k k -- if l mod k = 0
```

> The padding can be removed unambiguously since all input is padded and no padding string is a suffix of another. This padding method is well-defined if and only if k < 256; methods for larger k are an open issue for further study.

---

So specifically, this line

```
k k ... k k -- if l mod k = 0
```

is how this padding works (rule of thumb/when in doubt in AES-CBC: 'we are always padding' => when it's a multiple (k=0) we append a full block.

The original code is/was doing the right thing, adding an extra block when k=0.

christo4ferris:
so, the only think that might need addressing is the empty message panic? Or, should this PR simply be closed.

JonathanLevi:
@christo4ferris I'm looking...

core/crypto/primitives/aes.go
```diff
@@ -39,108 +38,101 @@ func GenAESKey() ([]byte, error) {
...
 }
 
-// PKCS7UnPadding unpads as prescribed by the PKCS7 standard
-func PKCS7UnPadding(src []byte) ([]byte, error) {
-	length := len(src)
-	unpadding := int(src[length-1])
+//PKCS7UnPadding removes any padding from plain text
+func PKCS7UnPadding(plainText []byte) ([]byte, error) {
+	plaintextLen := len(plainText)
+	if plaintextLen == 0 {
```
JonathanLevi: @christo4ferris Are we referring to this panic here? That is to return the plaintext here?

core/crypto/primitives/aes.go
```diff
@@ -39,108 +38,101 @@ func GenAESKey() ([]byte, error) {
...
+		return nil, errors.New("length of plaintext is 0")
+	}
 
-	if unpadding > aes.BlockSize || unpadding == 0 {
-		return nil, fmt.Errorf("invalid padding")
+	removepadSize := int(plainText[plaintextLen-1])
+	if removepadSize == 0 || removepadSize > aes.BlockSize {
```
JonathanLevi, line 61: @christo4ferris Are we referring to this panic here? That is to return the plaintext here?

JonathanLevi:
@christo4ferris Assuming/if the below code is what was referred to as the panic... 
then the original code [core/crypto/primitives/aes.go#L54-L55](https://github.com/hyperledger/fabric/blob/ca8135b1a0f95127c96478518628a65a377ec123/core/crypto/primitives/aes.go#L54-L55), returning nil with an error:

```
if unpadding > aes.BlockSize || unpadding == 0 {
        return nil, fmt.Errorf("invalid padding")
```

is/was also correct.

Since the standard states/dictates that we always pad, we cannot 'not unpad'. Therefore, we should not return the plaintext, raising/flagging an error is the standard. It's not really a panic, it's more of a correct reporting.

p.s. There is another check in the CBCEncrypt/CBCDecrypt implementation as for the length, so the two tie up well together actually.

JonathanLevi:
@smithbk, @itp4IBM how about we discuss/approach the (re)naming separately (in different issues/PRs)?

I agree that we could/should do some housekeeping in the core/crypto section. I think we can safely `import` other libraries with the appropriate licensing. Will happily spend some time on this with you two.

itp4IBM:
@JonathanLevi . Thank you for the reference to PKCS #7 padding. 

I will close this PR and PR 2093. I will reach out to you regarding the renaming of the AES CBC functions.

PR #2093, opened 2016-07-01
Updated AES CBC encryption/decryption and padding

itp4IBM:
## Description

In the original code, the PKCS7Padding function always adds an extra block of padding for messages that are multiple of the AES blockSize. An "if statement" has been added to take care of that. 

Also, in the PKCS7UnPadding function, an "if statement" has been added to take care of the condition when a message is empty, e.g message := "". In the current code, an empty message will cause a panic.

Finally, in AESEncryptCBC and AESDecryptCBC functions, changes are made to variable declaration.
## Motivation and Context

Added code in PKCS7Padding function removes the need for padding for plaintext messages with length of multiple of the AES blocksize.

Added code in PKCS7UnPadding function to prevent users from encrypting an empty message which will trigger a panic.

To satisfy compliance requirements.

## How Has This Been Tested?

This only changes implementation of existing functions and existing unit test cases (primitives_test.go) cover these changes.
## Checklist:

Signed-off-by: intelmind787@gmail.com

### DIFF
--- a/core/crypto/primitives/aes.go
+++ b/core/crypto/primitives/aes.go
@@ -22,7 +22,6 @@ import (
 	"crypto/cipher"
 	"crypto/rand"
 	"errors"
-	"fmt"
 	"io"
 )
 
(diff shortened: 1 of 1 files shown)


JonathanLevi:
Hi @itp4IBM, we have been talking/working exactly on this with @starsheriff as part of #1780 (and following PR #2051).

We might have a bunch of tests that can be useful to test and support this change...

JonathanLevi:
@itp4IBM Do you want to discuss this and #2092 over Slack, or shall I review these here? These need some work... but along with some other work that's in the making, we can do these together.

itp4IBM:
hello @JonathanLevi . Let us discuss over Slack. My slack id is: vanle032016 at https://openchain.slack.com.

Thank you.

mastersingh24:
@JonathanLevi  - what's your take on our exposure here?  looking to see if we should merge this in to the developer preview branch as well but not sure we have a huge risk with this?

JonathanLevi:
[Good morning @mastersingh24] Hmmm, yes, I believe that this change requires some more work before merging to either branch, and definitely tests.

It does not make sense that changing such a fundamental primitive (AES-CBC mode + Padding) does not break any test. In addition to my suggestions regarding the current/proposed changes, my view is that we need to test this (if not before, then definitely after the change).

JonathanLevi:
Can quickly cast a quick review & add a few comments to the code in #2092 to assist in assessing the risk.

itp4IBM:
I am closing PR 2093 as it is related to PR 2092.

Issue #1791, opened 2016-06-10
Appending the timestamp in the nonce generation for the TCertIndex


gromeroar:
## Description
- add the timestamp type to the nonce attibute in the TCert index.
## Description
- append the timestamp to the nonce during the Tcert Index generation.

PR #2099, opened 2016-07-04
Addition of the timestamp to the nonce calculation of the TcertIndex

raullaprida:
The current timestamp was appended to the nonce field for the TcertIdex during its generation.
## Description

In tcap.go, line 197, the nonce for the TCertIdex is generated.

We added line 199:
binary.LittleEndian.PutUint64(nonce[8:], uint64(in.Ts.Seconds))

Which basically appends the timestamp to the nonce, to comply with the following format:
nonce := make([]byte, 16) // 8 bytes rand, 8 bytes timestamp
## Motivation and Context

The timestamp is required in the nonce format for the TCertIndex. This fixes issue #1791
## How Has This Been Tested?

Running make unit-test  behave tests and node sdk tests
## Checklist:

Signed-off-by:
Raul laprida: raullaprida laprida@ar.ibm.com


srderson:
@raullaprida should there be a unit test to check for this?

raullaprida:
@srderson  there's no need for a unit test for this change. And there's no original Unit test for the TcertIndex value to modify.
Thanks

srderson:
Thanks, LGTM

PR #1404, opened 2016-05-08
fix 1403

christo4ferris:
fixes #1403

fix incorrect document references to former openblockchain/obc-docs to correctly reference the hyperledger/fabric/docs documentation

Also addressed a couple of lint errors in asset_management_with_roles.go

Signed-off-by: Christopher Ferris chrisfer@us.ibm.com

PR #2100, opened 2016-07-05
Fixes #1401

### LINKED ISSUE #1401: Would someone please document the ins-and-outs of peer addressing?
## Description

The _fabric/peer/core.yaml_ file has 4 settings that control the network connectivity of the peer. 

```
    # The Address this Peer will listen on
    listenAddress: 0.0.0.0:30303
    # The Address this Peer will bind to for providing services
    address: 0.0.0.0:30303
    # Whether the Peer should programmatically determine the address to bind to.
    # This case is useful for docker containers.
    addressAutoDetect: false

    # Peer port to accept connections on
    port:    30303
```

What is the difference between the address "the peer will listen on" and the address used "for providing services". Can these be different? If so, why might one want to make them different? What does `addressAutoDetect` do exactly, why is it useful for Docker, and is it useful for Docker when True or False? My preference would be for these questions to be answered in comments in the YAML file.

Note that the `peer.port` variable is redundant (never used by the code) and should be removed.

Thanks, Bishop

brad-gorman:
## Description

Updated peer/core.yaml configuration descriptions to be more consistent with their actual use and behaviour. Changed peer.addressAutoDetect=true so that it will work out-of-the-box in the simplest cases.
## Motivation and Context

The previous description implied peer.address was used as the listening / binding address of the peer. This is what peer.listenAddress is used for.
The default of peer.addressAutoDetect=false also makes little sense as for the "false" flag to work, you MUST have defined peer.address to a valid address. When it is "true" then at least it will work in simple deployments.
Fixes #1404 
## How Has This Been Tested?

Tested with `make unit-test`. No new features, so no new tests.
## Checklist:

Signed-off-by: Brad Gorman bgorman@au1.ibm.com

### DIFF
--- a/peer/core.yaml
+++ b/peer/core.yaml
@@ -97,11 +97,14 @@ peer:
 
     # The Address this Peer will listen on
     listenAddress: 0.0.0.0:30303
-    # The Address this Peer will bind to for providing services
+
+    # The Address advertised to other services that will connect to this Peer
+    # Chaincode deployment will use this to create chaincode docker containers
+    # (This is NOT a binding address for the listening port)
     address: 0.0.0.0:30303
-    # Whether the Peer should programmatically determine the address to bind to.
-    # This case is useful for docker containers.
-    addressAutoDetect: false
+
+    # Whether to override the IP/hostname portion of peer.address with a local, non-loopback IP address 
+    addressAutoDetect: true
 
     # Setting for runtime.GOMAXPROCS(n). If n < 1, it does not change the current setting
     gomaxprocs: -1


brad-gorman:
reverify jenkins

peer/core.yaml
```diff
@@ -97,11 +97,14 @@ peer:
 
     # The Address this Peer will listen on
     listenAddress: 0.0.0.0:30303
-    # The Address this Peer will bind to for providing services
+
+    # The Address advertised to other services that will connect to this Peer
+    # Chaincode deployment will use this to create chaincode docker containers
+    # (This is NOT a binding address for the listening port)
     address: 0.0.0.0:30303
```
gabre, line 104: I like the more detailed info on those parameters but it is still clear for me what address is used for. I mean if it is listening on listenAddress how can other services connect to it on address? Would it make sense to keep listenAddress but change address to something more descriptive? E.g. xyzAddress (where xyz is something useful of course :))

christo4ferris:
Agree with @gabre comment on renaming 'address' to be more self-descriptive. Maybe advertisedAddress? Otherwise, LGTM

jeffgarratt:
@gabre @christo4ferris I would simply advise thinking carefully about the auto detect to true.  This may confuse folks who set the listenAddress and then are surprised that the logic grabbed the first non-loopback.

christo4ferris:
@brad-gorman where does this stand? Has it been migrated to Gerrit? Should it be closed? Thanks!

christo4ferris:
@brad-gorman we will be transferring the fabric and fabric-api repository and all the corresponding issues and pull requests to the hyperledger-archives org on GH and replacing this repository with a read-only mirror of our Gerrit source. As a result, any automatic GH redirects will be broken. If you haven't already migrated your PR to Gerrit, and intend to do so, you will need to change your git remote to reference the transferred repository (e.g. git@github.com:hyperledger-archives/fabric) on your local clone.

Issue #1631, opened 2016-05-27
Next Consensus Architecture discussion


vukolic:
Use the comment field to discuss the [next consensus architecture proposal](http://github.com/hyperledger/fabric/wiki/Next-Consensus-Architecture-Proposal).

gregmisiorek:
for section  2.1, we should try to use a UTC timestamp of possible: tsUTC.

gregmisiorek:
for section 2.2, 1st alternative design, HASH(txPayload) seems preferable as we should try to avoid clear text wherever we can without losing the clarity of the specification.

adecaro:
Just some clarifications:
In Section 1.2.i, shall 'Clients create and thereby invoke transactions.' be '' Clients create and thereby invoke chaincodes'?
In Section 2.1, shall 'To invoke a transaction' be 'To invoke a chaincode'?

binhn:
@gregmisiorek yes, hash would be preferred, especially in the case of confidential transactions.
@adecaro at this level, we should use transaction instead of chaincode to be more generic.

bmandler:
A couple of comments / thoughts
1.  The struggle to keep a blurry distinction between the different kinds of peers seems to add unnecessary confusion without having a real need (examples of how this complicates the text can be seen in the last paragraph of section 1.2: "Nodes other than committing ..., and section 2.3 first bullet, second paragraph). For simplicity I'd unify these entities and state that there is an entity called peer and it has 3 distinct capabilities, namely endorser, submitter, and committer. To my understanding all peers can play all roles and their specific role is determined per chaincode (so a peer can be endorser for one chaincode, and a submitter for another, but I think that all need to be committers). Since I think that all peers need to maintain the state and all peers should be able to communicate with clients (see 1.2.i: the client may connect to any peer of its choice), I don't see a real value in the separation. 
   • I believe an endorsing peer has to maintain the state as well, first because it may very well be needed for endorsements, and second because it's required for simulating the transaction execution (and is required to respond to a query request).
   • In section 1.2.ii.a it's stated that a committing peer may give clients read access to the blockchain and state. Once again. It means that all committing peers are submitting peers; and in addition if read access to the state means responding to chaincode queries, than only the endorsers can respond (and potentially the submitting peer), since they are the only ones that have the actual chaincode in question.
[...]

PR #2101, opened 2016-07-05
New architecture

gabre:
In this PR, we tried to start the conversion of fabric towards the new architecture (https://github.com/hyperledger/fabric/wiki/Next-Consensus-Architecture-Proposal).
## Description

The PR and the implementation is not complete, we would like to get feedback.
The PR contains the following changes:
- A proto file was added describing all the new message types and services needed by endorsers (endorser-api) and consenter peers (consensus-api).
- A consensus service was introduced. For a node, whether to act as a consenter can be enabled/disabled from configuration. Other nodes use consensus-api to connect to the consensus layer.
- Nodes ask for endorsements before forwarding a transaction to the consensus. Enabled/disabled status of the endorsement service on a node can be configured in core.yaml. Endorsers can be reached using endorser-api.
- Nodes observe the consensus-layer using a consensus observer and consensus-api and commit to the ledger if a new consensus is made. (Ledger related functionalities were moved from consensus to core)

Functionalities missing:
- Real endorsement checking is needed (not just the endorsement count but signature verification). It is a question how we could handle private/public keys and signatures (Do we need a CA? Do we need preconfigured key lists? What about using Schnorr signatures?)
- Some ledger related code needs to be removed from consensus and maybe moved to core if it is still needed. (The TX execution and commitment are done in consensus-observer)
- All the fields of all the records from consensus-api.proto needs to be populated with real, meaningful values.)
## Motivation and Context

The motivation is to begin work on new architecture and start discussion on implementation level problems.

Fixes #
## How Has This Been Tested?

BDD tests will be added.
## Checklist:

Signed-off-by:
Gabor Hosszu gabor@digitalasset.com


christo4ferris:
@gabre this has some linter errors that need to be resolved. see the Travis output.

srderson:
For the read set and write set, what about

```

message VerDep {
      repeated VarRead readSet = 1;
      repeated VarWrite writeSet = 2;
 }

message VarRead {
    string key = 1;
    bytes version = 2;
} 

message VarWrite {
    string key = 1;
    bytes value = 2;
    bytes version = 3;
}
```

srderson:
Making a note to myself to consider the following transaction tomorrow

```
Read key A version 1
Write key A version 2
Read key A version 2
Write key A version 3
```

Need to consider if that order needs to be preserved which I cannot be in the above data structure.

gabre:
@christo4ferris Yes I know about that, I will fix it. This is still a proposal, we would like to have more feedback on this (see my questions in the files, the proto definitions and the "lower level" implementation details). I would like to encourage everybody to give us feedback and suggestions.

gabre:
@srderson What about making just one set of repeated VarUsage s (I was unable to find a better name, Usage = Read U Write U EveryThingElseOneCanThinkOf, U stands for union) and this way we could preserve the order if needed? Just came into my mind, haven't considered it deeply..

srderson:
Something @manish-sethi mentioned this morning based on the conversation at https://github.com/hyperledger/fabric/issues/1631 is that maybe `versionDependencies` should just be bytes

```
message TransactionProposal {
    enum TxContentType {
        TRANSPARENT = 0;
        OPAQUE = 1;
        COMMITMENT = 2;
        NONE = 3;
    }

    string submitingPeerID = 1;
    string clientID = 2;
    bytes chaincodeID = 3;
    TxContentType txContentType = 4;
    bytes txContent = 5;
    bytes versionDependencies = 6;
}
```

This allows different ledger implementations to handle the simulation/results differently. This is very powerful as it allows for unlimited flexibility and does not lock us into the readSet/writeSet concept.

The default implementation would store a read/write set in these bytes. It would look something like

```
message VerDep {
      repeated VarRead readSet = 1;
      repeated VarWrite writeSet = 2;
 }
```

Also @manish-sethi and I confirmed that only the first reads and last writes of a key in a transaction need to be stored so we don't need to worry about capturing intermediate reads/writes.

Thoughts?

christo4ferris:
@gabre @srderson have we migrated this discussion to Gerrit?

christo4ferris:
@gabre we will be transferring the fabric and fabric-api repository and all the corresponding issues and pull requests to the hyperledger-archives org on GH and replacing this repository with a read-only mirror of our Gerrit source. As a result, any automatic GH redirects will be broken. If you haven't already migrated your PR to Gerrit, and intend to do so, you will need to change your git remote to reference the transferred repository (e.g. git@github.com:hyperledger-archives/fabric) on your local clone.

PR #2113, opened 2016-07-06
fix 2112

### LINKED ISSUE #2112: fix documentation link in README.md
## Description

The README.md links to the markdown README documentation page rather than the http://hyperledger-fabric.readthedocs.io/en/latest/ 
## Describe How to Reproduce

click on the documentation link in the main README

christo4ferris:
## Description

fix documentation link
## Motivation and Context

Fixes #2112
## How Has This Been Tested?

tested the markdown link in local repo
## Checklist:

Fixes #2112 

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ We use [Hyperledger Slack](https://slack.hyperledger.org/) for communication and
 [Installation](docs/dev-setup/install.md): Describes how to install the blockchain fabric and use project tools.
 
 ## Documentation
-Project documentation can be found [here](docs/README.md).
+Project documentation can be found [here](http://hyperledger-fabric.readthedocs.io/en/latest/).
 
 ## Still Have Questions?
 For general purpose questions, please use [StackOverflow](http://stackoverflow.com/questions/tagged/hyperledger).


gabre:
I guess it fixes this: https://github.com/hyperledger/fabric/issues/2112 (not 2121)

christo4ferris:
good catch, thanks @gabre my dyslexia sometimes catches me up;-)

christo4ferris:
reverify Jenkins

christo4ferris:
reverify Jenkins

srderson:
LGTM

stephyee:
@jyellick 

> Certainly it looks nice to have the operations and channel participation APIs broken out. But since this is a top level Fabric artifact, do we need to worry about server namespacing? My concern here is that someone will end up trying to use the channel participation APIs against the peer for instance, though perhaps my concerns there are overblown.

I misunderstood your comment before, you concern is not overblown. You mean some APIs are for the orderer but not the orderer etc.. As a temp solution I can add to the descriptions to specify which server supports specific endpoints. We can open a separate issue to investigate namespacing servers. Also I added external doc links which will clarify the API usage. 

> It's great to get at least some basic validation that it's possible to build a client which works for some operation using the swagger definition. Looking at your branch, it certainly is a lot of additional tooling and code, so I agree that it's not obvious we want to pull of that in. I hate advertising support for a feature, and having no way to tell if we break it (beyond the obvious syntactical failures), but if it would be too onerous to integrate, I suppose we might have to skip it this time. Still, I feel much better about integrating this with the manual testing done.

Fair points. If we could minimize the dependencies and code used to validate client generation then it'd be worthwhile to have a build check for it.

I'm going to rebase but like Brett says, the build check is an invalid here. 
This PR isn't exactly straightforward so thanks for all the reviewing pre and post holidays.

Issue #802, opened 2016-03-08
Log level does not  follow the defination in openchain.yaml


samding01:
The latest master code (commit ee4056e5f799527fefc285e9e70d54ee7c543d71) does not follow log level definition in `openchain.yaml`.

Here is part of definition of the `openchain.yaml`:

```
peer:      error
crypto:    error
status:    warning
stop:      warning
login:     error
vm:        warning
chaincode: warning
```

However, after starting 4 peers with security enabled, consensus=obcpbft,  the container (vp0-3) outputs:

```
  ...
 14:47:59.484 [consensus/handler] HandleMessage -> DEBU 0c2 Did not handle message of type DISC_GET_PEERS, passing on to next MessageHandler
 14:47:59.484 [peer] HandleMessage -> DEBU 0c3 Handling OpenchainMessage of type: DISC_GET_PEERS
 14:47:59.484 [peer] beforeGetPeers -> DEBU 0c4 Sending back DISC_PEERS
 14:47:59.484 [peer] SendMessage -> DEBU 0c5 Sending message to stream of type: DISC_PEERS
  ...
```

If run 1 peer in [`sandbox`](https://github.com/openblockchain/obc-docs/blob/master/api/SandboxSetup.md), it follows the log level definition, but outputs

```
 block 2d 2d 2d 2d 2d 42 4 .....
```

2016Nishi:
Found similar issue when testing for consensus on 4 peers.
Is this a duplicate of #763?

juliancarrivick-ibm:
I am happy to have a look at this. I am assuming that `openchain.yaml` has become `peer/core.yaml`.

Glancing at it, it seems the logging levels in the example `core.yaml` do not necessarily correspond with the commands available (e.g. vm is not used at this point in time). Should they only correspond with the top level commands (e.g. node, network, chaincode) or are they intended to be able to be nested (e.g. node status).

Finally, INFO seems to be the default logging level. This seems a bit verbose. Thoughts?

juliancarrivick-ibm:
This was due to it looking for `logging.node` in `core.yaml` which didn't exist at the time of this issue. PR #2076 fixed this and this can now be closed.

JonathanLevi:
Closed by #2076 as per the above note. Thank you @bcbrock, @srderson and of course @juliancarrivick-ibm.

PR #2121, opened 2016-07-07
Add test cases for the logging.go file

juliancarrivick-ibm:
## Description

Add test cases for the Logging Initialisation code which parses the logging configuration in core.yaml
## Motivation and Context

There were previously no tests for parsing the config in core.yaml. In the process of looking at issue #802 I decided it would be easier to fiddle with if there were tests in place.
## How Has This Been Tested?

This pull request consists of tests.
## Checklist:

### DIFF
--- a/core/logging_test.go
+++ b/core/logging_test.go
@@ -1,0 +1,157 @@
+/*
+Copyright IBM Corp. 2016 All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+		 http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package core
+
+import (
+	"testing"
+
+	"github.com/op/go-logging"
+	"github.com/spf13/viper"
+)
+
+func TestLoggingLevelDefault(t *testing.T) {
+	viper.Reset()
+
+	LoggingInit("")
+
+	assertDefaultLoggingLevel(t, DefaultLoggingLevel())
+}
+
+func TestLoggingLevelOtherThanDefault(t *testing.T) {
+	viper.Reset()
+	viper.Set("logging_level", "warning")
+
+	LoggingInit("")
+
+	assertDefaultLoggingLevel(t, logging.WARNING)
+}
+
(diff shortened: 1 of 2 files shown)


gabre:
For me, it seems to be OK. Travis/Jenkins errors seem to be totally unrelated. (protoc not found, peer_basic.feature:1267 fails)

gabre:
reverify jenkins

core/logging_test.go
```diff
@@ -0,0 +1,157 @@
...
+	viper.Reset()
+	viper.Set("logging_level", "test=warning:info")
+
+	LoggingInit("")
+
+	assertDefaultLoggingLevel(t, logging.INFO)
+	assertModuleLoggingLevel(t, "test", logging.WARNING)
+}
+
+func TestLoggingLevelForSpecificCommand(t *testing.T) {
```
christo4ferris, line 92: shouldn't each command for which a distinct logging level be tested?

core/logging_test.go
```diff
@@ -0,0 +1,157 @@
...
+func TestLoggingLevelOtherThanDefault(t *testing.T) {
+	viper.Reset()
+	viper.Set("logging_level", "warning")
+
+	LoggingInit("")
+
+	assertDefaultLoggingLevel(t, logging.WARNING)
+}
+
+func TestLoggingLevelForSpecificModule(t *testing.T) {
```
christo4ferris, line 43: should we be testing for each valid module?

srderson:
Travis is rerunning now. Will investigate if the failure shows up again

core/logging_test.go
juliancarrivick-ibm, line 92: The current logging logic simply uses the command given to do a lookup in viper:

``` Go
spec := viper.GetString("logging_level")
if spec == "" {
    spec = viper.GetString("logging." + command)
}
// Parse spec
```

Doing multiple different commands would just exercise the same logic but I can add some extra ones if you would like.

I think the logging level logic needs a look at which I was going to do for issue 802 but I thought it would be best to get some tests in place first.

core/logging_test.go
juliancarrivick-ibm, line 43: A 'module' in this case is just an arbitrary string that gets passed to the logger which can be used when formatting the log statement. Doing it with different values would just exercise the same logic with the different value.

I can add some different values though if you would like.

srderson:
@juliancarrivick-ibm Shall I go ahead and merge this PR or do you plan to make changes based on @christo4ferris's comments? Seems those can be made in a second PR if that works better for you.

christo4ferris:
I could live with follow-on PR.

juliancarrivick-ibm:
@srderson Go ahead and merge it!
Depending on how I go with issue #802 I might create another pull request for those extra tests.

srderson:
LGTM

PR #2128, opened 2016-07-07
Behave test scenario making use of attributes.

### LINKED ISSUE #2116: A behave test scenario making use of attributes is needed
## Description

In order to increase the Behave test coverage, a scenario making use of attributes is required 
## Describe How to Reproduce

Run behave tests.

### LINKED ISSUE #2126: The node SDK hfc command should support the "peer network/chaincode" commands
## Description

The "hfc" command is installed by running
     npm install -g hfc
The hfc command currently only supports a
    hfc test <test-input.json>
command, which uses the node SDK to drive load against the fabric.
This command should also support the "peer network" and "peer chaincode" commands.

This means fabric/sdk/node/bin/main.js must be enhanced to support something similar to:
    hfc login <user> <pass>
    hfc list
    hfc deploy ...
    hfc invoke
    hfc query
Options:
    -c <json-config-file>                       (check for '$HOME/hfc.json' if not specified)
    -ms <memberServicesUrl>           (default value is "grpc://localhost:50051")
    -peer <peerUrl>                            (default value is "grpc://localhost:30303", and this option can be used multiple times for failover)

The content of the <json-config-file> might be as follows:
   {
           "memberServicesUrl": "grpc://localhost:50051",
           "peers": [ "grpc://localhost:30303" ],
   }
## Describe How to Reproduce

Not an issue

andresgaragiola:
## Description

These changes add the following:
- A behave scenario to tests chaincode making use of attributes.
- An upgrade of the behave test's steps, in order to use "/chaincode" rest API endpoint instead of the deprecated "/devops/invoke|query|deploy" ones.
- A way to assert hierarchical attributes in behave tests, for example:
  
   _Then I should get a JSON response with **"result.message"** = "10"_
## Motivation and Context

Fixes #2116 
Makes use of "/chaincode" endpoint in behave tests.
Hierarchical attributes are required in order to support "/chaincode" service.
## How Has This Been Tested?

Behave tests were executed without errors. 
## Checklist:

- [X] this change requires no new documentation.
- [X] I have either added behave tests to cover my changes.

Signed-off-by:
Andrés Garagiola garagiol@ar.ibm.com / andresgaragiola@gmail.com


andresgaragiola:
@srderson I have seen that travis failed. It is because 'The job exceeded the maximum time limit for jobs, and has been terminated.'. I have added two scenarios 1) Using attributes in clear, 2) Using attributes with encryption enabled. Attributes encryption is not available yet, so in this moment in the second scenario is only setting the property "attribute-encryption.enabled = true" but the behavior is the same.  I could momentarily comment this second scenario in order to reduce the build time. Anyway I think that soon this scenario should be included again. Let me know.

jeffgarratt:
@srderson @andresgaragiola Looks good to me!! Nice job!

rameshthoomu:
Behave test case  (  `Scenario: java SimpleSample chaincode example single peer` ) failed in Jenkins x86 with below error message: `Assertion Failed: Failed to POST to http://172.17.0.31:5000/chaincode:  {"jsonrpc":"2.0","error":{"code":-32700,"message":"Parse error","data":"Error unmarshalling chaincode request payload: json: cannot unmarshal string into Go value of type protos.ChaincodeSpec_Type"},"id":null}`. 
Jenkins Log file:
https://jenkins.hyperledger.org/job/fabric-github-verify-x86_64/527/artifact/gopath/src/github.com/hyperledger/fabric/bddtests/testsummary.log
Same test case is failed in `Travis` too.
Below is the log captured from Travis Raw log file:
Scenario java SimpleSample chaincode example single peer failed. Getting container logs (This indicates fail). @andresgaragiola please check this..

andresgaragiola:
@rameshthoomu after merge java shim scenario some errors have appeared. 
- The most important issue is that "/chaincode" service doesn't support pass attribute 'type' of      chaincode spec by value. 
  
   If 'JAVA' is passed as type the following error is raised during Unmarshal 
  
  **"Error unmarshalling chaincode request payload: json: cannot unmarshal string into Go value of      type protos.ChaincodeSpec_Type"**
  
  to avoid this the ID must be passed ("4" for the mentioned example), 
  
  ```
     enum Type {
       UNDEFINED = 0;
       GOLANG = 1;
       NODE = 2;
       CAR = 3;
       JAVA = 4;
  }
  ```
  
  This is a problem of the "/chaincode" services and is out of the scope of this PR. I see two options    here:
    a) Covert values to ids in the behave test scenario.
    b) Use deprecated API again in the behave tests scenario.
  
  I chose option a) to fix the error because in my opinion test non deprecated functions is more   relevant than use type as string, but I think this problem should be analyzed in a new issue in order to fix "/chaincode" services.
- A second issue is due that while "/devops/query" service returns JAVA chaincode value encoded as [JSON](https://github.com/hyperledger/fabric/blob/master/core/rest/rest_api.go#L1145#L1159) in this way "{u'Amount': u'100', u'Name': u'a'}",  "/chaincode" service returns as a [string](https://github.com/hyperledger/fabric/blob/master/core/rest/rest_util.go#L49) in this way "{\"Name\": \"a\", \"Amount\", \"100\"}". Additionally is not possible escape double quotes in this way '\"' in behave test scenarios:
  
  `Then I should get a JSON response with "result.message" = "{\"Name\":\"a\",\"Amount\":\"100\"}"` 
  **Don't works**
  
  To solve this problem I have chosen format values replacing "\"" by "'" so the scenario was wrote in this way:
  
  `Then I should get a JSON response with "result.message" = "{'Name':'a','Amount':'100'}"`
  
  I think a new issue should be open again to fix "/chaincode" service in order to be consistent between "/devops/..." and "/chaincode" formats.

xspeedcruiser:
@srderson @andresgaragiola The original tests can be modified to pass 4 for JAVA language type.

andresgaragiola:
Thanks @xspeedcruiser in my opinion pass JAVA is easier than pass 4. As is coded now in the scenario JAVA is passed and it is translated to 4 in the step.

srderson:
Rerunning Travis. It timed out earlier.

rameshthoomu:
reverify jenkins

andresgaragiola:
@rameshthoomu  jenkins/x86_64 is working without errors and travis-ci still failing by time out (jenkins/z  I think is not working because has failed in all the PRs that I have checked).  @srderson Let me know if you need me to remove temporarily the second scenario in order to reduce the build times in travis.

rameshthoomu:
reverify jenkins

srderson:
Will rerun Travis now that PR #2184 has been merged.

srderson:
Nice, looks like Travis is passing now that it has been split into multiple jobs. LGTM

Issue #1959, opened 2016-06-23
Print version info


kchristidis:
We'll need a way to print the version info using the `peer` binary.

The expected behavior is that `peer version`, `peer -v`, and `peer --version` should return version info by reading the `peer.version` key in the `core.yaml` file and quit.

This is a fairly straightforward task and a good opportunity for someone to get their hands wet with submitting patches to the project. They'll be using the [cobra](https://github.com/spf13/cobra), [viper](https://github.com/spf13/viper), and [pflag](https://github.com/ogier/pflag) packages.

@srderson If you agree, can we add the "Help wanted" tag to invite participation?

GrapeBaBa:
@kchristidis @srderson Assign to me:)

srderson:
Thanks @GrapeBaBa, you got it

christo4ferris:
Thinking about this... seems like the API should also report peer version info so we can detect incompatible versions... this could be independent, but seems like we would need this once we are live in the field.

JonathanLevi:
Yes, I may have an idea/suggestion of how to put some more info into the version which would tremendously help.

@GrapeBaBa, your issue ;-), but happy to collaborate, and suggest/would like to have it in Membership Services' server too. Will quickly add/introduce some command line/prompt parameters so that we can do/use this in the server too.

GrapeBaBa:
@christo4ferris Glad to add it to API once I fix this issue.

GrapeBaBa:
@JonathanLevi Will do. Ping me a slack when you are online.

GrapeBaBa:
Just note #1962 ,the decision will impact this issue.

mastersingh24:
@GrapeBaBa  - We should probably be a little more clear in #1962.  I did not mean that we get rid of every command line option, just the ones which specifically override / set configuration parameters.  Having a way to output version would be a good example, although I think you would have either `peer -v` and `peer --version`  or you would `peer version`.  I don't think we should mix styles

GrapeBaBa:
@mastersingh24 Clear. @kchristidis @srderson @JonathanLevi @christo4ferris what do you think about the advice which not mix styles?

kchristidis:
FWIW, `git` does both `git version` and `git --version`. And `docker` does all three. So there is definitely a mixing of styles going in on the industry. I am not an expert on this, so I don't have a strong take.

GrapeBaBa:
[...]

PR #2140, opened 2016-07-08
change init to get

ckeyer:
## Description

a bug , init() is before main(),so, can not get Environment Variables

### DIFF
--- a/core/container/dockercontroller/dockercontroller.go
+++ b/core/container/dockercontroller/dockercontroller.go
@@ -32,7 +32,7 @@ import (
 
 var (
 	dockerLogger = logging.MustGetLogger("dockercontroller")
-	hostConfig   = new(docker.HostConfig)
+	hostConfig   *docker.HostConfig
 )
 
 //DockerVM is a vm. It is identified by an image id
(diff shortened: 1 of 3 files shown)


srderson:
@yeasy Does this solve #1269 for you? Someone on Slack mentioned that this still was not working.

GrapeBaBa:
@srderson this can partial fix what i found in #1959

yeasy:
@srderson  seems not work for me.
btw, it's a little bit strange for vipper's design. Its doc suggests put config/init operations into init(), while mostly we have to put in main().

@ckeyer In the title, maybe you mean "chang" --> "change"?

ckeyer:
the Environment Variables not be used. `viper.UnmarshalKey`

srderson:
@ckeyer Is it possible to add a unit test for the `getDockerHostConfig` function?

ckeyer:
@srderson PTAL

srderson:
LGTM, thanks! It seems like this is a partial fix, so I'll merge. @yeasy Could you open PR with a test case for the problem that you're still seeing? Then we can work there to resolve.

yeasy:
With latest code, I've successfully ported hyperledger fabric on top of my container clouds.
Thanks!

christo4ferris:
nice!

PR #2142, opened 2016-07-08
Use Txid instead of UUID

gabre:
This PR changes the ID of a TX to be an SHA256 hash. We use the UTF8 byte form of the arguments from the Chaincode Invocation to generate the hash.
Note that there will be another PR changing Chaincode API to operate with only one byte array argument. This will make hashing easier (no need for UTF8 or any conversion). <PR_number_placeholder>
## Description

The following modifications are included:
- SHA256 hash for TX ID (not only the first 128 bits)
- Removal of previously used random generated UUID
- Renaming of UUID to TxID where it was needed
- It is still possible to add other custom TX ID generation algorithms (using TX payload) but this is the default.
## Motivation and Context

UUID cannot be calculated by client. This PR makes it possible:
- the user has to create byte arrays from the arguments of the chain code invocation
- the byte arrays has to be used to generate the SHA256 hash
## Fixes

No fixes, only new functionality.
## How Has This Been Tested?

Unit tests were run.
## Checklist:

Signed-off-by:
Gabor Hosszu gabor@digitalasset.com


bddtests/syschaincode/noop/chaincode.go
```diff
@@ -30,7 +30,7 @@ import (
 var logger = shim.NewLogger("noop")
 
 type ledgerHandler interface {
-	GetTransactionByUUID(txUUID string) (*protos.Transaction, error)
+	GetTransactionByID(txUUID string) (*protos.Transaction, error)
```
esgott: I think the parameters should be renamed as well.

core/chaincode/handler.go
```diff
@@ -331,7 +331,7 @@ func (handler *Handler) processStream() error {
 				chaincodeLogger.Debug("Next state nil message, ending chaincode support stream")
 				return err
 			}
-			chaincodeLogger.Debugf("[%s]Move state message %s", shortuuid(in.Uuid), in.Type.String())
+			chaincodeLogger.Debugf("[%s]Move state message %s", shorttxid(in.Uuid), in.Type.String())
```
esgott: why does in has Uuid?

core/chaincode/handler.go
```diff
@@ -581,7 +581,7 @@ func (handler *Handler) afterGetState(e *fsm.Event, state string) {
 		e.Cancel(fmt.Errorf("Received unexpected message type"))
 		return
 	}
-	chaincodeLogger.Debugf("[%s]Received %s, invoking get state from ledger", shortuuid(msg.Uuid), pb.ChaincodeMessage_GET_STATE)
+	chaincodeLogger.Debugf("[%s]Received %s, invoking get state from ledger", shorttxid(msg.Uuid), pb.ChaincodeMessage_GET_STATE)
```
esgott: Why does msg has Uuid?

gregmisiorek:
if all 3 checks have failed, but the branch has no conflict, then the fork is successful, isn't it?

srderson:
@gabre @esgott Yes on those other UUID references. These point to https://github.com/hyperledger/fabric/blob/master/protos/chaincode.proto#L152 which likely should be changed to `txID` also. Confirmed with @muralisrini. Of course this is a breaking API change so that needs to be considered.

srderson:
@gregmisiorek A couple of behave tests failed in Travis. Sometimes these flap due to timeouts so I will rerun ton confirm. Jenkins is not reliable yet.

gabre:
@rameshthoomu You are right, those might be real failures. The PR changes the previous UUID mechanism to another which is called 'sha256base64' that assumes the 1st argument of the chaincode invocation to be a base64 encoded string. This should be changed/fixed somehow.

gabre:
We could do the following changes to this PR:
- sha256 calculation: it can remain as it is (calculated from the 1st argument of the chaincode invocation... but this does not seem to be general) OR we can calculate that hash from e.g. the wire format version of the chaincode invocation (a byte array made by protobuf Marshal)
- base64: we should remove this. In case we use protobuf wire format for hash calculation, it is trivially removable.
  @tamasblummer @esgott What is your opinion?

bddtests/syschaincode/noop/chaincode.go
gabre: Done.

core/chaincode/handler.go
gabre: Done.

core/chaincode/handler.go
gabre: Done.

gabre:
@srderson Could you check this please?

gabre:
@srderson I rebased it.

gabre:
'chaincode authorizable_counter with 4 peers, two stopped, bring back both' is this unstable? Before rebasing, it was OK.

christo4ferris:
reverify Jenkins

christo4ferris:
@gabre we will be transferring the fabric and fabric-api repository and all the corresponding issues and pull requests to the hyperledger-archives org on GH and replacing this repository with a read-only mirror of our Gerrit source. As a result, any automatic GH redirects will be broken. If you haven't already migrated your PR to Gerrit, and intend to do so, you will need to change your git remote to reference the transferred repository (e.g. git@github.com:hyperledger-archives/fabric) on your local clone.

PR #2156, opened 2016-07-10
Move cc events out of transactionresults#issue2118

### LINKED ISSUE #2118: move chaincode events from TransactionResults into NonHashData
## Description

With PR #1719 chaincode events need a new home. They should be moved directly into NonHashData.

This potentially will go away in New Consensus when we could move chaincode events directly into transaction (which HAS to happen if NonHashData goes).
## Describe How to Reproduce

This needs to be done once PR #1719 is accepted. Without it chaincode events won't be sent out and https://github.com/hyperledger/fabric/tree/master/examples/events/block-listener will fail.

muralisrini:
With PR #1719 TransactionResults was removed from NonHashData and with that we needed to find a new home for ChaincodeEvents. This PR moves ChaincodeEvents from TransactionResults directly into NonHashData.
## Description

Following changes were applied
- create a new ChaincodeEvents field in NonHashData protobuf
- copy chaincode events passed to the ledger to the NonHashData->ChaincodeEvents field at the time of commit to ledger
- modified the block-listener example to listen for chaincode events.
- Updated block-listener README
## Motivation and Context

PR #1719 breaks sending and storing of chaincode events. This PR fixes that.
Fixes #2118
## How Has This Been Tested?

Test case in events test even framworks and chaincode events. Test case in chaincode tests chaincode event generation. Enhanced examples/events/block_listener (and updated README in that folder).
Followed README to test event generation (good tx, bad tx, chaincode events) against PBFT network.
## Checklist:


core/ledger/ledger.go
```diff
@@ -159,7 +159,25 @@ func (ledger *Ledger) CommitTxBatch(id interface{}, transactions []*protos.Trans
 	writeBatch := gorocksdb.NewWriteBatch()
 	defer writeBatch.Destroy()
 	block := protos.NewBlock(transactions, metadata)
-	block.NonHashData = &protos.NonHashData{}
+
+	ccEvents := []*protos.ChaincodeEvent{}
+
+	if transactionResults != nil {
+		ccEvents = make([]*protos.ChaincodeEvent, len(transactionResults))
+		for i := 0; i < len(transactionResults); i++ {
```
gabre, line 167: It is just a nit but isn't it more "go-ish" to iterate using for i, elem := range.. ?

core/ledger/ledger.go
```diff
@@ -159,7 +159,25 @@ func (ledger *Ledger) CommitTxBatch(id interface{}, transactions []*protos.Trans
...
+			if transactionResults[i].ChaincodeEvent != nil {
+				ccEvents[i] = transactionResults[i].ChaincodeEvent
+			} else {
+				//protobuf does not like elements of a repeated array to be nil
+				//we should discard empty events without chaincodeID when sending
+				//out events
+				ccEvents[i] = &protos.ChaincodeEvent{}
```
gabre, line 174: If I am right you don't want to create a ccEvent for TXs which have no ChaincodeEvent in their TXresult. Consider the following: 
- ccEvents = make([]*protos.ChaincodeEvent, 0, len(transactionResults)) <-- init an array with zero length but enough capacity for the case when there is a ccEvent for each TX/TXresult.
- Then use 'if txRes[i].CCEvent != nil { ccEvents = append(ccEvents, txRes[i].CCEvent) }' to add an element if needed.
- This results in no nil / empty CCEvents and it is 'go-ish'. Btw the rest of the code gets simpler.

core/ledger/ledger.go
muralisrini, line 167: I thought over that a bit....I like that I know exactly what i'm getting when iterating old fashioned way (definitely no extra instructions). For this purpose its exactly what's needed - nothing more or less.

core/ledger/ledger.go
muralisrini, line 174: Will take another look and update.

muralisrini, line 174: The problem with appending non-nil events is that we lose indexes, ie, we lose tx <->chaincodeevent association. Persisting chaincode event without that association is useless.

gabre, line 174: Yes, you are right, you will lose indexes that way. Btw it is a bit 'weird' that we have to use empty ChaincodeEvents as null values (they are semantically different). I will try to find some better/nicer solution for the future but now this is OK.

core/ledger/ledger.go
gabre, line 167: With for i, item := range someArray {} you know what you get - the same when you iterate the old fashioned way just with a better syntax.

muralisrini:
moving to gerrit... need to rebase, etc. Closing.

internal/cryptogen/ca/ca_test.go
```diff
@@ -62,6 +64,8 @@ func TestLoadCertificateECDSA(t *testing.T) {
 		testOrganizationalUnit,
 		testStreetAddress,
 		testPostalCode,
+		365*15*24*time.Hour,
+		365*24*time.Hour,
```
sykesm: Why are we repeating these constants at every call site instead of using the pattern that has already been established of using constants declared earlier in the file? This is pretty fragile.

internal/cryptogen/ca/ca_test.go
```diff
@@ -8,11 +8,13 @@ package ca_test
 import (
 	"crypto/ecdsa"
 	"crypto/x509"
+	"github.com/stretchr/testify/assert"
```
sykesm: Why do we need assert when the rest of the tests use require?

Also, the import grouping (which should actually be enforced in ci - which I'll follow up on) should have standard library packages first and external packages towards the end.

Regardless, assert should be removed and require used in the tests.

internal/cryptogen/ca/ca_test.go
```diff
@@ -205,7 +215,8 @@ func TestGenerateSignCertificate(t *testing.T) {
 	// KeyUsage should be x509.KeyUsageDigitalSignature | x509.KeyUsageKeyEncipherment
 	require.Equal(t, x509.KeyUsageDigitalSignature|x509.KeyUsageKeyEncipherment,
 		cert.KeyUsage)
-	require.Contains(t, cert.ExtKeyUsage, x509.ExtKeyUsageAny)
+	assert.Contains(t, cert.ExtKeyUsage, x509.ExtKeyUsageAny)
+	validateCertExpiry(t, cert, signedCertExpiry)
```
sykesm: The expiry assertions should really be in a separate test.

internal/cryptogen/ca/ca_test.go
```diff
@@ -205,7 +215,8 @@ func TestGenerateSignCertificate(t *testing.T) {
 	// KeyUsage should be x509.KeyUsageDigitalSignature | x509.KeyUsageKeyEncipherment
 	require.Equal(t, x509.KeyUsageDigitalSignature|x509.KeyUsageKeyEncipherment,
 		cert.KeyUsage)
-	require.Contains(t, cert.ExtKeyUsage, x509.ExtKeyUsageAny)
+	assert.Contains(t, cert.ExtKeyUsage, x509.ExtKeyUsageAny)
```
sykesm: require was intentional and should be preserved.

internal/cryptogen/ca/ca_test.go
```diff
@@ -223,9 +234,10 @@ func TestGenerateSignCertificate(t *testing.T) {
 	ous := []string{"TestOU", "PeerOU"}
 	cert, err = rootCA.SignCertificate(certDir, testName, ous, nil, &priv.PublicKey,
 		x509.KeyUsageDigitalSignature, []x509.ExtKeyUsage{})
-	require.NoError(t, err)
-	require.Contains(t, cert.Subject.OrganizationalUnit, ous[0])
-	require.Contains(t, cert.Subject.OrganizationalUnit, ous[1])
+	assert.NoError(t, err)
```
sykesm: require is intentional and should be preserved.

internal/cryptogen/ca/ca_test.go
```diff
@@ -263,3 +275,11 @@ func checkForFile(file string) bool {
 	}
 	return true
 }
+
+func validateCertExpiry(t *testing.T, cert *x509.Certificate, expiry time.Duration) {
+	notBefore := time.Now().Round(time.Minute).Add(-5 * time.Minute).UTC()
+	notAfter := notBefore.Add(expiry)
+
+	assert.WithinDuration(t, notBefore, cert.NotBefore, 1*time.Minute, "Certificate NotBefore not in expected range")
+	assert.WithinDuration(t, notAfter, cert.NotAfter, 1*time.Minute, "Certificate NotAfter not in expected range")
```
sykesm: please use require; no need to multiple with a constant of `1`.

internal/cryptogen/msp/msp_test.go
```diff
@@ -6,10 +6,12 @@ SPDX-License-Identifier: Apache-2.0
 package msp_test
 
 import (
+	"github.com/stretchr/testify/assert"
```
sykesm: assert instead of require; please reorder imports.

cmd/cryptogen/main.go
```diff
@@ -67,13 +68,17 @@ type UsersSpec struct {
...
-	Template      NodeTemplate `yaml:"Template"`
-	Specs         []NodeSpec   `yaml:"Specs"`
-	Users         UsersSpec    `yaml:"Users"`
+	Name               string        `yaml:"Name"`
+	Domain             string        `yaml:"Domain"`
+	EnableNodeOUs      bool          `yaml:"EnableNodeOUs"`
+	CACertExpiry       time.Duration `yaml:"CACertExpiry"`
+	IdentityCertExpiry time.Duration `yaml:"IdentityCertExpiry"`
+	TLSCertExpiry      time.Duration `yaml:"TLSCertExpiry"`
+
```
sykesm: nit: Please remove the blank line.

Issue #2158, opened 2016-07-11
Untracked Go files are not added to the docker images


juliancarrivick-ibm:
## Description

For readability and maintainability it would be nice to be able to be split the functionality in `peer/main.go` into multiple files. However, once anything is moved into another file (still in package `main`), the peer image is unable to be create is unable to be created.

However, the executable can still be compiled: `make peer`
## Describe How to Reproduce

Create `peer/test.go` with contents:

``` go
package main

testVar := 1
```

In `peer/main.go` add a print statement referencing `testVar`:

``` go
func main() {
     fmt.Println(testVar)
     ...
```

Run `make peer-image`:

``` bash
opt/gopath/src/github.com/hyperledger/fabric/peer/main.go:229: undefined: testVar
```

juliancarrivick-ibm:
Stumbled onto this when refactoring for issue #802, I am going to attempt to fix this before continuing with it.

juliancarrivick-ibm:
It seems that only files tracked by Git are currently copied into the docker images which means compilation fails when new files are added but not checked in yet.
I will modify the Makefile accordingly.

PR #2159, opened 2016-07-11
Add untracked Go files to docker images

juliancarrivick-ibm:
## Description

Allows untracked Go files to be added to the docker src image so that local changes do not have to be checked in to be used during development
## Motivation and Context

Fixes issue #2158.
During development new Go files will be created, this allows Devs to use these files in the docker containers without checking them in.
## How Has This Been Tested?

This is not a code change, so I have not added any new test cases. I have tested this by running `make checks` which has verified this hasn't broken any of the build process.
## Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -152,7 +152,9 @@ build/image/src/.dummy: build/image/base/.dummy $(PROJECT_FILES)
 	@echo "Building docker src-image"
 	@mkdir -p $(@D)
 	@cat images/src/Dockerfile.in > $(@D)/Dockerfile
-	@git ls-files | tar -jcT - > $(@D)/gopath.tar.bz2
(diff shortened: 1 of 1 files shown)


ghaskins:
Hello Julian,

Thank you for the patch.  I am a little concerned this will have unintended consequences (e.g. pulling in intermediate files).  Note that git ls-files will include files that have been added-but-not-yet-committed.  Therefore, wouldn't a sensible workaround be to leave the Makefile as is and require users to git-add files before it would be expected for them to work properly?

juliancarrivick-ibm:
@ghaskins I did not realise that adding it but not committing it would do that. 
I agree, that is an acceptable and sensible workaround.

PR #2160, opened 2016-07-11
Fix bugs in hlc.ts-The this notation in function private newNetModeTr…

### LINKED ISSUE #2157: A bug in hlc sdk file hlc.ts
## Description

I  set up a network with 4 peers running on dockers. I tried to deploy the transaction on specific peer with peer.sendTransaction() method. The code went wrong in hlc.js in the function TransactionContext.prototype.newNetModeTransaction:
/opt/gopath/src/github.com/hyperledger/fabric/sdk/node/lib/hlc.js:1342
                        this.binding = new Buffer(self.chain.cryptoPrimitives.
                                     ^
TypeError: Cannot set property 'binding' of undefined

It is solved by changing it to self.binding. The "self" is defined above in the same function. You need also to change several other lines in the same block. That is, this-> self. You need to revise the hlc.ts file and rebuild it. To revise the hlc.ts, goes to line 1685 to line 1702. To rebuild, go to node/ folder and make. 
## How to reproduce

N/A

zhoyy007:
## Description

In the file sdk/node/src/hfc.ts, in the function newNetModeTransaction, in the if block "if (request.userCert) {}", change "this" to "self". 
## Motivation and Context

Fixes #2157
The "this" is not referred to the correct object. Which lead to the following error:
/opt/gopath/src/github.com/hyperledger/fabric/sdk/node/lib/hlc.js:1342
                        this.binding = new Buffer(self.chain.cryptoPrimitives.
                                     ^
TypeError: Cannot set property 'binding' of undefined
## How Has This Been Tested?

First, go to /opt/gopath/src/github.com/hyperledger/fabric/sdk/node/ and make to build.
Then, start membersrvc and peer in net mode. Enroll a user and deploy a chaincode. In the deployrequest, add the userCert gotten from getNextTCert of member object.
The error is removed with the above changes.
## Checklist:

Signed-off-by: zhoy zhang.4746@osu.edu
…ansaction is not recognized. Change it to self.Fixed #2157

### DIFF
--- a/sdk/node/src/hfc.ts
+++ b/sdk/node/src/hfc.ts
@@ -1769,11 +1769,11 @@ export class TransactionContext extends events.EventEmitter {
                         let nonceRaw = new Buffer(self.nonce);
                         let bindingMsg = Buffer.concat([certRaw, nonceRaw]);
                         // debug('========== Binding Msg [%s]', bindingMsg.toString('hex'));
-                        this.binding = new Buffer(self.chain.cryptoPrimitives.hash(bindingMsg), 'hex');
-                        // debug('========== Binding [%s]', this.binding.toString('hex'));
+                        self.binding = new Buffer(self.chain.cryptoPrimitives.hash(bindingMsg), 'hex');
+                        // debug('========== Binding [%s]', self.binding.toString('hex'));
                         let ctor = chaincodeSpec.getCtorMsg().toBuffer();
                         // debug('========== Ctor [%s]', ctor.toString('hex'));
-                        let txmsg = Buffer.concat([ctor, this.binding]);
+                        let txmsg = Buffer.concat([ctor, self.binding]);
                         // debug('========== Payload||binding [%s]', txmsg.toString('hex'));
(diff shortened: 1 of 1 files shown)


zhoyy007:
Signed-off-by: zhoy zhang.4746@osu.edu

srderson:
@smithbk or @angrbrd please review, thanks

angrbrd:
Hi @zhoyy007 

I see what you're doing in this change, but I'd like to better understand how it causes issue #2157 without you observing it in other ways. As this same usage of this --> self appears in two other places, one of which is invoke/query. Could you please elaborate in issue #2157 how you are triggering the deploy transaction for your test? Thanks.

zhoyy007:
Hi @angrbrd 
This issue occurs when I try to deploy chaincode with request which specify the userCert field. That is, to make the block "if (request.userCert) {}" in function newNetModeTransaction to run. It doesn't happen if userCert field is not specified. You can observe that this is the only place where "this" appears in a callback function. I think this is the reason that "this" can not refer to the object expected in a callback function. And, self is used since it is defined outside. 
To reproduce the problem, start membersrvc and peer in net mode. login a user and write a deployrequest with the userCert field. use member.deploy method to deploy.

srderson:
Thanks, LGTM

PR #2175, opened 2016-07-12
Fix #769 - Remove crypto package log level handling

### LINKED ISSUE #769: Confusion about or new specification of module-level logging setup [re: crypto]?

juliancarrivick-ibm:
## Description

Remove log level retreval from crypto_settings.go and the configuration value from core.yaml
Add tests to prevent regressions
## Motivation and Context

Fixes #769. Prevents confusion on how logging is done by removing the override of the crypto package logging level.
## How Has This Been Tested?

Added test cases to ensure that the logging level doesn't get overridden. 
Manually verified that the crypto logs were following the level of the peer command.
## Checklist:

### DIFF
--- a/core/crypto/crypto_settings.go
+++ b/core/crypto/crypto_settings.go
@@ -29,22 +29,7 @@ var (
 // Init initializes the crypto layer. It load from viper the security level
 // and the logging setting.
 func Init() (err error) {
-	// Init log
-	log.ExtraCalldepth++
-
-	level, err := logging.LogLevel(viper.GetString("logging.crypto"))
-	if err == nil {
-		// No error, use the setting
-		logging.SetLevel(level, "crypto")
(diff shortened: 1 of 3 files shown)


srderson:
LGTM, thanks!

PR #2186, opened 2016-07-13
Fix #2185 make wrong command return non-zero value

### LINKED ISSUE #2185: Inconsistent cli document about wrong command execution will return non-zero value
## Description

Run any wrong command, then run `echo $?`, will return 0.
## Describe How to Reproduce

```
peer dsdsd
echo $?
```

GrapeBaBa:
## Description

Uncomment os.Exit(1) in main.go and remove main_test.go
## Motivation and Context

Fixes #2185 
1.Now after run wrong command will return 0 value and it is inconsistent with the document https://github.com/hyperledger/fabric/blob/master/docs/API/CoreAPI.md#cli
2.Now the main_test.go actually will run a wrong command `peer -test.timeout=20`, that is not a good test
## How Has This Been Tested?

make checks successfully
## Checklist:


peer/main_test.go
```diff
@@ -39,5 +39,5 @@ func TestSystem(t *testing.T) {
 	// if *systemTest {
 	// 	main()
 	// }
-	main()
+	//main()
```
gabre: This is "dead" code (commented). Should we preserve dead code? (Git implicitly preserves it in the history)

gabre:
Just a nit: maybe adding some motivation (why this fixes that, OK I know it is trivial) would be good for documentational purposes.

peer/main_test.go
GrapeBaBa: yes, i'd like to remove main_test.go. what about your thoughts?

srderson:
LGTM, thanks

PR #2189, opened 2016-07-13
 clarify setup instructions

joequant:
Clarifies doc to as for what needs to be changed in membersrvc.yaml
## Description

Doc change to allow new users to see what has be to added to membersrvc.yaml for a working secure network
## Motivation and Context

This clarifies the setup instructions so that new users can see what
exactly must be changed to add a secure node.

Signed-off-by: Joseph C Wang joequant@gmail.com


gabre:
IMO this can be merged.

srderson:
@joequant LGTM, thanks. @gabre Thanks for the review

PR #2216, opened 2016-07-15
Remove accounts from ecerts

diegomasini:
Accounts were introduced before the incorporation of attributes to TCerts. Since accounts have a more dynamic aspect it makes more sense to put them into TCerts.
## Description

Removed the inclusion of accounts into ECerts during ECert generation. Accounts will be handled as attributes on TCerts.
## Motivation and Context

Accounts have a more dynamic aspect and there should be more than one account per user for asset transfer to avoid linkability. Given that ECerts are long term certificates it has more sense to handle accounts as attributes on TCerts.
## How Has This Been Tested?

All the previous tests run correctly. No additional tests were added since this change implies the removal of a particular field from the ECerts, avoiding the need to pass the value for that field as a parameter in the request for ECerts.
## Checklist:

Signed-off-by: Diego Masini damasini@us.ibm.com / diego.masini@gmail.com


diegomasini:
Hello @angrbrd, could you take a look at these changes? I made some modifications to the client SDK since there is no need to pass the account as part of an ECert request. Thanks!

angrbrd:
Hi @diegomasini 

I reviewed this change and everything looks good to me in general as well as the modifications relating to the client SDK. I have also verified the Node UTs in my environment on this PR and I can confirm that they pass. 

The only thing I am not 100% sure on is the modifications you made to the `membersrvc/protos/ca.proto`. I see that you have removed the field and renumbered the other fields to decrement them by 1 to account for the fact that you removed a field in the middle. I am wondering if someone else knows better the convention here. Should unused fields be removed like this? Or just left untouched? @srderson or @jeffgarratt could you please comment?

srderson:
Renumbering fields will break backwards compatibility if that's a concern. You can just remove the field and leave the others numbered at they currently are.

diegomasini:
@srderson I made the change you suggested but Travis CI is failing. I run the behave tests locally and in the Travis instance associated to my fork and they are passing, it seems to be a timeout issue. Could you relaunch the build? Thanks!

srderson:
They are passing now. Thanks, LGTM

PR #2220, opened 2016-07-15
[WIP] Issue #2119

### LINKED ISSUE #2119: Ability to unit test Go chaincode
## Description

Unit testing chaincode is difficult today as there's no standard way to mock the shim.ChaincodeStub object. This is needed by anyone writing a chaincode. This needs code (possibly there's an existing Go framework that could be used) in addition to documentation.

brad-gorman:
## Description

This pull request is only partially complete. I would dearly like feedback before progressing futher with this.
It provides an interface that allows chaincode developers to easily unit test their own code.
Implemented functions are: Init() Invoke() Query() RangeQueryState() GetState() PutState() DelState() InvokeChaincode() QueryChaincode().
Unimplemented functions: anything related to tables, certificates and security.
This pull request does away with the shim.ChaincodeStub pointers used in most chaincode and replaces them with a shim.ChaincodeStubInterface - so there will be matching changes needed on the user side as well.
## Motivation and Context

The current implementation does not allow unit testing of chaincode, which will be important for complex contracts.
Partially fixes #2119 
## How Has This Been Tested?

There are unit tests implemented for chaincode_example02 through 05
## Checklist:

Signed-off-by:


ghaskins:
Hi Brad,

I support this effort, so thank you.  I am curious: Why did you opt to force chaincode to change from ChaincodeStub to ChaincodeStubInterface?  Could we do it the other way around (leave the chaincode unmodified and with "ChaincodeStub" as the abstraction, and then manage the objects behind the scenes?

brad-gorman:
@ghaskins Thanks for taking the time to comment!

I chose the ChaincodeStub so that the unit testing would only be testing the chaincode written by the users. That and core/chaincode/shim/chaincode.go seems to have quite a lot of code in it. I see most of the function is externalised into a Handler object, but there's still a lot of functionality in there.

Instead, I could insert an interface between ChaincodeStub and Handler if you think that would be a better approach. That should let us avoid changing any of the existing functions of ChaincodeStub.

christo4ferris:
@brad-gorman I also support this direction - thanks. 

You might use the convention of editing the PR title to prefix with '[WIP]' so that the maintainers can bypass until you are ready for another review.

brad-gorman:
@christo4ferris sorry, I updated the title. Also it's not clear to me. Are you in support of the interface to replace ChaincodeStub or to replace Handler?

brad-gorman:
Mocking Handler out seems considerably more complicated than doing the same thing at the ChaincodeStub level. As well as dealing with the actual state being stored, it maintains a state machine and passes the heavily structured ChaincodeMessage(s) around as its main form of communication. I feel like I'm going to have to re-implement a lot of the Handler code more-or-less the same to get a MockHandler working. I feel like I should also be hiding the complexity of ChaincodeMessage from the unit-test user.

christo4ferris:
I'm a little torn, because this will be disruptive to any currently written chaincode. 

HOWEVER, I am a yooge fan of 'everything gets an interface', and reminded just how burned we have collectively been by broken language design (I'm looking at you Java). When Sun decided to be led by the nose by some fantasy that early applets of juggling Duke somehow needed to have priority over getting the language right at around the 1.0 timeframe. They could and should have finished the interfaces for everything work and disrupted a few early applets to make the language that much more powerful for the 25 years that followed).

There isn't enough meaningful chaincode out there. I would prefer to see us do this right.

That said, it isn't up to me. I think that the @hyperledger/fabric-maintainers need to make this decision.

brad-gorman:
@christo4ferris thanks. I think I'll park this for now until the fabric maintainers make a call. I really do feel like chaincode should be written to and tested by an interface. If I need to do something to get this looked at more, please let me know as I'm not familiar with the process here.

I will try and assess how much impact this would be to existing chaincode though. From a code point of view it's pretty simple to update - change your *shim.ChaincodeStub arguments to shim.ChaincodeStubInterface. I'll try to check if this would change hashes or anything else.

brad-gorman:
Here is what I can tell would impact a user. The list is probably incomplete.

Users will need to update their chaincode to have it compile. Function signatures that use *shim.ChaincodeStub will need to be shim.ChaincodeStubInterface.

```
-func (t *SimpleChaincode) Init(stub *shim.ChaincodeStub, function string, args []string) ([]byte, error) {
+func (t *SimpleChaincode) Init(stub shim.ChaincodeStubInterface, function string, args []string) ([]byte, error) {
```

It looks like the chaincode hash has changed as a result of this update (observation of chaincode_example02). So if a user was to redeploy their existing chaincode (after modifying the signatures), anything that relied on the chaincode hash would also need to be updated (see chaincode_example04). References via chaincode names or URLs would obviously be unaffected.

christo4ferris:
@brad-gorman I think that this is the right thing to do, despite the disruption to existing chaincode. Would you please draft up release note entry that provides users with precise changes that they will need to make, suitable for publishing. Thanks.

@srderson think we discussed this in slack... would you might weighing in? Thanks. @binhn @muralisrini as well. I do think it important that we get this right rather than have it difficult to test chaincode by imposing a requirement to run integration tests rather than unit tests.

brad-gorman:
@christo4ferris Are there any examples of release notes for this project? I'm not entirely sure what sort of format or length they should take.

christo4ferris:
@brad-gorman release notes are in the wiki. I'd link but that link will be broken shortly.

we will be transferring the fabric and fabric-api repository and all the corresponding issues and pull requests to the hyperledger-archives org on GH and replacing this repository with a read-only mirror of our Gerrit source. As a result, any automatic GH redirects will be broken. If you haven't already migrated your PR to Gerrit, and intend to do so, you will need to change your git remote to reference the transferred repository (e.g. git@github.com:hyperledger-archives/fabric) on your local clone.

PR #2162, opened 2016-07-11
Fix proto build

gabre:
## Description

This PR fixes proto build ('make protos') which gave an error cause of a wrong import.
## Motivation and Context

To motivation is to make protos compile again on master.

## Checklist:

- [x] I have added a [Signed-off-by](https://github.com/hyperledger/fabric/blob/master/CONTRIBUTING.md#legal-stuff).
- [x] I have either added documentation to cover my changes or this change requires no new documentation.
- [x] I have either added unit tests to cover my changes or this change requires no new tests.
- [x] I have run [golint](https://github.com/golang/lint) and have fixed valid warnings in code I have added or modified. This tool generates false positives so you may choose to ignore some warnings. The goal is clean, consistent, and readable code.

Signed-off-by: Gabor Hosszu gabor@digitalasset.com

PR #2224, opened 2016-07-15
Recompile protos with new script

srderson:
## Description

This PR contains the recompiled protos using the new script in PR #2162 
## Motivation and Context

The proto build script was broken with the introduction of Java chaincode. PR #2162 resolved this and simplified the script. Due to the script simplification, there were a few differences in the results `.pb.go` files. These must be committed. It was also necessary to change the package name of `dah.proto` in the utxo example chaincode to avoid compile errors after compiling with the updated script.
## How Has This Been Tested?

No new tests required as there is no new functionality.
Ran `make checks`. 
## Checklist:


gabre:
Seems OK to me, can be merged @tamasblummer / @srderson

gabre:
On the other side, shouldn't we add proto compilation to the CI? (I guess it is not part of it now.) @ghaskins

srderson:
@gabre Yes, I agree we should add the compilation to the CI. Could we a quick task during the hackfest next week.

ghaskins:
One way to do it is to just get rid of checked in artifacts and have the build generate them from the proto.  This has the advantages of never allowing the .proto stuff to get out of whack, but has the downside of requiring everyone to have the protoc suite in an operating and, ideally, identical configuration to everyone else.

Another way to do it would be to have the CI check things like whether the protos were changed and test that the intermediates were updated too.

Which is preferred?

srderson:
Closing as the PR has been moved to Gerrit. https://gerrit.hyperledger.org/r/#/c/373/

@ghaskins I guess I'd prefer CI to handle this. I think there was a thread on the mailing list where we can continue the discussion.

PR #2234, opened 2016-07-17
Have PrePrepare carry non-opaque RequestBatch message

kchristidis:
## Motivation and Context

Request batches are currently marshaled and stored as opaque blocks. This means redundant marshaling and unmarshaling in order to process them, and also means we don't have transparency on the actual request level when handling resubmissions. This is the first in a series of pull requests that aim to merge the `pbft-batch` and `pbft-core` logic in the `pbft` package. Thanks to @jyellick for the feedback while creating this PR.
## Description

The key change is the introduction of a`request_batch` protobuf message which holds a slice of `request` messages and is now used as the container for all request batches. This change rippled into the entire package; all logic and data structures that dealt with the old request batches have been modified accordingly; same goes for all tests.

I've also removed the useless `validate()` method in the `pbft` package, and cut down on existing code duplication in the tests by adding functions to create commonly-used messages.
## How Has This Been Tested?

All existing tests have been modified accordingly. Passes all unit and behave tests locally.
## Checklist:


consensus/pbft/batch.go
```diff
@@ -128,18 +127,13 @@ func (op *obcBatch) Close() {
...
-
+	op.broadcastMsg(&BatchMessage{Payload: &BatchMessage_Request{Request: req}})
 	op.logAddTxFromRequest(req)
 	op.reqStore.storeOutstanding(req)
 	op.startTimerIfOutstandingRequests()
-
-	// if we believe we are the leader, then process this request
-	leader := op.pbft.primary(op.pbft.view)
-	if leader == op.pbft.id && op.pbft.activeView {
+	if op.pbft.primary(op.pbft.view) == op.pbft.id && op.pbft.activeView {
```
gabre, line 134: For me (with no so much knowledge about this part of the code), the former version was easier to understand.

jyellick, line 134: My vote is to keep this change, I find it more clear on a single line.

consensus/pbft/batch.go
```diff
@@ -261,31 +228,16 @@ func (op *obcBatch) leaderProcReq(req *Request) events.Event {
...
-	// process internally
-	logger.Infof("Creating batch with %d requests", len(reqBlock.Requests))
+	logger.Infof("Creating batch with %d requests", len(reqBatch.Batch))
 	return pbftMessageEvent{
-		msg: &Message{&Message_Request{&Request{
-			Payload:   reqsPacked,
-			Timestamp: earliestRequest.Timestamp,
-			ReplicaId: op.pbft.id},
-		}},
+		msg:    &Message{Payload: &Message_RequestBatch{RequestBatch: reqBatch}},
```
jyellick: Is this necessary to wrap in a message? Could this nor be returned simply as a `RequestBatch`?

consensus/pbft/batch.go
```diff
@@ -449,23 +401,17 @@ func (op *obcBatch) ProcessEvent(event events.Event) events.Event {
...
-			if cert.prePrepare.Request == nil {
-				logger.Warningf("Batch replica %d found a non-null prePrepare with no request, ignoring")
-				continue
-			}
-
-			reqs := &RequestBlock{}
-			if err := proto.Unmarshal(cert.prePrepare.Request.Payload, reqs); err != nil {
-				logger.Warningf("Batch replica %d could not unmarshal request block: %s", op.pbft.id, err)
+			if cert.prePrepare.RequestBatch == nil {
+				logger.Warningf("Replica %d found a non-null prePrepare with no request batch, ignoring")
```
jyellick, line 407: This seems like it should be more than a warning, more critical.  We should probably check for this condition before preparing.

consensus/pbft/messages.proto
```diff
@@ -36,15 +36,15 @@ package pbft;
...
         pre_prepare pre_prepare = 2;
         prepare prepare = 3;
         commit commit = 4;
         checkpoint checkpoint = 5;
         view_change view_change = 6;
         new_view new_view = 7;
-        fetch_request fetch_request = 8;
-        request return_request = 9;
+        fetch_request_batch fetch_request_batch = 8;
+        request_batch return_request_batch = 9;
```
jyellick, line 47: I'm not 100% convinced on modifying this message.  In some ways, I understand that it's possibly more efficient to fetch whole batches at a time, but simultaneously, because backups/clients do broadcast their requests to the network, it seems like an exceptional scenario when we are missing  a full block of requests.

consensus/pbft/viewchange.go
```diff
@@ -435,23 +435,23 @@ func (instance *pbftCore) processNewView() events.Event {
...
 		}
 	}
 
-	if len(instance.missingReqs) == 0 {
+	if len(instance.missingReqBatches) == 0 {
 		return instance.processNewView2(nv)
-	} else if newRequestMissing {
-		instance.fetchRequests()
+	} else if newReqBatchMissing {
+		instance.fetchRequestBatches()
```
jyellick, line 454: I understand this is not a new mechanism, but since it's been touched in this changeset, I think it's at least worth some discussion.  I'm not sure why we generally expect that we can fetch missing requests from other replicas.  In general, I would expect that if other replicas already have the requests, then they will quickly execute to a checkpoint and garbage collect the old requests.

In many ways, it almost seems like the new view message should include the requests which are being reprocessed in the new view.  It would balloon the new view message, but, view change should be a rare case, and, with periodic view change enabled, things are lined up on checkpoint boundaries, expressly to avoid having large xSet/pSet/qSets or having to fetch requests.

consensus/pbft/messages.proto
kchristidis, line 47: I did consider this. So, the digests of the requests comprising a batch need to be communicated somehow. This will allow a replica processing a new-view can to identify the exact requests it's missing from a, err, missing batch. The simplest approach I thought of was to modify the `PQ` message definition so that it carries an array of the digests of the requests included in the batch. Let me know if you had something simpler in mind.

Ultimately I decided against it, to keep the invasiveness of the changes to a minimum.

But also, because I think an argument for efficiency and consistency applies here as well. If we were to avoid exchanging batches in this stage, could we argue that we'd be inconsistent, since we treat the `sendBatch/recvRequestBatch` path differently? You can use the "since backups broadcast their requests to the network, it seems like an exceptional scenario when we are missing a full block of requests" argument to question why we have a leader submit a batch of requests in `sendBatch`. (And IIRC  at one point early during this project the thinking was that the leader should indeed only send digests, and the replicas would ask for the missing individual requests. But I think that we decided that just sending the batch right away was a more efficient/practical way of going at it.)

consensus/pbft/batch.go
kchristidis, line 407: Does this change modify the existing behavior? If it doesn't, do you agree that this issue is better off being addressed in a separate PR? If it does, I'll be glad to tackle it here.

consensus/pbft/batch.go
kchristidis: Right on, fixed in https://github.com/hyperledger/fabric/pull/2234/commits/d38b48105bdff0ae7cca7456b68d9fae21823b77.

consensus/pbft/viewchange.go
kchristidis, line 454: I'm letting that one sink in, but your logic doesn't raise any flags on me. At any rate, do you agree that this is better suited in a separate PR? I'll be happy to tackle it. @corecode What's your take on this?

consensus/pbft/batch.go
jyellick, line 407: Yes, I agree, maybe open an issue for this and defer.

consensus/pbft/messages.proto
jyellick, line 47: Agreed this is the least invasive way for now, but something to consider for future work.

jyellick:
Looks good to me.

consensus/pbft/viewchange.go
corecode, line 454: fetchRequest is required for the new primary.

kchristidis, line 454: But if we add the requests referenced in the `Xset`, as Jason suggests, in the new-view message, then the `fetchRequest` path can be skipped.

corecode, line 454: I'd prefer to stick with the PBFT protocol as described in the paper.

jyellick, line 454: ~~Well, we are already deviating by fetching missing requests, which will not necessarily succeed.~~

I just saw that this is in the paper... I'm less sold now.

corecode, line 454: I thought this was standard protocol behavior.

jyellick, line 454: Sorry if you missed my complaint amendment, I was under the mistaken impression that this was a deviation.  There are certainly scenarios where a replica may never be able to fetch a request, but this is only when 2f+1 other nodes are healthy enough to generate a new stable checkpoint, so I'm willing to write this off as okay.

srderson:
Rerunning Travis

PR #2237, opened 2016-07-18
Improve attribute padding using hash function

andresgaragiola:
## Description

Before this PR a padding of 16 0xFF bytes was added to attribute value in order to verify consistency during decryption. 
As a block cipher is used, if the consistency problem was in a non-padding block, then  the problem was not detected. To avoid this in this PR we are using a sha256 padding truncated to 16 bytes.
## Motivation and Context

Improve consistency check in attribute values. 
## How Has This Been Tested?

I have run unit tests without errors. 
I have run behave tests without errors.
I have added unit tests in attributes_tests.go.
## Checklist:

- [X] no new documentation.
- [X] I have either added unit tests to cover my changes.

Signed-off-by: Andres Garagiola garagiol@ar.ibm.com / andresgaragiola@gmail.com


andresgaragiola:
I have close this PR, I'll send this changes again once attributes encryption will be merged.

PR #2250, opened 2016-07-19
Reduce db access lock competition

GrapeBaBa:
## Description

1.Open the db only in peer start
2.Close the db in peer stop
## Motivation and Context

Currently db.GetDBHandle() has a sync block which actually no need, remove it and only open the db when peer start.
## How Has This Been Tested?

unit-test and behave test successfully
## Checklist:

### DIFF
--- a/core/db/db.go
+++ b/core/db/db.go
@@ -22,7 +22,6 @@ import (
 	"os"
 	"path"
 	"strings"
-	"sync"
 
 	"github.com/op/go-logging"
 	"github.com/spf13/viper"
@@ -45,13 +44,6 @@ var columnfamilies = []string{
 	persistCF,    // persistent per-peer state (consensus)
 }
 
-type dbState int32
-
-const (
-	closed dbState = iota
-	opened
-)
-
 // OpenchainDB encapsulates rocksdb's structures
 type OpenchainDB struct {
 	DB           *gorocksdb.DB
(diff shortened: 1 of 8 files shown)


rameshthoomu:
reverify jenkins

srderson:
@manish-sethi Could you review this one please?

GrapeBaBa:
@manish-sethi  Are you available for reviewing this PR?

manish-sethi:
@GrapeBaBa yes, I plan to go through this today.

GrapeBaBa:
@manish-sethi Thanks.

core/db/db.go
```diff
@@ -73,10 +74,22 @@ func Create() *OpenchainDB {
...
+		openchainDB = Create()
+	})
+	openchainDB.Open()
+}
+
+// Stop the db
+func Stop() {
+	openchainDB.Close()
+}
+
```
manish-sethi, line 78: - I think that both pairs of methods (i.e., start/stop and open/close) are confusing. Can you please make the methods open/close private and use start/stop outside this file (including the files db_test.go and db_test_exports.go).
- Document that now method `GetDBHandle()` requires that db.Start() should have been called before. In other words, method `GetDBHandle()` and `Start()` are not mutually thread-safe.
- Now (see above point) locks inside `Open()` and `Close()` does not make sense and can be removed. Because, `GetHandle()` is without locks (I guess, this is what you intent to achieve). In fact, this is how it was before you had introduced locks in `Open()` and `Close()` :-)

GrapeBaBa, line 78: OK, maybe use document to indicate not guarantee thread safe is simpler.

GrapeBaBa:
Migrated to Gerrit.

PR #2286, opened 2016-07-22
[WIP] Fix Issue 2283

### LINKED ISSUE #2283: Port number conflict with UPnP on Windows (5000)
## Description

Port used for the peer's REST API is 5000. This is apparently used by a number of services, most notably UPnP on Windows. To make things simpler, we should choose a port that is not commonly used.
## Describe How to Reproduce

try any of the default examples on Windows.

christo4ferris:
## Description

change peer REST port from 5000 to 7050 to avoid conflict with UPnP on Windows. Also reassigned all other ports as follows:

7050 REST
7051 peer gRPC
7052 peer CLI
7053 peer events
7054 eCAP
7055 eCAA
7056 tCAP
7057 tCAA
7058 tlsCAP
7059 tlsCAA

NOTE: when this PR lands, it will require people to rebuild their Vagrant development environment(s), so we will want to make an announcement.
## Motivation and Context

Fixes #2283 
Also @bcbrock recommended that we eventually seek a block of ports from IANA. We should hold off on that until we sort out Consensus next.
## How Has This Been Tested?

ran all tests
## Checklist:


bcbrock:
@christo4ferris Is it time for us to apply for official ports from IANA? I believe we could apply for a range of several numbers and then define them later as needed.

Note that port 5075 is assigned to an "Experimental Physics and Industrial Control System" so it is probably a safe choice. However there are also many ports in the 5000 range that are not assigned. 
See: https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.csv

christo4ferris:
I didn't check IANA, but we certainly could do that. There's a sizable block here 7041-7069. I think that we use five presently. What is less clear is that we can simply reserve a block. There are some fairly rigorous rules. What is also unclear to me is why we need so many ports. Seems like the fewer ports we need to expose the better.

bcbrock:
I see that 5075 is also claimed to be used by some Skype product: https://technet.microsoft.com/en-us/library/gg398833.aspx

You are right that we use 5: gRPC, HTTP (REST), Profiling, CLI, and events. 5 ports is not unheard of for one service (see ports 5985-5990). X-windows reserves a block of 64.

If an "Experimental Physics and Industrial Control System" can get a port, then I would think that the "Cobol of the 21st Century" should be able to get a few!

bcbrock:
On second thought, we should probably wait until at least the next architecture before requesting permanent port assignments. We may end up needing many more than 5.

christo4ferris:
I can change this to poach the unassigned range above. It was unclear to me why we went to 30k range for membership services and events. @srderson is there some history?

srderson:
I don't know the history of the membership services port choice. Perhaps @diegomasini may know

christo4ferris:
I'll be updating with the block I highlighted above soon as the build/tests pass

christo4ferris:
NOTE: when this PR lands, it will require people to rebuild their Vagrant development environment(s), so we will want to make an announcement.

corecode:
I think we should aspire to use as few ports as possible.  Collect all GRPC services on one port, and all HTTP services on another.

christo4ferris:
I don't disagree but Think that should be an independent refactor

lehors:
I've tested this on Windows and I have not detected any port conflicts. It appears to work out of the box, which is a nice improvement!
I'd be happier if the all tests passed but unfortunately I can't remember the last time that was true on my machine so that's independent...

christo4ferris:
migrated PR to Gerrit - closing

PR #2291, opened 2016-07-22
Clarify peer command.

thomasleplus:
## Description

This is only a small documentation change.
## Motivation and Context

It clarifies that the command to start a node is `peer node start` and not just `node start`.
## How Has This Been Tested?

This is only documentation but I checked that the correct command is indeed `peer node start`.
## Checklist:

Signed-off-by: Thomas Leplus thomas.leplus@gmail.com

Node.js developers are used to run commands like:

$ node start.js

So their first reaction reading this doc is to run:

$ node start

instead of:

$ peer node start

### DIFF
--- a/docs/dev-setup/build.md
+++ b/docs/dev-setup/build.md
@@ -41,7 +41,7 @@ You should see the following output:
     Use "peer [command] --help" for more information about a command.
 ```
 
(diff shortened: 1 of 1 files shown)


srderson:
Thanks, I can see how this might be confusing. LGTM

thomasleplus:
Thanks!

PR #2287, opened 2016-07-22
Update Consensus plugin changes

kamraanki:
## Description

In documentation, if we want to enable consensus plugin there are changes in 'consensus/obcpbft/config.yaml' it should be 'consensus/pbft/config.yaml'
## Motivation and Context

Fixes #
## How Has This Been Tested?

Setup Dev network
## Checklist:

- [x] I have added a [Signed-off-by](https://github.com/hyperledger/fabric/blob/master/CONTRIBUTING.md#legal-stuff).
- [x] I have either added documentation to cover my changes or this change requires no new documentation.
- [x] I have either added unit tests to cover my changes or this change requires no new tests.
- [ ] I have run [golint](https://github.com/golang/lint) and have fixed valid warnings in code I have added or modified. This tool generates false positives so you may choose to ignore some warnings. The goal is clean, consistent, and readable code.

Signed-off-by:

There must be 'consensus/pbft/config.yaml' instead of 'consensus/obcpbft/config.yaml' as there is no directory named obcpbft under consensus in hyperledger fabric developer preview v0.5 release.

Signed-off-by: Ankit

PR #2293, opened 2016-07-22
Revert "Update Consensus plugin changes"

srderson:
Reverts hyperledger/fabric#2287


srderson:
See comment https://github.com/hyperledger/fabric/pull/2287#issuecomment-234645321

PR #2360, opened 2016-08-04
Fixed the link

EvanCarroll:
Link is not working, this fixes it.

Signed-off-by:

Link is not working, fixed.


christo4ferris:
@EvanCarroll thanks for submitting this. However, we have moved development to [Gerrit](https://gerrit.hyperledger.org/r/#/admin/projects/fabric), and this problem has already been addressed. Thanks again.

